PR #22, opened 2018-07-27
Drop the fixed list of `knative/serving` components from logging.

mattmoor:


logging/config.go
```diff
@@ -118,8 +118,8 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 }
 
 // NewConfigFromConfigMap creates a LoggingConfig from the supplied ConfigMap
```
tcnghia: Update doc string to mentions `components` ?

mattmoor: Done

PR #37, opened 2018-08-02
Use Annotations for common knative validation.

n3wscott:
Here is the start of me looking at replacing some of the validation we do with annotations.

Please let me know if you like this direction of if I should stop.


n3wscott:
Alternates: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/#advanced-topics Heading: Validation

We could use OpenAPI to do validation. But I think the downside is it is all or nothing with OpenAPI.

n3wscott:
/hold

@n3wscott to make a PR with the same functionality but as an OpenAPI config.

n3wscott:
/hold cancel

I think this is still a good idea... 

/assign @mattmoor

apis/validators/k8s_qualified_name.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (v *K8sQualifiedNameValidator) OnParent() bool {
+	return false
+}
+
+func (v *K8sQualifiedNameValidator) OnField() bool {
+	return true
+}
+
+func (v *K8sQualifiedNameValidator) Validate(fieldName string, value interface{}) (bool, *apis.FieldError) {
```
ImJasonH: Could `Validate` just return err, and absence of error means it's valid?

n3wscott:
I am gonna close this now. Learning about annotations in go was interesting but the inability to annotate internal fields from refed objects makes this not useful.

PR #40, opened 2018-08-09
Move VolatileTime into knative/pkg

mattmoor:
`VolatileTime` is a type that enables us to exclude certain volatile timestamp fields from causing semantic equality to report a difference when only that field changes.

We use this in `knative/serving` for `LastTransitionTime`, which can drift forward during reconciliation if different phases of the reconcile flip-flop a condition.

### DIFF
--- a/apis/volatile_time_test.go
+++ b/apis/volatile_time_test.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


apis/volatile_time_test.go
```diff
@@ -0,0 +1,87 @@
+/*
+Copyright 2017 The Knative Authors
```
jonjohnsonjr: 2018?

mattmoor: Done

PR #41, opened 2018-08-09
Support multiple group/kind contexts in shared webhook.

mattmoor:
In order to have a single webhook support multiple domain contexts, this reworks the `Handlers` argument to embed the `schema.GroupVersion` by wrapping the existing keys with it as a `schema.GroupVersionKind`.

This is mostly straightforward, but one oddity is that I discovered that `AdmissionRequest` gets this same tuple as the less capable `metav1.GroupVersionKind`, so there's a silly conversion we have to do.

I tried this manually vendored in serving with KPA and things worked great.

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -107,11 +107,10 @@ type ResourceDefaulter func(patches *[]jsonpatch.JsonPatchOperation, crd Generic
 // AdmissionController implements the external admission webhook for validation of
 // pilot configuration.
 type AdmissionController struct {
-	Client       kubernetes.Interface
-	Options      ControllerOptions
-	GroupVersion schema.GroupVersion
-	Handlers     map[string]runtime.Object
-	Logger       *zap.SugaredLogger
+	Client   kubernetes.Interface
+	Options  ControllerOptions
+	Handlers map[schema.GroupVersionKind]runtime.Object
+	Logger   *zap.SugaredLogger
 }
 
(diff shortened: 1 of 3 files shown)


webhook/webhook.go
```diff
@@ -476,12 +488,19 @@ func (ac *AdmissionController) admit(ctx context.Context, request *admissionv1be
...
+	// Why, oh why are these different types...
+	gvk := schema.GroupVersionKind{
+		Group:   kind.Group,
+		Version: kind.Version,
+		Kind:    kind.Kind,
+	}
+
 	logger := logging.FromContext(ctx)
-	handler, ok := ac.Handlers[kind]
+	handler, ok := ac.Handlers[gvk]
```
n3wscott: fancy

webhook/webhook.go
```diff
@@ -329,30 +328,43 @@ func (ac *AdmissionController) register(
...
+				Resources:   []string{plural},
+			},
+		})
 	}
-	resources.Sort()
+
+	// Sort the rules by Group, Version, Kind so that things are deterministically ordered.
+	sort.Slice(rules, func(i, j int) bool {
+		lhs, rhs := rules[i], rules[j]
+		if lhs.APIGroups[0] != rhs.APIGroups[0] {
```
n3wscott: `...[0]` looks like a bug in the waiting?

mattmoor: Possibly, but all of the singletons are built a handful of lines above.

PR #44, opened 2018-08-14
Remove strict name metadata check to be shareable in eventing

n3wscott:
Eventing uses names like "dev.knatave.github.pullrequest" as a event type name. This fails the generic validation added for serving. Moving this logic to serving and removing from pkg.


mattmoor:
/lgtm
/approve

I'd like to get this check into `knative/serving` asap, so we don't import this and lose coverage.

mattmoor:
poke (for gubernator)

PR #45, opened 2018-08-17
Add common test logging module

srinivashegde86:


test/logging/logging.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2018 The Knative Authors
-
```
adrcunha: nit: the boiletplate **does** contain the blank lines you removed

srinivashegde86: updated.

PR #48, opened 2018-08-23
Share Controller and Reconciler test tooling.

mattmoor:
These are copied directly from https://github.com/knative/serving/pull/1930

### DIFF
--- a/controller/testing/hooks.go
+++ b/controller/testing/hooks.go
@@ -1,0 +1,161 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package testing includes utilities for testing controllers.
+package testing
+
+import (
+	"fmt"
+	"sync/atomic"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	kubetesting "k8s.io/client-go/testing"
+)
+
+// HookResult is the return value of hook functions.
+type HookResult bool
+
+const (
+	// HookComplete indicates the hook function completed, and WaitForHooks should
+	// not wait for it.
+	HookComplete HookResult = true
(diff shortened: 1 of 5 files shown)


dprotaso:
I think prior to pulling the table testing into this repo I think there's more experimentation that could be done within serving.

My concerns of the current state are:
- creating test fixtures is a pain and verbose
- lots of duplication of similar objects in table rows
- table tests are huge
- the current patterns evolved from writing the tests after the fact

### Two potential areas of focus

#### Improve testing infrastructure
- builders for fixtures
  - (upside) readability improved
  - (downside) api specific so these wouldn't be moved to pkg
- provide higher abstractions that make writing table rows more easier & succinct
  - (downside) adds additional complexity

This approach addresses the symptoms but not the cause

#### Improve reconciler design (suggested)

Right now a reconciler has the responsibility of reconciling an object through creating and updating child resources. I think for the most part managing of the resources is independent of one another.

Given this observation we pull out the specific responsibility of managing a child resource to a new component.

Thus a Revision's reconciler would be composed

Thus you could consider reconciliation of say a Revision to be the collection of the following sub-reconcilations

mattmoor:
@dprotaso Ack, I don't by any means think moving this freezes it.  My hope is simply that as folks improve it in these ways that all of the repos can move forward.

I'd love to see more PoCs like your builder PR, to illustrate the improvements you have in mind.  I by no means think any of this is perfect. :)

dprotaso:
> I don't by any means think moving this freezes it. My hope is simply that as folks improve it in these ways that all of the repos can move forward.

I'm not worried about freezing but adopters going through churn

> I'd love to see more PoCs like your builder PR

Let me get started

mattmoor:
@dprotaso My naive hope is that we can get all of those into the mold I keep outlining:
```golang
func (r *Reconciler) reconcileBar(foo *vN.Foo) error {
  barName := names.Bar(foo)
  bar, getBarErr := r.barLister.Get(foo.Namespace, barName)
  if IsNotFound(getBarErr) {
    bar, err = r.createBar(foo)
    if err != nil {
      return err
    }
  } else if getBarErr != nil {
    return getBarErr
  }

  // Take action based on the state of foo/bar
  ...

  // Reflect status from bar into foo.
  ...
}
```

In my perfect world, I'd define a generic interface through which to drive this, but without generics this seems like it would be a glorious pain to make generic.

IMO everything that this rough pattern reaches out to should be encapsulated and tested separately.

This PR feels like an awkward place for this discussion.  Do you want to open an issue outlining what you have in mind, and we can discuss there?

dprotaso:
> This PR feels like an awkward place for this discussion. Do you want to open an issue outlining what you have in mind, and we can discuss there?

Sure

dprotaso:
FYI: I made some minor changes here: https://github.com/knative/serving/pull/1977 that affect items in this PR

PR #51, opened 2018-08-23
Update WaitForEndpointState to return response

srinivashegde86:
Return the spoof response when checking endpoint state. Some serving e2e tests use this.

### DIFF
--- a/test/request.go
+++ b/test/request.go
@@ -95,6 +95,5 @@ func WaitForEndpointState(kubeClient *KubeClient, logger *logging.BaseLogger, do
 		return nil, err
 	}
 
-	res, err := client.Poll(req, inState)
-	return res, err
+	return client.Poll(req, inState)
 }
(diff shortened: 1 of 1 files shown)


test/request.go
```diff
@@ -80,21 +80,21 @@ func EventuallyMatchesBody(expected string) spoof.ResponseChecker {
...
 
 	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s", domain), nil)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	_, err = client.Poll(req, inState)
-	return err
+	res, err := client.Poll(req, inState)
```
ImJasonH: `return client.Poll(req, inState)`

ImJasonH: This isn't a big thing, just a nit. Feel free to ignore.

srinivashegde86: thanks. updated

PR #55, opened 2018-08-28
Adding ViaIndex, ViaKey, ViaFieldIndex, ViaFieldKey to FieldErrors

n3wscott:
I have found situations where I want to do the following:

```
  for i, c := range spec.Collection {
    if err := doValidation(c); err != nil {
      return err.ViaIndex(i).ViaField("collection")
    }
  }
```

OR

```
  for k, v := range spec.Bag. {
    if err := doValidation(v); err != nil {
      return err.ViaKey(k).ViaField("bag")
    }
  }
```

This adds helpers to do that.

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -104,7 +104,7 @@ func (fe *FieldError) ViaIndex(index ...int) *FieldError {
 // For example, if a type recursively validates a parameter that has a collection:
 //  for k, v := range spec.Bag. {
 //    if err := doValidation(v); err != nil {
-//      return err.ViaField("bag").ViaKey(k)
+//      return err.ViaKey(k).ViaField("bag")
 //    }
 //  }
 func (fe *FieldError) ViaKey(key ...string) *FieldError {
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -73,15 +73,18 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 				strings.Join(append(prefix, oldPath), "."))
 		}
 	}
+
+	newPaths = flattenIndexes(newPaths)
+
 	fe.Paths = newPaths
 	return fe
 }
 
-// ViaIndex is used to attach an index to the last field provided.
+// ViaIndex is used to attach an index to the next via field provided.
 // For example, if a type recursively validates a parameter that has a collection:
 //  for i, c := range spec.Collection {
 //    if err := doValidation(c); err != nil {
-//      return err.ViaField("collection").ViaIndex(i)
+//      return err.ViaIndex(i).ViaField("collection")
 //    }
 //  }
 func (fe *FieldError) ViaIndex(index ...int) *FieldError {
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -73,10 +73,84 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
...
 }
 
+// ViaIndex is used to attach an index to the next via field provided.
+// For example, if a type recursively validates a parameter that has a collection:
+//  for i, c := range spec.Collection {
+//    if err := doValidation(c); err != nil {
+//      return err.ViaIndex(i).ViaField("collection")
+//    }
+//  }
+func (fe *FieldError) ViaIndex(index ...int) *FieldError {
```
mattmoor: I feel like this would be more natural if combined given the inverted ordering:
```go
  return err.ViaIndex("collection", i)
```

Is Index the most idiomatic term?  Subscript?

apis/field_error.go
```diff
@@ -73,10 +73,84 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
...
+}
+
+// ViaKey is used to attach a key to the next via field provided.
+// For example, if a type recursively validates a parameter that has a collection:
+//  for k, v := range spec.Bag. {
+//    if err := doValidation(v); err != nil {
+//      return err.ViaKey(k).ViaField("bag")
+//    }
+//  }
+func (fe *FieldError) ViaKey(key ...string) *FieldError {
```
mattmoor: Same as above, combine with `ViaField`?

Alternatively we could say `ViaFieldIndex(foo, bar)` as sugar for:
```
  return err.ViaIndex(bar).ViaField(foo)
```

apis/field_error.go
```diff
@@ -73,10 +73,84 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
...
+	return fe.ViaField(prefix...)
+}
+
+// flattenIndexes takes in a set of paths and looks for chances to flatten
+// objects that have index prefixes, examples:
+//   err([0]).ViaField(bar).ViaField(foo) -> foo.bar.[0] converts to foo.bar[0]
+//   err(bar).ViaIndex(0).ViaField(foo) -> foo.[0].bar converts to foo[0].bar
+//   err(bar).ViaField(foo).ViaIndex(0) -> [0].foo.bar converts to [0].foo.bar
+//   err(bar).ViaIndex(0).ViaIndex[1].ViaField(foo) -> foo.[0].[1].bar converts to foo[1][0].bar
+func flattenIndexes(paths []string) []string {
```
mattmoor: Hmm, let's see if we can rework to kill this?

apis/field_error.go
n3wscott: I intend to use this by using a helper function to validate a collection, so in my above example I have a simple case, in the real usage I have something like the following:
```
    if err := doCollectionValidation(spec.Collection); err != nil {
      return err.ViaField("collection")
    }
```

```
func doCollectionValidation(collection []Foos) {
   for i, c := range collection {
      if err := c.Validate(); err != nil {
        return err.ViaIndex(i)
      }
    }
}
```

`doCollectionValidation` is not aware of the field it is assigned to in the parent object, so making it a tuple does not help me in this way.

apis/field_error.go
n3wscott: Index is the path into an array https://en.wikipedia.org/wiki/Index_notation

Subscript seems to imply the full path in the nth-dimensional array: http://mathworld.wolfram.com/Subscript.html

I think most people would understand index, and not subscript

apis/field_error.go
n3wscott: I tried to not add this, but the issue is the index has to come in first, and then the field is attached to it so the flattening logic has to be somewhere...

apis/field_error.go
```diff
@@ -73,10 +73,90 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 				strings.Join(append(prefix, oldPath), "."))
 		}
 	}
+
+	newPaths = flattenIndices(newPaths)
```
mattmoor: I'm wondering if all of `flattenIndices` can be reduced to the following in the loop above?
```go
		if oldPath == CurrentField || strings.HasPrefix(oldPath, "[") {
			newPaths = append(newPaths, strings.Join(prefix, "."))
			newPaths = newPaths + oldPath
		} else {
			newPaths = append(newPaths,
				strings.Join(append(prefix, oldPath), "."))
		}
```

PR #57, opened 2018-08-30
Move interfaces into single file.

vaikas-google:
I found it a bit odd that Defaultable interface was in a single file, yet the other interfaces immutable / validatable were in field_error.go. Obviously no functional changes, just thought this is cleaner. Thoughts?


n3wscott:
makes sense to me!

/LGTM

PR #63, opened 2018-09-06
The test logger will now log the correct caller

dprotaso:
Previous, since we weren't skipping, the logged caller would be 'logging.go'

For example see conformance logs [here](https://gubernator.knative.dev/build/knative-prow/logs/ci-knative-serving-continuous/1037642038863990784)

```
info TestBlueGreenRoute logging/logging.go:141 Creating a Configuration
```

Will become
```
info TestBlueGreenRoute  conformance/blue_green_test.go:183 Creating a Configuration
```

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -34,8 +34,8 @@ import (
 // If configuration is empty, a fallback configuration is used.
 // If configuration cannot be used to instantiate a logger,
 // the same fallback configuration is used.
-func NewLogger(configJSON string, levelOverride string) (*zap.SugaredLogger, zap.AtomicLevel) {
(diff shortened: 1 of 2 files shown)


srinivashegde86:
can u add some context on what is this for?

dprotaso:
@srinivashegde86 I updated the PR & commit messages to provide better context

Assigning @mdemirhan (as well) since I cleaned up the implementation by allowing zap.Options to be accepted by the pkg/logging library

/assign @mdemirhan

srinivashegde86:
thanks for doing this. I was wondering how to do this, but did not get time to look into it.

This is great and will be helpful in debugging :)

PR #66, opened 2018-09-11
Adding common condition status.

n3wscott:
Standardize how we set conditions on a `Status`.

Fixes https://github.com/knative/serving/issues/1824


apis/condition_status_type.go
```diff
@@ -0,0 +1,118 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ConditionType string
+
+const (
+	// ConditionReady specifies that the resource is ready.
+	ConditionReady ConditionType = "Ready"
```
ImJasonH: "Ready" isn't a meaningful term in every Knative project. `knative/build`'s condition type is `BuildSuccessful`. Should we remove this value from `knative/pkg` and copy it in `knative/serving` and `knative/eventing`?

n3wscott: The majority of resources will have a "Ready", and those that do not don't have to use this. that controller can always create their own ConditionType strings. (or just send a string in.)

n3wscott: Possibly "Ready" for build could mean a controller is watching the resource...

ImJasonH: Maybe, but that's not what `Ready` means for other knative resources, so it would probably be confusing. I'm fine with having `Ready` defined in this shared package, so long as it's clear it has no special meaning to the package except that it's a commonly used condition type across knative resources.

n3wscott: What if you did use `Ready` and `Reason` was `BuildSuccessful`

ImJasonH: That would be a pretty big semantic shift from what our condition means today. Unlike a service or event stream, a build is run to completion and then is just "done" -- maybe it could be considered "ready" for some downstream thing to act on it, but I don't think we can make that statement about all builds.

evankanderson: https://github.com/knative/serving/blob/master/docs/spec/errors.md#conditions defines two 'top-level "happy state" conditions':

* For long-running resources, this condition type should be `Ready`.
* For objects which run to completion, the condition type should be `Succeeded`.

evankanderson:
/assign @jonjohnsonjr @evankanderson

I think Jon might have started looking at this.

apis/condition_status_type.go
ImJasonH: Great, if that's already part of our API spec, then maybe this package should just define both condition types, with comments pointing to the spec.

apis/condition_status_type.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+	}
+	new.LastTransitionTime = VolatileTime{metav1.NewTime(time.Now())}
+	conditions = append(conditions, *new)
+	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
+	cs.Conditions = conditions
+}
+
+// InitializeConditions updates the Ready Condition to unknown if not set.
```
evankanderson: This should iterate over a set of conditions which are present on the resource. I'm not sure how you define the set of conditions across all resources, though.

https://github.com/knative/serving/blob/master/pkg/apis/serving/v1alpha1/revision_types.go#L310

apis/condition_status_type.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, cond := range cs.Conditions {
+		if cond.Type == t {
+			return &cond
+		}
+	}
+	return nil
+}
+
+// SetCondition sets or updates the Condition on ConditionStatus for Condition.Type.
+func (cs *ConditionStatus) SetCondition(new *Condition) {
```
evankanderson: It would be nice to be able to follow the pattern in `knative/serving` where you can aggregate all the Conditions to determine the status of the `Ready` condition. See example here:

https://github.com/knative/serving/blob/master/pkg/apis/serving/v1alpha1/service_types.go#L289

Note that this requires knowing the types of Conditions which can be set on a resource -- this is currently encoded as a list in the source file.

apis/condition_status_type.go
n3wscott: I have added it

n3wscott:
> I think there are some substantial unanswered questions about how to manage multiple Condition 

Turns out this is tricky because IsReady wants to look at all conditions (i think) AND the Ready (or Succeeded) condition. The trick was in if Ready is not present, it is also not ready even if all present things are all true.

mattmoor:
@n3wscott I'd recommend looking at https://github.com/knative/serving/issues/1824 and how we're managing conditions in knative/serving.

I think extracting a `[]Condition` type alias in `knative/pkg` that generalizes the implementation from `knative/serving` would be very useful to reduce duplication and get semantics consistent with our `errors.md` across the repos.

n3wscott:
OK PTAL, I have updated this to match updates and learnings from https://github.com/knative/serving/pull/2015

apis/duck/v1alpha1/condition_set.go
```diff
@@ -0,0 +1,280 @@
...
+	// InitializeConditions updates all Conditions in the ConditionSet to Unknown
+	// if not set.
+	InitializeConditions()
+
+	// InitializeCondition updates a Condition to Unknown if not set.
+	InitializeCondition(t ConditionType)
+}
+
+// NewLivingConditionSet returns a ConditionSet to hold the conditions for the
+// ongoing resource. ConditionReady is used as the happy.
```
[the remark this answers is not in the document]
n3wscott: it is defined on `ConditionSet` -> ```It also holds the "happy condition"
// for that resource, which we define to be one of Ready or Succeeded depending
// on whether it is a Living or Batch process respectively.```

PR #67, opened 2018-09-13
Create a kmeta package for libraries acting on ObjectMeta.

mattmoor:
This bootstraps this package with a useful collection of label methods for managing sub-resources.  I'd initially written this as a way of managing versioned DaemonSet resources as part of WarmImage ([see here](https://github.com/mattmoor/warm-image/blob/62cad8045a6bb4bbc42bc1556beba47a73f081b9/pkg/reconciler/warmimage/resources/meta.go#L28)).

I am upstreaming this here because I want to take advantage of it in Build for managing Image cache subresources of [Cluster]BuildTemplate resources.

### DIFF
--- a/kmeta/labels_test.go
+++ b/kmeta/labels_test.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


kmeta/labels_test.go
```diff
@@ -0,0 +1,116 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestMakeVersionLabels(t *testing.T) {
+	tests := []struct {
+		name string
+		om   metav1.ObjectMeta
+		s    string
+	}{{
+		name: "simple translation",
```
ImJasonH: I don't understand the benefit of doing this twice. Is 1234 or ABCD semantic in some way? If not I think it's fine to just do it once to demonstrate that input -> output.

PR #70, opened 2018-09-13
Rename kmeta.NewOwnerReference to NewControllerRef

jonjohnsonjr:
This is more accurate, since we're creating a controller ref (and keeps
the naming consistent with metav1).

Also, adds a test!


jonjohnsonjr:
/hold cancel

I'll keep the frobber, I'd have to move some stuff around to use Resource in that test and avoid dependency cycles. Happy to revisit that later.

Issue #72, opened 2018-09-17
No way to express multiple validation errors


grantr:
## Expected Behavior

I have an array of objects. I want to validate that each of those objects has at least one field. Ideally, I'd like to return a message to the user saying that index 1 and 2 in the array are both missing fields.

## Actual Behavior

I can't do that today because there's no concept of a non-fatal validation error. It might be useful for validation methods to take a parameter like `testing.T` that exposes richer error collection behavior than is possible with a return value.

/cc @n3wscott

PR #77, opened 2018-09-18
Introduce FieldErrors

n3wscott:
Fixes: https://github.com/knative/pkg/issues/72

Testing an idea for field errors, PTAL and then I can add more tests.


n3wscott:
/cc @grantr 

Please take a look at this now.

apis/field_error.go
```diff
@@ -69,12 +80,29 @@ func (fe *FieldError) ViaIndex(index int) *FieldError {
 	if fe == nil {
 		return nil
 	}
-	return fe.ViaField(fmt.Sprintf("[%d]", index))
+	var newErrs []FieldError
+	for _, e := range fe.getNormalizedErrors() {
+		e.ViaField(fmt.Sprintf("[%d]", index))
+		newErrs = append(newErrs, e)
```
mattmoor: `newErrs = append(newErrs, e.ViaField(...))`

apis/field_error.go
```diff
@@ -50,11 +52,20 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
...
+		for _, oldPath := range e.Paths {
+			newPaths = append(newPaths, flatten(append(prefix, oldPath)))
+		}
+		e.Paths = newPaths
+
+		// Append the mutated error to the errors list.
+		newErrs = append(newErrs, e)
 	}
-	fe.Paths = newPaths
+	fe.clear()
```
mattmoor: I wonder if these assorted methods should stop mutating themselves and simply return a new `FieldError`.

n3wscott: I looked at that and I could make the assertment that a field error is immutable but that was not how the original class was so I made it mutable

apis/field_error.go
n3wscott: Yeah sounds good

apis/field_error.go
```diff
@@ -88,12 +116,79 @@ func (fe *FieldError) ViaKey(key string) *FieldError {
...
+	if fe == nil {
+		fe = &FieldError{}
+	}
+	newErrs = append(newErrs, fe.getNormalizedErrors()...)
+
+	for _, e := range errs {
+		newErrs = append(newErrs, e.getNormalizedErrors()...)
+	}
+
+	fe.clear()
```
mattmoor: I think I'm uncomfortable with the self-modifying FieldError pattern if we have paths where the receiver should become non-nil (like this).  I think we should switch to returning a modified FieldError and leaving the receiver intact.

apis/field_error_test.go
```diff
@@ -139,7 +139,7 @@ can not use @, do not try`,
 			if test.want != "" {
 				got := fe.Error()
 				if got != test.want {
-					t.Errorf("Error() = %v, wanted %v", got, test.want)
+					t.Errorf("&FieldError() = %v, wanted %v", got, test.want)
```
mattmoor: nit: undo this, we called `Error()` :)

n3wscott: Woops!!!

apis/field_error.go
n3wscott: I had a version of this that used .Also to collect the errors internally and it was really neat. I will update to that tomorrow

apis/field_error.go
```diff
@@ -50,12 +53,19 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 	if fe == nil {
 		return nil
 	}
-	var newPaths []string
-	for _, oldPath := range fe.Paths {
-		newPaths = append(newPaths, flatten(append(prefix, oldPath)))
+	newErr := &FieldError{}
+	for _, e := range fe.getNormalizedErrors() {
+		// Prepend the Prefix to existing errors.
+		var newPaths []string
```
mattmoor: nit: you could avoid resizing with `make([]string, 0, len(e.Paths))`

apis/field_error.go
```diff
@@ -122,12 +169,17 @@ func isIndex(part string) bool {
 	return strings.HasPrefix(part, "[") && strings.HasSuffix(part, "]")
 }
 
-// Error implements error
+// FieldError implements error
```
mattmoor: nit: `Error`

mattmoor: /golint

apis/field_error.go
```diff
@@ -85,17 +92,57 @@ func (fe *FieldError) ViaFieldIndex(field string, index int) *FieldError {
...
+	// and then collect the passed in errors
+	for _, e := range errs {
+		newErr.errors = append(newErr.errors, e.getNormalizedErrors()...)
+	}
+	return newErr
+}
+
+func (fe *FieldError) getNormalizedErrors() []FieldError {
+	if fe == nil {
+		return []FieldError(nil)
```
mattmoor: this cast feels superfluous, if intentional then add a comment explaining?

n3wscott: Ah. this is guarding `(*FieldError).Error()` case

Issue #76, opened 2018-09-18
Support Spec mutations in pkg/apis/duck


mattmoor:
Today, `pkg/apis/duck` is almost exclusively useful for **reading** duck-typed data.  However, it is also possible for us to leverage duck typing for **read-modify-write** operations on duck-typed data.

The basic idea is that we:
1. Define a partial schema for spec,
1. Parse the opaque resource we want to mutate as our Duck type,
1. DeepCopy the original resource,
1. Modify the copy via the duck type's spec,
1. Compute a JSON Patch of the delta,
1. Issue a PATCH against the API via the dynamic client.

A great example to help us drive this is the way we adjust Generations in our mutating webhook, which does all but the last of the above (the JSON Patch is returned from the mutating webhook instead).

/assign @mattmoor
cc @vaikas-google @n3wscott 

Kudos to @ImJasonH for the inspiration.

PR #78, opened 2018-09-18
Eliminate custom methods from GenericCRD

mattmoor:
This leverages a couple of duck types to eliminate the need for resource authors to expose custom methods to support generation bumping in the shared webhook.

The first duck type we use is the **first "spec" duck type** (see #76), and duck types the generation field, which enables us to read-modify-write the portion of the resource we want by leveraging a JSON patch for the update.  Here we use it in the context of the mutating webhook, which deals in JSON patches directly, so the dynamic client isn't needed here.  To verify a resource implements the Generation duck type resource authors should write:
```go
// This is needed because of Go's inability to do so inline.
var emptyGen duckv1alpha1.Generation
// Perform the actual validation that MyResource implements Generation.
var _ = duck.VerifyType(&MyResource, &emptyGen)
```

The second duck type this uses is a schema-less duck type (doesn't use our common infrastructure) for extracting the `"spec"` key from arbitrary JSON messages, which enables us to drop `GetSpecJSON()`.

Fixes: https://github.com/knative/pkg/issues/76


n3wscott:
This is awesome!!

Can you also add the Get/SetConditions methods too?

PR #80, opened 2018-09-19
Use legitimate methods to enable get and set conditions.

n3wscott:
It kinda sucks that I am asking everyone to implement `Get/SetConditions` on their status. Lets just try to get it for them.


n3wscott:
/cc @evankanderson 

Used reflection like you asked :D

/assign @mattmoor

apis/duck/v1alpha1/condition_set.go
```diff
@@ -126,14 +118,23 @@ var _ ConditionManager = (*conditionsImpl)(nil)
...
 	ConditionSet
-	accessor ConditionsAccessor
+	status interface{}
 }
 
-// Manage creates a ConditionManager from an object that implements
-// ConditionsAccessopr using the original ConditionSet as a reference.
-func (r ConditionSet) Manage(accessor ConditionsAccessor) ConditionManager {
+// Manage creates a ConditionManager from a status object using the original
+// ConditionSet as a reference. Status must be or point to a struct.
```
evankanderson: Does the struct need to contain any particular members of any particular types?

apis/duck/v1alpha1/condition_set.go
```diff
@@ -126,14 +118,23 @@ var _ ConditionManager = (*conditionsImpl)(nil)
...
+func (r ConditionSet) Manage(status interface{}) ConditionManager {
+	accessorValue := reflect.Indirect(reflect.ValueOf(status))
+
+	// If status is not a struct, don't even try to use it.
```
evankanderson: Should this return `error` instead of a thing that doesn't do things?

apis/duck/v1alpha1/condition_set.go
```diff
@@ -150,10 +151,11 @@ func (r conditionsImpl) IsHappy() bool {
 // GetCondition finds and returns the Condition that matches the ConditionType
 // previously set on Conditions.
 func (r conditionsImpl) GetCondition(t ConditionType) *Condition {
-	if r.accessor == nil {
+	if r.status == nil {
```
evankanderson: Again, I'd encourage not allowing a `conditionsImpl` to have a `nil` status.

apis/duck/v1alpha1/condition_set.go
```diff
@@ -278,3 +280,24 @@ func (r conditionsImpl) InitializeCondition(t ConditionType) {
 		})
 	}
 }
+
+// getConditions uses reflection to return the field on the status called "Conditions".
+func (r conditionsImpl) getConditions() Conditions {
+	conditionsField := reflect.Indirect(reflect.ValueOf(r.status)).FieldByName("Conditions")
```
evankanderson: Can you assert that the struct contains a "Conditions" field in the constructor (i.e. `Manage()`)?

apis/duck/v1alpha1/condition_set.go
```diff
@@ -126,14 +118,23 @@ var _ ConditionManager = (*conditionsImpl)(nil)
...
+	status interface{}
 }
 
-// Manage creates a ConditionManager from an object that implements
-// ConditionsAccessopr using the original ConditionSet as a reference.
-func (r ConditionSet) Manage(accessor ConditionsAccessor) ConditionManager {
+// Manage creates a ConditionManager from a status object using the original
+// ConditionSet as a reference. Status must be or point to a struct.
+func (r ConditionSet) Manage(status interface{}) ConditionManager {
+	accessorValue := reflect.Indirect(reflect.ValueOf(status))
```
evankanderson: Why not store `accessorValue` in the return value? Do we need to deref the pointer each time if it is one? If so, please include a test which would fail if we stored `accessorValue`.

apis/duck/v1alpha1/condition_set.go
```diff
@@ -184,7 +186,7 @@ func (r conditionsImpl) SetCondition(new Condition) {
 	conditions = append(conditions, new)
 	// Sorted for convince of the consumer, i.e.: kubectl.
```
evankanderson: No ":" after "id est" ("that is")

apis/duck/v1alpha1/condition_set.go
evankanderson: Also, check that `conditionsField.Interface()` can be cast to `Conditions` in the constructor?

apis/duck/v1alpha1/condition_set_impl_test.go
```diff
@@ -28,28 +28,24 @@ import (
...
 	Conditions Conditions
 }
 
-func (ts *TestConds) GetConditions() Conditions {
-	return ts.Conditions
-}
-
-func (ts *TestConds) SetConditions(conditions Conditions) {
-	ts.Conditions = conditions
+type TestFatFingers struct {
```
evankanderson: Add a comment about what this tests?

apis/duck/v1alpha1/condition_set_impl_test.go
```diff
@@ -123,11 +119,30 @@ func TestSetCondition(t *testing.T) {
 			Type:   ConditionReady,
 			Status: corev1.ConditionTrue,
 		},
+	}, {
+		name: "invalid status struct, should be nil",
```
evankanderson: If you switch the type checks to constructor tests, then you won't need to supply a specific `Condition` here to detect the problem.

apis/duck/v1alpha1/condition_set.go
n3wscott: if we do that then you can't do the `thingy.Manage(anotherthingy).DoSomething()`

apis/duck/v1alpha1/condition_set.go
n3wscott: I could add in the nil handling in the `conditionsImpl` to allow `nil.DoSomething()` like we did with `FieldError`. But it will still be a silent error unless you check for nil

apis/duck/v1alpha1/condition_set.go
n3wscott: > Can you assert that the struct contains a "Conditions" field in the constructor (i.e. Manage())?

I could use the `duck` type testing that was added I think. I was having issues trying that because it was not clear how to use a generic `{}interface` but I can try again.

>Also, check that conditionsField.Interface() can be cast to Conditions in the constructor?

I do the check on line 289.

apis/duck/v1alpha1/condition_set.go
n3wscott: interesting idea. I can fold that in. Thanks!

apis/duck/v1alpha1/condition_set.go
evankanderson: It seems like the problem here is that the use of reflection introduces the possibility of error (if the supplied type does not match), but we don't want to litter all of our code with error checks for that case (also reasonable).

The current code silently fails if you pass in the wrong `anotherthingy`, particularly in the case of `SetCondition` or the like.

This may be an appropriate use case for `panic` -- we expect that with something like generics (or the ability to declare an interface on field names rather than methods) we could use the type system to enforce this.

apis/duck/v1alpha1/condition_set.go
evankanderson: That's what I'm worried about. Someone writes some reasonable-looking code, but passes in the wrong object to SetConditions. It would be nice if something would flag that there was a problem.

apis/duck/v1alpha1/condition_set.go
evankanderson: >> Also, check that conditionsField.Interface() can be cast to Conditions in the constructor?
>
>I do the check on line 289.

If you do the check in the constructor before storing `r.status`, you do the check once per construction. If you do the check in `getCondition`/`setCondition`, you do the check once per method call.

apis/duck/v1alpha1/condition_set.go
```diff
@@ -126,14 +118,23 @@ var _ ConditionManager = (*conditionsImpl)(nil)
...
-	accessor ConditionsAccessor
+	status interface{}
 }
 
-// Manage creates a ConditionManager from an object that implements
-// ConditionsAccessopr using the original ConditionSet as a reference.
-func (r ConditionSet) Manage(accessor ConditionsAccessor) ConditionManager {
+// Manage creates a ConditionManager from a status object using the original
+// ConditionSet as a reference. Status must be or point to a struct.
+func (r ConditionSet) Manage(status interface{}) ConditionManager {
```
mattmoor: (Surfacing offline discussion)

I think I'd rather have the `conditionsImpl` still operate in terms of a `ConditionsAccessor`, but also supporting the synthesis of a `ConditionsAccessor` via reflection.

That means `Manage` could start taking an `interface{}`, and checking the type via:
```go
  ca, ok := foo.(ConditionsAccessor)
  if !ok {
    // Synthesize a ConditionsAccessor from foo into ca via reflection.
  }
```

mattmoor: I think that this separation makes cleanly unit testing the two pieces much easier, and creates a nice functional contract.

PR #83, opened 2018-09-20
Adding also always to return nil for nil errors

n3wscott:
Found an issue integrating with serving that `nil.also(nil)` should be `nil`, but it was returning an empty object. Fixed that.


n3wscott:
/assign @mattmoor 

I am sorry, I found a bug

PR #84, opened 2018-09-21
Create a common way for turning GVK into GVR.

mattmoor:
We've got a few of these evolving, and so I wanted to create a helper (in the hopes that someone will solve our pluralization properly).

[knative/eventing](https://github.com/knative/eventing/pull/437/files#diff-df3ef4686fadf4bf98714d6163f81654R277)
[knative/serving](https://github.com/knative/serving/blob/69811e7a9158732d29963ebb13193c33b51dd313/pkg/reconciler/v1alpha1/autoscaling/kpa_scaler.go#L119)

### DIFF
--- a/apis/kind2resource.go
+++ b/apis/kind2resource.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


rahulchheda:
I know this is an old PR, but just wanted to inform you guys, there are better and foolproof methods to do this,
for eg: In cases of Kubernetes native resource which is singular as `endpoints` , and plural as `endpoints`, this code might fail to perform it tasks.
Now with Dynamic Client, we have `DiscoveryCache` and `RESTMapper` which can help form relation between GVR and GVK of a particular resource.https://godoc.org/k8s.io/client-go/restmapper

PR #85, opened 2018-09-21
Define an interface for interacting with Listable objects.

mattmoor:
This will help us build informers from the dynamic API client.


n3wscott:
that test coverage though....

PR #86, opened 2018-09-22
Informers / Listers for Duck Types.

mattmoor:
This starts to sketch common libraries for instantiating informers/listers for a particular GroupVersionResource as one of our duck types.

You can instantiate a duck.InformerFactory like so:
```go
        dynaClient, err := dynamic.NewForConfig(cfg)
        if err != nil {
                logger.Fatalf("Error building dynamic clientset: %v", err)
        }

        dif := &duck.CachedInformerFactory{
                Delegate: &duck.TypedInformerFactory{
                        Client:       dynaClient,
                        Type:         &duckv1alpha1.Target{},
                        ResyncPeriod: 30 * time.Second,
                        StopChannel:  stopCh,
                },
        }
```

Then, as you come across new GroupVersionResources that you want to handle:
```go
        informer, lister, err := dif.Get(gvr)
        if err != nil {
               logger.Fatalf("Error starting shared index informer: %v", err)
        }
```

With the `duck.TypedInformerFactory` the objects will be returned as the provided `Type:`, so in this example, you could safely write:
```go
        elt, err := lister.ByNamespace(ns).Get(name)
        if err != nil { ... }

        target := elt.(*duckv1alpha1.Target)
        // Stuff involving target.
```

WIP because I still need tests for the `TypedInformerFactory`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -415,11 +415,13 @@
   version = "kubernetes-1.11.0"
 
 [[projects]]
-  digest = "1:7fa94edf47d2e91cffec1efe594a114f104199992f015086d673db5c776392cf"
+  digest = "1:292821230607e7e954ed569d52ede3006c2ecc5f44bb6af0e9eec79777316fee"
   name = "k8s.io/client-go"
   packages = [
     "discovery",
     "discovery/fake",
+    "dynamic",
+    "dynamic/fake",
     "informers",
     "informers/admissionregistration",
     "informers/admissionregistration/v1alpha1",
@@ -666,6 +668,8 @@
     "k8s.io/apimachinery/pkg/watch",
     "k8s.io/client-go/discovery",
     "k8s.io/client-go/discovery/fake",
+    "k8s.io/client-go/dynamic",
+    "k8s.io/client-go/dynamic/fake",
     "k8s.io/client-go/informers",
     "k8s.io/client-go/informers/core/v1",
     "k8s.io/client-go/kubernetes",
--- a/apis/duck/cached.go
+++ b/apis/duck/cached.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package duck
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+)
+
+// CachedInformerFactory implements InformerFactory by delegating to another
+// InformerFactory, but memoizing the results.
+type CachedInformerFactory struct {
+	Delegate InformerFactory
+
+	m     sync.Mutex
+	cache map[schema.GroupVersionResource]*result
+}
+
+// Check that CachedInformerFactory implements InformerFactory.
+var _ InformerFactory = (*CachedInformerFactory)(nil)
+
+// Get implements InformerFactory.
+func (cif *CachedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.SharedIndexInformer, cache.GenericLister, error) {
+	cif.m.Lock()
+	if cif.cache == nil {
+		cif.cache = make(map[schema.GroupVersionResource]*result)
+	}
+	elt, ok := cif.cache[gvr]
+	if !ok {
+		elt = &result{}
+		elt.init = func() {
+			elt.inf, elt.lister, elt.err = cif.Delegate.Get(gvr)
+		}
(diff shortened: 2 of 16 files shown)


mattmoor:
Using the dynamic fake, I am hitting: https://github.com/kubernetes/client-go/issues/465

It actually makes the informer hang when initially sync, but the linked PR fixes things.

mattmoor:
@dprotaso I was thinking about adding a patch we would apply as part of `./hack/update-deps.sh` based on the PR (Grant did something similar previously), but I quickly hit another hang when I add an initial resource to seed the fake client with.

I'll keep digging :)

mattmoor:
Ok, so the new issue doesn't appear to be addressed by the linked PR, but I haven't looked to see whether it is either fixed at HEAD or if there is another PR open to fix the issue.

Basically the test fixture sets up a catchall `testing.ObjectReaction(o)` [here](https://github.com/kubernetes/client-go/blob/03bfb9bdcfe5482795b999f39ca3ed9ad42ce5bb/dynamic/fake/simple.go#L45) to fake interactions.

For `List()` interactions, we come through [here](https://github.com/kubernetes/client-go/blob/03bfb9bdcfe5482795b999f39ca3ed9ad42ce5bb/testing/fixture.go#L84) and extract a `GroupVersionResource` and `GroupVersionKind` from the `action`, in our case these are:
 * `GroupVersionResource{Group:"pkg.knative.dev", Version:"v2", Resource:"resources"}`
 * `GroupVersionKind{Group:"fake-dynamic-client-group", Version:"v1", Kind:""}`

The implementation of `List()` on the `ObjectTracker` clearly assumes that the above are compatible because [here](https://github.com/kubernetes/client-go/blob/03bfb9bdcfe5482795b999f39ca3ed9ad42ce5bb/testing/fixture.go#L222) it calls `meta.SetList` which fails [here](https://github.com/kubernetes/apimachinery/blob/99c5fa21f87204c2875a7f7ba977895d79b35c44/pkg/api/meta/help.go#L200) on the assignability of these two types.

I am able to reproduce the error I am seeing with the following trivial test:
```go
package duck_test

import (
        "testing"

        "k8s.io/client-go/dynamic/fake"

        . "github.com/knative/pkg/testing"
)

func TestList(t *testing.T) {
        scheme := runtime.NewScheme()
        AddToScheme(scheme)

        client := fake.NewSimpleDynamicClient(scheme, &Resource{
                ObjectMeta: metav1.ObjectMeta{
                        Namespace: "foo",
                        Name:      "bar",
                },
        })

        _, err := client.Resource(SchemeGroupVersion.WithResource("resources")).List(metav1.ListOptions{})
        if err != nil {
                t.Errorf("List() = %v", err)
        }
}
```

Which prints:
```
--- FAIL: TestList (0.00s)
    typed_test.go:42: List() = item[0]: can't assign or convert testing.Resource into unstructured.Unstructured
```

I believe the only changes needed are a `register.go` to register `testing/resource.go` (provides the `AddToScheme` method, but this is boilerplate).

mattmoor:
Judging from the unit tests for the PR with the fix for the first issue, perhaps I should be passing only `unstructured.Unstructured` resources to the fake client (as it does [here](https://github.com/kubernetes/kubernetes/pull/66078/files#diff-27d280d7cd56f876fc6e278721c90d98R47)).

mattmoor:
Looks like the error I get when using `unstructured.Unstructured` was due to my laziness copying over the patch.  I'd thought [this](https://github.com/kubernetes/kubernetes/pull/68552/files#diff-db7b2ce019d631cdfba8d5e4e221ddd9R306) bit was cosmetic, but it turns out to be semantically important because `unstructure.Unstructured` doesn't implement `metav1.Object`, however, `*unstructured.Unstructured` does 🎉 

I'll see how far I can get writing tests with `unstructured.Unstructured`, but IMO if `NewSimpleDynamicClient` takes `[]runtime.Object` it should support it properly.

mattmoor:
I'm going to keep experimenting with testing for the informer path, but I'm not super optimistic the fakes are that solid, so removing the WIP.

mattmoor:
I added a `duck.InformerFactory` implementation to help integrate with controller enqueuing, here's the updated commit message example:
```go
            dynaClient, err := dynamic.NewForConfig(cfg)
            if err != nil {
                    logger.Fatalf("Error building dynamic clientset: %v", err)
            }
    
            // Cache as the outermost layer so we only register the EventHandler once.
            dif := &duck.CachedInformerFactory{
                    Delegate: &duck.EnqueueInformerFactory{
                            Delegate: &duck.TypedInformerFactory{
                                    Client:       dynaClient,
                                    Type:         &duckv1alpha1.Target{},
                                    ResyncPeriod: 30 * time.Second,
                                    StopChannel:  stopCh,
                            },
                            EventHandler: cache.ResourceEventHandlerFuncs{
                                    AddFunc: func(obj interface{}) {
                                            // Enqueue, obj is: *duckerv1alpha1.Target
                                    },
                                    UpdateFunc: func(old, new interface{}) {
                                            // Enqueue, old and new are: *duckerv1alpha1.Target
                                    },
                            },
                    },
            }
```

dprotaso:
> Then as the Reconciler reconciles resources and discovers new types implementing Targettable it can:

Does this mean there's a subtle bug where restarting a controller could imply that 'certain' gvk informer's aren't restart potentially for the entire resync period?

imikushin:
@mattmoor Please take a look if I'm using this intended ☝️

n3wscott:
/lgtm 

thanks for all the tests. this will be neat to use for general informers of these duck types :D

PR #87, opened 2018-09-24
Create a helper for synthesizing patches.

mattmoor:
This creates a helper for synthesizing JSON patches from before/after objects.

This aliases `[]jsonpatch.JsonPatchOperation` as `JSONPatch` to provide `MarshalJSON` as a convenience, since this is what needs to be passed to `Patch()` on K8s clients.

### DIFF
--- a/apis/duck/patch_test.go
+++ b/apis/duck/patch_test.go
@@ -1,0 +1,191 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


apis/duck/patch_test.go
```diff
@@ -0,0 +1,169 @@
+/*
+Copyright 2017 The Knative Authors
```
n3wscott: it is 2018

apis/duck/patch_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestCreatePatch(t *testing.T) {
```
n3wscott: Please add some complicated tests, like with an array?

n3wscott: Also something that is not a number

apis/duck/patch_test.go
mattmoor: Done

apis/duck/patch_test.go
mattmoor: I have a bool!

mattmoor: I added a fancy add-in-the-middle array test.

n3wscott:
/lgtm

Thanks for that array test :D

PR #90, opened 2018-09-24
Sort keys and paths for FieldError.

n3wscott:
I need to make the FieldError object be deterministic. Sorting on `Message` and `Paths`.

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -18,6 +18,7 @@ package apis
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -137,6 +139,7 @@ func (fe *FieldError) getNormalizedErrors() []FieldError {
 	for _, e := range fe.errors {
 		errors = append(errors, e.getNormalizedErrors()...)
 	}
+	sort.Slice(errors, func(i, j int) bool { return errors[i].Message < errors[j].Message })
```
mattmoor: What about same message different fields?

n3wscott: That will have to be something added to flatten. I have been thinking about that...

PR #91, opened 2018-09-24
Merge FieldErrors into a single error when all but the path matches.

n3wscott:
Implemented merge for squashing field errors together that are the same except for path.

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -20,6 +20,9 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 // CurrentField is a constant to supply as a fieldPath for when there is
@@ -139,8 +142,40 @@ func (fe *FieldError) getNormalizedErrors() []FieldError {
 	for _, e := range fe.errors {
 		errors = append(errors, e.getNormalizedErrors()...)
 	}
-	sort.Slice(errors, func(i, j int) bool { return errors[i].Message < errors[j].Message })
+
+	errors = flattenErrors(errors)
+
 	return errors
+}
+
+// flattenErrors will combine errors that differ only in their paths.
+// This assumes that errors has been normalized.
+func flattenErrors(errors []FieldError) []FieldError {
+	if len(errors) <= 1 {
+		return errors
+	}
+
+	ignoreArguments := cmpopts.IgnoreFields(FieldError{}, "Paths")
+	ignoreUnexported := cmpopts.IgnoreUnexported(FieldError{})
+
+	// Sort first.
+	sort.Slice(errors, func(i, j int) bool { return errors[i].Message < errors[j].Message })
+
+	newErrors := make([]FieldError, 0, len(errors))
+	i := 0
+	curr := errors[i]
+	newErrors = append(newErrors, curr)
+	for j := 1; j < len(errors); j++ {
+		next := errors[j]
+		if diff := cmp.Diff(curr, next, ignoreArguments, ignoreUnexported); diff == "" {
+			// they match, merge the paths.
+			curr.Paths = append(curr.Paths, next.Paths...)
+		} else {
+			curr = next
+			newErrors = append(newErrors, curr)
+		}
+	}
+	return newErrors
 }
 
 func asIndex(index int) string {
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -139,10 +144,62 @@ func (fe *FieldError) getNormalizedErrors() []FieldError {
...
+			}
+			// only add new ones.
+			curr.Paths = append(curr.Paths, nextPaths...)
+		} else {
+			// moving the pointer, save the current object.
+			newErrors = append(newErrors, curr)
+			curr = next
+		}
+	}
+	if diff := cmp.Diff(curr, next, ignoreArguments, ignoreUnexported); diff != "" {
```
vdemeester: @n3wscott why is this one required ?

n3wscott:
@vdemeester thanks for the review, I have updated the merge method and added a perf test.

vdemeester:
/lgtm 

@n3wscott by curiosity, why not doing a `func Benchmark*` for the perf test ? :angel:

n3wscott:
The benchmark was interesting:

```
⦿ go test ./apis/... -bench=.
goos: darwin
goarch: amd64
pkg: github.com/knative/pkg/apis
BenchmarkFieldError/no_merge/1-4         	 1000000	      1935 ns/op
BenchmarkFieldError/no_merge/2-4         	  300000	      4020 ns/op
BenchmarkFieldError/no_merge/4-4         	  200000	      9057 ns/op
BenchmarkFieldError/no_merge/8-4         	   50000	     26859 ns/op
BenchmarkFieldError/no_merge/16-4        	   20000	     77719 ns/op
BenchmarkFieldError/no_merge/32-4        	   10000	    227855 ns/op
BenchmarkFieldError/no_merge/64-4        	    2000	    780173 ns/op
BenchmarkFieldError/no_merge/128-4       	     500	   2610421 ns/op
BenchmarkFieldError/no_merge/256-4       	     100	  10452324 ns/op
BenchmarkFieldError/no_merge/512-4       	      50	  30228864 ns/op
BenchmarkFieldError/no_merge/1024-4      	      10	 124929299 ns/op
BenchmarkFieldError/merge/1-4            	  500000	      2084 ns/op
BenchmarkFieldError/merge/2-4            	  200000	      6752 ns/op
BenchmarkFieldError/merge/4-4            	  100000	     18472 ns/op
BenchmarkFieldError/merge/8-4            	   30000	     51968 ns/op
BenchmarkFieldError/merge/16-4           	   10000	    148259 ns/op
BenchmarkFieldError/merge/32-4           	    3000	    473297 ns/op
BenchmarkFieldError/merge/64-4           	    1000	   1688618 ns/op
BenchmarkFieldError/merge/128-4          	     200	   6231113 ns/op
BenchmarkFieldError/merge/256-4          	     100	  21168326 ns/op
BenchmarkFieldError/merge/512-4          	      20	  77858865 ns/op
BenchmarkFieldError/merge/1024-4         	       3	 380925478 ns/op
```

I read this as merge decreases performance by 3 in the low end and ~1.25 in the high end.

I wonder if there are gains if the keys are all similar...

apis/field_error_test.go
```diff
@@ -503,6 +616,113 @@ func TestFlatten(t *testing.T) {
...
 }
 
+func TestFieldErrorPerformance(t *testing.T) {
+	limit := int64(3) // limit performance decrease to at most 3x slower with merge
+	count := 100
+
+	var withoutMerge time.Duration
+	var withMerge time.Duration
+
+	{
```
mattmoor: This is weird (though I used to do this in C++ commenting scopes of code), use `t.Run` to describe the scopes?

apis/field_error.go
```diff
@@ -139,10 +153,54 @@ func (fe *FieldError) getNormalizedErrors() []FieldError {
...
 
+// merge will combine errors that differ only in their paths.
+// This assumes that errors has been normalized.
+func merge(errors []FieldError) []FieldError {
+	if len(errors) <= 1 {
+		return errors
+	}
+
+	// Sort first.
+	sort.Slice(errors, func(i, j int) bool { return errors[i].Message < errors[j].Message })
```
mattmoor: I suspect this would be more clearly implemented as a `map[string /* Message */]FieldError` to build up the deduplicated `FieldErrors` and then sorting the result.

Would it be useful to *store* things as a `map` or "set" internally?

n3wscott: Let me try.

n3wscott:
/hold

going to go try to replace the internal error array as a map

PR #92, opened 2018-09-25
Access Conditions on KResourceStatus

imikushin:
Needed to conveniently access Conditions from controllers working with 
KResource duck type.

### DIFF
--- a/apis/duck/v1alpha1/conditions_types.go
+++ b/apis/duck/v1alpha1/conditions_types.go
@@ -119,6 +119,32 @@ type KResourceStatus struct {
 	Conditions Conditions `json:"conditions,omitempty"`
 }
 
+var condSet = NewBatchConditionSet()
+
+// GetCondition returns the Condition matching the given type.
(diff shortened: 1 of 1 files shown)


apis/duck/v1alpha1/conditions_types.go
```diff
@@ -119,6 +119,32 @@ type KResourceStatus struct {
 	Conditions Conditions `json:"conditions,omitempty"`
 }
 
+var condSet = NewBatchConditionSet()
```
n3wscott: this will not do what you want for all things that might use this. 
I have a competing PR to do this same thing:  https://github.com/knative/pkg/pull/80

imikushin: I copied over from Build :)

It'd probably make sense to remove the two methods depending on `condSet` and maybe use `condSet.Manage(krs).GetCondition(t)` directly - I just need `GetCondition(t)`.

imikushin: Done!

imikushin:
@mattmoor PTAL

PR #93, opened 2018-09-25
Add an example / test that we can Duck type Core K8s resources.

mattmoor:
This defines a simple test duck type called PodSpecable, which has the same shape for embedding a corev1.PodTemplateSpec as numerous core Kubernetes resource: ReplicaSet, Deployment, StatefulSet, DaemonSet, and Job.

### DIFF
--- a/apis/duck/podspec_test.go
+++ b/apis/duck/podspec_test.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package duck_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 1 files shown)


apis/duck/podspec_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+type WithPodSpec struct {
+	Template PodSpecable `json:"template,omitempty"`
+}
+
+// Check that our canonical type implements PodSpecable.
+var _ = duck.VerifyType(&WithPod{}, &PodSpecable{})
+
+// Check that several Kubernetes built-in type implement PodSpecable.
+var _ = duck.VerifyType(&appsv1.ReplicaSet{}, &PodSpecable{})
```
n3wscott: could this all be 

```
var _ = duck.VerifyType(&PodSpecable{}, &appsv1.ReplicaSet{}, &appsv1.Deployment{},  &appsv1.StatefulSet{} &appsv1.DaemonSet{} &batchv1.Job{})
```

mattmoor: I think I'd rather make the inverse easier:
```
var _ = duck.VerifyType(&MyRsc, &iface1{}, &iface2{}, &iface3{})
```

This pattern I expect to be more exceptional, but we already have a half dozen where the above would help.

mattmoor: We could even do it in a backwards compatible way.  Wanna send a PR? :D

n3wscott: > `var _ = duck.VerifyType(&MyRsc, &iface1{}, &iface2{}, &iface3{})`

That is what I tried to write super fast...

Issue #94, opened 2018-09-25
Generalized BuildTracker


mattmoor:
There are a class of cases where object Foo is watching object Bar and we want to trigger level-based reconciliation of Foo when Bar changes.  There are two cases of this in `knative/serving` today:
1. `Revision` -> `Build`, which is done via the specialized build tracker (see https://github.com/knative/serving/issues/1267)
1. `Route` -> `Configuration`, which is done by relying on the labels we introduced for enforcing 1:N and hope to retain the capacity to relax (see https://github.com/knative/serving/issues/1495).

The former is complicated by our desire to loosen the coupling between our repos.

---

We effectively want a generalized `Tracker` that during reconciliation enables us to create edges from `corev1.ObjectReference` to a `key` that should be enqueued for reconciliation.

For where we are with `Revision` -> `Build` (with `spec.buildName`) we can use something like:
```go
tracker.Track(corev1.ObjectReference{
  APIVersion: "build.knative.dev/v1alpha1",
  Kind: "Build",
  Namespace: rev.Namespace,
  Name: rev.Spec.BuildName,
}, rev)
```

For where we want to go (with `spec.buildRef`) we can use something like:
```go
buildRef := rev.Spec.BuildRef.DeepCopy()
buildRef.Namespace = rev.Namespace 
tracker.Track(buildRef, rev)
```

For `Route` -> `Configuration` and `Revision` we want something similar to what's outlined for `Build` where we are today, but needs some facility for aging out records (effectively requiring resyncs to refresh leases).

---

I'm going to experiment with usable APIs for this, and circle back here to summarize anything I come up with that I like.

cc @tcnghia @imikushin

mattmoor:
So here's what I'm thinking:
```go
package tracker

// Interface defines the interface through which an object can register
// that it is tracking another object by reference.
type Interface interface {
        // Track tells us that "obj" is tracking changes to the referenced object.
        Track(ref corev1.ObjectReference, obj interface{}) error

        // OnChanged is a callback to register with the InformerFactory
        // so that we are notified for appropriate object changes.
        OnChanged(obj interface{})
}

// New returns an implementation of Interface that lets a Reconciler
// register a particular resource as watching an ObjectReference for
// a particular lease duration.  This watch must be refreshed
// periodically (e.g. by a controller resync) or it will expire.
//
// When OnChanged is called by the informer for a particular
// GroupVersionKind, the provided callback is called with the "key"
// of each object actively watching the changed object.
[...]

PR #95, opened 2018-09-26
This change implements a more generalized form of the BuildTracker.

mattmoor:
To use this with a typical informer, e.g. the way `Route` would monitor `Configurations` and `Revisions`, you'd do something like:
```go
	c := &Reconciler{
		Base:                 reconciler.NewBase(opt, controllerAgentName),
		...
	}
	impl := controller.NewImpl(c, c.Logger, "Routes")

        t := tracker.New(impl.EnqueueKey, 30*time.Minute)
        configInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
                AddFunc:    t.OnChanged,
                UpdateFunc: controller.PassNew(t.OnChanged),
        })
        revisionInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
                AddFunc:    t.OnChanged,
                UpdateFunc: controller.PassNew(t.OnChanged),
        })

        // Now use c.tracker.Track(revisionRef, route) or c.tracker.Track(configRef, route)
        // each Reconcile(route) to refresh the cross-object leases.
        c.tracker = t
```

To use this with a `duck.InformerFactory`, e.g. the way `Revision` will monitor `BuildRef`s you'd do something like:
```go
	c := &Reconciler{
		Base:             reconciler.NewBase(opt, controllerAgentName),
		...
	}
	impl := controller.NewImpl(c, c.Logger, "Revisions")

        t := tracker.New(impl.EnqueueKey, 30*time.Minute)
        cif := &duck.CachedInformerFactory{
                Delegate: &duck.EnqueueInformerFactory{
                        Delegate: buildInformerFactory,
                        EventHandler: cache.ResourceEventHandlerFuncs{
                                AddFunc:    t.OnChanged,
                                UpdateFunc: controller.PassNew(t.OnChanged),
                        },
                },
        }

        // Now use: `c.buildInformerFactory.Get()` to access ObjectReferences.
        c.buildInformerFactory = buildInformerFactory

        // Now use: `c.tracker.Track(rev.Spec.BuildRef, rev)` to queue `rev`
        // each time `rev.Spec.BuildRef` changes (expires 30 minutes after BuildRef
        // points to something else).
        c.tracker = t
```

Fixes: https://github.com/knative/pkg/issues/94

### DIFF
--- a/tracker/doc.go
+++ b/tracker/doc.go
@@ -1,0 +1,21 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package tracker defines a utility to enable Reconcilers to trigger
+// reconciliations when objects that are cross-referenced change, so
+// that the level-based reconciliation can react to the change.  The
+// prototypical cross-reference in Kubernetes is corev1.ObjectReference.
+package tracker
--- a/tracker/enqueue.go
+++ b/tracker/enqueue.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


mattmoor:
cc @tcnghia as well.

I think I am going to experiment with using this for the Route controller today

mattmoor:
@dprotaso We want to retain the ability to drop those labels, and allow several Route resources to reference a Configuration / Revision.  I want to move the labeling into a separate Reconciler, but have been reluctant to because of the semantic dependency.  We also aren't labeling Revisions, which makes informing on them harder without a lookup.

Definitely open to suggestions on API.

dprotaso:
Has anyone explored routes being an owner reference on configurations & revisions?

mattmoor:
@dprotaso Not in this context, and I'm also not sure it's right.

We juggled owner references to share a fluentd configmap per-namespace in the Revision controller at one point and it was pretty messy (maybe Patch would have helped).

However, even if we could, I'm not sure you can set OwnerReferences without enabling GC, and we don't want Route deletion to trigger Configuration deletion.

mattmoor:
FYI this is what this would look like integrating into `knative/serving` today: https://github.com/knative/serving/pull/2089

Issue #97, opened 2018-09-26
Move the use of `VerifyType` in consumer packages to unit tests


dprotaso:
We're using `duck.VerifyType` to enforce that serialization formats match. These statements are being placed in packages  as follows:

https://github.com/knative/pkg/blob/2b6bdaba3b747205139b1202168523306cdfd333/apis/duck/v1alpha1/targetable_types.go#L37

This technique means verify is invoked at runtime during package initialization.

As we add more potential types that need to be verified this package initialization technique has an affect on a program's startup.

To avoid this startup time cost, we can simply move these assertions to unit tests that invoke `VerifyType`

```go
func TestTargetImplementsTargetable(t *testing.T) {
  duck.VerifyType(&Target{}, &Targetable{}) 
}
```

dprotaso:
Uses in other repos

Serving: https://github.com/knative/serving/search?q=%22VerifyType%22&unscoped_q=%22VerifyType%22
Eventing: https://github.com/knative/eventing/search?q=%22VerifyType%22&unscoped_q=%22VerifyType%22
Build: https://github.com/knative/build/search?q=%22VerifyType%22&unscoped_q=%22VerifyType%22

mattmoor:
This is done.

PR #98, opened 2018-09-26
Move the use of `VerifyType` in tests

vdemeester:
Those calls to `duck.VerifyType` are done at runtime and thus could be
costly at program startup. Putting them under tests ensure we still
assert those types but during unit testing.

This is part of #97 (not clusing it as there will be PRs on `serving`, `eventing` and `build` to follow :wink: 

cc @dprotaso 

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

### DIFF
--- a/apis/duck/podspec_test.go
+++ b/apis/duck/podspec_test.go
@@ -42,16 +42,6 @@ type WithPodSpec struct {
 	Template PodSpecable `json:"template,omitempty"`
 }
 
-// Check that our canonical type implements PodSpecable.
-var _ = duck.VerifyType(&WithPod{}, &PodSpecable{})
-
-// Check that several Kubernetes built-in type implement PodSpecable.
-var _ = duck.VerifyType(&appsv1.ReplicaSet{}, &PodSpecable{})
-var _ = duck.VerifyType(&appsv1.Deployment{}, &PodSpecable{})
-var _ = duck.VerifyType(&appsv1.StatefulSet{}, &PodSpecable{})
-var _ = duck.VerifyType(&appsv1.DaemonSet{}, &PodSpecable{})
-var _ = duck.VerifyType(&batchv1.Job{}, &PodSpecable{})
-
 var _ duck.Populatable = (*WithPod)(nil)
 var _ duck.Implementable = (*PodSpecable)(nil)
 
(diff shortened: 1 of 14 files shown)


apis/duck/v1alpha1/channelable_types_test.go
```diff
@@ -0,0 +1,27 @@
...
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/knative/pkg/apis/duck"
+)
+
+func TestChannelImplementsChannelable(t *testing.T) {
+	duck.VerifyType(&Channel{}, &Channelable{})
```
ImJasonH: Maybe `VerifyType` should return an error that we can check and call `t.Error` on, rather than panicking as it does today.

apis/duck/v1alpha1/channelable_types_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/knative/pkg/apis/duck"
+)
+
+func TestChannelImplementsChannelable(t *testing.T) {
```
ImJasonH: Any reason not to move all of these new tests into a single file within this package, that just checks that all the given types satisfy the expected duck types? It could be a pretty simple table-driven test.

apis/duck/v1alpha1/channelable_types_test.go
vdemeester: @ImJasonH That's a good point :+1:

apis/duck/v1alpha1/channelable_types_test.go
vdemeester: @ImJasonH didn't thought of that initially, work for me :wink:

apis/duck/v1alpha1/channelable_types_test.go
vdemeester: That's gonna make the change way bigger though… Should I do that in the same PR or create a follow-up / different PR for that change ?

n3wscott:
I think the point was we wanted them run at runtime to catch type mismatches? You could deploy code that never ran a unit test

What if they are called at runtime in the init block in a go func? So they would spin a thread up and not block

vdemeester:
@n3wscott hum :thinking: right, but anything released should have ran the unit tests at some point (and unit tests are ran on any PR, …)… so I would think it is safe to not call those at runtime.

> What if they are called at runtime in the init block in a go func? So they would spin a thread up and not block

That would work if we think it should be called at runtime yeah.

dprotaso:
> I think the point was we wanted them run at runtime to catch type mismatches? You could deploy code that never ran a unit test
>
> What if they are called at runtime in the init block in a go func? So they would spin a thread up and not block

I'm not really concerned about the performance enough to warrant spinning up go-routines. Having these as tests should provide confidence that the concrete type satisfies the duct type. So you shouldn't need runtime checks.

For those deploying code without running unit tests whether duck typing is satisfied will be the least of their concerns.

apis/duck/v1alpha1/channelable_types_test.go
mattmoor: Let's do it all at once

vdemeester: ok fair enough, I'll work on that :wink:

mattmoor:
So vendoring could lead to skew between the version tested and the version in use, so we'll need to be careful about how we version these.

PR #99, opened 2018-09-27
Simplify FieldError merging.

n3wscott:
I realized that there was no need to pre-optimize the error object because in all likely-hood this error is only going to contain a few objects, and the complex merging is simpler (and less wasteful) if done at the very end.

I removed the usage of normalizing over and over again, and this is done only once when `.Error()` is called. `merge()` also only happens when `.Error()` is called. The upside is that we are not constantly recreating FieldError objects when we don't need to. Sorting has been moved inside the merge method.

The downside is each time `.Error()` is called now, the complex part of this code needs to run. We could choose to expose a `.Simplify()` method if we find a common case where an error has `.Error()` called on it many times.

One other downside is that the end result of a FieldError object is as messy as it's validation callstack. This should be ok though, we don't intend to inspect them directly.

Please take a look, sorry my first attempt was very complicated. 🤷‍♂️ I am learning 😸


n3wscott:
/assign @mattmoor 

/cc vdemeester

Superseeds https://github.com/knative/pkg/pull/91

apis/field_error.go
```diff
@@ -106,19 +105,31 @@ func (fe *FieldError) ViaFieldKey(field string, key string) *FieldError {
 func (fe *FieldError) Also(errs ...*FieldError) *FieldError {
 	newErr := &FieldError{}
 	// collect the current objects errors, if it has any
-	if fe != nil {
-		newErr.errors = fe.getNormalizedErrors()
+	if fe.isEmpty() == false {
```
vdemeester: nit: why not `!fe.isEmpty()` ?

vdemeester: (same below)

apis/field_error.go
```diff
@@ -55,17 +55,16 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
...
-		e.Paths = newPaths
-
-		// Append the mutated error to the errors list.
-		newErr = newErr.Also(&e)
+	// Prepend the Prefix to existing errors.
+	newPaths := make([]string, 0, len(fe.Paths))
+	for _, oldPath := range fe.Paths {
+		newPaths = append(newPaths, flatten(append(prefix, oldPath)))
+	}
+	newErr.Details = fe.Details
```
vdemeester: nit: could be extracted in a function : 

```go
func From(fe *FieldError, ops ...func(*FieldError)) *FieldError {
    newErr := &FieldError{
        Details: fe.Details,
        Message: fe.Message,
        Paths: fe.Paths,
    }
    for _, op := range ops {
        op(fe)
    }
    return newErr
}
// […]
newErr := From(fe, func(e *FieldError) { e.Paths = newPaths })
// this func could even be extracted in a `withPaths` func
```

apis/field_error.go
n3wscott: I find the `== false` easier to read for these small cases

vdemeester: @n3wscott fair enough :stuck_out_tongue_closed_eyes:

apis/field_error.go
n3wscott: I am not sure I follow the example above. I have updated the code to use deepcopy on .Also, and simplified the construction of the newFe inside of .ViaField. Maybe this is in-line with what you are suggesting

apis/field_error.go
n3wscott: I updated to BANG

apis/field_error.go
vdemeester: @n3wscott yeah even better, forgot about those generation and deepcopy stuff :wink: 
:+1:

n3wscott:
@mattmoor fresh from 🍹 🏖 🌞 ready to 👀 and then 😻 this PR?

apis/field_error.go
```diff
@@ -173,23 +207,80 @@ func flatten(path []string) string {
...
 }
 
-func isIndex(part string) bool {
-	return strings.HasPrefix(part, "[") && strings.HasSuffix(part, "]")
+// mergePaths takes in two string slices and returns the combination of them
+// without any duplicate entries.
+func mergePaths(a, b []string) []string {
+	newPaths := make([]string, 0, len(a)+len(b))
+	newPaths = append(newPaths, a...)
+	for p := 0; p < len(b); p++ {
```
mattmoor: why not `for _, bi := range b {`?

n3wscott: Hm, mistake. fixed.

PR #104, opened 2018-09-28
The channelable ducktype added an extra layer of abstraction

n3wscott:
The channelable ducktype added an extra layer of abstraction and embedded the type into the spec, removing that.

So the object goes from:

```
foo.spec.channelable.subscribers[]
```

to 

```
foo.spec.subscribers[]
```


mattmoor:
I'll defer to @vaikas-google.  @grantr In my mental model for how this would dovetail with custom subresources it looks something like:
```
subresources:
    status: {}
    # Writeable
    channelable:
      specPath: .spec.channelable  # Has type Channelable
    # Readable
    targetable:
      statusPath: .status.targetable  # Has type Targetable
```

`ScaleTargetRef`'s replica counts would be degenerate cases of `int`s as the embedded type.

cc @pwittrock since we were talking about this fantasy yesterday :)

n3wscott:
Could it not be 

```
subresources:
    status: {}
    # Writeable
    channelable:
      specPath: .spec.subscribers
```

I suppose there is the hope that there might be more than the subscribers array in the object one day? I ran into this today and it seemed like a mistake.

vaikas-google:
By having a channelable as a real field, it groups them under the Channelable field. Yes, today there's only one field there, but if there are others in the future, it seems better than having at the same level all the fields to avoid name collisions.

PR #118, opened 2018-10-08
Support 'hostname' and 'ip' when communicating with ingressgateway.

markusthoemmes:
When E2E tests talk to the revisions under test, they use the `knative-ingressgateway` to talk to them. Given there's no resolvable domain, the IP of said ingressgateway is taken and the hostname of the revision to talk to is spoofed in via the `Host` header. Under some circumstances the ingressgateway loadbalancer does not get an IP but a hostname though (namely when using Docker for Mac's Kubernetes support).

Falling back to the hostname value if no IP is set supports this use-case.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -101,16 +101,21 @@ func New(kubeClientset *kubernetes.Clientset, logger *logging.BaseLogger, domain
 		if err != nil {
 			return nil, err
 		}
+		ingresses := ingress.Status.LoadBalancer.Ingress
 
-		if len(ingress.Status.LoadBalancer.Ingress) != 1 {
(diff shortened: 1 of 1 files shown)


bbrowning:
How does this change relate to usage of the resolvable flag? It looks like the case this covers is when the ingress does not get assigned a specific IP but instead assigned a resolvable hostname.

Is there an example of where this logic would be useful but setting the resolvable flag would not? Or, does this change render the resolvable flag no longer needed?

markusthoemmes:
@bbrowning I might be mistaken here but I think the resolvable hostname needs to be the hostname of the service itself whereas I'm technically spoofing a host to another host. I could make my test-setup bind against "localhost" I suppose, I haven't actually tried that to be honest.

The `curl` needed for my service is something like: `curl -H "Host: something.example.com" localhost`

PR #119, opened 2018-10-08
Adding Zipkin tracing support for SpoofingClient

dushyanthsc:
Change adds Zipkin Tracing header to each HTTP request made through
SpoofingClient to knit together Zipkin traces on the backend to requests
made through the client. To use the support tests needs to call into
SetupZipkinTracing(..) in their Setup method to setup Zipkin tracing.
This method setups Port-Forwarding from the machine(Port 9411) where tests are run
to Zipkin Pod on the Knative setup.  TearDown/Cleanup method needs to call KillZipkinPortFowardingProcess(..) to cleanup Zipkin tracing setup.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -19,6 +19,9 @@ limitations under the License.
 package spoof
 
 import (
+	"bytes"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -26,9 +29,14 @@ import (
 	"time"
 
 	"github.com/knative/pkg/test/logging"
+	zipkin "github.com/knative/pkg/test/zipkin"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/plugin/ochttp/propagation/b3"
+	"go.opencensus.io/trace"
 )
 
 const (
@@ -82,7 +90,7 @@ type SpoofingClient struct {
 // If that's a problem, see test/request.go#WaitForEndpointState for oneshot spoofing.
 func New(kubeClientset *kubernetes.Clientset, logger *logging.BaseLogger, domain string, resolvable bool) (*SpoofingClient, error) {
 	sc := SpoofingClient{
-		Client:          http.DefaultClient,
+		Client:          &http.Client{Transport: &ochttp.Transport{Propagation: &b3.HTTPFormat{}}}, // Using ochttp Transport required for zipkin-tracing
 		RequestInterval: requestInterval,
 		RequestTimeout:  requestTimeout,
 		logger:          logger,
@@ -127,6 +135,7 @@ func New(kubeClientset *kubernetes.Clientset, logger *logging.BaseLogger, domain
 
 // Do dispatches to the underlying http.Client.Do, spoofing domains as needed
 // and transforming the http.Response into a spoof.Response.
+// Each response is augmented with "ZipkinTraceID" header that identifies the zipkin trace corresponding to the request.
 func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 	// Controls the Host header, for spoofing.
 	if sc.domain != "" {
@@ -138,12 +147,18 @@ func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 		req.URL.Host = sc.endpoint
 	}
 
-	resp, err := sc.Client.Do(req)
+	// Starting span to capture zipkin trace.
+	traceContext, span := trace.StartSpan(req.Context(), "SpoofingClient-Trace")
+	defer span.End()
+
+	resp, err := sc.Client.Do(req.WithContext(traceContext))
 	if err != nil {
 		return nil, err
 	}
 
 	defer resp.Body.Close()
+
+	resp.Header.Add(zipkin.ZipkinTraceIDHeader, span.SpanContext().TraceID.String())
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
@@ -178,4 +193,35 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
 	})
 
 	return resp, err
+}
+
+// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
+func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+	if err := zipkin.CheckZipkinPortAvailability(); err == nil {
+		return errors.New("port-forwarding for Zipkin is not-setup. Failing Zipkin Trace retrieval")
+	}
+
+	sc.logger.Infof("Logging Zipkin Trace: %s", traceID)
+
+	zipkinTraceEndpoint := zipkin.ZipkinTraceEndpoint + traceID
+	// Sleep to ensure all traces are correctly pushed on the backend.
+	time.Sleep(5 * time.Second)
+	resp, err := http.Get(zipkinTraceEndpoint)
+	if err != nil {
+		return fmt.Errorf("Error retrieving Zipkin trace: %v", err)
+	}
+
+	trace, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("Error reading Zipkin trace response: %v", err)
+	}
+
+	var prettyJSON bytes.Buffer
+	error := json.Indent(&prettyJSON, trace, "", "\t")
+	if error != nil {
+		return fmt.Errorf("JSON Parser Error while trying for Pretty-Format: %v, Original Response: %s", error, string(trace))
+	}
+	sc.logger.Infof(prettyJSON.String())
+
+	return nil
 }
--- a/test/zipkin/util.go
+++ b/test/zipkin/util.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//util has constants and helper methods useful for zipkin tracing support.
+
+package zipkin
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"os/exec"
+
+	"github.com/knative/pkg/test/logging"
+	"go.opencensus.io/trace"
+	"k8s.io/client-go/kubernetes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+
+	//ZipkinTraceIDHeader HTTP response header key to be used to store Zipkin Trace ID.
+	ZipkinTraceIDHeader = "ZIPKIN_TRACE_ID"
+
+	// ZipkinPort port exposed by the Zipkin Pod
+	// https://github.com/knative/serving/blob/master/config/monitoring/200-common/100-zipkin.yaml#L25 configures the Zipkin Port on the cluster.
+	ZipkinPort = 9411
+
+	// ZipkinTraceEndpoint port-forwarded zipkin endpoint
+	ZipkinTraceEndpoint = "http://localhost:9411/api/v2/trace/"
+
+	// ZipkinNamespace namespace where zipkin pod runs.
+	// https://github.com/knative/serving/blob/master/config/monitoring/200-common/100-zipkin.yaml#L19 configures the namespace for zipkin.
+	ZipkinNamespace = "istio-system"
+)
+
+var zipkinPortForwardPID int
+
+// SetupZipkinTracing sets up zipkin tracing which involves a) Setting up port-forwarding from localhost to zipkin pod on the cluster (pid of the process doing Port-Forward is stored in a global variable).
+// b) enable AlwaysSample config for tracing.
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) error {
+	logger := logging.GetContextLogger("SpoofUtil")
+
+	if err := CheckZipkinPortAvailability(); err != nil {
+		return fmt.Errorf("Zipkin port not available on the machine: %v", err)
+	}
+
+	zipkinPods, err := kubeClientset.CoreV1().Pods(ZipkinNamespace).List(metav1.ListOptions{LabelSelector: "app=zipkin"})
+	if err != nil {
+		return fmt.Errorf("Error retrieving Zipkin pod details : %v", err)
+	}
+
+	if len(zipkinPods.Items) == 0 {
+		return errors.New("No Zipkin Pod found on the cluster. Ensure monitoring is switched on for your Knative Setup")
+	}
+
+	portForwardCmd := exec.Command("kubectl", "port-forward", "--namespace=" + ZipkinNamespace, zipkinPods.Items[0].Name, fmt.Sprintf("%d:%d", ZipkinPort, ZipkinPort))
+	if err = portForwardCmd.Start(); err != nil {
+		return fmt.Errorf("Error starting kubectl port-forward command : %v", err)
+
+	}
+	zipkinPortForwardPID = portForwardCmd.Process.Pid
+	logger.Infof("Zipkin port-forward process started with PID: %d", zipkinPortForwardPID)
+
+	// Applying AlwaysSample config to ensure we propagate zipkin header for every request made by this client.
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+
+	logger.Infof("Successfully setup SpoofingClient for Zipkin Tracing")
+
+	return nil
+}
+
+// CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+func CleanupZipkinTracingSetup() error {
+	logger := logging.GetContextLogger("SpoofUtil")
+
+	ps := os.Process{Pid: zipkinPortForwardPID}
+	if err := ps.Kill(); err != nil {
+		return fmt.Errorf("Encoutered error killing port-forward process in CleanupZipkingTracingSetup() : %v", err)
+	}
+
+	logger.Infof("Successfully killed Zipkin port-forward process")
+	return nil
+}
+
+// CheckZipkinPortAvailability checks to see if Zipkin Port is available on the machine.
+// returns error if the port is not available.
+func CheckZipkinPortAvailability() error {
+	server, err := net.Listen("tcp", fmt.Sprintf(":%d", ZipkinPort))
+	if err != nil {
+		// Port is likely taken
+		return err
+	}
+	server.Close()
+	return nil
+}


dushyanthsc:
/assign @adrcunha @srinivashegde86 

cc: @josephburnett

test/spoof/zipkin_util.go
```diff
@@ -0,0 +1,122 @@
...
+const (
+	ZipkinPortForwardPid = "ZIPKIN_PORTFWD_PID"
+	ZipkinTraceIdHeader = "ZIPKIN_TRACE_ID"
+	ZipkinPort = 9411
+	ZipkinTraceEndpoint = "http://localhost:9411/api/v2/trace/%s"
+	ZipkinNamespace = "istio-system"
+)
+
+// Sets up Port-Forward from localhost to Zipkin pod on the cluster. pid of the process doing Port-Forward is stored in an env variable.
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) {
```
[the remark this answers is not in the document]
dushyanthsc: Fixed.

test/spoof/zipkin_util.go
```diff
@@ -0,0 +1,122 @@
...
+
+	logger := logging.GetContextLogger("SpoofUtil")
+	if !CheckZipkinPortAvailability() {
```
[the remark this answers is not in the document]
dushyanthsc: Fixed.

test/spoof/zipkin_util.go
```diff
@@ -0,0 +1,122 @@
...
+		logger.Errorf("Error retrieving Zipkin pod details. Zipkin Setup failed", err)
+		return
+	}
+
+	if len(zipkinPods.Items) == 0 {
+		logger.Errorf("No Zipkin Pod found. Ensure monitoring is switched on for your Knative Setup. Zipkin Setup failed")
+		return
+	}
+
+	portForwardCmd := exec.Command("kubectl", "port-forward", fmt.Sprintf("--namespace=%s", ZipkinNamespace), zipkinPods.Items[0].Name, fmt.Sprintf("%d:%d", ZipkinPort, ZipkinPort))
```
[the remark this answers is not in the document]
dushyanthsc: Output() would call in Run() which blocks the caller as port-forward process does not terminate for entirety of the test. Hence Start() is needed here.

test/spoof/zipkin_util.go
```diff
@@ -0,0 +1,122 @@
...
+	err = portForwardCmd.Start()
+
+	if err != nil {
+		logger.Errorf("Error starting kubectl port-forward command", err)
+		return
+	} else {
+		os.Setenv(ZipkinPortForwardPid, strconv.Itoa(portForwardCmd.Process.Pid))
+
+		// Applying AlwasySample config to ensure we propagate zipkin header for every request made by this client.
```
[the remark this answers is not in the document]
dushyanthsc: Fixed.

test/spoof/zipkin_util.go
```diff
@@ -0,0 +1,122 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Helper methods supporting SpoofClient's Zipkin tracing.
+package spoof
```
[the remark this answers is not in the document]
dushyanthsc: Zipkin itself is enabled inside SpoofingClient when transport for HttpClient is setup to use ochttp transport. But your meta point makes sense. Moved these helper methods and constants to zipkin/util.

test/logging/logging.go
```diff
@@ -67,7 +67,7 @@ func (e *zapMetricExporter) ExportView(vd *view.Data) {
 func (e *zapMetricExporter) ExportSpan(vd *trace.SpanData) {
 	duration := vd.EndTime.Sub(vd.StartTime)
 	// We will start the log entry with `metric` to identify it as a metric for parsing
-	e.logger.Infof("metric %s %d %d %s", vd.Name, vd.StartTime.UnixNano(), vd.EndTime.UnixNano(), duration)
+	e.logger.Debugf("metric %s %d %d %s", vd.Name, vd.StartTime.UnixNano(), vd.EndTime.UnixNano(), duration)
```
adrcunha: This affects our metric aggregation jobs. What's the reason for the change?

test/spoof/spoof.go
```diff
@@ -174,3 +189,33 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+	if zipkin.CheckZipkinPortAvailability() {
+		return errors.New("Zipkin Port(9411) is not-setup. failing Zipkin Trace retrieval")
+	}
+
+	sc.logger.Infof("Logging Zipkin Trace: %s", traceID)
+
+	zipkinTraceEndpoint := fmt.Sprintf(zipkin.ZipkinTraceEndpoint, traceID)
+
+	// Sleep to ensure all traces are correctly pushed on the backend.
+	time.Sleep(5 * time.Second)
```
adrcunha: We had a discussion about the danger of sleeping here. What are the alternatives?

test/spoof/spoof.go
```diff
@@ -174,3 +189,33 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+	resp, err := http.Get(zipkinTraceEndpoint)
+
+	if err != nil {
+		return errors.New("Error retrieving Zipkin trace")
```
adrcunha: Augment the existing error message instead of providing a generic one without details.

test/spoof/spoof.go
```diff
@@ -174,3 +189,33 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+	}
+
+	trace, err := ioutil.ReadAll(resp.Body)
+
+	if err != nil {
+		errors.New("Error reading Zipkin trace response")
```
adrcunha: Augment the existing error message instead of providing a generic one without details.

adrcunha: Missing `return`.

test/spoof/spoof.go
```diff
@@ -174,3 +189,33 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+	}
+
+	sc.logger.Infof(string(trace))
```
adrcunha: What's the expected log output here? Wouldn't be better formatting it somehow?

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	ZipkinTraceIDHeader = "ZIPKIN_TRACE_ID"
+
+	// ZipkinPort port exposed by the Zipkin Pod
+	ZipkinPort = 9411
+
+	// ZipkinTraceEndpoint port-forwarded zipkin endpoint
+	ZipkinTraceEndpoint = "http://localhost:9411/api/v2/trace/%s"
+
+	// ZipkinNamespace namespace where zipkin pod runs.
+	ZipkinNamespace = "istio-system"
```
adrcunha: Why on `istio-system`? Can it be namespace agnostic? Can it run on a non-Knative namespace?

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+
+	// ZipkinPortForwardPid environment variable name to store Process ID of the process started for zipkin port-foward.
+	ZipkinPortForwardPid = "ZIPKIN_PORTFWD_PID"
+
+	//ZipkinTraceIDHeader HTTP response header key to be used to store Zipkin Trace ID.
+	ZipkinTraceIDHeader = "ZIPKIN_TRACE_ID"
+
+	// ZipkinPort port exposed by the Zipkin Pod
+	ZipkinPort = 9411
```
adrcunha: This sounds like would be better if configurable.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	logger.Infof("Successfully killed Zipkin port-forward process")
+
+	return nil
+}
+
+// CheckZipkinPortAvailability checks to see if Zipkin Port is available on the machine.
+// If due to error it is not possible to determine if the port is available the method returns "false"
+func CheckZipkinPortAvailability() bool {
```
adrcunha: Return an error instead of a `bool`.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+
+// CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+// Method expects the port-forwarding process to be stored in an environment variable.
+func CleanupZipkinTracingSetup() error {
+
+	logger := logging.GetContextLogger("SpoofUtil")
+	pid := os.Getenv(ZipkinPortForwardPid)
+
+	if len(pid) == 0 {
+		return errors.New("Zipking port-forward Process ID not found")
```
adrcunha: Make the error message clearer, you're expecting a hardcoded environment variable.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	ZipkinNamespace = "istio-system"
+)
+
+// SetupZipkinTracing sets up zipkin tracing which involves a) Setting up port-forwarding from localhost to zipkin pod on the cluster (pid of the process doing Port-Forward is stored in an env variable).
+// b) enable AlwaysSample config for tracing.
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) error {
+
+	logger := logging.GetContextLogger("SpoofUtil")
+	if !CheckZipkinPortAvailability() {
+		return errors.New("Zipkin port not available on the machine. Zipkin Setup failed")
```
adrcunha: Return the error (see my comment below).

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	zipkinPods, err := kubeClientset.CoreV1().Pods(ZipkinNamespace).List(metav1.ListOptions{LabelSelector: "app=zipkin"})
+
+	if err != nil {
+		return errors.New("Error retrieving Zipkin pod details. Zipkin Setup failed")
```
adrcunha: Return the error, not a generic failure message.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if len(zipkinPods.Items) == 0 {
+		return errors.New("No Zipkin Pod found. Ensure monitoring is switched on for your Knative Setup. Zipkin Setup failed")
```
adrcunha: Make the error message clearer, you're expecting a few hardcoded things here, make it actionable.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	portForwardCmd := exec.Command("kubectl", "port-forward", fmt.Sprintf("--namespace=%s", ZipkinNamespace), zipkinPods.Items[0].Name, fmt.Sprintf("%d:%d", ZipkinPort, ZipkinPort))
+	err = portForwardCmd.Start()
+
+	if err != nil {
+		return errors.New("Error starting kubectl port-forward command")
```
adrcunha: Return the error, not a generic failure message.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/knative/pkg/test/logging"
+	"go.opencensus.io/trace"
+	"k8s.io/client-go/kubernetes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+
+	// ZipkinPortForwardPid environment variable name to store Process ID of the process started for zipkin port-foward.
```
adrcunha: I only see this being used in the setup/cleanup methods. Why does it need to be an environment variable? Why not a global variable?

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	if len(pid) == 0 {
+		return errors.New("Zipking port-forward Process ID not found")
+	}
+
+	logger.Infof("Found Zipkin port-forward Process ID: %s", pid)
+
+	pidInt, err := strconv.Atoi(pid)
+
+	if err != nil {
+		return errors.New("Error converting Zipkin port-forward Process ID into an Int")
```
adrcunha: This will be unnecessary if you use a global variable for the PID.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	ps := os.Process{Pid: pidInt}
+	err = ps.Kill()
+
+	if err != nil {
+		return errors.New("Encoutered error killing process")
```
adrcunha: Again, return the error.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+	// ZipkinTraceEndpoint port-forwarded zipkin endpoint
+	ZipkinTraceEndpoint = "http://localhost:9411/api/v2/trace/%s"
+
+	// ZipkinNamespace namespace where zipkin pod runs.
+	ZipkinNamespace = "istio-system"
+)
+
+// SetupZipkinTracing sets up zipkin tracing which involves a) Setting up port-forwarding from localhost to zipkin pod on the cluster (pid of the process doing Port-Forward is stored in an env variable).
+// b) enable AlwaysSample config for tracing.
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) error {
```
adrcunha: In the design doc we discussed about having the zipkin traces on by default. Your PR makes it optional. What was the conclusion?

test/logging/logging.go
dushyanthsc: The reason was because for Zipkin tracing we are enabling Tracing for every request, which results in too many of these log statements especially for tests that use Poll(..) and E2E tests that make concurrent requests. Is it possible to change the metric aggregation jobs to look for Debug logs and enable verbose logging for those jobs, if not I can revert this change.

test/spoof/spoof.go
dushyanthsc: Discussion in the design doc was around requiring sleep for the client side spans to be pushed to zipkin. However we are no longer pushing client side spans(doc was updated) as only adding the trace header suffices to get details about a request. While testing I discovered that there is a lag even for back-end components(esp istio-policy) where spans were not pushed for few seconds. A 5 second sleep guaranteed that we see all the details for a request hence it is added here in the LogZipkinTrace(..) method. Since these are backend component we don't have any control over them once the request is complete.

test/spoof/spoof.go
dushyanthsc: Fixed.

test/spoof/spoof.go
dushyanthsc: Fixed.

dushyanthsc: Ugg sorry bad one. Fixed it.

test/spoof/spoof.go
dushyanthsc: The trace log itself is a JSON payload representing different traces. Not sure if you thinking about pretty printing of JSON otherwise there is not much we can format.

test/zipkin/util.go
dushyanthsc: The thinking here was that since this process would be outside the Go process (test) it would be better to store the information about it in an environment variable rather than a variable inside the Go process. Especially if the test is somehow cut short for any reason and Cleanup(..) method doesn't get a chance to run. Having it in an environment variable would be more easy to find(still would require a manual kill).

test/zipkin/util.go
dushyanthsc: The port itself is set by the Zipkin setup on the cluster here: https://github.com/knative/serving/blob/master/config/monitoring/200-common/100-zipkin.yaml#L25 So backend side cannot be modified by the test setup. But if you are referring to client side port that is mapped, I fee it would be better there is one port on both ends all controlled by this package rather giving user the flexibility. But if you feel strongly I can change it.

test/zipkin/util.go
dushyanthsc: From the comment here this is a limitation on the Istio side: https://github.com/knative/serving/blob/master/config/monitoring/200-common/100-zipkin.yaml#L19

test/zipkin/util.go
dushyanthsc: As discussed in the design doc, zipkin traces would be enabled by default. Only thing is that following this change I would be updating the Setup and TearDown methods in test/e2e.go. The reason for this is the port-forward process required for Zipkin traces. There is no clean way in SpoofingClient to clean up this process. Adding it in the client would mean each test has to be changed, hence having it the test/e2e.go would be cleaner way to enable zipkin tracing by default and also clean up after test completion.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: I have updated this message but not sure if I have correctly addressed your concern. Please let me know otherwise.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Is this more a Go pattern? coz in my opinion, error would not be appropriate as not having the zipkin port available is not essentially an error, rather the state of the system.

test/spoof/spoof.go
adrcunha: I would challenge the "guaranteed" statement. Let me insist one more time: is it really impossible to  check (for sure) that all traces were pushed?

test/spoof/spoof.go
adrcunha: I'm concerned that this would print a blob hard to read, thus useless. Pretty printing is a possible solution, but it depends on the objective of this log entry: is it intended to be read by a human?

test/zipkin/util.go
adrcunha: I don't see where you're spawning a child process in order to use that, but maybe I'm missing something?

test/zipkin/util.go
adrcunha: Thanks for clarifying. Please refer to the yaml file here, explaining where it's configured, so the reader won't think this was an arbitrary choice.

test/zipkin/util.go
adrcunha: Thanks. Please clarify it here as well.

test/zipkin/util.go
adrcunha: SGTM.

test/zipkin/util.go
adrcunha: You might have several reasons for the port not being available, thus you return the error explaining what was that instead of a generic "it didn't work" (line 130).

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error retrieving Zipkin pod details : %v. Zipkin Setup failed", err)
+	}
+
+	if len(zipkinPods.Items) == 0 {
+		return errors.New("No Zipkin Pod found on the cluster. Ensure monitoring is switched on for your Knative Setup. Zipkin Setup failed")
+	}
+
+	portForwardCmd := exec.Command("kubectl", "port-forward", fmt.Sprintf("--namespace=%s", ZipkinNamespace), zipkinPods.Items[0].Name, fmt.Sprintf("%d:%d", ZipkinPort, ZipkinPort))
```
adrcunha: You don't need `fmt` for composing the namespace flag, you can just concatenate strings.

test/zipkin/util.go
```diff
@@ -0,0 +1,135 @@
...
+
+// CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+// Method expects the port-forwarding process to be stored in an environment variable.
+func CleanupZipkinTracingSetup() error {
+
+	logger := logging.GetContextLogger("SpoofUtil")
+	pid := os.Getenv(ZipkinPortForwardPid)
+
+	if len(pid) == 0 {
+		return fmt.Errorf("Zipking port-forward Process ID not found. CleanupZipkingTracingSetup(..) expects %s environment variable to to be pointing to Process ID perform port-forwarding", ZipkinPortForwardPid)
```
adrcunha: nit: "to to" and unnecessary ".."

test/zipkin/util.go
dushyanthsc: This statement : err = portForwardCmd.Start() starts the child process which is tracked in the environment variable.  Start() returns by creating the child process whose Process ID is stored in Cmd.Process.Pid

test/spoof/spoof.go
dushyanthsc: So the process for trace propagation is:

1. On start of trace the trace ID is posted to Zipkin backend without any spans associated with the trace.
2. After the request is completed, as an out-of-band operation, zipkin instrumentations pushes the span information to Zipkin backend.

Since the span propagation is out-of-band and this is handled by the Zipkin instrumentation library, client doesn't have access to it nor there is an API call to force it. (Zipkin API doc: https://zipkin.io/zipkin-api/#/) As the traceId itself is pushed in Step 1, if a GET call is made on that traceId to Zipkin backend, the call itself will not fail but returns only spans for that trace that have been pushed by the participating components.

I think I loosely used the word "guaranteed" it should be more "Safe to say" that spans will be propagated. We are also logging the TraceID in case if the spans are not pushed in 5 secs.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Got your point and makes sense. So to make sure we also have information about the reason for port-unavailability I am returning both error and bool and logging the error where-ever it makes sense.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed

test/zipkin/util.go
dushyanthsc: Fixed.

test/spoof/spoof.go
srinivashegde86: Is 5 sec the recommended timeout period? Maybe we can create a polling operation that does the GET call till a timeout for every second or so, instead?

test/zipkin/util.go
adrcunha: Yes, I'm aware of all that. But as I said, you're setting the variable later, so `kubectl` is not using it. I still don't see where and why you're using an environment variable here instead of a simple global variable.

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+
+	logger := logging.GetContextLogger("SpoofUtil")
+
+	zipkinPortAvailable, err := CheckZipkinPortAvailability()
+	if !zipkinPortAvailable {
+		return fmt.Errorf("Zipkin port not available on the machine: %v. Zipkin Setup failed", err)
+	}
+
+	zipkinPods, err := kubeClientset.CoreV1().Pods(ZipkinNamespace).List(metav1.ListOptions{LabelSelector: "app=zipkin"})
+
```
srinivashegde86: [nit] dont add blank lines in these cases. they are part of the same group

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return fmt.Errorf("Error retrieving Zipkin pod details : %v. Zipkin Setup failed", err)
+	}
+
+	if len(zipkinPods.Items) == 0 {
+		return errors.New("No Zipkin Pod found on the cluster. Ensure monitoring is switched on for your Knative Setup. Zipkin Setup failed")
```
srinivashegde86: I dont think we need to add Zipkin setup failed to the error messages. The trace will metion the function name SetupZipkin

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	portForwardCmd := exec.Command("kubectl", "port-forward", "--namespace=" + ZipkinNamespace, zipkinPods.Items[0].Name, fmt.Sprintf("%d:%d", ZipkinPort, ZipkinPort))
+	err = portForwardCmd.Start()
+
```
srinivashegde86: remove blank line

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+// CheckZipkinPortAvailability checks to see if Zipkin Port is available on the machine.
+// If due to error it is not possible to determine if the port is available the method returns "false"
+func CheckZipkinPortAvailability() (bool, error) {
+
+	host := ":" + strconv.Itoa(ZipkinPort)
+	server, err := net.Listen("tcp", host)
```
srinivashegde86: can be combined.

if server, err := net.Listen(); err != nil {
return false, err
} else {
server.Close()
return true, nil
}

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+
+	pidInt, err := strconv.Atoi(pid)
+
+	if err != nil {
+		return fmt.Errorf("Error converting Zipkin port-forward Process ID into an Int : %v", err)
+	}
+
+	ps := os.Process{Pid: pidInt}
+	err = ps.Kill()
+
```
srinivashegde86: remove blank line

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+// CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+// Method expects the port-forwarding process to be stored in an environment variable.
+func CleanupZipkinTracingSetup() error {
+
+	logger := logging.GetContextLogger("SpoofUtil")
+	pid := os.Getenv(ZipkinPortForwardPid)
+
```
srinivashegde86: remove blank line

test/zipkin/util.go
```diff
@@ -0,0 +1,139 @@
...
+	if len(pid) == 0 {
+		return fmt.Errorf("Zipking port-forward Process ID not found. CleanupZipkingTracingSetup() expects %s environment variable to be pointing to Process ID perform port-forwarding", ZipkinPortForwardPid)
+	}
+
+	logger.Infof("Found Zipkin port-forward Process ID: %s", pid)
+
+	pidInt, err := strconv.Atoi(pid)
+
```
srinivashegde86: remove blank line

test/spoof/spoof.go
adrcunha: We need to make sure (2) happens before shutting down the test environment. Waiting and hoping that it's done should be the last resort. This sounds like a too common task to me, so that we're the only ones affected by the lack of a deterministic behavior. If the API doesn't provide the feature, maybe there's a signal you can send, or a graceful shutdown that ensures the completion? As a last resort, there's Srinivas' suggestion of polling the server to ensure the operations were completed.

test/zipkin/util.go
adrcunha: This is unnecessary complexity. Return the error if something failed, return `nil` if no error.

test/zipkin/util.go
adrcunha: I'll ignore the missed "perform" because I'm sure you'll abandon this env var approach, marking as resolved. :)

test/zipkin/util.go
adrcunha: As I mentioned before, this is unnecessary complexity. Just return the error or `nil`. You don't even need `host` here, because it's used just once and is a simple concatenation (I would even use `fmt.Sprintf` for clarity).

test/spoof/spoof.go
dushyanthsc: I don't think poll will help here, as GET even with no spans would not fail because of Step 1.  Having said that I will dig further and also reach out to the monitoring team.

test/spoof/spoof.go
dushyanthsc: Yes, it would be read by human. I have pretty printed the JSON output.

test/zipkin/util.go
dushyanthsc: You win :)

test/zipkin/util.go
dushyanthsc: Sorry was just trying to answer your question "I don't see where you're spawning a child process". The "kubectl" itself is the child process as the port-forward command doesn't terminate till the process that started itself is killed. So what I was doing is storing the process id of the kubectl command in the environment variable and then using that in the cleanup method to kill it. 

The rationale behind of using an environment variable instead of global variable was explained in my earlier comment - The test could be killed before cleanup method gets a chance to run. In such cases having it in env variable would be easier then a global variable.

I have updated to use global variable. But also wanted to make sure I explain the rationale behind.

test/zipkin/util.go
dushyanthsc: Its updated to use GlobalVariable.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/util.go
adrcunha: Unless I'm missing something here, I don't see how the environment variable is helping you here. If the test can be killed before the cleanup phase, why the cleanup phase survives? And if it survives, it should be able to read the PID from a local resource. You're not using the env var in a subprocess, you're just using it as a storage area. I would understand that in a sibling UNIX process context, which doesn't seem to be the case here. Correct me if I'm wrong.

dushyanthsc: Yes, if the test is killed before the cleanup phase, the cleanup phase would not survive. In such a scenario it would require a manual kill of the port-forward process. This might come up when developer is testing in local box and cut short the test by Crtl+C. My rationale was in such scenario it would be easier to find the Process to kill if the process-id is stored in an environment variable. Hence in my first comment I mentioned : "Having it in an environment variable would be more easy to find(still would require a manual kill)."

test/spoof/spoof.go
adrcunha: Nice. Can you please add an example of the output as a comment to this PR?

test/zipkin/util.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	logger.Infof("Successfully killed Zipkin port-forward process")
+	return nil
+}
+
+// CheckZipkinPortAvailability checks to see if Zipkin Port is available on the machine.
+// returns error if the port is not available.
+func CheckZipkinPortAvailability() error {
+
```
adrcunha: Please remove this blank line.

test/zipkin/util.go
```diff
@@ -0,0 +1,112 @@
...
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+
+	logger.Infof("Successfully setup SpoofingClient for Zipkin Tracing")
+
+	return nil
+}
+
+// CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+func CleanupZipkinTracingSetup() error {
+
```
adrcunha: Please remove this blank line.

test/zipkin/util.go
```diff
@@ -0,0 +1,112 @@
...
+	// https://github.com/knative/serving/blob/master/config/monitoring/200-common/100-zipkin.yaml#L19 configures the namespace for zipkin.
+	ZipkinNamespace = "istio-system"
+)
+
+var zipkinPortForwardPID int
+
+// SetupZipkinTracing sets up zipkin tracing which involves a) Setting up port-forwarding from localhost to zipkin pod on the cluster (pid of the process doing Port-Forward is stored in a global variable).
+// b) enable AlwaysSample config for tracing.
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) error {
+
```
adrcunha: Please remove this blank line.

test/spoof/spoof.go
```diff
@@ -179,3 +196,35 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
 
 	return resp, err
 }
+
+// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
+func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+
```
adrcunha: Please remove this blank line.

test/spoof/spoof.go
```diff
@@ -127,7 +135,9 @@ func New(kubeClientset *kubernetes.Clientset, logger *logging.BaseLogger, domain
 
 // Do dispatches to the underlying http.Client.Do, spoofing domains as needed
 // and transforming the http.Response into a spoof.Response.
+// Each response is augmented with "ZipkinTraceID" header that identifies the zipkin trace corresponding to the request.
 func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
+
```
adrcunha: Please remove this blank line.

test/zipkin/util.go
adrcunha: You won't be able to inspect such environment variable in the given scenario because your process will be in different shells. In this case, logging the PID is the best option.

test/spoof/spoof.go
```diff
@@ -138,12 +148,19 @@ func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 		req.URL.Host = sc.endpoint
 	}
 
-	resp, err := sc.Client.Do(req)
+	// Starting span to capture zipkin trace.
+	traceContext, span := trace.StartSpan(req.Context(), "SpoofingClient-Trace")
+	defer span.End()
+
+	resp, err := sc.Client.Do(req.WithContext(traceContext))
+
```
adrcunha: Please remove this blank line, it's part of the same logical block.

test/zipkin/util.go
dushyanthsc: Ah got it. I have added a log statement.

test/spoof/spoof.go
adrcunha: Edit your comment and add the proper markdown formatting (3 backticks before and after the block) and it should be rendered correctly.

adrcunha: Is that a single log entry? I was expecting something way smaller. I'm concerned that this will flood the test logs. Can you log just the relevant fields (or assemble, say, an inspection URL with the relevant fields)?

josephburnett: I recommend dumping the whole trace since I'm not sure what part we'll need.  But only do it once.  E.g. if we get 10 502's, just dump the trace for the first one and then log how many 502's we received in total.

dushyanthsc: The call to LogZipkinTrace() would be made from the test, so the logic to aggregate the number of 502s (or other condition depending on the test)would reside within the test.

test/spoof/spoof.go
dushyanthsc: After investigating further and reaching out to OpenZipkin community forum ((https://gitter.im/openzipkin/zipkin?at=5bacc7b1cdc50131725f4abd) and talking to monitoring team, it was identified that currently, there is no way to force server side components to publish spans. This mainly stems from the fact that Zipkin being distributed there is no centralized authority that can enforce this behavior. Due to this limitation, waiting by sleep to allow components to push their spans is the best we can currently do.

test/spoof/spoof.go
dushyanthsc: Based on the comments above, leaving the current logging statement as is.

test/logging/logging.go
dushyanthsc: I have reverted this back Infof. Enabling tracing for every request is required by Zipkin. But moving to Debug does affect other jobs, it makes sense to leave it as Infof.

adrcunha:
@dushyanthsc you need further approval due to the changes to the Gopkg files. Which leads me to ask: you updated the lock/toml files but not anything in vendor, is it expected?

Issue #124, opened 2018-10-11
Expose ClientAuth as an option for webhook


zrss:
## Expected Behavior

Expose ClientAuth as an option for webhook

## Actual Behavior

https://github.com/knative/pkg/blob/bb9898e8ebb2e8a76927ae4778c30fecfb32377d/webhook/webhook.go#L152-L160

/assign

zrss:
:smiley: and kindly ping TODOer @mattmoor

PR #125, opened 2018-10-11
Make clientAuth of webhook server into a configuration option

zrss:
Fixes: #124 

Make clientAuth of webhook server into a configuration option

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -94,6 +94,11 @@ type ControllerOptions struct {
 	// potential races where registration completes and k8s apiserver
 	// invokes the webhook before the HTTP server is started.
 	RegistrationDelay time.Duration
+
+	// ClientAuthType declares the policy the webhook server will follow for
+	// TLS Client Authentication.
+	// The default value is tls.NoClientCert.
+	ClientAuth tls.ClientAuthType
 }
 
 // ResourceCallback defines a signature for resource specific (Route, Configuration, etc.)
@@ -142,7 +147,7 @@ func getAPIServerExtensionCACert(cl kubernetes.Interface) ([]byte, error) {
 }
 
 // MakeTLSConfig makes a TLS configuration suitable for use with the server
-func makeTLSConfig(serverCert, serverKey, caCert []byte) (*tls.Config, error) {
+func makeTLSConfig(serverCert, serverKey, caCert []byte, clientAuthType tls.ClientAuthType) (*tls.Config, error) {
 	caCertPool := x509.NewCertPool()
 	caCertPool.AppendCertsFromPEM(caCert)
 	cert, err := tls.X509KeyPair(serverCert, serverKey)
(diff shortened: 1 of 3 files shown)


webhook/webhook.go
```diff
@@ -94,6 +94,10 @@ type ControllerOptions struct {
 	// potential races where registration completes and k8s apiserver
 	// invokes the webhook before the HTTP server is started.
 	RegistrationDelay time.Duration
+
+	// ClientAuthType declares the policy the webhook server will follow for
+	// TLS Client Authentication.
+	ClientAuth tls.ClientAuthType
```
n3wscott: Can you add a comment to say what this default is when not set from the config? It took me awhile to understand that `0 == tls.NoClientCert` OR I am misunderstanding.

zrss: yes, the default value of tls.ClientAuthType is tls.NoClientCert. i have added a comment to illustrate it. PTAL

zrss:
thanks @mattmoor, and kindly ping @n3wscott for reviewing

n3wscott:
Grant for approval:
/assign @grantr

zrss:
thanks @n3wscott (also ping @mattmoor ) and can u add me as a knative member. currently, i am working hard in this repo (means that `knative` or `knative/pkg` i am not sure) for my study, and yep, finds some bug and try my best to fix it

webhook/webhook_test.go
```diff
@@ -367,6 +367,20 @@ func TestCertConfigurationForGeneratedSecret(t *testing.T) {
...
 }
 
+func TestSettingWebhookClientAuth(t *testing.T) {
+	opts := newDefaultOptions()
+	if opts.ClientAuth != tls.NoClientCert {
+		t.Fatalf("Expected default ClientAuth to be NoClientCert (%v) but got (%v)",
+			tls.NoClientCert, opts.ClientAuth)
+	}
+
+	opts.ClientAuth = tls.RequireAndVerifyClientCert
```
grantr: This test isn't necessary since it's just testing native Go assignment. Remove this one and keep the default check above.

webhook/webhook.go
```diff
@@ -235,15 +236,14 @@ func SetDefaults(ctx context.Context) ResourceDefaulter {
 
 func configureCerts(ctx context.Context, client kubernetes.Interface, options *ControllerOptions) (*tls.Config, []byte, error) {
 	apiServerCACert, err := getAPIServerExtensionCACert(client)
-	if err != nil {
+	if (options.ClientAuth >= tls.VerifyClientCertIfGiven) && err != nil {
```
grantr: Two questions:

Why is this a `>=` check? Are we checking for either `VerifyClientCertIfGiven` or `RequireAndVerifyClientCert`? If so I suggest changing this to an OR.

Why do we only check the error value if the ClientAuth option is set?

grantr: I don't know enough about Kubernetes webhook certs to say for sure. Paging @mattmoor :smile: 

/cc @mattmoor

mattmoor: This is @vaikas-google territory

zrss: right, that's the point that i try to do, and push some new changes, PTAL

PR #126, opened 2018-10-11
Add stackdriver exporter support in pkg.

akyyy:


metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+	"contrib.go.opencensus.io/exporter/stackdriver"
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/stats/view"
+	"go.uber.org/zap"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ObservabilityConfigName = "config-observability"
```
mattmoor: This looks unused and fwiw probably doesn't belong in `knative/pkg`

metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+	backendDestinationKey   = "metrics.backend-destination"
+	stackdriverProjectIdKey = "metrics.stackdriver-project-id"
+	metricsPath             = "/etc/config-observability"
```
mattmoor: also unused.

metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+
+type metricsConfig struct {
+	// The metrics domain. e.g. "serving.knative.dev" or "build.knative.dev".
+	domain string
+	// The component that emits the metrics. e.g. "activator", "autoscaler".
+	component string
+	// The metrics backend destination.
+	backendDestination MetricsBackend
+	// The stackdriver project ID.
+	stackdriverProjectId string
```
mattmoor: Go style wants `ProjectID` (capitalize `d`)

metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if exporter != nil {
+		view.UnregisterExporter(exporter)
+	}
+	if config.backendDestination == Stackdriver {
+		exporter, err = newStackdriverExporter(config, logger)
+	} else {
+		exporter, err = newPrometheusExporter(config, logger)
+	}
+	if err == nil {
```
mattmoor:
```
if err != nil {
  return err
}
...
return nil
```

Consider doing this directly following the two places it's set.

metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+		view.RegisterExporter(exporter)
+		view.SetReportingPeriod(10 * time.Second)
+		logger.Info("Registered the exporter.")
+		logger.Infof("Successfully updated the metrics exporter; old config: %v; new config %v", mConfig, config)
```
mattmoor: nit: two `Infof` back to back seems spammy.

metrics/config.go
```diff
@@ -0,0 +1,198 @@
...
+	"go.opencensus.io/stats/view"
+	"go.uber.org/zap"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ObservabilityConfigName = "config-observability"
+	backendDestinationKey   = "metrics.backend-destination"
+	stackdriverProjectIdKey = "metrics.stackdriver-project-id"
```
mattmoor: Capitalize `d` (see below).

Do we expect there to be cases where this is something other than the cluster's project?

If not, consider using: https://godoc.org/cloud.google.com/go/compute/metadata#ProjectID

Gopkg.toml
```diff
@@ -37,6 +37,10 @@ required = [
   name = "go.uber.org/zap"
   revision = "67bc79d13d155c02fd008f721863ff8cc5f30659"
 
+[[override]]
+  name = "github.com/census-ecosystem/opencensus-go-exporter-stackdriver"
+  version = "0.6.0"
```
mattmoor: Ouch, this pulls in a lot of stuff :-/

metrics/config.go
akyyy: yes, this is the stackdriver project id, not necessarily the GCP project id.

akyyy:
Thanks for taking a look!

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+	} else {
+		mb = Prometheus
+		if !strings.EqualFold(backend, "prometheus") {
+			logger.Infof("Unsupported metrics backend value \"%s\". Use the default metrics backend prometheus.", backend)
+		}
+	}
+	mc.backendDestination = mb
+
+	sdProj, ok := m[stackdriverProjectIDKey]
+	if strings.EqualFold(backend, "stackdriver") && !ok {
```
mdemirhan: why not mb.backendDestination == Stackdriver? Would be cheaper to do and easier to understand. Also, following might look better (two ifs combined):
```
    sdProj, ok := m[stackdriverProjectIDKey]
    if !ok || sdProj == "" {
        return mc, some error
    }
}
```

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+		return mc, errors.New("metrics.stackdriver-project-id key is missing when the backend-destination is set to stackdriver.")
+	}
+	if strings.EqualFold(backend, "stackdriver") && sdProj == "" {
+		return mc, errors.New("metrics.stackdriver-project-id field cannot be empty.")
```
mdemirhan: why return mc at all when returning an error?

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+		mb = Prometheus
+		if !strings.EqualFold(backend, "prometheus") {
+			logger.Infof("Unsupported metrics backend value \"%s\". Use the default metrics backend prometheus.", backend)
+		}
+	}
+	mc.backendDestination = mb
+
+	sdProj, ok := m[stackdriverProjectIDKey]
+	if strings.EqualFold(backend, "stackdriver") && !ok {
+		return mc, errors.New("metrics.stackdriver-project-id key is missing when the backend-destination is set to stackdriver.")
```
mdemirhan: why return mc at all when returning an error?

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+		promSrv.ListenAndServe()
+	}()
+	return e, nil
+}
+
+func getMetricsConfig(m map[string]string, domain string, component string, logger *zap.SugaredLogger) (metricsConfig, error) {
+	var mc metricsConfig
+	backend, ok := m[backendDestinationKey]
+	if !ok {
+		return mc, errors.New("metrics.backend-destination key is missing")
```
mdemirhan: why return mc at all when returning an error?

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+	sdProj, ok := m[stackdriverProjectIDKey]
+	if strings.EqualFold(backend, "stackdriver") && !ok {
+		return mc, errors.New("metrics.stackdriver-project-id key is missing when the backend-destination is set to stackdriver.")
+	}
+	if strings.EqualFold(backend, "stackdriver") && sdProj == "" {
+		return mc, errors.New("metrics.stackdriver-project-id field cannot be empty.")
+	}
+	mc.stackdriverProjectID = sdProj
+
+	if domain == "" {
```
mdemirhan: If domain is wrong, this should error out.

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+		logger.Info("Metrics domain name missing. Use \"domain\"")
+		mc.domain = "domain"
+	} else {
+		mc.domain = domain
+	}
+
+	if component == "" {
```
mdemirhan: If component is missing, this should error out.

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+			if exporter == nil {
+				// Fail the process if there doesn't exist an exporter.
+				logger.Fatal("Failed to get a valid metrics config")
+			} else {
+				logger.Error("Failed to get a valid metrics config; Skip updating the metrics exporter", zap.Error(err))
+				return
+			}
+		}
+		if newConfig.backendDestination != mConfig.backendDestination ||
+			newConfig.stackdriverProjectID != mConfig.stackdriverProjectID {
```
mdemirhan: If destination is prometheus, but sdprojectid changes, this will recreate the exporter, which is not needed. I would restructure this for easier reading:
```
changed bool
if newConfig.backendDestination != mConfig.backendDestination {
    changed = true
} else if newConfig.backendDestination == stackdriver && newConfig.stackdriverProjectID != mConfig.stackdriverProjectID) {
    changed = true
}

if changed {
   err = newMetricsExporter...
...
}
```

metrics/config.go
```diff
@@ -0,0 +1,200 @@
...
+		mc.component = component
+	}
+	return mc, nil
+}
+
+// UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
+// when a config map is updated
+func UpdateExporterFromConfigMap(domain string, component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	return func(configMap *corev1.ConfigMap) {
+		newConfig, err := getMetricsConfig(configMap.Data, domain, component, logger)
```
mdemirhan: You should get a lock on the mutex above. Otherwise, another call to this method could change contents of exporter or mConfig while in the middle of this method which will have unintended affects.

But then locking here and not unlocking before newMetricsExporter will cause a deadlock. It might be best if the lock is removed from newMetricsExporter and is just added here (as long as newMetricsExporter is not called elsewhere).

metrics/config_test.go
```diff
@@ -0,0 +1,187 @@
...
+		component:            testComponent,
+		backendDestination:   Stackdriver,
+		stackdriverProjectID: testProj}, TestLogger(t))
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestGetMetricsConfig(t *testing.T) {
+	_, err := getMetricsConfig(map[string]string{
```
mdemirhan: Config tests should be mostly table tests. See https://github.com/knative/serving/blob/master/pkg/autoscaler/config_test.go as an example.

metrics/config.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+func getMetricsConfig(m map[string]string, domain string, component string, logger *zap.SugaredLogger) (*metricsConfig, error) {
+	var mc metricsConfig
+	backend, ok := m[backendDestinationKey]
+	if !ok {
+		return nil, errors.New("metrics.backend-destination key is missing")
+	}
+	lb := strings.ToLower(backend)
+	switch lb {
```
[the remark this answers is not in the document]
akyyy: Thanks for the casting.
I don't want to say we only support "stackdriver", not "Stackdriver", or "STACKDRIVER".

metrics/config_test.go
```diff
@@ -0,0 +1,175 @@
...
+		domain:    servingDomain,
+		component: testComponent,
+		expectedConfig: metricsConfig{
+			domain:               servingDomain,
+			component:            testComponent,
+			backendDestination:   Stackdriver,
+			stackdriverProjectID: anotherProj},
+	}, {
+		name: "validCapitalStackdriver",
+		cm: map[string]string{"metrics.backend-destination": "STACKDRIVER",
```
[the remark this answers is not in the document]
akyyy: "Stackdriver" is more real.

metrics/config.go
```diff
@@ -0,0 +1,128 @@
...
+
+// UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
+// when a config map is updated
+func UpdateExporterFromConfigMap(domain string, component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	return func(configMap *corev1.ConfigMap) {
+		newConfig, err := getMetricsConfig(configMap.Data, domain, component, logger)
+		if err != nil {
+			if exporter == nil {
+				// Fail the process if there doesn't exist an exporter.
+				logger.Fatal("Failed to get a valid metrics config")
```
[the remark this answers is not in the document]
akyyy: For this func, we need to implement Observer.
https://github.com/knative/pkg/blob/403c2256ff0ebcfc34748c3be743c2fa61f4ffc5/configmap/watcher.go#L27

metrics/exporter.go
```diff
@@ -0,0 +1,109 @@
...
+	mConfig     *metricsConfig
+	promSrvChan chan *http.Server = make(chan *http.Server, 1)
+)
+
+// newMetricsExporter gets a metrics exporter based on the config.
+// This function is not thread safe. It is client's responsibility to synchronize any
+// multithreaded access.
+func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) error {
+	select {
+	case svr := <-promSrvChan:
```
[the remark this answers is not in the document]
akyyy: Thanks for your suggestion. Finally got rid of promSrvChan.

metrics/exporter.go
```diff
@@ -0,0 +1,121 @@
...
+	"contrib.go.opencensus.io/exporter/stackdriver"
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/stats/view"
+	"go.uber.org/zap"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+var (
+	exporter   view.Exporter
+	mConfig    *metricsConfig
```
mdemirhan: The naming is inconsistent with above. I recommend using something like metricsConfig or even clearer currentConfig or currentMetricsConfig

metrics/exporter.go
```diff
@@ -0,0 +1,121 @@
...
+	promSrv    *http.Server
```
mdemirhan: same comment as above. a better naming is needed here. activePromSrv would be better.

metrics/config.go
```diff
@@ -0,0 +1,126 @@
...
+			} else {
+				logger.Error("Failed to get a valid metrics config; Skip updating the metrics exporter", zap.Error(err))
+				return
+			}
+		}
+		changed := false
+		mux.Lock()
```
mdemirhan: See my comments on the exporter.go

metrics/config.go
```diff
@@ -0,0 +1,126 @@
...
+		if mConfig == nil || newConfig.backendDestination != mConfig.backendDestination {
```
mdemirhan: This would be better served as a separate func. `isConfigChanged(oldConfig, newConfig *MetricsConfig)` or something similar. You should also write dedicated tests for that func.

metrics/config.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// isMetricsConfigChanged compares the non-nil newConfig against curMetricsConfig. When backend changes,
+// or stackdriver project ID changes for stackdriver backend, we need to update the metrics exporter.
+func isMetricsConfigChanged(newConfig *metricsConfig) bool {
+	changed := false
+	cc := getCurMetricsConfig()
+	if cc == nil || newConfig.backendDestination != cc.backendDestination {
+		changed = true
```
mdemirhan: nit: you can remove changed variable and short circuit trues with a return. Example:
```
	cc := getCurMetricsConfig()
	if cc == nil || newConfig.backendDestination != cc.backendDestination {
		return true
	} else if newConfig.backendDestination == Stackdriver && newConfig.stackdriverProjectID != cc.stackdriverProjectID {
		return true
	}
	return false
```

metrics/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+func resetCurPromSrv() {
+	metricsMux.Lock()
+	defer metricsMux.Unlock()
+	if curPromSrv != nil {
+		curPromSrv.Close()
+		curPromSrv = nil
+	}
+}
+
+func startCurPromSrv(e *prometheus.Exporter) *http.Server {
```
mdemirhan: nit: This should be named startNewPromSrv as it is settings the curPromSrv and not using it.

metrics/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+	case Prometheus:
+		e, err = newPrometheusExporter(config, logger)
+	default:
+		err = fmt.Errorf("Unsupported metrics backend %v", config.backendDestination)
+	}
+	if err != nil {
+		return err
+	}
+	existingConfig := getCurMetricsConfig()
+	logger.Infof("Successfully updated the metrics exporter; old config: %v; new config %v", existingConfig, config)
```
mdemirhan: Nit: the log should be after the setCurMetricsExporterAndConfig call

metrics/exporter.go
```diff
@@ -0,0 +1,146 @@
...
+		curPromSrv = nil
+	}
+}
+
+func startCurPromSrv(e *prometheus.Exporter) *http.Server {
+	sm := http.NewServeMux()
+	sm.Handle("/metrics", e)
+	metricsMux.Lock()
+	defer metricsMux.Unlock()
+	curPromSrv = &http.Server{
```
mdemirhan: There is a chance that another thread sets the curPromSrv right before this one as well. I recommend the following:
```
	if curPromSrv != nil {
		curPromSrv.Close()
	}
	curPromSrv = &http.Server{
		Addr:    ":9090",
		Handler: sm,
	}
	return curPromSrv
```

mattmoor:
Can you rebase to fix the merge conflict?

metrics/config.go
```diff
@@ -0,0 +1,126 @@
...
+	// The component that emits the metrics. e.g. "activator", "autoscaler".
+	component string
+	// The metrics backend destination.
+	backendDestination MetricsBackend
+	// The stackdriver project ID where the stats data are uploaded to. This is
+	// not the GCP project ID.
+	stackdriverProjectID string
+}
+
+var domainRE = regexp.MustCompile("^(serving|build|eventing)\\.knative\\.dev$")
```
mattmoor: This feels like a layering violation.  I haven't traced through how this is used, but could we pass this in from the consuming repo?

akyyy:
@mattmoor I merged the conflict and removed the regex in pkg repo. Please take a look. Thanks,

mattmoor:
/approve

For the new `OWNERS` file and the `Gopkg.lock` changes.  Will defer to @mdemirhan for `/lgtm`

Issue #128, opened 2018-10-12
Allow unknown JSON fields


mattmoor:
Makes things less bad wrt API versioning.

PR #129, opened 2018-10-12
Allow unknown JSON fields

jonjohnsonjr:
This enables us to make additive changes without breaking downgrades.

Unfortunate side effect is that we don't get a nice typo checker in the
webhook :(

Fixes: #128


grantr:
/cc @mattmoor 
I think you added this in the first place (probably to get the typo checker)

jonjohnsonjr:
@grantr see https://github.com/knative/pkg/issues/128 + https://github.com/knative/serving/issues/2224 -- this PR came out of an offline conversation with Matt re: our ability to downgrade after we make additive changes.

PR #133, opened 2018-10-19
 fix bug of generating plural in webhook

lichuqiang:
Fixes: https://github.com/knative/serving/issues/2262

/assign @tcnghia @evankanderson

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -38,6 +38,7 @@ import (
 	"github.com/knative/pkg/logging"
 	"github.com/knative/pkg/logging/logkey"
 
+	"github.com/markbates/inflect"
 	"github.com/mattbaird/jsonpatch"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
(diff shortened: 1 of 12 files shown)


webhook/webhook.go
```diff
@@ -326,7 +327,7 @@ func (ac *AdmissionController) register(
 	var rules []admissionregistrationv1beta1.RuleWithOperations
 	for gvk := range ac.Handlers {
 		// Lousy pluralizer
-		plural := strings.ToLower(gvk.Kind) + "s"
+		plural := strings.ToLower(inflect.Pluralize(gvk.Kind))
```
tcnghia: Do we know if this will change existing behavior of other types?

lichuqiang: The tool has a set of rules for plural conversion “smartly”.
We didn't hit this issue before because all of the existing resources did not need "es" as suffix when converting to plural.

In fact I borrow this logic from [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/resource/crd.go#L56). so k8s api should follow the same rules. And we won't break backwards compatibility.

mattmoor: Can you get rid of the comment, since this is no longer a "Lousy pluralizer" :)

mattmoor:
Noticed one more nit, if you fix I'll stamp it in the morning here.

lichuqiang:
@mattmoor ,updated to resolve the conflict. PTAL :)

webhook/webhook.go
lichuqiang: done

PR #136, opened 2018-10-25
Clean useless unregister mutatingwebhookconfiguration process

zrss:

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -294,11 +294,6 @@ func (ac *AdmissionController) Run(stop <-chan struct{}) error {
 			logger.Error("Failed to register webhook", zap.Error(err))
 			return err
 		}
-		defer func() {
-			if err := ac.unregister(ctx, cl); err != nil {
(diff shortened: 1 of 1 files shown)


zrss:
hi @akyyy , maybe there is a flaky test about `TestNewStackdriverExporter` and i can't reproduce in my local machine, but actually it fail in `pull-knative-pkg-unit-tests`, i have to work around by adding `resetCurPromSrc` at the top of `TestNewStackdriverExporter`

metrics/exporter_test.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 func TestNewStackdriverExporter(t *testing.T) {
+	resetCurPromSrv()
```
vaikas-google: What's this change about?

akyyy: That's the workaround for a flaky test. I'm trying to address it in another pr
https://github.com/knative/pkg/pull/138

zrss: Mhm., waiting for fixing :smirk:, i have removed the work around commit. PTAL

Issue #139, opened 2018-10-26
Add webhook svc name to DNSnames of cert


zrss:
https://github.com/knative/pkg/blob/5c809f170b42301a276080f4fd884a30188ca1a5/webhook/certs.go#L49-L50

lack of the name of svc, full connectable DNSnames are

1. {svcName}
1. {svcName}.{Namespace}
1. {svcName}.{Namespace}.svc
1. {svcName}.{Namespace}.svc.cluster.local

/assign @zrss

PR #140, opened 2018-10-26
Add webhook svc name to DNSNames of cert

zrss:
Fixes: #139

### DIFF
--- a/webhook/certs.go
+++ b/webhook/certs.go
@@ -47,7 +47,12 @@ func createCertTemplate(name, namespace string) (*x509.Certificate, error) {
 	}
 
 	serviceName := name + "." + namespace
-	serviceNames := []string{serviceName, serviceName + ".svc", serviceName + ".svc.cluster.local"}
+	serviceNames := []string{
+		name,
+		serviceName,
+		serviceName + ".svc",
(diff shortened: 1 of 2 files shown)


webhook/certs.go
```diff
@@ -47,7 +47,12 @@ func createCertTemplate(name, namespace string) (*x509.Certificate, error) {
 	}
 
 	serviceName := name + "." + namespace
-	serviceNames := []string{serviceName, serviceName + ".svc", serviceName + ".svc.cluster.local"}
+	serviceNames := []string{
```
vaikas-google: What's the reasoning for allowing non-namespaced address here?

zrss: that's the point, we should add the svc name to DNSNames field.

as the kubernetes service docs said [what-things-get-dns-names](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#what-things-get-dns-names)

```
Assume a Service named **foo** in the Kubernetes namespace **bar**. A Pod running in namespace bar can look up this service by simply doing a DNS query for **foo**. A Pod running in namespace quux can look up this service by doing a DNS query for foo.bar.
```

we can directly use svc name (foo) to access current namespace's svc (foo.bar)

WDYT

PR #145, opened 2018-10-26
Add support for StrategicMergePatch to duck

### LINKED ISSUE #144: Fake clients do not support our duck.patch format
## Expected Behavior
I should be able to pass a patch created from our apis.duck.patch to a fake client.

## Actual Behavior
This fails due to https://github.com/knative/pkg/blob/master/vendor/k8s.io/client-go/testing/fixture.go#L140 - k8s fake clients only support StrategicMergePatch's (https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md) and we produce a JSONPatch.

greghaynes:
K8s fake clients only support StrategicMergePatches so we should be
creating patches in this format.

Fixes #144

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -77,6 +77,14 @@
   revision = "782f4967f2dc4564575ca782fe2d04090b5faca8"
 
 [[projects]]
+  digest = "1:32598368f409bbee79deb9d43569fcd92b9fb27f39155f5e166b3371217f051f"
+  name = "github.com/evanphx/json-patch"
+  packages = ["."]
+  pruneopts = "NUT"
+  revision = "72bf35d0ff611848c1dc9df0f976c81192392fa5"
+  version = "v4.1.0"
+
+[[projects]]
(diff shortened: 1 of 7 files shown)


vaikas-google:
FYI, I added json patch support couple of weeks back. I'd like to figure out which format we really want to be using not only based on what the fake clients support or not, My limited understanding is that they both have some strengths and limitations.
@mattmoor 
@greghaynes 

https://github.com/kubernetes/client-go/commit/70926af6e803a02d5c4cc3d3567de40d3c095cab

greghaynes:
Looks like we barely missed that fix for us :) - Barring any other knative-specific limitations we hit IMO we should follow k8s and use StrategicMergePatch as I expect they will choose the optimal option for mutating k8s resources. We may want to look in to using https://godoc.org/k8s.io/apimachinery/pkg/util/strategicpatch#CreateTwoWayMergePatch instead for generating these patches for the same reason.

PR #146, opened 2018-10-29
update EnqueueControllerOf to properly handle Delete

lichuqiang:
First step to address https://github.com/knative/serving/issues/2343

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -123,19 +123,72 @@ func (c *Impl) Enqueue(obj interface{}) {
 // EnqueueControllerOf takes a resource, identifies its controller resource,
 // converts it into a namespace/name string, and passes that to EnqueueKey.
 func (c *Impl) EnqueueControllerOf(obj interface{}) {
-	// TODO(mattmoor): This will not properly handle Delete, which we do
-	// not currently use.  Consider using "cache.DeletedFinalStateUnknown"
-	// to enqueue the last known owner.
 	object, err := meta.Accessor(obj)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -139,6 +146,52 @@ func (c *Impl) EnqueueControllerOf(obj interface{}) {
...
+			if !ok {
+				c.logger.Errorf("Couldn't get object from tombstone %#v", obj)
+				return
+			}
+			object, ok = tombstone.Obj.(metav1.Object)
+			if !ok {
+				c.logger.Errorf("The object that Tombstone contained is not of metav1.Object %#v", obj)
+				return
+			}
+		}
```
mattmoor: Feels like we are going to want our own wrapper for the boilerplate up to this point.

controller/controller.go
```diff
@@ -139,6 +146,52 @@ func (c *Impl) EnqueueControllerOf(obj interface{}) {
...
+
+		labels := object.GetLabels()
+		controllerKey, ok := labels[nameLabel]
+		if !ok {
+			c.logger.Errorf("Object %s/%s does not have a referring name label %s",
+				object.GetNamespace(), object.GetName(), nameLabel)
```
mattmoor: I'd use `Infof` here and below.  I suspect that this would be because we have a mix of objects that are owned by our controller and not (e.g. Deployment).

controller/controller.go
lichuqiang: done

controller/controller.go
lichuqiang: done

PR #147, opened 2018-10-29
Relax the requirement for stackdriver project id for metrics exporter.

akyyy:
When metrics.stackdriver-project-id is not provided and the deployment is in GKE, there is application default credentials for the stackdriver exporter.

This is to address @mattmoor 's comment in https://github.com/knative/serving/pull/2337


metrics/config.go
```diff
@@ -67,10 +67,9 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
 
 	if mc.backendDestination == Stackdriver {
 		sdProj, ok := m[stackdriverProjectIDKey]
-		if !ok || sdProj == "" {
-			return nil, errors.New("For backend stackdriver, metrics.stackdriver-project-id field must exist and cannot be empty")
+		if ok && sdProj != "" {
+			mc.stackdriverProjectID = sdProj
```
mdemirhan: When this is left empty, do OpenCensus pick the current GKE project as the SD project id?

mattmoor: The question is my mind is: how discoverable would that be for users?

Noting that `metadata.ProjectID()` returns an optional `error` (e.g. not running on GCE), I would contend that we should explicitly use it here and surface the error (rejecting the configmap update).

akyyy: @mdemirhan yes, OpenCensus is able to use the current GCP project as the SD project ID in my testing.

@mattmoor  If OpenCensus is not able to use the default account, it'll return an error when creating a metrics exporter: 
errors.New("stackdriver: no project found with application default credentials")"
Is this sufficient?

mattmoor: This error would surface [here](https://github.com/knative/pkg/pull/147/files#diff-0f986afc14030fe191c861028eeb004dR105)?

akyyy: Yes.

akyyy:
Gentle ping. Could you please take a look when you get a chance? After this pr is in, I can update https://github.com/knative/serving/pull/2337.

metrics/config.go
yanweiguo: Maybe add a comment for what will happen in case stackdriverProjectIDKey is not provided and OpenCensus is not able to use the default account.

akyyy: Added a comment

metrics/config.go
```diff
@@ -67,10 +67,9 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
 
 	if mc.backendDestination == Stackdriver {
 		sdProj, ok := m[stackdriverProjectIDKey]
-		if !ok || sdProj == "" {
-			return nil, errors.New("For backend stackdriver, metrics.stackdriver-project-id field must exist and cannot be empty")
+		if ok && sdProj != "" {
```
mdemirhan: We could simply do:
```
mc.stackdriverProjectID = m[stackdriverProjectIDKey]
```
If the project doesnt exist, map will return "" anyway.

PR #149, opened 2018-10-31
requeue keys on error in controller

lichuqiang:
Fixes: https://github.com/knative/serving/issues/2350

Here I also introduce an error wrapper `permanentError`, which allow users to decide to requeue/not requeue a key on certain error(I meant to make it `transientError`, but considering the fact we will have to wrap existing errors in serving controllers that way, I make it the opposite).

/assign @mattmoor @tcnghia

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -37,6 +37,10 @@ import (
 const (
 	falseString = "false"
 	trueString  = "true"
+
+	// maxRetries is the number of times a key will be retried before it is dropped out of the queue.
+	// TODO(lichuqiang): make this configurable?
+	maxRetries = 15
 )
 
 // Reconciler is the interface that controller implementations are expected
@@ -275,6 +279,7 @@ func (c *Impl) processNextWorkItem() bool {
 		// Run Reconcile, passing it the namespace/name string of the
 		// resource to be synced.
 		if err = c.Reconciler.Reconcile(ctx, key); err != nil {
+			c.handleErr(err, key)
 			return fmt.Errorf("error syncing %q: %v", key, err)
 		}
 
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -37,6 +37,10 @@ import (
 const (
 	falseString = "false"
 	trueString  = "true"
+
+	// maxRetries is the number of times a key will be retried before it is dropped out of the queue.
+	// TODO(lichuqiang): make this configurable?
+	maxRetries = 15
```
mattmoor: Hmm, this seems strange for a declarative controller.  Are you aware of any other K8s controllers that do this?  rate limiting and back-off, sure, but giving up seems somewhat antithetical.

controller/controller.go
```diff
@@ -275,6 +279,7 @@ func (c *Impl) processNextWorkItem() bool {
 		// Run Reconcile, passing it the namespace/name string of the
 		// resource to be synced.
 		if err = c.Reconciler.Reconcile(ctx, key); err != nil {
+			c.handleErr(err, key)
```
mattmoor: Based on the comment below, I didn't think requeues were necessary (until `Forget` is explicitly called)

controller/controller.go
lichuqiang: Yeah, some of the k8s controller have this.
Personally I'm fine with both, as the delay would be long enough in the end.

controller/controller.go
lichuqiang: The key point is that it does not behavior as it commented, as I described in knative/serving#2350, the key will not get requeued  automatically, unless we call `AddRateLimited` explicitly.
as a validation, you can scale up the sleep period [here in the test](https://github.com/knative/pkg/blob/master/controller/controller_test.go#L449), and you can see `NumRequeues` did not grow accordingly, which means the key is not indeed requeued and step into the reconcile look again.

lichuqiang: I guess the code here might be originally borrowed from sample-controller in k8s.
The behavior here has been fixed in the functional controllers. but no one noticed this in the sample one I think

controller/controller.go
lichuqiang: removed

controller/controller.go
mattmoor: Cool, thanks for explaining

PR #153, opened 2018-11-01
 Clean up eventing duck types

evankanderson:
Clean up eventing duck types.
* Introduce Addressable to unify delivery ducks.
* Remove Subscribable and Channelable (moved to knative/eventing).
* Mark Sinkable and Targetable as retired in favor of Addressable &
  annotations for Callable.

See https://docs.google.com/document/d/1xu18lprM8EFknqrsyZeNgYX2sLE13hfT2UKlYfdxv1g/edit#
for interface/naming details.

/assign @vaikas-google @scothis @n3wscott 

I selected `status.address.hostname` as the equivalent of `status.domainInternal` previously; this seemed like the right mix of explicit and structured.

We'll need to merge this into `knative/serving` and update the Route and Service, hopefully for 0.2.1 (0.2 will continue to work with the old `status.domainInternal`).


apis/duck/v1alpha1/addressable_types.go
```diff
@@ -0,0 +1,98 @@
...
+	Status AddressStatus `json:"status"`
+}
+
+// AddressStatus shows how we expect folks to embed Addressable in
+// their Status field.
+type AddressStatus struct {
+	Address *Addressable `json:"address,omitempty"`
+}
+
+// In order for Targetable to be Implementable, Target must be Populatable.
```
n3wscott: Targetable, Target --> Addressable, Addressee ?

apis/duck/v1alpha1/addressable_types.go
```diff
@@ -0,0 +1,98 @@
...
+var _ duck.Populatable = (*AddressableType)(nil)
+
+// Ensure Target satisfies apis.Listable
```
n3wscott: AddressableType

apis/duck/v1alpha1/retired_targetable_types.go
```diff
@@ -24,11 +24,15 @@ import (
...
-// Targetable is the schema for the targetable portion of the payload
+// Targetable is an earlier version of the Callable interface.
+// Callable is a higher-level interface which implements Addressable
+// but further promises that the destination may synchronously return
+// response messages in reply to a message.
+//
+// Targetable implementations should instead implement Addressable and
+// include an `eventing.knative.dev/returns=any` annotation.
+
+// Targetable is retired; implement Addressable for now.
```
n3wscott: The language in other parts of the repo is `// Deprecated: Use XYZ` but perhaps this is not deprecated as much as it is deleted?

apis/duck/v1alpha1/addressable_types.go
```diff
@@ -0,0 +1,98 @@
...
+// definition to indicate a destination for message delivery.
+// (Currently, only hostname is supported, and HTTP is implied. In the
+// future, additional schemes may be supported, and path components
+// ala UI may also be supported.)
+
+// Addressable is the schema for the destination information. This is
+// typically stored in the object's `status`, as this information may
+// be generated by the controller.
+type Addressable struct {
+	Hostname string `json:"hostname,omitempty"`
```
vaikas-google: no port, etc.? How come this is not URI?

apis/duck/v1alpha1/legacy_targetable_types.go
```diff
@@ -24,19 +24,19 @@ import (
...
 //
-// Do not use this for anything new, use Targetable
+// Do not use this for anything new, use Addressable
 
-// LegacyTargetable is the old schema for the targetable portion
+// LegacyTargetable is the old schema for the addressable portion
 // of the payload
 //
-// For new resources use Targetable.
+// For new resources use Addressable.
```
vaikas-google: If we have two things that are legacy, let's just get rid of this at least?

apis/duck/v1alpha1/retired_targetable_types.go
vaikas-google: I think in Eventing repo so many things have not been deprecated but outright changed, since these support them, how about we just delete things instead of leave cruft here.

apis/duck/v1alpha1/addressable_types.go
evankanderson: I'm trying to get this into the existing serving for the 0.2.1 cut, so minimizing change. We can add scheme and port later if needed.

vaikas-google: ack

apis/duck/v1alpha1/addressable_types.go
evankanderson: AddressableType.

Cleaned up this comment altogether, since it was simply describing the contracts being verified on the following lines.

apis/duck/v1alpha1/retired_targetable_types.go
evankanderson: I was thinking that I'd update deps in serving and then switch the types, but I think I can just do the `dep` and caller update in one pass.

apis/duck/v1alpha1/legacy_targetable_types.go
evankanderson: I think we want to keep the legacy_ version, since 0.1 only implements the Legacy version. I'd rather ditch the abortive naming that's only in 0.2.0.

apis/duck/v1alpha1/addressable_types.go
```diff
@@ -0,0 +1,96 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status AddressStatus `json:"status"`
+}
+
+// AddressStatus shows how we expect folks to embed Addressable in
+// their Status field.
+type AddressStatus struct {
+	Address *Addressable `json:"address,omitempty"`
```
scothis: 🙇 for not calling the field `addressable`

apis/duck/v1alpha1/legacy_targetable_types.go
scothis: `Flow` still uses domainInternal, not that I expect it to live long, but for the moment it's a thing.

apis/duck/v1alpha1/register.go
```diff
@@ -45,20 +45,14 @@ var (
...
 	scheme.AddKnownTypes(
 		SchemeGroupVersion,
-		&Channel{},
-		(&Channel{}).GetListType(),
 		&KResource{},
 		(&KResource{}).GetListType(),
 		&Generational{},
 		(&Generational{}).GetListType(),
-		&Target{},
-		(&Target{}).GetListType(),
```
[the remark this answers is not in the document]
evankanderson: My very next move was going to be to re-vendor pkg into serving and change Targetable to Addressable.

But, it seems safer to leave Targetable in until I've done that, so let me revert a few things.

evankanderson:
Updated, thanks.

mattmoor:
Thanks, LGTM, but will leave final `/lgtm` to @vaikas-google or @n3wscott

PR #161, opened 2018-11-02
Update the ConditionSet to support non-terminal conditions.

mattmoor:
This change updates the `ConditionSet` datatype to treat the set of conditions that it is initially supplied with as the set of "terminal" conditions, where as before:
1. If all `True`, results in `Ready=True`
2. If any `False`, results in `Ready=False`
3. Otherwise, `Ready=Unknown`

However, in additional to this initial "terminal" set, other conditions may be set that are "non-terminal" and have no influence over the "happy" condition (e.g. `Ready`).

Related to: https://github.com/knative/serving/issues/2394

I'm putting a `/hold` on this until we discuss at next week's API WG.

### DIFF
--- a/apis/duck/v1alpha1/condition_set.go
+++ b/apis/duck/v1alpha1/condition_set.go
@@ -211,13 +211,30 @@ func (r conditionsImpl) SetCondition(new Condition) {
 	r.accessor.SetConditions(conditions)
 }
 
+func (r conditionsImpl) isTerminal(t ConditionType) bool {
+	for _, cond := range append(r.dependents, r.happy) {
+		if cond == t {
+			return true
+		}
+	}
+	return false
+}
+
+func (r conditionsImpl) severity(t ConditionType) ConditionSeverity {
+	if r.isTerminal(t) {
+		return ConditionSeverityError
+	}
+	return ConditionSeverityInfo
+}
+
 // MarkTrue sets the status of t to true, and then marks the happy condition to
 // true if all other dependents are also true.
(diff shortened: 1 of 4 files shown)


apis/duck/v1alpha1/condition_set.go
```diff
@@ -257,23 +260,32 @@ func (r conditionsImpl) MarkUnknown(t ConditionType, reason, messageFormat strin
 			}
 			return
 		}
+		if cond == t {
+			isDependent = true
+		}
 	}
 
-	// set the happy condition
-	r.SetCondition(Condition{
```
n3wscott: I was so close!

apis/duck/v1alpha1/condition_set.go
```diff
@@ -257,23 +260,32 @@ func (r conditionsImpl) MarkUnknown(t ConditionType, reason, messageFormat strin
...
 // MarkFalse sets the status of t and the happy condition to False.
 func (r conditionsImpl) MarkFalse(t ConditionType, reason, messageFormat string, messageA ...interface{}) {
-	for _, t := range []ConditionType{
-		t,
-		r.happy,
-	} {
+	types := []ConditionType{t}
+	for _, cond := range r.dependents {
+		if cond == t {
+			types = append(types, r.happy)
```
n3wscott: good catch!

mattmoor:
I've updated this with `Severity` and some basic management thereof.

/hold

I am putting a hold on this because I'd like to coordinate this going in with an update to `knative/serving`, since I suspect the table tests will explode, and I'd like to get in front of that.

PR #166, opened 2018-11-13
Update EnqueueLabelOf to handle resources in same namespace

lichuqiang:
Update the `EnqueueLabelOf` func to be able to replace funcs like `EnqueueEndpointsKPA`, `EnqueueEndpointsRevision`

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -136,11 +136,11 @@ func (c *Impl) EnqueueControllerOf(obj interface{}) {
 	}
 }
 
-// EnqueueLabelOf returns with an Enqueue func that takes a resource,
-// identifies its controller resource through given namespace and name labels,
-// converts it into a namespace/name string, and passes that to EnqueueKey.
-// It would pass through given object's namespace if an empty namespace label is specified.
-func (c *Impl) EnqueueLabelOf(namespaceLabel, nameLabel string) func(obj interface{}) {
+// EnqueueLabelOfNamespaceScopedResource returns with an Enqueue func that
+// takes a resource, identifies its controller resource through given namespace
+// and name labels, converts it into a namespace/name string, and passes that
+// to EnqueueKey. The controller resource must be of namespace-scoped.
+func (c *Impl) EnqueueLabelOfNamespaceScopedResource(namespaceLabel, nameLabel string) func(obj interface{}) {
 	return func(obj interface{}) {
 		object, err := kmeta.DeletionHandlingAccessor(obj)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -165,10 +164,18 @@ func (c *Impl) EnqueueLabelOf(namespaceLabel, nameLabel string) func(obj interfa
...
-			controllerKey = fmt.Sprintf("%s/%s", controllerNamespace, controllerKey)
-		}
+			c.EnqueueKey(fmt.Sprintf("%s/%s", controllerNamespace, controllerKey))
+			return
 
-		c.EnqueueKey(controllerKey)
+		}
+		// Pass through namespace of the object itself if no namespace label specified.
+		// This is for the scenario that object and the parent resource are of same namespace,
+		// e.g. to enqueue the reivison of an endpoint.
```
mattmoor: typo: revision

controller/controller.go
```diff
@@ -165,10 +164,18 @@ func (c *Impl) EnqueueLabelOf(namespaceLabel, nameLabel string) func(obj interfa
...
+		// Note that for objects whose parent resource are cluster-scoped, here it adds the
+		// namespace part that we don't indeed need. Callers should be able to call
```
mattmoor: Perhaps we should treat this as the "namespaced" variety, and have a separate version for "cluster"-scoped?

I would expect that this would almost certainly have to be used with a GroupVersionKind-filtered function, so the logic setting it up should know which they are dealing with.

controller/controller.go
```diff
@@ -165,7 +164,36 @@ func (c *Impl) EnqueueLabelOf(namespaceLabel, nameLabel string) func(obj interfa
...
+			c.EnqueueKey(fmt.Sprintf("%s/%s", controllerNamespace, controllerKey))
+			return
+		}
+
+		// Pass through namespace of the object itself if no namespace label specified.
+		// This is for the scenario that object and the parent resource are of same namespace,
+		// e.g. to enqueue the revision of an endpoint.
+		c.EnqueueKey(fmt.Sprintf("%s/%s", object.GetNamespace(), controllerKey))
+	}
+}
```
mattmoor: Could you leave a `// DEPRECATED ... ` alias with the name `EnqueueLabelOf` until we migrate the downstream repos?  Otherwise this LGTM (though the method name is a mouthful).

lichuqiang: That would be friendly to backward compatibility, while the shortcoming is that we'll need to cleanup it later.

To avoid conflict, I was intending to update the upstream usage once this is merged.
But let's leave the deprecated one since you feel that better too.

lichuqiang: done

controller/controller.go
```diff
@@ -141,6 +141,10 @@ func (c *Impl) EnqueueControllerOf(obj interface{}) {
 // converts it into a namespace/name string, and passes that to EnqueueKey.
 // Callers should pass in an empty string as namespace label key for obj
 // whose controller is of cluster-scoped resource.
+//
+// Deprecated: this will be removed once all of its upstream usage get removed.
+// Use "EnqueueLabelOfNamespaceScopedResource" or "EnqueueLabelOfClusterScopedResource"
+// instead.
```
mattmoor: I meant just delegate to `EnqueueLabelOfNamespaceScopeResource` here.

mattmoor: Let's go with #2, and rip the bandaid off, sorry for the churn.

lichuqiang: done

PR #167, opened 2018-11-13
Add ErrOutOfBoundsValue for value boundary validation

k4leung4:


apis/field_error.go
```diff
@@ -335,3 +335,12 @@ func ErrInvalidKeyName(value, fieldPath string, details ...string) *FieldError {
 		Details: strings.Join(details, ", "),
 	}
 }
+
+// ErrOutOFBoundsValue constructs a FieldError for a field that has received an
+// out of bound value.
+func ErrOutOfBoundsValue(value, lower, upper float64, fieldPath string) *FieldError {
```
jonjohnsonjr: I think this name is a bit too general for the signature being specifically float64.

k4leung4: change it to take string.
ptal

apis/field_error.go
```diff
@@ -335,3 +335,12 @@ func ErrInvalidKeyName(value, fieldPath string, details ...string) *FieldError {
 		Details: strings.Join(details, ", "),
 	}
 }
+
+// ErrOutOFBoundsValue constructs a FieldError for a field that has received an
+// out of bound value.
+func ErrOutOfBoundsValue(value, lower, upper, fieldPath string) *FieldError {
+	return &FieldError{
+		Message: fmt.Sprintf("expected %q <= %q <= %q", lower, value, upper),
```
mattmoor: I'd drop the quotes, it makes it look funny.

k4leung4: removed.

PR #169, opened 2018-11-21
Quieting controller log

lichuqiang:
Fixes: https://github.com/knative/serving/issues/2529

/assign @mattmoor


lichuqiang:
I've seen flake in reconciler test. I think it's [here when waiting for the item to get processed](https://github.com/knative/pkg/blob/master/controller/controller_test.go#L543), it closed the channel before the instance got processed as expect.

Maybe we should consider a more reliable mechanism to replace `time.Sleep`

PR #170, opened 2018-11-21
update gateway

lichuqiang:
It's dependency of https://github.com/knative/serving/pull/2434

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -43,7 +43,7 @@ const (
 	requestInterval = 1 * time.Second
 	requestTimeout  = 5 * time.Minute
 	// TODO(tcnghia): These probably shouldn't be hard-coded here?
-	ingressName      = "knative-ingressgateway"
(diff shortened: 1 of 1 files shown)


lichuqiang:
/hold
It would break compatibility. We should not merge this until knative/serving#2434 is ready.

lichuqiang:
I update the spoof client to:
1. visit with knative-ingressgateway if it exists
2. fallback to istio-ingressgateway otherwise.

/cc @tcnghia @jonjohnsonjr Can you help clarify if this can work around the upgrade test issue?

tcnghia:
@lichuqiang thanks for making the change.  I think this would work.

If you have a remote with your `knative/pkg` version you can update your `Gopkg.toml` to something like this to try testing against your remote
```
[[override]]
  name = "github.com/knative/pkg"
  source = "https://github.com/lichuqiang/pkg.git"
  revision = "your-commit-id"
```

PR #171, opened 2018-11-26
Run gofmt

mattmoor-sockpuppet:
Produced via: `gofmt -s -w $(find -name '*.go' | grep -v vendor)`


mattmoor:
@vdemeester as discussed in slack, I'd love to leverage Github's suggested-edit functionality for this (as I do in my typo-bot), but right now it's limited to single-line suggestions.

All you really have to do is comment with:
<pre>
```suggestion
replacement text
```
</pre>

Apparently this was one of the top requests, so **hopefully** we'll see it soon.

I'd be fine with adding a check that this doesn't regress as well, but in the limit suggested edits >> CI failures due to linting.

/approve

PR #172, opened 2018-11-26
encoding: expose HeaderExtensionPrefix

srvaroa:
So that the header prefix can be used in downstream dependencies.

See a concrete example here: https://github.com/knative/eventing-sources/pull/129


matzew:
/lgtm 

I like the exposure of the `CE-X` extention header prefix!

PR #173, opened 2018-11-26
Make controller more responsive to events

alculquicondor:
Only delay the events when errors occur.

Also, clean up unnecessary code and misleading comments.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -25,7 +25,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
@@ -201,7 +200,7 @@ func (c *Impl) EnqueueLabelOfClusterScopedResource(nameLabel string) func(obj in
 
 // EnqueueKey takes a namespace/name string and puts it onto the work queue.
 func (c *Impl) EnqueueKey(key string) {
-	c.WorkQueue.AddRateLimited(key)
+	c.WorkQueue.Add(key)
 }
 
 // Run starts the controller's worker threads, the number of which is threadiness.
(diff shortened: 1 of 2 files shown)


alculquicondor:
friendly ping

mattmoor:
@alculquicondor I've spent some time digging into this, and I believe I've identified our bottleneck.

I believe that the problem we're facing is a rate-limit built into the Kubernetes API clients to avoid controllers DoSing the API server.  I'm guessing that these limits are reasonably chosen for a single controller and we are running a half dozen using shared clients (and therefore likely shared limits!).

If I bump the `QPS` and `Burst` settings on the [`rest.Config`](https://godoc.org/k8s.io/client-go/rest#Config) object (otherwise unspecified, so they pick up defaults) then I see things become considerably more performant.

Given this finding, I'm actually growing skeptical that the changes here really move the needle.  I'd really like to see us have proper benchmarks for this scenario (I'm just using a trivial shell script for now).

My fix exacerbates a bug, which makes measurement hard.  I'm going to look at fixing that bug, and then I'll try to circle back with some results.

mattmoor:
Time to 50 ready services

w/ essentially HEAD
Start: 22:22:59
End:   22:25:54

w/ #173
Start: 22:31:39
End:   22:34:41

w/ bumped QPS/Burst=20/50
Start: 22:41:49
End:   22:42:57

w/ bumped QPS/Burst=20/50, 10 workers
Start: 22:47:34
End:   22:48:41

w/ #173, 10 workers, bumped QPS/Burst=20/50
Start: 22:17:02
End:   22:18:25

---
In summary, this change doesn't make an observable difference for me at HEAD (or stacked with my changes).

The only thing that seems to move the dial is bumping the QPS/Burst used by our controller pod.  This takes the 3 min @ HEAD down to about 70 sec.

Here are the two lines I added to `./cmd/controller/main.go`:
```go
       // We run a good number of controllers, so bump the defaults.
       cfg.QPS = 20
       cfg.Burst = 50
```

FWIW, I tried raising these even higher and got comparable timings, so it is unlikely that we're being throttled at the above limits.

mattmoor:
Time to 100 ready services:

w/ essentially HEAD
Start: Sat Dec  1 23:34:24 UTC 2018
End:   Sat Dec  1 23:40:14 UTC 2018

w/ bumped QPS/Burst=20/50
Start: Sat Dec  1 23:27:44 UTC 2018
End:   Sat Dec  1 23:30:22 UTC 2018


So this makes things pretty consistently ~2x faster for bursty creations.

PR #182, opened 2018-12-01
Use different reporting periods based on the metrics backend

mdemirhan:
Use different reporting periods based on the metrics backend. Original implementation was setting this value to 1 min to accommodate the minimum allowed value for stackdriver, but that is a sub optimal behavior for Prometheus. The change also allows operator to override the default values.


metrics/config.go
```diff
@@ -72,6 +78,25 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
...
 	}
 
+	// If reporting period is specified, use the value from the configuration.
+	// If not, set a default value based on the selected backend.
+	// Each exporter makes different promises about what the lowest supported
+	// reporting period is. For Stackdriver, this value is 1 minute.
+	// For Prometheus, we will use a lower value since the exporter doesn't
+	// push anything but just responds to pull requests, and shorter durations
+	// do not really hurt the performance and we rely on the scraping configuration.
+	if repStr, ok := m[reportingPeriodKey]; ok && len(repStr) != 0 {
```
yanweiguo: Change `len(reqStr) != 0` to `reqStr != ""`?

metrics/config.go
```diff
@@ -72,6 +78,25 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
...
+		if repInt, err := strconv.Atoi(repStr); err == nil {
+			mc.reportingPeriod = time.Duration(repInt) * time.Second
+		} else {
+			return nil, fmt.Errorf("Invalid reporting-period-seconds value \"%s\"", repStr)
+		}
+	} else if mc.backendDestination == Stackdriver {
+		mc.reportingPeriod = 60 * time.Second
+	} else if mc.backendDestination == Prometheus {
```
yanweiguo: What happen if  `backendDestination` is neither `Stackdriver` nor `Prometheus`.

metrics/exporter.go
```diff
@@ -137,7 +136,12 @@ func setCurMetricsExporterAndConfig(e view.Exporter, c *metricsConfig) {
 	metricsMux.Lock()
 	defer metricsMux.Unlock()
 	view.RegisterExporter(e)
-	view.SetReportingPeriod(60 * time.Second)
+	if c != nil {
+		view.SetReportingPeriod(c.reportingPeriod)
```
yanweiguo: Could `c.reportingPeriod` be `nil`?

metrics/config.go
mdemirhan: It will return an error in the if block above (line 71).

metrics/exporter.go
mdemirhan: c.reportingPeriod is time.Duration (type Duration int64) - so the default value is 0 and setting reporting period to 0 is same as leaving it as its default value in Opencensus.

mdemirhan:
@yanweiguo This is ready for another look.

controller/stats_reporter.go
```diff
@@ -123,7 +123,7 @@ func (r *reporter) ReportReconcile(duration time.Duration, key, success string)
 	}
 
 	stats.Record(ctx, reconcileCountStat.M(1))
-	stats.Record(ctx, reconcileLatencyStat.M(int64(duration)))
+	stats.Record(ctx, reconcileLatencyStat.M(int64(duration/time.Millisecond)))
```
yanweiguo: Is this change relative?

mdemirhan: We were accidentally reporting nanoseconds (duration is in terms of nanoseconds). We wanted to report milliseconds instead. This change fixes that behavior.

controller/controller.go
```diff
@@ -267,13 +267,14 @@ func (c *Impl) processNextWorkItem() bool {
 	// resource to be synced.
 	if err = c.Reconciler.Reconcile(ctx, key); err != nil {
 		c.handleErr(err, key)
+		logger.Errorf("Reconcile failed. Time taken: %d ms.", time.Now().Sub(startTime)/time.Millisecond)
```
mattmoor: note that if you just pass the duration with `%v` it will serialize with units e.g. `3s`

PR #191, opened 2018-12-05
This exempts defaulting from the rules governing field immutability.

mattmoor:
Immutable fields with default values may now be changed iff they change is to populate their default value.  This is to support defaulting in the scenario where an object was created long ago and a new field (with a default!) is added.  When controllers attempt to mutate the object status today, this would create a webhook rejection!  With this change, we compare against a freshly defaulted "old" object to exclude newly defaulted fields from the immutability check.

We saw this in knative/serving for the newly added TimeoutSeconds field in Revision (otherwise immutable), which I believe it leading to upgrade testing flakes since post-upgrade Revision status updates will fail.

### DIFF
--- a/apis/duck/patch.go
+++ b/apis/duck/patch.go
@@ -18,6 +18,7 @@ package duck
 
 import (
 	"encoding/json"
+	"sort"
 
 	jsonmergepatch "github.com/evanphx/json-patch"
 	"github.com/mattbaird/jsonpatch"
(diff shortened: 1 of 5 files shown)


vaikas-google:
Looks like we also introduced ordering in the patch (which I think is a fine change, just not obvious from looking at the title of the PR), it might have been good to have that in the description, but it got merged already :)

mattmoor:
@vaikas-google Oops, sorry I meant to.  Yeah, this was driving me nuts as my testing was super flaky as a result.  Sorry for not calling it out.

PR #194, opened 2018-12-06
Some clean ups in the FieldError code and its tests.

vagababov:
- rename FieldError.getNormalizedError to be normalized:
  - Go kind of discourages GetA() vs A()
  - it's an error, so normalizing itself is also an error, so
    normalizedError() is a tautology at this point
- flatten was doing a[x]=fmt.Sprintf("%s%s", a[x],y), which is a very
  expensive substitution for a[x]+=y
- some minor test cleanups (fix strings, streamline `got, want:=`
  pattern, etc.

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -133,8 +133,8 @@ func (fe *FieldError) isEmpty() bool {
 	return fe.Message == "" && fe.Details == "" && len(fe.errors) == 0 && len(fe.Paths) == 0
 }
 
-func (fe *FieldError) getNormalizedErrors() []FieldError {
-	// in case we call getNormalizedErrors on a nil object, return just an empty
+func (fe *FieldError) normalized() []FieldError {
+	// In case we call normalized on a nil object, return just an empty
 	// list. This can happen when .Error() is called on a nil object.
 	if fe == nil {
 		return []FieldError(nil)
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -133,34 +133,36 @@ func (fe *FieldError) isEmpty() bool {
...
 	if fe == nil {
-		return []FieldError(nil)
+		return []*FieldError(nil)
 	}
-	var errors []FieldError
-	// if this FieldError is a leaf,
+
+	// Allocate errors with at least as many objects as we'll get on the first pass.
+	errors := make([]*FieldError, 0, len(fe.errors)+1)
+	// If this FieldError is a leaf,
```
vaikas-google: don't leave me hanging, what happens?

apis/field_error.go
```diff
@@ -235,27 +237,26 @@ func containsString(slice []string, s string) bool {
 // FiledErrors. FieldErrors have their Paths combined (and de-duped) if their
```
vaikas-google: while here, FiledErrors -> FieldErrors?

apis/field_error.go
vagababov: well :) it was like this. I'll change to ...

apis/field_error.go
vagababov: Eh, didn't notice this one.
Unforts, the PR automerged, so I'll send a new one.

PR #196, opened 2018-12-07
Fix some comments about code comments left from the previous PR.

vagababov:


vagababov:
The merge confglicts should be merged by now

PR #197, opened 2018-12-07
Add kmp, a safe wrapper for google/go-cmp

jonjohnsonjr:
* Returns an error instead of panicking.
* Holds common options that we use.

See: https://github.com/knative/serving/issues/2622

### DIFF
--- a/kcmp/diff.go
+++ b/kcmp/diff.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


jonjohnsonjr:
@mattmoor this is what I had in mind. I think we'll eventually want to get rid of any non-test usage of cmp, but we can use this in the meantime.

jonjohnsonjr:
Great naming suggestions from @tcnghia:
* `kmp` instead of `cmp`
* `SafeDiff` instead of `Diff`

Happy to 🚲 🏠 but I like those 😄

vaikas-google:
would it be useful to try to get these upstreamed?

jonjohnsonjr:
I suspect they would be amenable to a non-panicky variant of `Diff`.

I just found [this comment](https://github.com/google/go-cmp/pull/97#issuecomment-438820774) that proposes a new option: `cmpopts.AllowUnexported(cmpopts.PackagePath("k8s.io"))`. If both of those would land, we could ditch this package.

@dsnet any thoughts on returning an error instead of panicking? Also 👍 to a `PackagePath` option.

PR #200, opened 2018-12-11
Bump "safe" dependencies

vdemeester:
This bump some "safe" dependencies atomically (as 1 update per commit). By "safe" dependency, I means it tried to update dependency that are not constrained, nor overridden, and that do not track `master`.

This has been done using [`sec`](https://github.com/vdemeester/sec), which means:
- It tried to update the dependency
- Ran tests to validate it doesn't break (at least the build) (`go test ./...` or `go test ./pkg/...`, …).
- Commit if the test command was successful.

`sec` is really experimental, but my end-goal is to have some kind of bot that could update those safe deps for us :angel:.
Also, this *never* update to `latest` either (if `latest` is the latest release, even if it's a major version bump). It's only updating minor or fixes versions.

/assign @mattmoor


evankanderson:
Looks like this will need a merge (sorry, holidays were a busy time)!

/approve

PR #206, opened 2018-12-19
Add durable websocket client from serving.

markusthoemmes:
The durable websocket client has been proven valuable in other projects too (namely an eventing websocket source), thus moving it here makes it possible to broaden its scope and make it accessible for all projects more easily.

It's coming from https://github.com/knative/serving/tree/master/pkg/websocket and contains the changes I proposed in https://github.com/knative/serving/pull/2635

/cc @josephburnett


markusthoemmes:
/assign @mattmoor 

If you'd do me the honors 🙏

PR #207, opened 2019-01-02
update k8s version

zouyee:
bump apimachinery、api、code-generator、client-golibrary to 1.13.1


zouyee:
https://github.com/knative/build/pull/503 
We need to use the new version of the `apimachinery`library, so we bump it, if there is a correspondence and compatibility matrix doc, it might be better.

mattmoor:
cc @pmorie who may have more firsthand knowledge here.

I think that generally picking up 1.12 would mean dropping support for 1.10 because Kubernetes generally targets supporting +/-1 version of compatibility.

1.13 would mean dropping support for 1.11, which is the latest GA version on both GKE and EKS IIUC.

If we merge CRD sub-resource support in 0.3, I could see us move to 1.12 since CRD sub-resource support when beta in 1.11, but until we drop support for 1.10, even moving to 1.12 seems bad.

PR #210, opened 2019-01-07
Change Stackdriver resource type for metrics

yanweiguo:
`knative_revision` is the resource type will be used for knative.

Bonus: Add `os.Exit()` to `TestMain` and fix already broken tests.


metrics/exporter.go
```diff
@@ -64,13 +64,54 @@ func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) error
...
+			NamespaceName:     getResourceLabelValue(metricskey.LabelNamespaceName, tags),
+			ServiceName:       getResourceLabelValue(metricskey.LabelServiceName, tags),
+			ConfigurationName: getResourceLabelValue(metricskey.LabelConfigurationName, tags),
+			RevisionName:      getResourceLabelValue(metricskey.LabelRevisionName, tags),
+		}
+
+		var newTags []tag.Tag
+		for _, t := range tags {
+			// Keep the metrics labels that are not resource labels
+			fmt.Println(t.Key.Name() + ":" + t.Value)
```
mdemirhan: We probably should get rid of this.

metrics/exporter.go
```diff
@@ -64,13 +64,54 @@ func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) error
...
+			if _, ok := metricskey.KnativeRevisionLabels[t.Key.Name()]; !ok {
+				newTags = append(newTags, t)
+			}
+		}
+
+		return newTags, kr
+	}
+}
+
+func getResourceLabelValue(key string, tags []tag.Tag) string {
```
mdemirhan: I wonder if it would be more efficient to get all these four values in one iteration. something like:
```
func updateResourceLabels(kr *KnativeRevision, tags []tag.Tag) {
	for _, t := range tags {
                switch t.Key,.Name() {
                case metricskey.LabelNamespaceName:
                    kr.NamespaceName = t.Value
                case metricskey.LabelRevisionName:
                    kr.RevisionName = t.Value
                case ...:
...
                }
	}
}
```

metrics/exporter.go
```diff
@@ -111,6 +152,29 @@ func resetCurPromSrv() {
...
+	}
+}
+
+func setMonitoredResourceFunc(config *metricsConfig) {
+	metricsMux.Lock()
+	defer metricsMux.Unlock()
+	if getMonitoredResourceFunc == nil {
+		gm := retrieveGCPMetadata()
+		metricsPrefix := config.domain + "/" + config.component
+		fmt.Println("metrics prefix", metricsPrefix)
```
mdemirhan: We should use the proper logger here. not fmt.Println.

metrics/exporter.go
yanweiguo: Done

metrics/exporter.go
yanweiguo: Done.

metrics/exporter.go
yanweiguo: Changed to use a map since we need to handle key missing case.

PR #211, opened 2019-01-08
Add a utility to check against a minimum required version number.

markusthoemmes:
Knative meanwhile relies on at least Kubernetes 1.11 and I've been bitten by this requirement at least once. This utility can be built into each of our controllers, webhooks etc. to notify/crash if the version is not compatible.

Initially proposed at https://github.com/knative/build/pull/503.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Knative Authors
+Copyright 2019 The Knative Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/version/version.go
+++ b/version/version.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2018 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


version/version.go
```diff
@@ -0,0 +1,51 @@
...
+package version
+
+import (
+	"fmt"
+
+	"github.com/rogpeppe/go-internal/semver"
+	"k8s.io/apimachinery/pkg/version"
+)
+
+type serverVersion interface {
```
ImJasonH: This interface should be exported otherwise callers of `CheckMinimumVersion` won't know what's expected of them.

markusthoemmes: Good one, fixed.

version/version.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2018 The Knative Authors
```
vaikas-google: nit, apparently there was New Years :)

version/version_test.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2018 The Knative Authors
```
vaikas-google: ditto

markusthoemmes:
@vaikas-google fair enough! Fixed the headers.

PR #216, opened 2019-01-09
Implement version check without dependence on go-internal.

markusthoemmes:
The released version of go-internal that we transitively depend on does not contain a LICENSE file. Therefore I cannot include the currently implemented version into knative/serving for example, because license checks will fail.

Options are:
- Override the constraint for go-internal to a never commit. Felt unclean to me.
- Use a different package to parse SemVer. I tried https://github.com/blang/semver, it worked but it doesn't parse `v` prepended versions, so I had to sanitize the output of `GitVersion` from the Kubernetes server.

I, therefore, opted to implement it without any dependency on a SemVer tool and use the `Major`, `Minor` fields instead and compared manually.

I also forgot to run update-deps the last time round, so here we go.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -995,7 +995,9 @@
   analyzer-name = "dep"
   analyzer-version = 1
   input-imports = [
+    "cloud.google.com/go/compute/metadata",
     "contrib.go.opencensus.io/exporter/stackdriver",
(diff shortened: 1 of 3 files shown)


markusthoemmes:
/hold

trying a different route to fix the license checking. Might make this obsolete.

PR #217, opened 2019-01-09
Fix license check for presubmit tests.

markusthoemmes:
While licenses are indeed checked during the test run, and an error is printed, the test does not fail as it should. This fixes that condition.


markusthoemmes:
This confirms the check now correctly fails with:

```
Error identifying licenses for transitive dependencies: unable to find license for "github.com/knative/pkg/vendor/github.com/rogpeppe/go-internal/modfile"
```

hack/update-deps.sh
```diff
@@ -25,6 +25,8 @@ cd ${REPO_ROOT_DIR}
 # Ensure we have everything we need under vendor/
 dep ensure
 
+check_licenses $(go list ./...)
```
mattmoor: I don't think this is particularly fast (not that `dep` is), and we already run it in presubmit tests...

markusthoemmes: Took it out of here for the presubmit test to catch.

PR #218, opened 2019-01-09
Read metrics backend from environment variable

yanweiguo:
Propose:

1. Set metrics backend to the value from an environment variable first. Then override it to the value from config map, if any. This gives the ability to set metrics backend when rolling out a new knative serving components while keep existing config map unchanged.

1. Do not panic the process if the backend is not supported. Activator and autoscaler should continute to work without emitting metrics to any backend.

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -70,8 +70,7 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
 		backend = string(Prometheus)
 	}
 	// Override backend if it is setting in config map.
-	backendFromConfig, ok := m[backendDestinationKey]
-	if ok {
+	if backendFromConfig, ok := m[backendDestinationKey]; ok {
 		backend = backendFromConfig
 	}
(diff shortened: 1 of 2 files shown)


metrics/config.go
```diff
@@ -60,10 +63,17 @@ type metricsConfig struct {
...
-	backend, ok := m[backendDestinationKey]
-	if !ok {
+	// Read backend setting from environment variable first
+	backend := os.Getenv(defaultBackendEnvName)
+	if backend == "" {
+		// Use Prometheus if DEFAULT_METRICS_BACKEND does not exist or is empty
 		backend = string(Prometheus)
 	}
+	// Override backend if it is setting in config map.
+	backendFromConfig, ok := m[backendDestinationKey]
```
vagababov: join two lines if bfc, ok := ...;ok {...}

yanweiguo: Done.

PR #220, opened 2019-01-09
🦆 Duck Typing - add a ConformsToType helper

dprotaso:
Unlike `VerifyType`, `ConformsToType` will return the following:
- an error when any marshalling/unmarshalling fails
- false when the concrete type does not implement the duck type
- true when the concrete type implements the duck type

### DIFF
--- a/apis/duck/verify.go
+++ b/apis/duck/verify.go
@@ -56,23 +56,8 @@ func VerifyType(instance interface{}, iface Implementable) error {
 	// that we will compare at the end.
 	input, output := iface.GetFullType(), iface.GetFullType()
 
-	// Populate our input resource with values we will roundtrip.
-	input.Populate()
-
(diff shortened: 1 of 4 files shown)


apis/duck/verify.go
```diff
@@ -56,30 +56,53 @@ func VerifyType(instance interface{}, iface Implementable) error {
 	// that we will compare at the end.
 	input, output := iface.GetFullType(), iface.GetFullType()
 
+	if err := roundTrip(instance, input, output); err != nil {
+		return err
+	}
+
+	// Now verify that we were able to roundtrip all of our fields through the type
+	// we are checking.
+	if diff := cmp.Diff(input, output); diff != "" {
```
mattmoor: Can you use Jon's safe version of this under `kmp`?

dprotaso: Sure - was wondering if it didn’t use it for a reason

PR #224, opened 2019-01-11
Update test-infra

adrcunha:
Bring in the latest features and fixes.

Bonus: use default build, unit and integration test runners.


adrcunha:
/hold

There's one package that's giving me trouble with the license file.

PR #229, opened 2019-01-16
Conditions implementation no longer mutates the underlying dependents array

dprotaso:
Addresses: https://github.com/knative/serving/issues/2849

This is because the condition set in serving is a global variable.

### DIFF
--- a/apis/duck/v1alpha1/condition_set.go
+++ b/apis/duck/v1alpha1/condition_set.go
@@ -212,11 +212,16 @@ func (r conditionsImpl) SetCondition(new Condition) {
 }
 
 func (r conditionsImpl) isTerminal(t ConditionType) bool {
-	for _, cond := range append(r.dependents, r.happy) {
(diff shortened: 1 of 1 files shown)


apis/duck/v1alpha1/condition_set.go
```diff
@@ -212,11 +212,16 @@ func (r conditionsImpl) SetCondition(new Condition) {
 }
 
 func (r conditionsImpl) isTerminal(t ConditionType) bool {
-	for _, cond := range append(r.dependents, r.happy) {
+	for _, cond := range r.dependents {
```
vaikas-google: side note. I think the name of this function is confusing... not your doing...

I thought append returned the new slice so is this really any different from what was there before?

vaikas-google: oops, I started reviewing, went back to the golang docs, and re-educated myself about append happiness, and forgot to get rid of this comment :)

PR #231, opened 2019-01-17
Fix another dependents mutation in the conditions.

markusthoemmes:
Hopefully finally addresses: https://github.com/knative/serving/issues/2849

Similar to #229.


dprotaso:
Thanks -  I missed this one.

PR #233, opened 2019-01-18
Grep fix formatting issues

vagababov:
Cleanup the formatting function usage.

/cc @vaikas-google

### DIFF
--- a/cloudevents/handler.go
+++ b/cloudevents/handler.go
@@ -19,6 +19,7 @@ package cloudevents
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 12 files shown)


cloudevents/handler.go
```diff
@@ -114,7 +115,7 @@ func validateOutParamSignature(fnType reflect.Type) error {
 // of allowed types. If successful, returns the expected in-param type, otherwise panics.
 func validateFunction(fnType reflect.Type) errAndHandler {
 	if fnType.Kind() != reflect.Func {
-		return &failedHandler{err: fmt.Errorf("Must pass a function to handle events")}
+		return &failedHandler{err: errors.New("Must pass a function to handle events")}
```
vaikas-google: since you're cleaning these up (Thanks!), would it be useful to fix these to also conform to 'errors should not be capitalized')? 
Don't mean do all in this PR, but like this one since you're touching it, and in zipkin util?

https://github.com/golang/go/wiki/Errors

I'm fine with this as is, this is bit of a nit, but just wanted to throw it in here...

vagababov:
Can I also get /lgtm, plz? :roll_eyes:

cloudevents/handler.go
vagababov: Done.

PR #236, opened 2019-01-22
Move crd checks from serving in to test

greghaynes:
These are generally usable test cases for K8s and as such should live in
pkg/test

### DIFF
--- a/test/kube_checks.go
+++ b/test/kube_checks.go
@@ -29,6 +29,7 @@ import (
 	apiv1beta1 "k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	k8styped "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


srinivashegde86:
thanks for doing this.

/lgtm
/approve

test/kube_checks.go
```diff
@@ -73,3 +74,15 @@ func WaitForPodListState(client *KubeClient, inState func(p *corev1.PodList) (bo
 		return inState(p)
 	})
 }
+
+// GetConfigMap gets the knative serving config map.
+func GetConfigMap(client *KubeClient) k8styped.ConfigMapInterface {
+	return client.Kube.CoreV1().ConfigMaps("knative-serving")
```
dprotaso: Is there a reason why is the namespace hardcoded?

And does this helper apply to anything beyond `knative-serving`?

dprotaso:
Also - https://github.com/knative/serving/issues/2703

test/kube_checks.go
greghaynes: good point: https://github.com/knative/pkg/pull/237

PR #239, opened 2019-01-24
Add 'RetryingRouteUpdate' which retries conditions around route creation/update.

markusthoemmes:
Some of our tests only retry 404s, some also 503s and it seems fairly random which do which. This adds a helper method to retry conditions around the creation/update of routes specifically and keeps them as narrow as possible.

See https://github.com/knative/serving/pull/2986 for the planned knative/serving change.


markusthoemmes:
Closing since we decided to continue to dig before retrying.

PR #241, opened 2019-01-25
Add CloudEvents Client to sdk

n3wscott:
Adds a new client and builder to cloudevents and documents how to use them.

### DIFF
--- a/cloudevents/README.md
+++ b/cloudevents/README.md
@@ -113,17 +113,17 @@ func main() {
 
 ### Binary
 
-Changes to the producer code required to leverage binary request format:
+This is default, but to leverage binary request format:
 
 ```go
-req, err := cloudevents.Binary.NewRequest(target, data, ctx)
+c.Builder.Encoding = cloudevents.BinaryEncoding
 
 ```
 
 ### Structured
 
-Changes to the producer code to leverage structured request format:
+To leverage structured request format:
 
 ```go
-req, err := cloudevents.Structured.NewRequest(target, data, ctx)
+c.Builder.Encoding = cloudevents.StructuredEncoding
 ```
--- a/cloudevents/README.md
+++ b/cloudevents/README.md
@@ -116,6 +116,12 @@ func main() {
 This is default, but to leverage binary request format:
 
 ```go
+
+c := cloudevents.NewClient(
+    "dev.knative.cloudevent.example",
+    "https://github.com/knative/pkg#cloudevents-example",
+    "http://localhost:8080",
+)
 c.Builder.Encoding = cloudevents.BinaryEncoding
 
 ```
@@ -125,5 +131,11 @@ c.Builder.Encoding = cloudevents.BinaryEncoding
 To leverage structured request format:
 
 ```go
+
+c := cloudevents.NewClient(
+    "dev.knative.cloudevent.example",
+    "https://github.com/knative/pkg#cloudevents-example",
+    "http://localhost:8080",
+)
 c.Builder.Encoding = cloudevents.StructuredEncoding
 ```
--- a/cloudevents/README.md
+++ b/cloudevents/README.md
@@ -1,0 +1,149 @@
+# Knative CloudEvents SDK
+
+This library produces CloudEvents in version 0.1 compatible form. To learn more
+about CloudEvents, see the [Specification](https://github.com/cloudevents/spec).
+
+There are two roles the SDK fulfills: the [producer](#producer) and the 
+[consumer](#consumer). The producer creates a cloud event in either 
+[Binary](#binary) or [Structured](#structured) request format. The producer
+assembles and sends the event through an HTTP endpoint. The consumer will
+inspect the incoming HTTP request and select the correct decode format.
+
+This SDK should be wire-compatible with any other producer or consumer of the
+supported versions of CloudEvents.   
+
+## Getting Started
+
+CloudEvents acts as the envelope in which to send a custom object. Create a
+CloudEvent type for the events you will be producing. 
+
+Example CloudEvent Type: `dev.knative.cloudevent.example`
+
+Select a source to identify the originator of this CloudEvent. It should be a 
+valid URI. 
+
+Example CloudEvent Source: `https://github.com/knative/pkg#cloudevents-example` 
+
+And finally, create a struct that will be the data inside the CloudEvent, 
+example:
+
+```
+type Example struct {
+	Sequence int    `json:"id"`
+	Message    string `json:"message"`
+}
+``` 
+
+### Producer
+
+The producer will create a new `Example` object, fill out the CloudEvent struct,
+and post the event via a [Binary](#binary) or [Structured](#structured) request
+format.
+
+```go
+
+package main
+
+import (
+	"github.com/google/uuid"
+	"github.com/knative/pkg/cloudevents"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"time"
+)
+
+type Example struct {
+	Sequence int    `json:"id"`
+	Message  string `json:"message"`
+}
+
+func main() {
+	target := "http://localhost:8080"
+	eventType := "dev.knative.cloudevent.example"
+	eventSource := "https://github.com/knative/pkg#cloudevents-example"
+	data := &Example{
+		Sequence: 0,
+		Message:  "hello, world!",
+	}
+	ctx := cloudevents.EventContext{
+		CloudEventsVersion: cloudevents.CloudEventsVersion,
+		EventType:          eventType,
+		EventID:            uuid.New().String(),
+		Source:             eventSource,
+		EventTime:          time.Now(),
+	}
+	req, err := cloudevents.Binary.NewRequest(target, data, ctx)
+	if err != nil {
+		log.Printf("failed to create http request: %s", err)
+		return
+	}
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		log.Printf("failed to do POST: %v", err)
+		return
+	}
+	defer resp.Body.Close()
+	log.Printf("response Status: %s", resp.Status)
+	body, _ := ioutil.ReadAll(resp.Body)
+	log.Printf("response Body: %s", string(body))
+}
+
+```
+
+### Consumer
+
+The consumer will listen for a post and then inspect the headers to understand
+how to decode the request. 
+
+```go
+
+package main
+
+import (
+	"context"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/knative/pkg/cloudevents"
+)
+
+type Example struct {
+	Sequence int    `json:"id"`
+	Message  string `json:"message"`
+}
+
+func handler(ctx context.Context, data *Example) {
+	metadata := cloudevents.FromContext(ctx)
+	log.Printf("[%s] %s %s: %d,%q", metadata.EventTime.Format(time.RFC3339), metadata.ContentType, metadata.Source, data.Sequence, data.Message)
+}
+
+func main() {
+	log.Print("ready and listening on port 8080")
+	log.Fatal(http.ListenAndServe(":8080", cloudevents.Handler(handler)))
+}
+
+
+```
+
+## Request Formats
+
+### Binary
+
+Changes to the producer code required to leverage binary request format:
+
+```go
+req, err := cloudevents.Binary.NewRequest(target, data, ctx)
+
+```
+
+### Structured
+
+Changes to the producer code to leverage structured request format:
+
+```go
+req, err := cloudevents.Structured.NewRequest(target, data, ctx)
+```
+
--- a/cloudevents/README.md
+++ b/cloudevents/README.md
@@ -28,10 +28,12 @@ And finally, create a struct that will be the data inside the CloudEvent,
 example:
 
 ```go
+
 type Example struct {
     Sequence int    `json:"id"`
-    Message    string `json:"message"`
+    Message  string `json:"message"`
 }
+
 ```
 
 ### Producer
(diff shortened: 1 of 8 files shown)


evankanderson:
FYI, I've started working on converting the library to support both v0.1 and v0.2. This will probably manifest as:

* `V01EventContext` and `V02EventContext` classes, which implement:
  * `BinaryContext` interface with `AsHeaders() http.Headers` and `FromHeaders(http.Headers) error` methods
  * `json.Marshaler` and `json.Unmarshaler` for structured format.

And then a refactor of most of the remaining code to support both v0.1 and v0.2 use cases.

I'll update this README once that lands. FWIW, it looks like knative/eventing doesn't use this version of the library from knative/pkg yet, so that should probably be another item for 0.4.

vaikas-google:
Thanks so much for doing this!! Just couple of small nits.

cloudevents/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+CloudEvents acts as the envelope in which to send a custom object. Create a
+CloudEvent type for the events you will be producing.
+
+Example CloudEvent Type: `dev.knative.cloudevent.example`
+
+Select a source to identify the originator of this CloudEvent. It should be a
+valid URI.
+
+Example CloudEvent Source: `https://github.com/knative/pkg#cloudevents-example`
```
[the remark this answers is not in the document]
n3wscott: That is my source URI.

cloudevents/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+And finally, create a struct that will be the data inside the CloudEvent,
+example:
+
+```go
+type Example struct {
+    Sequence int    `json:"id"`
```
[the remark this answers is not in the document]
n3wscott: arg, yeah some db added a test to prevent tabs in markdown files. 👎 so I had to search/replace...

cloudevents/builder.go
```diff
@@ -0,0 +1,65 @@
...
+	Source    string
+	EventType string
+	Encoding  CloudEventEncoding
+}
+
+// Build produces a http request with the constant data embedded in the builder
+// merged with the new data provided in the build function. The request will
+// send a pre-assembled cloud event to the given target. The target is assumed
+// to be a URL with a scheme, ie: "http://localhost:8080"
+func (b *Builder) Build(target string, data interface{}) (*http.Request, error) {
```
evankanderson: Is this interface meant to be exposed externally?

If so, can you make this take 0-1 additional arguments (using varargs) to supply a CloudEventsContext. For example, you might want to be able to set the `time` or `id` (v0.2 names) fields on the CloudEvent from upstream data (like data from the PubSub envelope).

If this is meant to be internal and used with the Client, then it might make more sense to de-capitalize this and have the Builder fields be part of the Client.

cloudevents/client.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		Target: target,
+	}
+	return c
+}
+
+// Send creates a request based on the client's settings and sends the data
+// struct to the target set for this client. It returns error if there was an
+// issue sending the event, otherwise nil means the event was accepted.
+func (c *Client) Send(data interface{}) error {
```
evankanderson: Same comment here about optionally supplying an EventContext.

cloudevents/client.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil
+	}
+	return fmt.Errorf("error sending cloudevent: %s", status(resp))
+}
+
+// RequestSend uses the internal builder to make a request with the provided
+// data struct using the previously set parameters of event type and target.
+// Use this instead of client.Send() if processing of the http response directly
+// is required.
+func (c *Client) RequestSend(data interface{}) (*http.Response, error) {
```
evankanderson: Why are both `Send` and `RequestSend` exposed publicly? (Event more strangely, they seem to only differ in return type, where `RequestSend` also returns an `http.Response`.

Can we change the `Send` contract to return `(*http.Response, error)`, and return both the Response and a non-nil error when the response is not `accepted()`? Users who aren't interested in the Response can simply write:

```golang
if _, err := c.Send(myStuff); err != nil {
  return err
}
```

cloudevents/client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	client := &http.Client{}
+	return client.Do(req)
+}
+
+// Channel returns a channel that can be used to invoke Client.Send via a chan.
+// This method has a side effect of another thread to monitor the send channel.
+// Call client.Close() to shutdown the monitor thread.
+// Experimental, error handling is not fully developed.
+func (c *Client) Channel() chan<- interface{} {
```
evankanderson: Is this part of the public interface? This seems like it should be private, since sharing around a `Client` could cause random panics if someone decides to `close(c.Channel())`.

cloudevents/client.go
```diff
@@ -0,0 +1,131 @@
...
+		case <-c.done:
+			return
+		}
+	}
+}
+
+// accept is a helper method to understand if the respose from the target
+// accepted the CloudEvent.
+func accepted(resp *http.Response) bool {
+	if resp.StatusCode == 204 {
```
evankanderson: This should probably also include 200 and 201.

cloudevents/client.go
```diff
@@ -0,0 +1,131 @@
...
+		return true
+	}
+	return false
+}
+
+// status is a helper method to read the response of the target.
+func status(resp *http.Response) string {
```
evankanderson: Why not store the `*http.Response` in the Error as an additional field?

```golang
type RemoteError struct {
  resp *http.Response
  body *string
  bodyErr error
}

func (e *RemoteError) Error() string {
  if e == nil {
    return "Not an error"
  }
  if e.resp != nil {
    if e.body == nil && e.bodyErr == nil {
      e.body, e.bodyErr := ioutil.ReadAll(e.resp.Body)
    }
    if e.bodyErr != nil {
      return fmt.Sprintf("Status [%s] %s": e.Status, e.body)
    }
    return fmt.Sprintf("Status [%s] error reading body: %v", e.Status, e.bodyErr)
  }
  return fmt.Sprintf("Request failed with no response object")
}
```

cloudevents/encoding_binary.go
```diff
@@ -147,13 +147,11 @@ func (binary) ToHeaders(context *EventContext, h http.Header) error {
...
-	if ceVersion == "" {
-		ceVersion = CloudEventsVersion
+	if context.CloudEventsVersion == "" {
+		context.CloudEventsVersion = CloudEventsVersion
 	}
-	contentType := context.ContentType
-	if contentType == "" {
-		contentType = contentTypeJSON
+	if context.ContentType == "" {
+		context.ContentType = contentTypeJSON
```
evankanderson: You should document that you are changing the context supplied as an argument, which is surprising. Otherwise, you want to make a copy of the context before writing to it here.

(You should probably write a test to verify that the the supplied arguments to a binary transform are not themselves modified by the call.)

cloudevents/builder.go
n3wscott: This work was in result of a thought process solving the first pain point, and un-covering a new one.. fixing that. I had thought the builder would solve my problem with the cloudevents lib but it did not, I saw new ways to make it cleaner.

Now that I see the client, I might want to go back and rewrite it with this new use in mind.

cloudevents/client.go
n3wscott: This leaves the Response open though right? My goal was to close the http request and read the body out for the user of the lib. So I did not want to return the `http.Response` back if that was the case. Plus I wrote some logic to understand the response and did not want to expose that to the client user. I just want them to know if it worked out not.

cloudevents/client.go
n3wscott: I was playing with this idea:

```
func main() {
	c := cloudevents.NewClient(
		"dev.knative.cloudevent.example",
		"https://github.com/knative/pkg#cloudevents-example",
		"http://localhost:8080",
	)
	ch := c.Channel()

	data := &Example{
		Message: "hello, world!",
	}

	for i := 0; i < 10; i++ {
		data.Sequence = i
		ch <- *data
	}

	time.Sleep(1 * time.Second)
	c.Done()
}
```

cloudevents/client.go
n3wscott: I don't think golang intends for http.Response to be sent around out of the context of the http client:

```
The client must close the response body when finished with it:

resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
body, err := ioutil.ReadAll(resp.Body)
// ...
```

Ref: https://golang.org/pkg/net/http/

cloudevents/encoding_binary.go
n3wscott: they are being modified, and this lack of defaulting is a bug in the current code, not setting the content type breaks binary encoding because golang can't encode empty content type

cloudevents/client.go
n3wscott: I am going to pull it for now, I can always add a client channel wrapper

cloudevents/README.md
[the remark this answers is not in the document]
n3wscott: I am trying to show what the link text is

cloudevents/README.md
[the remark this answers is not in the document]
n3wscott: fixed.

cloudevents/builder.go
n3wscott: I would like to leave Build as public just incase someone wants to build contexts and requests without my client.

cloudevents/client.go
n3wscott: done.

cloudevents/README.md
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func main() {
+    c := cloudevents.NewClient(
+        "dev.knative.cloudevent.example",
+        "https://github.com/knative/pkg#cloudevents-example",
+        "http://localhost:8080",
+    )
+    for i := 0; i < 10; i++ {
+        data := &Example{
```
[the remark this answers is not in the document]
n3wscott: it can be either. removed the ref.

cloudevents/README.md
```diff
@@ -0,0 +1,143 @@
...
+    Sequence int    `json:"id"`
+    Message  string `json:"message"`
+}
+
+func main() {
+    c := cloudevents.NewClient(
+        "dev.knative.cloudevent.example",
+        "https://github.com/knative/pkg#cloudevents-example",
+        "http://localhost:8080",
+    )
```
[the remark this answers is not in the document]
n3wscott: Can we leave this here and agree to break people later on?I suspect it might become more clear once there is a registry or something and things will have to change again.

cloudevents/README.md
```diff
@@ -0,0 +1,143 @@
...
+    Message  string `json:"message"`
+}
+
+func handler(ctx context.Context, data *Example) {
+    metadata := cloudevents.FromContext(ctx)
+    log.Printf("[%s] %s %s: %d, %q", metadata.EventTime.Format(time.RFC3339), metadata.ContentType, metadata.Source, data.Sequence, data.Message)
+}
+
+func main() {
+    log.Print("ready and listening on port 8080")
```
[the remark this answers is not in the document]
n3wscott: 👍

cloudevents/builder.go
```diff
@@ -0,0 +1,133 @@
...
+	// Structured v0.1
+	StructuredEncodingV01 CloudEventEncoding = 1
+)
+
+// Builder holds settings that do not change over CloudEvents. It is intended
+// to represent a builder of only a single CloudEvent type.
+type Builder struct {
+	// A URI describing the event producer.
+	Source string
+	// Timestamp when the event happened.
```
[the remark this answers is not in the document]
n3wscott: That was an oversight, removed! thanks.

cloudevents/builder.go
```diff
@@ -0,0 +1,133 @@
...
+	case StructuredEncodingV01:
+		return Structured.NewRequest(target, data, ctx)
+	default:
+		return nil, fmt.Errorf("unsupported encoding: %v", b.Encoding)
+	}
+}
+
+// cloudEventsContext creates a CloudEvent context object, assumes
+// application/json as the content type.
+func (b *Builder) cloudEventsContextV01(overrides *V01EventContext) V01EventContext {
```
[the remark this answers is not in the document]
n3wscott: I think that is a good idea! will follow up when integrating v02

cloudevents/client_test.go
```diff
@@ -0,0 +1,380 @@
...
+
+	"github.com/knative/pkg/cloudevents"
+)
+
+var (
+	// Headers that are added to the response, but we don't want to check in our assertions.
+	unimportantHeaders = map[string]struct{}{
+		"accept-encoding": {},
+		"content-length":  {},
+		"content-type":    {},
```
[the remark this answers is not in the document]
n3wscott: whoops!!! nice catch. that was me copying too much from eventing tests.

cloudevents/client_test.go
```diff
@@ -0,0 +1,380 @@
...
+		name      string
+		eventType string
+		source    string
+		target    string
+		want      *cloudevents.Client
+		opt       cmp.Option
+	}{
+		{
+			name: "Simple",
+			want: &cloudevents.Client{},
```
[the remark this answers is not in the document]
n3wscott: the call to NewClient does not validate. It is just a helper. This validates that the returned object is an empty client.

cloudevents/builder.go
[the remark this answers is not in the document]
evankanderson: I'll do that in my v02 PR, since I'll be in there changing types anyway.

cloudevents/client_test.go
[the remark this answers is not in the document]
evankanderson: The destination will be "", which seems like the client won't be useful, but will return errors much later.

evankanderson:
@mattmoor, since I don't have admin on `knative/pkg`.

Issue #193, opened 2018-12-05
CloudEvent 0.2 spec work


matzew:
/kind feature

## Expected Behavior

CNCF will announce 0.2 of the spec

## Actual Behavior

Knative's impl is compliant. Question: how much do we care about backwards comp. ? Or just move on ?

matzew:
:joy:  lot's of labels ... :man_facepalming:

PR #244, opened 2019-01-27
CloudEvents v0.2 support, step 1: interfaces and versioned types.

evankanderson:
In-progress: #193 

This PR renames `CloudEventsVersion` and `EventContext` to `V01...`, and leaves legacy aliases to allow seamless upgrades.

Additional changes:
* Introduces `SendContext`/`LoadContext` interfaces to support read/write of arbitrary CloudEvents versions.
* Makes `CloudEventsVersion` required per v0.1 spec.

The next step is to change the return type on `FromRequest` to `(LoadContext, error)` and introduce the `V02EventContext` file. This will also necessitate changing the tests (and therefore downstream code), so I'd prefer to do this in a separate PR (see 366 in `handler.go` for an example of the work needed -- basically calling `AsV01()` or `AsV02()` in locations which expect a particular version).

### DIFF
--- a/cloudevents/doc.go
+++ b/cloudevents/doc.go
@@ -19,5 +19,4 @@ limitations under the License.
 // https://github.com/cloudevents/spec/blob/v0.1/http-transport-binding.md
 // and
 // https://github.com/cloudevents/spec/blob/v0.1/spec.md
-
 package cloudevents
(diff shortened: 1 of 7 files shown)


evankanderson:
/assign @n3wscott 
/assign @vaikas-google 

Since I know both of you have been looking at this library recently...

PR #248, opened 2019-01-29
Add support for v0.2 CloudEvents "context attributes".

evankanderson:
In-progress: #193 

This PR introduces `V02...` versions of `V01EventContext`.

Additional changes:
* Clean up secondary interfaces to better support support read/write of arbitrary CloudEvents versions.
* Changes the return type of FromRequest to return a `LoadContext`, rather than a concrete `EventContext`. Callers should call `AsV01()` or `AsV02()` on the returned object.

### DIFF
--- a/cloudevents/client_test.go
+++ b/cloudevents/client_test.go
@@ -1,0 +1,389 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudevents_test
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/knative/pkg/cloudevents"
+)
+
+var (
+	// Headers that are added to the response, but we don't want to check in our assertions.
+	unimportantHeaders = map[string]struct{}{
+		"accept-encoding": {},
+		"content-length":  {},
+		"user-agent":      {},
+	}
+)
+
+func TestNewClient(t *testing.T) {
+	m := "NewClient"
+
+	for _, test := range []struct {
+		name      string
+		eventType string
+		source    string
+		target    string
+		want      *cloudevents.Client
+		opt       cmp.Option
+	}{
+		{
+			name: "Simple",
+			want: &cloudevents.Client{},
+		}, {
+			name:      "Full",
+			eventType: "event.type",
+			source:    "source",
+			target:    "target",
+			want: &cloudevents.Client{
+				Target: "target",
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			got := cloudevents.NewClient(test.target, cloudevents.Builder{Source: test.source, EventType: test.eventType})
+
+			if diff := cmp.Diff(test.want, got, test.opt, cmpopts.IgnoreUnexported(cloudevents.Client{})); diff != "" {
+				t.Errorf("%s (-want, +got) = %v", m, diff)
+			}
+		})
+	}
+}
+
+func TestClientSend(t *testing.T) {
+	now := time.Now()
+	eventID := "AABBCCDDEE"
+	doc := FirestoreDocument{
+		Name: "projects/demo/databases/default/documents/users/inlined",
+		Fields: map[string]interface{}{
+			"project": "eventing",
+			"handle":  "@inlined",
+		},
+		CreateTime: time.Date(1985, 6, 5, 12, 0, 0, 0, time.UTC),
+		UpdateTime: now.UTC(),
+	}
+
+	service := "firestore.googleapis.com"
+
+	for _, test := range []struct {
+		name            string
+		client          cloudevents.Client
+		override        *cloudevents.V01EventContext
+		fakeResponse    *http.Response
+		expectedRequest *requestValidation
+		errText         string
+	}{
+		{
+			name: "binary simple v0.1",
+			client: func() cloudevents.Client {
+				client := cloudevents.NewClient(
+					"server url",
+					cloudevents.Builder{
+						Source:    fmt.Sprintf("//%s/%s", service, doc.Name),
+						EventType: "google.firestore.document.create",
+						Encoding:  cloudevents.BinaryV01,
+					},
+				)
+				return *client
+			}(),
+			override: &cloudevents.V01EventContext{
+				EventTime: now,
+				EventID:   eventID,
+			},
+			fakeResponse: &http.Response{
+				StatusCode: http.StatusAccepted,
+			},
+			expectedRequest: &requestValidation{
+				Headers: map[string][]string{
+					"ce-cloudeventsversion": {"0.1"},
+					"ce-eventid":            {"AABBCCDDEE"},
+					"ce-eventtime":          {now.Format(time.RFC3339Nano)},
+					"ce-eventtype":          {"google.firestore.document.create"},
+					"ce-source":             {"//firestore.googleapis.com/projects/demo/databases/default/documents/users/inlined"},
+					"content-type":          {"application/json"},
+				},
+				Body: fmt.Sprintf(
+					`{"name":"projects/demo/databases/default/documents/users/inlined","fields":{"handle":"@inlined","project":"eventing"},"createTime":"1985-06-05T12:00:00Z","updateTime":%q}`,
+					now.UTC().Format(time.RFC3339Nano),
+				),
+			},
+		}, {
+			name: "binary full v0.1",
+			client: func() cloudevents.Client {
+				client := cloudevents.NewClient(
+					"source url",
+					cloudevents.Builder{
+						Source:           fmt.Sprintf("//%s/%s", service, doc.Name),
+						EventType:        "google.firestore.document.create",
+						EventTypeVersion: "v1beta2",
+						SchemaURL:        "http://type.googleapis.com/google.firestore.v1beta1.Document",
+						Encoding:         cloudevents.BinaryV01,
+						Extensions: map[string]interface{}{
+							"purpose": "tbd",
+						},
+					},
+				)
+				return *client
+			}(),
+			override: &cloudevents.V01EventContext{
+				EventTime: now,
+				EventID:   eventID,
+			},
+			fakeResponse: &http.Response{
+				StatusCode: http.StatusAccepted,
+			},
+			expectedRequest: &requestValidation{
+				Headers: map[string][]string{
+					"ce-cloudeventsversion": {"0.1"},
+					"ce-eventtypeversion":   {"v1beta2"},
+					"ce-eventid":            {"AABBCCDDEE"},
+					"ce-eventtime":          {now.Format(time.RFC3339Nano)},
+					"ce-eventtype":          {"google.firestore.document.create"},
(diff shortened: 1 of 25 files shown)


cloudevents/encoding_binary.go
```diff
@@ -76,14 +77,21 @@ type BinaryLoader interface {
 }
 
 // FromRequest parses event data and context from an HTTP request.
-func (binary) FromRequest(data interface{}, r *http.Request) (*EventContext, error) {
-	// TODO: detect version and create correct VXXEventContext
-	ec := &EventContext{}
+func (binary) FromRequest(data interface{}, r *http.Request) (LoadContext, error) {
+	var ec LoadContext
+	if r.Header.Get("CE-SpecVersion") == V02CloudEventsVersion {
```
vagababov: I really prefer `switch` in this case, especially i v03 is coming at some point.
```
switch {
 case r.Header....:
...
default:
  error
}
```

cloudevents/encoding_structured.go
```diff
@@ -64,13 +62,24 @@ func (structured) FromRequest(data interface{}, r *http.Request) (*EventContext,
...
+	var ec LoadContext
+	v := ""
+	if err := json.Unmarshal(raw["specversion"], &v); err == nil && v == V02CloudEventsVersion {
+		ec = &V02EventContext{}
+	} else if err := json.Unmarshal(raw["cloudEventsVersion"], &v); err == nil && v == V01CloudEventsVersion {
+		ec = &V01EventContext{}
+	} else {
+		// For historical reasons, this library treated unknown payloads as V01. Preserve for now.
+		ec = &V01EventContext{}
+		// TODO: change behavior to be an error:
```
vagababov: Issue#/Owner?

cloudevents/event_test.go
```diff
@@ -106,25 +121,29 @@ func TestValidRoundTrips(t *testing.T) {
...
-			}
-		})
+		for encoding, convert := range encoderSets {
+			t.Run(test.name+"-"+encoding, func(t *testing.T) {
+				req, err := test.encoder.NewRequest(webhook, doc, convert[0](context))
+				if err != nil {
+					t.Fatalf("Failed to encode event %s", err)
+				}
+
+				fmt.Printf("In %q\n", test.name+"-"+encoding)
```
vagababov: rm debug logging.

cloudevents/event_test.go
```diff
@@ -255,13 +277,13 @@ func TestExtensionExtraction(t *testing.T) {
...
-	expectedExtensions := map[string]interface{}{
+	expectedV01Extensions := map[string]interface{}{
 		"Prop1": "value1",
 		"Prop2": map[string]interface{}{
 			"nestedProp": "nestedValue",
 		},
 	}
-	if !reflect.DeepEqual(expectedExtensions, ctx.Extensions) {
-		t.Fatalf("Did not parse expected extensions. Wanted=%v; got=%v", expectedExtensions, ctx.Extensions)
+	if !reflect.DeepEqual(expectedV01Extensions, ctx.AsV01().Extensions) {
```
vagababov: Suggest: `cmp.Equal`?

cloudevents/event_test.go
```diff
@@ -255,13 +277,13 @@ func TestExtensionExtraction(t *testing.T) {
...
+		t.Fatalf("Did not parse expected extensions. Wanted=%v; got=%v", expectedV01Extensions, ctx.AsV01().Extensions)
```
vagababov: And here you can use `cmp.Diff` to print the diffs.

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
+/*
+Copyright 2018 The Knative Authors
```
vagababov: 2019

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
...
+
+// AsV02 implements the ContextTranslator interface.
+func (ec V02EventContext) AsV02() V02EventContext {
+	return ec
+}
+
+// AsHeaders implements the BinarySender interface.
+func (ec V02EventContext) AsHeaders() http.Header {
+	h := http.Header{}
+	h.Set("CE-SpecVersion", ec.SpecVersion)
```
vagababov: Any reason why those header names aren't constants? They seem to be shared throughout this change.

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
...
+		h.Set("Content-Type", ec.ContentType)
+	}
+	for k, v := range ec.Extensions {
+		// Per spec, map-valued extensions are converted to a list of headers as:
+		// CE-attrib-key
+		if mapVal, ok := v.(map[string]interface{}); ok {
+			for subkey, subval := range mapVal {
+				data := fmt.Sprint(subval)
+				encoded, err := json.Marshal(subval)
+				if err != nil {
```
vagababov: Is the logic correct here?

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
...
+					data = string(encoded)
+				}
+				h.Set("CE-"+k+"-"+subkey, data)
+			}
+			continue
+		}
+		data := fmt.Sprint(v)
+		encoded, err := json.Marshal(v)
+		if err != nil {
```
vagababov: ditto.

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
...
+		ContentType: in.Get("Content-Type"),
+		Extensions:  make(map[string]interface{}),
+	}
+	// Extensions and top-level fields are mixed under "CE-" headers.
+	// Extract them all here rather than trying to clear fields in headers.
+	for k, v := range in {
+		if strings.EqualFold(k[:len("CE-")], "CE-") {
+			key := strings.ToLower(string(k[len("CE-"):]))
+			value := v[0]
+			switch key {
```
vagababov: We can save 2 lines by:
```
switch key, value := ....;key {
...
}
```

cloudevents/event_v02.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	*ec = data
+	return nil
+}
+
+// AsJSON implementsn the StructuredSender interface.
+func (ec V02EventContext) AsJSON() (map[string]json.RawMessage, error) {
+	ret := make(map[string]json.RawMessage)
+	err := anyError(
+		encodeKey(&ret, "specversion", ec.SpecVersion),
```
vagababov: this function is elsewhere, but maps are passed by reference and there is no need to take pass around pointer to it.

cloudevents/encoding_binary.go
evankanderson: That would be easier if these two used the same field on the map. (Yes, they changed their version sentinel between v0.1 and v0.2...)

I can change this one over anyway.

cloudevents/encoding_structured.go
evankanderson: I'm going to do it and/or kill off V01 altogether.

If you want, I can make that part of this PR; I was feeling some "change fatigue" last night.

cloudevents/event_test.go
evankanderson: D'oh, thought I'd gotten it all.

cloudevents/encoding_structured.go
vagababov: If it's just around the corner, it doesn't matter, I guess.

cloudevents/event_test.go
evankanderson: Thanks for the hint on `cmp.Diff`! It helped me find a pair of bugs lurking in the code -- one on v1 JSON encoding, and one in V02 map extensions.

cloudevents/event_test.go
evankanderson: Switched to `cmp.Diff` and compare with empty string.

cloudevents/event_v02.go
evankanderson: I'm not sure number of lines is what I'm optimizing for here. I did combine the `key, value :=` assignment.

cloudevents/event_v02.go
evankanderson: Yes, thanks!

cloudevents/event_v02.go
evankanderson: Done.

cloudevents/event_v02.go
evankanderson: Done.

cloudevents/encoding_structured.go
evankanderson: I fixed this, it turned out to be 2-3 tests that needed the CloudEventsVersion field to be set.

cloudevents/client_test.go
```diff
@@ -168,7 +168,7 @@ func TestClientSend(t *testing.T) {
 					"ce-eventtype":          {"google.firestore.document.create"},
 					"ce-source":             {"//firestore.googleapis.com/projects/demo/databases/default/documents/users/inlined"},
 					"ce-schemaurl":          {"http://type.googleapis.com/google.firestore.v1beta1.Document"},
-					"ce-x-purpose":          {"tbd"},
+					"ce-x-purpose":          {"\"tbd\""},
```
vagababov:
```
`"tbd"`
```
is nicer :)

cloudevents/event_v01.go
```diff
@@ -90,15 +112,16 @@ func (ec V01EventContext) AsHeaders() http.Header {
 	for k, v := range ec.Extensions {
 		data := fmt.Sprint(v)
 		encoded, err := json.Marshal(v)
-		if err != nil {
+		if err == nil {
```
vagababov: We seem to swallow the error here.
In all honesty, I have yet manage to make json.Marshal() fail, but nonetheless, at least logging the error could be quite useful.

cloudevents/event_v02.go
```diff
@@ -0,0 +1,261 @@
...
+	if ec.ContentType != "" {
+		h.Set(headerContentType, ec.ContentType)
+	}
+	for k, v := range ec.Extensions {
+		// Per spec, map-valued extensions are converted to a list of headers as:
+		// CE-attrib-key
+		if mapVal, ok := v.(map[string]interface{}); ok {
+			for subkey, subval := range mapVal {
+				data := fmt.Sprint(subval)
+				if encoded, err := json.Marshal(subval); err == nil {
```
vagababov: Same error remark here.

cloudevents/event_v02.go
vagababov: Well, I am mostly in favor minimal contexts/scopes for variables (which in go often coincides with brevity). But it's more a personal preference.

cloudevents/event_v02.go
```diff
@@ -0,0 +1,261 @@
...
+				data.ID = value
+			case fieldSchemaURL:
+				data.SchemaURL = value
+			case fieldTime:
+				if data.Time, err = time.Parse(time.RFC3339Nano, value); err != nil {
+					return err
+				}
+			default:
+				var tmp interface{}
+				if err = json.Unmarshal([]byte(value), &tmp); err != nil {
```
vagababov: Do we care about the error here, or we expect them to happen?

cloudevents/event_v01.go
evankanderson: json.Marshal will fail if you put something non-convertible into the map[string]interface{}, like a channel.

I've changed `AsHeaders()` to return `(http.Header, error)`.

cloudevents/event_v02.go
evankanderson: The error will happen if you have an unquoted string, like `CE-SpecVersion: v0.2`.

Right now, it seems more helpful to interpret these as strings, rather than reject them.

evankanderson: I added some tests adjacent to this, but I'm not sure we have something which sends a raw HTTP request rather than something created by a marshaller.

Mind if I add the raw request testing in another PR (in part because the raw HTTP behavior is not well-defined in the CloudEvents 0.2 spec in terms of error handling)?

cloudevents/event_v01.go
```diff
@@ -110,15 +110,14 @@ func (ec V01EventContext) AsHeaders() http.Header {
...
 	}
 	for k, v := range ec.Extensions {
-		data := fmt.Sprint(v)
 		encoded, err := json.Marshal(v)
-		if err == nil {
-			data = string(encoded)
+		if err != nil {
+			return nil, err
 		}
 		// Preserve case in v0.1, even though HTTP headers are case-insensitive.
```
vagababov: Why not `h.Set` as above?

Issue #226, opened 2019-01-12
Move AppendRandomString to pkg


adrcunha:
## Expected Behavior

This useful helper function is shared across all repos through pkg.

## Actual Behavior

This function is duplicated in several repos, like in https://github.com/knative/build-pipeline/blob/master/test/randstring.go (with test) and https://github.com/knative/serving/blob/02d3c9b657fdefe76690a1d5751ae0b2c4e414ae/test/crd.go#L255

PR #256, opened 2019-01-30
test/helpers module to store AppendRandomString helper fn

odra:
Fixes: #226

I have added the helper function`AppendRandomString` into `test/helpers` module (helpers being a new module), sample usage:

```go
package pkg

import (
	"github.com/knative/pkg/test/helpers"
	"testing"
	"strings"
)


func TestAppendRandomString(t *testing.T) {
	s := helpers.AppendRandomString("myprefix")

	if !strings.HasPrefix(s, "myprefix") {
		t.Fatalf("Failed to verify prefix: %s", s)
	}

	if len(s) != 18 {
		t.Fatalf("Failed to verify string size: %d", len(s))
	}

	t.Logf("---%s---", s)
}
```

I was looking for a first good issue and this one looked simple enough :)

### DIFF
--- a/test/helpers/data.go
+++ b/test/helpers/data.go
@@ -1,0 +1,39 @@
+package helpers
+
+import (
+	"math/rand"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	letterBytes   = "abcdefghijklmnopqrstuvwxyz"
+	randSuffixLen = 8
+	sep = "-"
+)
+
+var (
+	r        *rand.Rand
+	rndMutex *sync.Mutex
+	once     sync.Once
+)
+
+func initSeed() {
+	seed := time.Now().UTC().UnixNano()
+	r = rand.New(rand.NewSource(seed))
+	rndMutex = &sync.Mutex{}
+}
+
+func AppendRandomString(prefix string) string {
+	once.Do(initSeed)
+	suffix := make([]byte, randSuffixLen)
+
+	rndMutex.Lock()
+	for i := range suffix {
+		suffix[i] = letterBytes[r.Intn(len(letterBytes))]
+	}
+	rndMutex.Unlock()
+
+	return strings.Join([]string{prefix, string(suffix)}, sep)
+}
--- a/test/helpers/data.go
+++ b/test/helpers/data.go
@@ -30,10 +30,11 @@ func AppendRandomString(prefix string) string {
 	suffix := make([]byte, randSuffixLen)
 
 	rndMutex.Lock()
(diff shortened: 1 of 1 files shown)


odra:
I signed it!

test/helpers/data.go
```diff
@@ -0,0 +1,39 @@
...
+
+func AppendRandomString(prefix string) string {
+	once.Do(initSeed)
+	suffix := make([]byte, randSuffixLen)
+
+	rndMutex.Lock()
+	for i := range suffix {
+		suffix[i] = letterBytes[r.Intn(len(letterBytes))]
+	}
+	rndMutex.Unlock()
```
markusthoemmes: Please `defer` this right after getting the Lock above. The minimal gain in concurrency is not worth the risk of a panic in between I think.

test/helpers/data.go
```diff
@@ -0,0 +1,39 @@
...
+	once     sync.Once
+)
+
+func initSeed() {
+	seed := time.Now().UTC().UnixNano()
+	r = rand.New(rand.NewSource(seed))
+	rndMutex = &sync.Mutex{}
+}
+
+func AppendRandomString(prefix string) string {
```
jessiezcc: the same function exists at https://github.com/knative/serving/blob/master/test/crd.go already. 
But it makes sense to move it out of serving to pkg. Will you pls remove it from serving?

jessiezcc:
the same function exists at https://github.com/knative/serving/blob/master/test/crd.go already.
But it makes sense to move it out of serving to pkg. Added comments in code.

test/helpers/data.go
odra: Sure, sorry I missed it

odra: Ah right, I was going to remove it right after the pr is in good shape :)

odra: Does it make sense to remove it from here as well? https://github.com/knative/build-pipeline/blob/master/test/randstring.go

test/helpers/data.go
odra: Done

test/helpers/data.go
jessiezcc: yes, all duplication

jessiezcc: can you pls open a git issue to track the next step?

odra: I've created one issue in this repository: https://github.com/knative/pkg/issues/274 and also two new ones on each repository to track the actual work:

https://github.com/knative/build-pipeline/issues/490
https://github.com/knative/serving/issues/3174

PR #257, opened 2019-01-31
Add `AtIndex` method to deal to attach index at the current error.

vagababov:
This is useful when we are dealing with primite leaf list (e.g.
a list of strings).
The alternative are
a) creating dummy error and then executing ViaFieldIndex(), which is
plain ugly
b) implementing Validate() on the primitive type via type aliasing and
then collecting errors, which seems an overkill.

In addition it seems ViaFieldIndex was not covered by unit tests,
so I've added one to take care of that.

/cc @evankanderson

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -73,7 +73,7 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 	return newErr
 }
 
-// ViaIndex is used to attach an index to the *next* ViaField provided.
+// ViaIndex is used to attach an index to the next ViaField provided.
 // For example, if a type recursively validates a parameter that has a collection:
 //  for i, c := range spec.Collection {
 //    if err := doValidation(c); err != nil {
@@ -82,19 +82,6 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 //  }
 func (fe *FieldError) ViaIndex(index int) *FieldError {
 	return fe.ViaField(asIndex(index))
-}
-
-
-// AtIndex is used to attach an index to the *previous* ViaField provided.
-func (fe *FieldError) AtIndex(index int) *FieldError {
-	if fe == nil {
-		return nil
-	}
-	ai := asIndex(index)
-	for i := range fe.Paths {
-		fe.Paths[i] = flatten([]string{fe.Paths[i], ai})
-	}
-	return fe
 }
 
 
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -84,6 +84,20 @@ func (fe *FieldError) ViaIndex(index int) *FieldError {
 	return fe.ViaField(asIndex(index))
 }
 
+// AtIndex attaches the index to the current field.
+// Which is different from ViaIndex, that attaches index to the
+// next field.
+func (fe *FieldError) AtIndex(index int) *FieldError {
```
n3wscott: how is this different than `ViaIndex`?

apis/field_error_test.go
```diff
@@ -338,55 +340,77 @@ can not use @, do not try`,
...
+			// 		return err
+			//   }
+			// }
+			// --> I expect path to be myList[i].foo
 
-			err = err.ViaIndex(0).ViaField("bar")
-			err = err.ViaIndex(2).ViaIndex(1).ViaField("baz")
-			err = err.ViaIndex(3).ViaIndex(4).ViaField("boof")
-			return err
+			return (&FieldError{
```
n3wscott: I think this is less readable, I am simulating errors in the way they would be called in real code.

apis/field_error.go
evankanderson: I think the docs make it pretty clear, but @vagababov was running into a problem indicating an error with the second item in a `[]string`, i.e. `releases[1]`. This ends up being the syntax for that:

```golang
 errs = errs.Also(apis.ErrInvalidValue(r, apis.CurrentField).ViaFieldIndex("revisions", i))
```

apis/field_error_test.go
evankanderson: I agree that the chaining is easier to read in 2-3 medium-length calls rather than chained more than 60 chars long.

(Purely a formatting issue.)

n3wscott:
How about this:

In `field_error.go`:

Add:

```
func ErrInvalidArrayValue(value string) *FieldError {
	return ErrInvalidValue(value, CurrentField)
}
```

And then in usage:

```
 errs = errs.Also(apis.ErrInvalidArrayValue(r).ViaFieldIndex("revisions", i))
```

And then in testing:

```
	{
		name: "use ErrInvalidArrayValue and viaFieldKey",
		err: func() *FieldError {
			return ErrInvalidArrayValue("args").ViaFieldIndex("foo", 5)
		}(),
		want: `invalid value "args": foo[5]`,
	},
```

apis/field_error_test.go
vagababov: Eh, matter of taste. Reverted.

PR #261, opened 2019-02-04
Use the k8s sets for the code that is actually set of strings

vagababov:
/cc @mattmoor

### DIFF
--- a/apis/duck/v1alpha1/condition_set.go
+++ b/apis/duck/v1alpha1/condition_set.go
@@ -280,7 +280,7 @@ func (r conditionsImpl) MarkUnknown(t ConditionType, reason, messageFormat strin
 			// Double check that the happy condition is also false.
 			happy := r.GetCondition(r.happy)
 			if !happy.IsFalse() {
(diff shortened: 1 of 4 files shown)


apis/duck/v1alpha1/condition_set.go
```diff
@@ -280,7 +280,7 @@ func (r conditionsImpl) MarkUnknown(t ConditionType, reason, messageFormat strin
 			// Double check that the happy condition is also false.
 			happy := r.GetCondition(r.happy)
 			if !happy.IsFalse() {
-				r.MarkFalse(r.happy, reason, messageFormat, messageA)
+				r.MarkFalse(r.happy, reason, messageFormat, messageA...)
```
vagababov: This change was required for compilation reasons, not sure how it was working before.

vagababov:
As part of https://github.com/knative/serving/issues/3049

apis/duck/v1alpha1/condition_set.go
mattmoor: 😲

PR #262, opened 2019-02-05
Hardening the websocket client.

markusthoemmes:
The websocket connection is under investigation of being an issue in production environments. This is a first stab at hardening it and gaining more insights.

- Add logging to the websocket client
- Recompose some of the inner works for better readability
- Add a test that actually connects to a websocket
- Improve shutdown behavior

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Knative Authors
+Copyright 2019 The Knative Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -56,7 +56,9 @@ type rawConnection interface {
 type ManagedConnection struct {
 	target     string
 	connection rawConnection
-	closeChan  chan struct{}
+
+	closeChan chan struct{}
+	closeOnce sync.Once
 
 	// If set, messages will be forwarded to this channel
 	messageChan chan []byte
@@ -95,14 +97,7 @@ func NewDurableConnection(target string, messageChan chan []byte) *ManagedConnec
 	go func() {
 		// If the close signal races the connection attempt, make
 		// sure the connection actually closes.
-		defer func() {
-			c.connectionLock.RLock()
-			defer c.connectionLock.RUnlock()
-
-			if conn := c.connection; conn != nil {
-				conn.Close()
-			}
-		}()
+		defer c.Shutdown()
 		for {
 			select {
 			default:
@@ -156,33 +151,35 @@ func (c *ManagedConnection) connect() (err error) {
 
 // keepalive keeps the connection open and reads control messages.
 // All messages are discarded.
-func (c *ManagedConnection) keepalive() (err error) {
+func (c *ManagedConnection) keepalive() error {
 	c.readerLock.Lock()
 	defer c.readerLock.Unlock()
 
 	for {
-		func() {
-			c.connectionLock.RLock()
-			defer c.connectionLock.RUnlock()
+		err := func() error {
+			c.connectionLock.Lock()
+			defer c.connectionLock.Unlock()
 
-			if conn := c.connection; conn != nil {
-				var reader io.Reader
-				var messageType int
-				messageType, reader, err = conn.NextReader()
-				if err != nil {
-					conn.Close()
-				}
-
-				// Send the message to the channel if its an application level message
-				// and if that channel is set.
-				if c.messageChan != nil && (messageType == websocket.TextMessage || messageType == websocket.BinaryMessage) {
-					if message, _ := ioutil.ReadAll(reader); message != nil {
-						c.messageChan <- message
-					}
-				}
-			} else {
-				err = ErrConnectionNotEstablished
+			if c.connection == nil {
+				return ErrConnectionNotEstablished
 			}
+
+			messageType, reader, err := c.connection.NextReader()
+			if err != nil {
+				c.connection.Close()
+				c.connection = nil
+				return err
+			}
+
+			// Send the message to the channel if its an application level message
+			// and if that channel is set.
+			if c.messageChan != nil && (messageType == websocket.TextMessage || messageType == websocket.BinaryMessage) {
+				if message, _ := ioutil.ReadAll(reader); message != nil {
+					c.messageChan <- message
+				}
+			}
+
+			return nil
 		}()
 
 		if err != nil {
@@ -196,8 +193,7 @@ func (c *ManagedConnection) Send(msg interface{}) error {
 	c.connectionLock.RLock()
 	defer c.connectionLock.RUnlock()
 
-	conn := c.connection
-	if conn == nil {
+	if c.connection == nil {
 		return ErrConnectionNotEstablished
 	}
 
@@ -210,17 +206,22 @@ func (c *ManagedConnection) Send(msg interface{}) error {
 		return err
 	}
 
-	return conn.WriteMessage(websocket.BinaryMessage, b.Bytes())
+	return c.connection.WriteMessage(websocket.BinaryMessage, b.Bytes())
 }
 
-// Close closes the websocket connection.
-func (c *ManagedConnection) Close() error {
-	c.closeChan <- struct{}{}
-	c.connectionLock.RLock()
-	defer c.connectionLock.RUnlock()
+// Shutdown closes the websocket connection.
+func (c *ManagedConnection) Shutdown() error {
+	c.connectionLock.Lock()
+	defer c.connectionLock.Unlock()
 
-	if conn := c.connection; conn != nil {
-		return conn.Close()
+	c.closeOnce.Do(func() {
+		close(c.closeChan)
+	})
+
+	if c.connection != nil {
+		err := c.connection.Close()
+		c.connection = nil
+		return err
 	}
 	return nil
 }
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -135,14 +135,15 @@ func newConnection(target string, messageChan chan []byte, connFactory func(targ
 
 // connect tries to establish a websocket connection.
 func (c *ManagedConnection) connect() (err error) {
+	c.connectionLock.Lock()
+	defer c.connectionLock.Unlock()
+
 	wait.ExponentialBackoff(c.connectionBackoff, func() (bool, error) {
 		var conn rawConnection
 		conn, err = c.connectionFactory(c.target)
 		if err != nil {
 			return false, nil
 		}
-		c.connectionLock.Lock()
-		defer c.connectionLock.Unlock()
 
 		c.connection = conn
 		return true, nil
@@ -164,15 +165,15 @@ func (c *ManagedConnection) keepalive() error {
 // If a messageChan is supplied and the current message type is not
 // a control message, the message is sent to that channel.
 func (c *ManagedConnection) read() error {
-	c.readerLock.Lock()
-	defer c.readerLock.Unlock()
-
 	c.connectionLock.Lock()
 	defer c.connectionLock.Unlock()
 
 	if c.connection == nil {
 		return ErrConnectionNotEstablished
 	}
+
+	c.readerLock.Lock()
+	defer c.readerLock.Unlock()
 
 	messageType, reader, err := c.connection.NextReader()
 	if err != nil {
@@ -215,12 +216,12 @@ func (c *ManagedConnection) Send(msg interface{}) error {
 
 // Shutdown closes the websocket connection.
 func (c *ManagedConnection) Shutdown() error {
-	c.connectionLock.Lock()
-	defer c.connectionLock.Unlock()
-
 	c.closeOnce.Do(func() {
 		close(c.closeChan)
 	})
+
+	c.connectionLock.Lock()
+	defer c.connectionLock.Unlock()
 
 	if c.connection != nil {
 		err := c.connection.Close()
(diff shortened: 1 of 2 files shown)


websocket/connection.go
```diff
@@ -78,38 +76,44 @@ type ManagedConnection struct {
...
 
 // NewDurableConnection creates a new websocket connection, that
 // passes incoming messages to the given message channel. It can also
 // send messages to the endpoint it connects to.
 // The connection will continuously be kept alive and reconnected
 // in case of a loss of connectivity.
-func NewDurableConnection(target string, messageChan chan []byte) *ManagedConnection {
-	c := newConnection(target, messageChan)
+func NewDurableConnection(target string, messageChan chan []byte, logger *zap.SugaredLogger) *ManagedConnection {
+	websocketConnectionFactory := func(target string) (rawConnection, error) {
```
vagababov: Interestingly enough `target` is redundant here, since its immutable and can be bound here (thought it can be kept for expressiveness).

websocket/connection.go
```diff
@@ -78,38 +76,44 @@ type ManagedConnection struct {
...
-				conn.Close()
-			}
-		}()
+		defer c.Shutdown()
 		for {
 			select {
 			default:
+				logger.Infof("Connecting to %q", target)
 				if err := c.connect(); err != nil {
+					logger.Errorw(fmt.Sprintf("Connecting to %q failed", target), zap.Error(err))
```
vagababov: We probably need a backoff of sorts here?

websocket/connection.go
```diff
@@ -120,11 +124,12 @@ func NewDurableConnection(target string, messageChan chan []byte) *ManagedConnec
...
 	conn := &ManagedConnection{
-		target:      target,
-		closeChan:   make(chan struct{}, 1),
-		messageChan: messageChan,
+		target:            target,
+		connectionFactory: connFactory,
+		closeChan:         make(chan struct{}),
+		messageChan:       messageChan,
 		connectionBackoff: wait.Backoff{
 			Duration: 100 * time.Millisecond,
```
vagababov: Can we comment why did we pick those exact values?

websocket/connection.go
```diff
@@ -138,14 +143,16 @@ func newConnection(target string, messageChan chan []byte) *ManagedConnection {
 
 // connect tries to establish a websocket connection.
 func (c *ManagedConnection) connect() (err error) {
+	c.connectionLock.Lock()
```
vagababov: Ah, I see, we backoff here.
So I see two problems here, if the target is unavailable:
- the lock is held basically "forever"  (we wait 20 exponential steps starting with 100ms, i.e. last wait is 19seconds, and if we consider jitter then even more, ~64 seconds in total without jitter). Lock should be held only for `c.connection=conn` part, rather than wholly.
- the <-c.closeChan won't be read during all this time. Considering we continue the loop if this returns an error, the backoff here should be minimal (or simply moved above).

websocket/connection.go
```diff
@@ -78,38 +76,44 @@ type ManagedConnection struct {
...
 			select {
 			default:
+				logger.Infof("Connecting to %q", target)
 				if err := c.connect(); err != nil {
+					logger.Errorw(fmt.Sprintf("Connecting to %q failed", target), zap.Error(err))
 					continue
 				}
-				c.keepalive()
+				logger.Infof("Connected to %q", target)
+				if err := c.keepalive(); err != nil {
```
vagababov: This in theory never ends, thus `c.closeChan` never gets the chance to be read below.
Is this what we intend?

websocket/connection.go
```diff
@@ -154,50 +161,53 @@ func (c *ManagedConnection) connect() (err error) {
 	return err
 }
 
-// keepalive keeps the connection open and reads control messages.
-// All messages are discarded.
-func (c *ManagedConnection) keepalive() (err error) {
+// keepalive keeps the connection open.
+func (c *ManagedConnection) keepalive() error {
+	for {
+		if err := c.read(); err != nil {
```
vagababov:
```
var err
for err = c.read(); err != nil; err=c.read() {
}
return err
```

websocket/connection.go
```diff
@@ -154,50 +161,53 @@ func (c *ManagedConnection) connect() (err error) {
...
-					if message, _ := ioutil.ReadAll(reader); message != nil {
-						c.messageChan <- message
-					}
-				}
-			} else {
-				err = ErrConnectionNotEstablished
-			}
-		}()
+	messageType, reader, err := c.connection.NextReader()
+	if err != nil {
```
vagababov: this block is repeated at least twice. Should probably be a function.

websocket/connection.go
```diff
@@ -210,17 +220,22 @@ func (c *ManagedConnection) Send(msg interface{}) error {
 		return err
 	}
 
-	return conn.WriteMessage(websocket.BinaryMessage, b.Bytes())
+	return c.connection.WriteMessage(websocket.BinaryMessage, b.Bytes())
```
vagababov: why don't we permit user to specify binary or text here?

websocket/connection.go
```diff
@@ -154,50 +161,53 @@ func (c *ManagedConnection) connect() (err error) {
...
+		return err
+	}
 
-		if err != nil {
-			return err
+	// Send the message to the channel if its an application level message
+	// and if that channel is set.
+	if c.messageChan != nil && (messageType == websocket.TextMessage || messageType == websocket.BinaryMessage) {
+		if message, _ := ioutil.ReadAll(reader); message != nil {
+			c.messageChan <- message
```
vagababov: Do we care to specify the type of the message? How the client is supposed to understand whether it was binary or text message?

websocket/connection.go
```diff
@@ -210,17 +220,22 @@ func (c *ManagedConnection) Send(msg interface{}) error {
 		return err
```
vagababov: I think it is better to encode outside the lock.
To hold lock for minimum possible time.

websocket/connection.go
vagababov: Now I read the whole file in detail, I can see how this *will* work, but I think my comment still stands. While `connect()` is spinning to return `Shutdown()` cannot finish, since it waits for this block to end and it's more than a minute in the worst case (i.e. k8s might force kill the pod by then).
I think select on closeChan and return error to the wait function is an appropriate way to deal with this.

websocket/connection_test.go
```diff
@@ -178,68 +184,51 @@ func TestCloseIgnoresNoConnection(t *testing.T) {
...
-	go conn.Close()
-	<-testConn.closeCalls
-
-	// Advance the reconnect loop until 'Close' is called.
-	<-testConn.nextReaderCalls
-	<-testConn.closeCalls
-
-	// Wait for the final call to 'Close' (when the loop is aborted)
-	<-testConn.closeCalls
+	var upgrader = websocket.Upgrader{}
```
vagababov:
```suggestion
	upgrader := websocket.Upgrader{}
```

websocket/connection.go
markusthoemmes: because until now we didn't needed. This was specifically built to encapsulate reconnecting logic for the queue-proxy -> autoscaler websocket connection. I'd say we add this if we need it?

websocket/connection.go
markusthoemmes: This reads super weird, I'd prefer to keep the explicit version implemented currently if you don't mind? The `c.read()` twice throws me off.

websocket/connection.go
markusthoemmes: Hm, I see what you mean. However the only part we can put into a function is

```golang
c.connection.Close()
c.connection = nil
```

And they theoretically need to hold a lock (which we'd need the caller to do). I'd prefer to keep the access to `c.connection` local to the functions they are in today for it makes clear that the lock is held by that function and thus this access is safe. WDYT?

websocket/connection.go
markusthoemmes: It isn't, added a "close" breaker here as well. The thinking was though, that a call to `Shutdown` closes the connection which would then cause keepalive to error out.

websocket/connection.go
markusthoemmes: Agreed, added a close signal to all long-running loops.

websocket/connection.go
markusthoemmes: I think at this point we don't care. I see the point but would rather leave a TODO for a later PR?

websocket/connection.go
markusthoemmes: The point of holding the lock in the outer scope though is to prevent the Send and Read to even try to do something while the connection is still being established. In theory that'd cause less dropped messages?

websocket/connection.go
markusthoemmes: I can't really. These have been based off what the hardcoded backoff was in the queue proxy before + some spreadsheeding to find a good fit here. The hockeystick seemed about right for this purpose.

markusthoemmes:
Thanks for the in-depth review @vagababov. Left comments + made changes accordingly. Let me know if there's more.

websocket/connection.go
vagababov: Do we expect to receive either, though?

websocket/connection.go
```diff
@@ -78,39 +79,46 @@ type ManagedConnection struct {
...
 
 // NewDurableConnection creates a new websocket connection, that
 // passes incoming messages to the given message channel. It can also
 // send messages to the endpoint it connects to.
 // The connection will continuously be kept alive and reconnected
 // in case of a loss of connectivity.
-func NewDurableConnection(target string, messageChan chan []byte) *ManagedConnection {
-	c := newConnection(target, messageChan)
+func NewDurableConnection(target string, messageChan chan []byte, logger *zap.SugaredLogger) *ManagedConnection {
+	websocketConnectionFactory := func(_ string) (rawConnection, error) {
```
vagababov: why even keep the parameter? `func() (rawConnection, error)` instead?

websocket/connection.go
vagababov: SG.

websocket/connection.go
vagababov: SG

websocket/connection.go
vagababov: I think the only interesting case that is remaining is:
- server sends messages in a stream
- this libraries client tries to shut down. The following things happen in parallel:
  - client stops consuming from `messagechan`
  - read consumes 2 messages (one is put in the channel the other is blocking `<-`)
- client calls `Shutdown`.

The race between `read()` blocked on writing to `messageChan` and `Shutdown`, causes `read()` to block forever and `Shutdown` to block forever as well, since it can't acquire the connection lock.

Probably the easiest way to solve this is to drain the channel in `Shutdown()` (we can presume library client is no longer interested in data, since Shutdown has been invoked).

markusthoemmes: I see your point, I'd rather not do something to that channel other than writing to it since we don't "own" it. WDYT about documenting that the user must drain the channel even after calling Shutdown to avoid deadlocks?

(Receiving messages really isn't used in any proper production code so I'm trying to not put too much effort into it since I want the logging fixes etc. out there)

vagababov: Well it's two things: `go c.Shutdown()` && `for {select {case <-data:default:}}`.
I need to think if/how we can solve this in a more elegant way, but for now, specifying an example usage in a comment should suffice.

websocket/connection.go
```diff
@@ -78,39 +78,53 @@ type ManagedConnection struct {
...
 // passes incoming messages to the given message channel. It can also
 // send messages to the endpoint it connects to.
 // The connection will continuously be kept alive and reconnected
 // in case of a loss of connectivity.
-func NewDurableConnection(target string, messageChan chan []byte) *ManagedConnection {
-	c := newConnection(target, messageChan)
+//
+// Note: The given channel needs to be drained after calling `Shutdown`
+// to not cause any deadlocks. If the channel's buffer is likely to be
+// filled, this needs to happen in seperate goroutines, i.e.
```
mattmoor-sockpuppet:
```suggestion
// filled, this needs to happen in separate goroutines, i.e.
```

websocket/connection.go
```diff
@@ -137,90 +151,111 @@ func newConnection(target string, messageChan chan []byte) *ManagedConnection {
...
+	var err error
 	wait.ExponentialBackoff(c.connectionBackoff, func() (bool, error) {
-		var conn rawConnection
-		conn, err = connFactory(c.target)
-		if err != nil {
-			return false, nil
-		}
-		c.connectionLock.Lock()
-		defer c.connectionLock.Unlock()
+		select {
```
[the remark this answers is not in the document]
markusthoemmes: This surfaced another deadlock potential. I relaxed the locking in `read` to only acquire a ReadLock and narrowed down the lock here again.

websocket/connection.go
```diff
@@ -137,90 +151,115 @@ func newConnection(target string, messageChan chan []byte) *ManagedConnection {
...
+	}
 
-		if err != nil {
-			return err
+	// Send the message to the channel if its an application level message
+	// and if that channel is set.
+	// TODO(markusthoemmes): Return the messageType along with the payload.
+	if c.messageChan != nil && (messageType == websocket.TextMessage || messageType == websocket.BinaryMessage) {
+		if message, _ := ioutil.ReadAll(reader); message != nil {
+			c.messageChan <- message
```
vagababov: Orthogonal: do we care about sending/receiving Ping/Pong messages to actually keep the connection alive?

markusthoemmes: Oh I haven't thought about this. This might be very helpful especially for the websocket connection of the activator which might go quite a while without ever sending a message. Great point, I'll open an issue to track that addition.

markusthoemmes:
@vagababov yes we can. Added that test and also threw in some test refactoring to cut down on LoC a bit there.

Issue #267, opened 2019-02-11
Istio VirtualService is missing tls definition


dicarlo2:
## Expected Behavior

knative-serving list of virtual services with a `tls` key works.

## Actual Behavior

It does not.

## Steps to Reproduce the Problem

1. Create a VirtualService with tls configuration

## Additional Info

https://istio.io/docs/reference/config/istio.networking.v1alpha3/#VirtualService

Related: https://github.com/knative/docs/issues/816

dicarlo2:
Also, weight is not required (though I can workaround this, of course): https://github.com/dicarlo2/pkg/blob/5db4e8e38e5682fd28c4849ffa3135da7787edb6/apis/istio/v1alpha3/virtualservice_types.go#L297

PR #268, opened 2019-02-11
Fix VirtualService tcp and tls definitions

dicarlo2:
Fixes a few issues with the VirtualService definition. 

I don't know how to build or test this repo, so no testing was done. https://github.com/knative/pkg/blob/master/DEVELOPMENT.md does not describe how, though it says: "Once you reach this point you are ready to do a full build and deploy as described below." FWIW - I'd guess that this is a pretty standard Go repo, but I'm a Go n00b, so I don't know what to do. Happy to take instructions/pointers to properly build/test this.

Reference: https://istio.io/docs/reference/config/istio.networking.v1alpha3

Fixes: https://github.com/knative/pkg/issues/267


apis/istio/v1alpha3/virtualservice_types.go
```diff
@@ -474,7 +476,48 @@ type TCPRoute struct {
 	// Currently, only one destination is allowed for TCP services. When TCP
```
vaikas-google: Is this comment still up to date? If so, what are the semantics now with this being an array?

dicarlo2: I think multiple destinations are allowed, I'll remove the comment.

dicarlo2:
ping @tcnghia @vaikas-google this should be good to go.

apis/istio/v1alpha3/virtualservice_types.go
```diff
@@ -484,22 +524,47 @@ type L4MatchAttributes struct {
 	// a.b.c.d/xx form or just a.b.c.d. This is only valid when the
 	// destination service has several IPs and the application explicitly
 	// specifies a particular IP.
```
vaikas-google: does this comment also need updating? 'particular IP'? Match L554-555?

apis/istio/v1alpha3/virtualservice_types.go
```diff
@@ -471,10 +473,48 @@ type TCPRoute struct {
...
+//         host: reviews.prod.svc.cluster.local
+// ```
+type TLSRoute struct {
+	// REQUIRED. Match conditions to be satisfied for the rule to be
+	// activated. All conditions inside a single match block have AND
+	// semantics, while the list of match blocks have OR semantics. The rule
+	// is matched if any one of the match blocks succeed.
+	Match []TLSMatchAttributes `json:"match"`
+
+	// The destination to which the connection should be forwarded to.
```
vaikas-google: destination => destinations
I assume also the weights have to add to 100%? Might be worth explicitly stating this.

apis/istio/v1alpha3/virtualservice_types.go
```diff
@@ -471,10 +473,48 @@ type TCPRoute struct {
...
+// traffic arriving at port 443 of gateway called mygateway to internal
+// services in the mesh based on the SNI value.
+//
+// ```yaml
+// kind: VirtualService
+// metadata:
+//   name: bookinfo-sni
+// spec:
+//   hosts:
+//   - &quot;*.bookinfo.com&quot;
```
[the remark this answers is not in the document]
dicarlo2: Oops, bad copy-paste from istio docs.

dicarlo2:
ping @tcnghia @vaikas-google let me know if there's anything else.

PR #269, opened 2019-02-11
Add ping message verification to the websocket connection.

markusthoemmes:
This should further increase the durability of our websocket connections. See https://github.com/knative/serving/issues/3138

Ping messages are sent to actively probe the websocket connection. By default, the websocket servers we use send a pong message back. This can be used to break indefinite read calls.

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -185,6 +185,10 @@ func (c *ManagedConnection) connect() error {
 				return false, nil
 			}
 
+			// Setting the read deadline will cause NextReader in read
+			// to fail if it is exceeded. This deadline is reset each
+			// time we receive a pong message so we know the connection
+			// is still intact.
 			conn.SetReadDeadline(time.Now().Add(pongTimeout))
 			conn.SetPongHandler(func(string) error {
 				conn.SetReadDeadline(time.Now().Add(pongTimeout))
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -140,6 +140,7 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 	// Keep sending pings 3 times per pongTimeout interval.
 	go func() {
 		ticker := time.NewTicker(pongTimeout / 3)
+		defer ticker.Stop()
 		for {
 			select {
 			case <-ticker.C:
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -40,6 +40,10 @@ var (
 
 	// errShuttingDown is returned internally once the shutdown signal has been sent.
 	errShuttingDown = errors.New("shutdown in progress")
+
+	// pongTimeout defines the amount of time allowed between two pongs to arrive
+	// before the connection is considered broken.
+	pongTimeout = 10 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


markusthoemmes:
/assign @vagababov 

This was your idea, so let's continue our review journey 🙂.

websocket/connection.go
```diff
@@ -130,6 +137,21 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 		}
 	}()
 
+	// Keep sending pings 3 times per pongTimeout interval.
+	go func() {
+		ticker := time.NewTicker(pongTimeout / 3)
```
vagababov: they don't divide evenly 😞

websocket/connection.go
```diff
@@ -130,6 +137,21 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
...
+		for {
+			select {
+			case <-ticker.C:
+				if err := c.write(websocket.PingMessage, []byte{}); err != nil {
+					logger.Errorw("Failed to send ping message", zap.Error(err))
+				}
+			case <-c.closeChan:
```
vagababov: `ticker.Stop()` here, to avoid resource leaks.
Or `defer ticker.Stop()` after creation.

websocket/connection.go
```diff
@@ -130,6 +137,21 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
...
 	}()
 
+	// Keep sending pings 3 times per pongTimeout interval.
+	go func() {
+		ticker := time.NewTicker(pongTimeout / 3)
+		for {
+			select {
+			case <-ticker.C:
+				if err := c.write(websocket.PingMessage, []byte{}); err != nil {
+					logger.Errorw("Failed to send ping message", zap.Error(err))
```
vagababov: should this trigger a reconnect?

websocket/connection.go
```diff
@@ -162,6 +184,12 @@ func (c *ManagedConnection) connect() error {
 				return false, nil
 			}
 
+			conn.SetReadDeadline(time.Now().Add(pongTimeout))
+			conn.SetPongHandler(func(string) error {
```
vagababov: Add a comment here about what's going on.

websocket/connection.go
markusthoemmes: in principle we could make it so. However, then we should also check the errors of usual `Send` calls and trigger reconnects accordingly.

Ultimately, the ping/pong mechanism should catch this reconnect desire in this case. Being quicker on Send errors seems a different can of worms I wouldn't want to open in this PR.

websocket/connection.go
markusthoemmes: Does it matter that much? 😂

markusthoemmes:
This was a legitimate race found on Shutdown. Added a WaitGroup to ensure everything's shut down after `Shutdown` finishes.

websocket/connection.go
vagababov: Well, it's a loss of 10% (1sec) of precision :) If the timeout was defined in ms it'd just 1ms, though :)

PR #273, opened 2019-02-12
Adding support for Zipkin tracing in SpoofClient

dushyanthsc:
Existing support for Zipkin tracing pushes the responsibility of logging zipkin traces to individual tests. This change moves the responsibility to SpoofClient, so that all tests get it by default. Individual tests
can disable this functionality by just disabling a ZipkinTracingEnabled flag in zipkin package.

Zipkin tracing in Spoofing client checks to see if the ZipkinTracingEnabled flag is enabled and only logs traces for requests that have encountered Server errors i.e if the http.Response.StatusCode between 500 to 599.

For SpoofingClient.Poll calls we only log trace of the final SpoofingClient.Do made.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Knative Authors
+Copyright 2019 The Knative Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -21,7 +21,6 @@ package spoof
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -47,6 +46,9 @@ const (
 	// TODO(tcnghia): These probably shouldn't be hard-coded here?
 	istioIngressNamespace = "istio-system"
 	istioIngressName      = "istio-ingressgateway"
+	// Name of the temporary HTTP header that is added to http.Request to indicate that
+	// it is a SpoofClient.Poll request. This header is removed before making call to backend.
+	pollReqHeader = "X-Kn-Poll-Request-Do-Not-Trace"
 )
 
 // Response is a stripped down subset of http.Response. The is primarily useful
@@ -185,6 +187,12 @@ func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 	traceContext, span := trace.StartSpan(req.Context(), "SpoofingClient-Trace")
 	defer span.End()
 
+	// Check to see if the call to this method is coming from a Poll call.
+	logZipkinTrace := true
+	if req.Header.Get(pollReqHeader) != "" {
+		req.Header.Del(pollReqHeader)
+		logZipkinTrace = false
+	}
 	resp, err := sc.Client.Do(req.WithContext(traceContext))
 	if err != nil {
 		return nil, err
@@ -198,12 +206,18 @@ func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 		return nil, err
 	}
 
-	return &Response{
+	spoofResp := &Response{
 		Status:     resp.Status,
 		StatusCode: resp.StatusCode,
 		Header:     resp.Header,
 		Body:       body,
-	}, nil
+	}
+
+	if logZipkinTrace {
+		sc.logZipkinTrace(spoofResp)
+	}
+
+	return spoofResp, nil
 }
 
 // Poll executes an http request until it satisfies the inState condition or encounters an error.
@@ -214,6 +228,10 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
 	)
 
 	err = wait.PollImmediate(sc.RequestInterval, sc.RequestTimeout, func() (bool, error) {
+		// As we may do multiple Do calls as part of a single Poll we add this temporary header
+		// to the request to indicate to Do method not to log Zipkin trace, instead it is
+		// handled by this method itself.
+		req.Header.Add(pollReqHeader, "True")
 		resp, err = sc.Do(req)
 		if err != nil {
 			if err, ok := err.(net.Error); ok && err.Timeout() {
@@ -226,13 +244,24 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
 		return inState(resp)
 	})
 
+	if resp != nil {
+		sc.logZipkinTrace(resp)
+	}
+
 	return resp, err
 }
 
-// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
-func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+// logZipkinTrace provides support to log Zipkin Trace for param: spoofResponse
+// We only log Zipkin trace for HTTP server errors i.e for HTTP status codes between 500 to 600
+func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
+	if !zipkin.ZipkinTracingEnabled || spoofResp.StatusCode < http.StatusInternalServerError || spoofResp.StatusCode >= 600 {
+		return
+	}
+
+	traceID := spoofResp.Header.Get(zipkin.ZipkinTraceIDHeader)
 	if err := zipkin.CheckZipkinPortAvailability(); err == nil {
-		return errors.New("port-forwarding for Zipkin is not-setup. Failing Zipkin Trace retrieval")
+		sc.logf("port-forwarding for Zipkin is not-setup. Failing Zipkin Trace retrieval")
+		return
 	}
 
 	sc.logf("Logging Zipkin Trace: %s", traceID)
@@ -242,20 +271,21 @@ func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
 	time.Sleep(5 * time.Second)
 	resp, err := http.Get(zipkinTraceEndpoint)
 	if err != nil {
-		return fmt.Errorf("Error retrieving Zipkin trace: %v", err)
+		sc.logf("Error retrieving Zipkin trace: %v", err)
+		return
 	}
 	defer resp.Body.Close()
 
 	trace, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return fmt.Errorf("Error reading Zipkin trace response: %v", err)
+		sc.logf("Error reading Zipkin trace response: %v", err)
+		return
 	}
 
 	var prettyJSON bytes.Buffer
 	if error := json.Indent(&prettyJSON, trace, "", "\t"); error != nil {
-		return fmt.Errorf("JSON Parser Error while trying for Pretty-Format: %v, Original Response: %s", error, string(trace))
+		sc.logf("JSON Parser Error while trying for Pretty-Format: %v, Original Response: %s", error, string(trace))
+		return
 	}
 	sc.logf("%s", prettyJSON.String())
-
-	return nil
 }
--- a/test/zipkin/doc.go
+++ b/test/zipkin/doc.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


test/spoof/spoof.go
```diff
@@ -47,6 +46,9 @@ const (
 	// TODO(tcnghia): These probably shouldn't be hard-coded here?
 	istioIngressNamespace = "istio-system"
 	istioIngressName      = "istio-ingressgateway"
+	// Name of the temporary http header that is added to http.Request to indicate that
```
vagababov:
```suggestion
	// Name of the temporary HTTP header that is added to http.Request to indicate that
```

test/spoof/spoof.go
```diff
@@ -47,6 +46,9 @@ const (
...
+	// it is a SpoofClient.Poll request. This header is removed before making call to backend.
+	pollRequest = "POLL_REQUEST"
```
vagababov: Don't we want this to look like canonical header? `Kn-Poll-Request`, e.g.?
In addition name it `do-not-trace` or something, since that is the purpose?

test/spoof/spoof.go
```diff
@@ -223,13 +241,23 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+
 	return resp, err
 }
 
-// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
-func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+// logZipkinTrace provides support to log Zipkin Trace for param: httpResponse
+// We only log Zipkin trace for http server errors i.e for http status codes between 500 to 600
+func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
+	if !zipkin.ZipkinTracingEnabled || !(spoofResp.StatusCode <= 500 && spoofResp.StatusCode < 600) {
```
vagababov: I think, something like this:
```suggestion
	if !zipkin.ZipkinTracingEnabled || spoofResp.StatusCode < http.StatusInternalServerError || spoofResp.StatusCode >= 600) {
```
As is the condition seems wrong.
And probably this has no unit tests, being a test client.

test/spoof/spoof.go
```diff
@@ -223,13 +241,23 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+	if resp != nil {
+		sc.logZipkinTrace(resp)
+	}
+
 	return resp, err
 }
 
-// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
-func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+// logZipkinTrace provides support to log Zipkin Trace for param: httpResponse
```
vagababov:
```suggestion
// logZipkinTrace provides support to log Zipkin Trace for param: spoofResponse
```

test/spoof/spoof.go
```diff
@@ -223,13 +241,23 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+// We only log Zipkin trace for http server errors i.e for http status codes between 500 to 600
```
vagababov:
```suggestion
// We only log Zipkin trace for http server errors i.e for HTTP status codes between 500 to 600
```

test/zipkin/util.go
```diff
@@ -48,13 +48,21 @@ const (
...
+	// ZipkinTracingEnabled global variable indicating if zipkin is enabled.
+	ZipkinTracingEnabled = false
+)
 
 // SetupZipkinTracing sets up zipkin tracing which involves:
 // 1. Setting up port-forwarding from localhost to zipkin pod on the cluster
 //    (pid of the process doing Port-Forward is stored in a global variable).
 // 2. Enable AlwaysSample config for tracing.
 func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) error {
+	if ZipkinTracingEnabled {
```
vagababov: This is racy and not safe.
I highly recommend to switch to `sync.Once` to do initialization like this.

test/zipkin/util.go
```diff
@@ -108,4 +122,4 @@ func CheckZipkinPortAvailability() error {
 	}
 	server.Close()
 	return nil
-}
+}
```
vagababov: Add a new line.

test/zipkin/util.go
dushyanthsc: Before I make the change, wouldn't protection through mutex be more suitable here.

The intended use here was to have this method be called logical by test.setup and CleanupZipkinTracing be called by test.Cleanup, hence synchronization protection was not added.

vagababov: So is this per-test or per test-suite?
If first, then mutex, if second then Once.

test/spoof/spoof.go
dushyanthsc: I am not sure if I have addressed the comment correctly. Partly because I didn't completely understand the header you were looking for. Please let know any changes required.

test/spoof/spoof.go
dushyanthsc: Yes, I discovered this while doing End to End test,  have fixed the condition check.

test/spoof/spoof.go
dushyanthsc: Fixed.

test/spoof/spoof.go
dushyanthsc: Fixed.

test/zipkin/util.go
dushyanthsc: It would be per-test. I have changed to using RWMutex.

test/zipkin/util.go
dushyanthsc: Fixed.

test/spoof/spoof.go
vagababov: Eh what I meant, is that doesn't look like a HTTP header. At least no a canonical one (e.g. `Content-Length`, or `Host`). What I recomment is using naming it something like `X-Knative-Dev-Poll-Request`.

vagababov: Yeah, I see what you did. Works for me. (`X-` prefix might be a good touch, but this is test, so meh).

test/zipkin/util.go
```diff
@@ -107,5 +130,17 @@ func CheckZipkinPortAvailability() error {
 		return err
 	}
 	server.Close()
+
 	return nil
 }
+
+// LockAndReadZipkinStatusFlag RLocks and returns zipkinTracingEnabled flag
+func LockAndReadZipkinStatusFlag() bool {
```
vagababov: why not just lock and unlock here? Having them separate is error prone
```
func GetZipkinStatusFlag() {
  rlock
  defer runlock
  return value
}
```

dushyanthsc: This would not work right because the read lock needs to be held until the client/method using the flag has completed retrieving zipkin trace. So for e.g. SpoofingClient reads the flag and sees that it is `true` and goes ahead and makes a Http Request. While the request is being satisfied some other routine can come can clean up the port-forwarding process because the Rlock was released and it was able to obtain a write lock. The read call on the Spoofingclient would fail as a result.

dushyanthsc: @vagababov I Hope I answered your question. Please let me know if I have missed anything.

test/spoof/spoof.go
```diff
@@ -223,13 +241,26 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
...
+
 	return resp, err
 }
 
-// LogZipkinTrace provides support to log Zipkin Trace for param: traceID
-func (sc *SpoofingClient) LogZipkinTrace(traceID string) error {
+// logZipkinTrace provides support to log Zipkin Trace for param: spoofResponse
+// We only log Zipkin trace for HTTP server errors i.e for HTTP status codes between 500 to 600
+func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
+	if !zipkin.LockAndReadZipkinStatusFlag() || spoofResp.StatusCode < http.StatusInternalServerError || spoofResp.StatusCode >= 600 {
```
vagababov: I think you need to do
```
x := zipking.LockAndRead...()
defer zipkin.Unlock...
// do stuff.
```

vagababov:
Sorry, somehow my response didn't get published.

test/zipkin/doc.go
```diff
@@ -0,0 +1,41 @@
...
+		zipkinTracingEnabled flag to false.
+
+A general flow for a Test Suite to use Zipkin Tracing support is as follows:
+
+		1. Call SetupZipkinTracing(*kubernetes.Clientset).
+		2. Use SpoofingClient to make HTTP requests.
+		3. Call CleanupZipkinTracingSetup on cleanup.
+
+ */
+package zipkin
```
mattmoor-sockpuppet:
```suggestion
package zipkin

```

test/zipkin/util.go
dushyanthsc: Thanks, if we are reverting to per-suite as per your earlier suggestion sync.Once makes more sense. I have updated the changeset with Sync.Once approach.

dushyanthsc:
@vagababov I have updated the PR addressing your comments. Please take a look when you get a chance. Thanks!

test/zipkin/util.go
```diff
@@ -48,54 +47,74 @@ const (
...
+func CleanupZipkinTracingSetup() {
+	teardownOnce.Do(func() {
+		if !ZipkinTracingEnabled {
+			return
+		}
+		logger := logging.GetContextLogger("SpoofUtil")
+
+		ps := os.Process{Pid: zipkinPortForwardPID}
+		if err := ps.Kill(); err != nil {
+			logger.Errorf("Encoutered error killing port-forward process in CleanupZipkingTracingSetup() : %v", err)
```
vagababov:
```suggestion
			logger.Errorf("Encountered error killing port-forward process in CleanupZipkingTracingSetup(): %v", err)
```

test/zipkin/doc.go
```diff
@@ -0,0 +1,41 @@
...
+		an internal flag zipkinTracingEnabled to true.
+
+	CleanupZipkinTracingSetup() error
+		CleanupZipkinTracingSetup cleans up zipkin tracing setup by cleaning up the
+		port-forwarding setup by call to SetupZipkinTracing. This method also sets
+		zipkinTracingEnabled flag to false.
+
+A general flow for a Test Suite to use Zipkin Tracing support is as follows:
+
+		1. Call SetupZipkinTracing(*kubernetes.Clientset).
```
vagababov: I think it'd be useful to say that it should be done in `TestMain` (which also means that Clientset  has to be initialized once, which is also not a bad idea).

test/zipkin/util.go
dushyanthsc: Fixed.

test/zipkin/doc.go
```diff
@@ -0,0 +1,41 @@
...
+		zipkinTracingEnabled flag to false.
+
+A general flow for a Test Suite to use Zipkin Tracing support is as follows:
+
+		1. Call SetupZipkinTracing(*kubernetes.Clientset) in TestMain.
+		2. Use SpoofingClient to make HTTP requests.
+		3. Call CleanupZipkinTracingSetup on cleanup after tests are executed.
+
+ */
+package zipkin
```
mattmoor-sockpuppet:
```suggestion
package zipkin

```

test/zipkin/doc.go
dushyanthsc: Makes sense modified the instructions.

test/zipkin/util.go
```diff
@@ -48,54 +47,74 @@ const (
...
-
-	// Applying AlwaysSample config to ensure we propagate zipkin header for every request made by this client.
-	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
-
-	logger.Info("Successfully setup SpoofingClient for Zipkin Tracing")
-
-	return nil
+func SetupZipkinTracing(kubeClientset *kubernetes.Clientset) {
+	setupOnce.Do(func() {
+		logger := logging.GetContextLogger("SpoofUtil")
```
srinivashegde86: we are using FormatLogger everywhere now instead of context logger - https://github.com/knative/pkg/blob/master/test/spoof/spoof.go#L79

We should use that here as well.

dushyanthsc: How  do you instantiate a FormatLogger, all places I see in the code base its through Test. Can you please point me to the documentation on how to use it, also it should not be Spoof Client? it should be in the logging package with proper documentation.

srinivashegde86: Maybe we can move it to logging. I am not sure if there is much documentation on it. You can check on existing tests on how to use it.

PR #275, opened 2019-02-13
Initial commit for the webhook to set the annotations about mutator.

vagababov:
The user that created or updated the resource will be set in the
annotations.

Part of knative/serving#1642.

/cc @mattmoor

### DIFF
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -47,3 +47,8 @@ type Listable interface {
 
 	GetListType() runtime.Object
 }
+
+// Annotatable indicates that a particular type applies various annotations.
+type Annotatable interface {
+	AnnotateUserInfo(previous Annotatable, userName string)
+}
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -17,6 +17,7 @@ limitations under the License.
 package apis
 
 import (
+	authenticationv1 "k8s.io/api/authentication/v1"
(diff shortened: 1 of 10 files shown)


apis/interfaces.go
```diff
@@ -47,3 +47,8 @@ type Listable interface {
 
 	GetListType() runtime.Object
 }
+
+// Annotatable indicates that a particular type applies various annotations.
+type Annotatable interface {
+	AnnotateUserInfo(previous Annotatable, userName string)
```
mattmoor: I'd pass in the [UserInfo](https://godoc.org/k8s.io/api/authentication/v1#UserInfo)

webhook/webhook.go
```diff
@@ -126,6 +127,7 @@ type AdmissionController struct {
 type GenericCRD interface {
 	apis.Defaultable
 	apis.Validatable
+	apis.Annotatable
```
mattmoor: I wouldn't require this, I'd check this like we do for `apis.Immutable`

vagababov: Done.

apis/interfaces.go
vagababov: Hm, from one point of view sure, from another, do we have ideas how to utilize other fields?

mattmoor: What if I just want the UID for PII reasons :)

PR #277, opened 2019-02-14
Don't close the http.Request body

dprotaso:
As per net.http package for `http.Request.Body`

```
The Server will close the request body. The ServeHTTP
Handler does not need to.
```


vagababov:
:-(

PR #278, opened 2019-02-14
Add status matchers to help unify test code.

markusthoemmes:
Some of our tests use a common pattern to match these, but some fail to provide all the debug information we want. This adds generalized versions to be used in tests so test output is uniform.

### DIFF
--- a/test/request.go
+++ b/test/request.go
@@ -50,20 +50,22 @@ func Retrying(rc spoof.ResponseChecker, codes ...int) spoof.ResponseChecker {
 	}
 }
 
-// IsStatusCode checks that the response code is equal to the given one.
-func IsStatusCode(code int) spoof.ResponseChecker {
+// IsOneOfStatusCodes checks that the response code is equal to the given one.
+func IsOneOfStatusCodes(codes ...int) spoof.ResponseChecker {
 	return func(resp *spoof.Response) (bool, error) {
-		if resp.StatusCode == code {
-			return true, nil
+		for code := range codes {
+			if resp.StatusCode == code {
+				return true, nil
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


test/request.go
```diff
@@ -50,6 +50,22 @@ func Retrying(rc spoof.ResponseChecker, codes ...int) spoof.ResponseChecker {
 	}
 }
 
+// IsStatusCode checks that the response code is equal to the given one.
+func IsStatusCode(code int) spoof.ResponseChecker {
```
vagababov: Would it be useful, to extend this to `codes []int` and check if any of matches? E.g. below we can check for 200 and 201.

markusthoemmes: In this case, yes.

I won't change the Matcher below though. I'd like us to write as tight checks as possible (I'm in the process of cleaning them all up). If we expect a 201, we should explicitly wait for that in the test.

test/request.go
```diff
@@ -50,6 +50,24 @@ func Retrying(rc spoof.ResponseChecker, codes ...int) spoof.ResponseChecker {
...
+// IsOneOfStatusCodes checks that the response code is equal to the given one.
+func IsOneOfStatusCodes(codes ...int) spoof.ResponseChecker {
+	return func(resp *spoof.Response) (bool, error) {
+		for code := range codes {
+			if resp.StatusCode == code {
+				return true, nil
+			}
+		}
+
+		return true, fmt.Errorf("status = %d, want one of: %v, body = %s", resp.StatusCode, codes, string(resp.Body))
```
srinivashegde86: return false, err?

markusthoemmes: @srinivashegde86 this is consistent with the other matchers (see `MatchesBody` below).

srinivashegde86: oops. this bool represents whether this is done and not indicate if it passes or fails..

PR #282, opened 2019-02-15
Simplify and comment AppenRandomString

vagababov:
Port and rewrite the comment from the original.
Remove the sync.Once(). Initializing during package initialization is just as good.
Mutex doesn't have to be a pointer

/cc @mattmoor @srinivashegde86

### DIFF
--- a/test/helpers/data.go
+++ b/test/helpers/data.go
@@ -3,7 +3,6 @@ package helpers
 import (
 	"math/rand"
 	"strings"
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


test/helpers/data.go
```diff
@@ -15,18 +15,20 @@ const (
...
-	once     sync.Once
+	rndMutex sync.Mutex
 )
 
-func initSeed() {
-	seed := time.Now().UTC().UnixNano()
-	r = rand.New(rand.NewSource(seed))
-	rndMutex = &sync.Mutex{}
+func init() {
+	r = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
```
mattmoor: I'd just `rand.Seed` here and save the mutex

test/helpers/data_test.go
```diff
@@ -0,0 +1,16 @@
+package helpers
```
mattmoor: license

vagababov:
Part of: knative/serving/#3174.

test/helpers/data_test.go
vagababov: Done.
Main file actually also lacked one

test/helpers/data.go
vagababov: Done.

PR #283, opened 2019-02-18
Fix range expression in 'IsOneOfStatusCodes' to actually match the codes.

markusthoemmes:
So sad the compiler did not catch this 😢


vagababov:
Ah , if I got a cent each time I did the same mistake :)

PR #285, opened 2019-02-18
Adapt the spoofing client to take a "FormatLogger" instead of a BaseLogger

mattmoor:
This is a breaking change, that can be fixed up by replacing `logger,` with `logger.Infof,` in the argument lists for these functions.

This is part of a larger cleanup to allow us to use alternate log methods like `t.Logf` in place of `BaseLogger`.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -29,7 +29,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/knative/pkg/test/logging"
 	"github.com/knative/pkg/test/zipkin"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


mattmoor:
/hold

Still need a review, but I will stage things so that this can land only once the downstream changes are also ready to land.

test/spoof/spoof.go
```diff
@@ -75,6 +74,8 @@ var _ Interface = (*SpoofingClient)(nil)
 // https://github.com/kubernetes/apimachinery/blob/cf7ae2f57dabc02a3d215f15ca61ae1446f3be8f/pkg/util/wait/wait.go#L172
 type ResponseChecker func(resp *Response) (done bool, err error)
 
+type FormatLogger func(template string, args ...interface{})
```
vagababov: I am not thrilled that we eliminate non formatting loggers, but for tests it probably doesn't matter much.

mattmoor: Yeah, if the methods were called the same things on `logger` and `testing.T` I would have defined an interface, but it's `Info[f]` and `Log[f]` respectively :(

PR #292, opened 2019-02-21
Decide whether to record metrics according to current metrics config

yanweiguo:
* Introduce a new observability flag `metrics.allow-stackdriver-custom-metrics` to decide whether to send none built-in metrics to Stackdriver as custom metrics, which will cause extra Stackdriver charge.
* Introduce a `Record` function wrapping the OpenCensus `Record` function. Make the decision whether to send metrics based on current config metrics.
* It should record a metric if one of the following conditions satisfies:
  * the backend is not Stackdriver
  * the backend is Stackdriver and it is allowed to use custom metrics
  * the backend is Stackdriver and the metric is "knative_revison" built-in metric.


metrics/config.go
```diff
@@ -53,16 +55,46 @@ type metricsConfig struct {
...
+	// uploaded to. This is not the GCP project ID.
+	stackdriverProjectID string
+	// allowStackdriverCustomMetrics indicates whether it is allowed to send metrics to
+	// Stackdriver using "global" resource type and custom metric type if the
+	// metrics are not supported by "knative_revision" resource type. Setting this
+	// flag to "true" could cause extra Stackdriver charge.
+	// If backendDestination is not Stackdriver, this is ignored.
+	allowStackdriverCustomMetrics bool
+	// True if backendDestination equals to "stackdriver". Store this in a variable
+	// to reduce string comparision operations.
```
mattmoor-sockpuppet:
```suggestion
	// to reduce string comparison operations.
```

controller/stats_reporter_test.go
```diff
@@ -79,6 +86,17 @@ func TestReportReconcile(t *testing.T) {
...
 }
 
+func initMetricsConfig(t *testing.T) {
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "config-observability",
+		},
+		Data: map[string]string{},
+	}
+	u := metrics.UpdateExporterFromConfigMap("testDomain", "testComponent", TestLogger(t))
```
vagababov: just  `metrics...(cm)`.

metrics/config.go
```diff
@@ -53,16 +55,46 @@ type metricsConfig struct {
...
+	// component, e.g. "custom.googleapis.com/knative.dev/serving/activator".
+	// Store this in a variable to reduce string join operations.
+	stackdriverCustomMetricTypePrefix string
 }
 
 func getMetricsConfig(m map[string]string, domain string, component string, logger *zap.SugaredLogger) (*metricsConfig, error) {
 	var mc metricsConfig
+
+	if domain == "" {
+		return nil, errors.New("Metrics domain cannot be empty")
```
vagababov:
```suggestion
		return nil, errors.New("metrics domain cannot be empty")
```

metrics/config.go
```diff
@@ -53,16 +55,46 @@ type metricsConfig struct {
...
+	}
+	mc.domain = domain
+
+	if component == "" {
+		return nil, errors.New("Metrics component name cannot be empty")
```
vagababov:
```suggestion
		return nil, errors.New("metrics component name cannot be empty")
```

metrics/config.go
```diff
@@ -86,6 +118,16 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
 	// metrics exporter.
 	if mc.backendDestination == Stackdriver {
 		mc.stackdriverProjectID = m[stackdriverProjectIDKey]
+		mc.isStackdriverBackend = true
+		mc.stackdriverMetricTypePrefix = path.Join(domain, component)
+		mc.stackdriverCustomMetricTypePrefix = path.Join(customMetricTypePrefix, component)
+		if ascmStr, ok := m[allowStackdriverCustomMetricsKey]; ok && ascmStr != "" {
+			ascmBool, err := strconv.ParseBool(ascmStr)
+			if err != nil {
+				return nil, fmt.Errorf("Invalid %s value \"%s\"", allowStackdriverCustomMetricsKey, ascmStr)
```
vagababov:
```suggestion
				return nil, fmt.Errorf("Invalid %s value %q", allowStackdriverCustomMetricsKey, ascmStr)
```

metrics/config.go
```diff
@@ -96,26 +138,17 @@ func getMetricsConfig(m map[string]string, domain string, component string, logg
 	// push anything but just responds to pull requests, and shorter durations
 	// do not really hurt the performance and we rely on the scraping configuration.
 	if repStr, ok := m[reportingPeriodKey]; ok && repStr != "" {
-		if repInt, err := strconv.Atoi(repStr); err == nil {
-			mc.reportingPeriod = time.Duration(repInt) * time.Second
-		} else {
-			return nil, fmt.Errorf("Invalid reporting-period-seconds value \"%s\"", repStr)
+		repInt, err := strconv.Atoi(repStr)
+		if err != nil {
+			return nil, fmt.Errorf("Invalid %s value \"%s\"", reportingPeriodKey, repStr)
```
vagababov:
```suggestion
			return nil, fmt.Errorf("Invalid %s value %q", reportingPeriodKey, repStr)
```

metrics/config.go
```diff
@@ -134,28 +167,30 @@ func UpdateExporterFromConfigMap(domain string, component string, logger *zap.Su
 			return
 		}
 
-		if isMetricsConfigChanged(newConfig) {
-			if err := newMetricsExporter(newConfig, logger); err != nil {
+		if isNewExporterRequired(newConfig) {
```
vagababov: Should this function return an error?

metrics/record.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/knative/pkg/metrics/metricskey"
+	"go.opencensus.io/stats"
+)
+
+// Record decides whether to record one measurement based on current
+// metrics config. If yes, it records measurements via OpenCensus.
+func Record(ctx context.Context, ms stats.Measurement) {
+	mc := getCurMetricsConfig()
+	if mc == nil {
+		logger := logging.FromContext(ctx)
```
vagababov: join these two lines?
Also is it useful at all?

metrics/record.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	// Should record if one of the following conditions satisfies:
+	// 1) the backend is not Stackdriver
+	// 2) the backend is Stackdriver and it is allowed to use custom metrics
+	// 3) the backend is Stackdriver and the metric is "knative_revison" built-in metric.
+	if !mc.isStackdriverBackend || mc.allowStackdriverCustomMetrics {
+		stats.Record(ctx, ms)
+	} else {
+		metricType := path.Join(mc.stackdriverMetricTypePrefix, ms.Measure().Name())
+		if metricskey.KnativeRevisionMetrics.Has(metricType) {
```
vagababov: can't you check before joining?

metrics/record_test.go
```diff
@@ -0,0 +1,120 @@
...
+		checkLastValueData(t, test.measurement.Measure().Name(), 3) // The value is still the last one of shouldReportCases
+	}
+}
+
+func checkLastValueData(t *testing.T, name string, wantValue float64) {
+	t.Helper()
+	if row := checkRow(t, name); row != nil {
+		if s, ok := row.Data.(*view.LastValueData); !ok {
+			t.Error("Reporter.Report() expected a LastValueData type")
+		} else if s.Value != (float64)(wantValue) {
```
vagababov: wantValue shouldn't be cast, it's already float64

metrics/record_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func checkRow(t *testing.T, name string) *view.Row {
+	t.Helper()
+	d, err := view.RetrieveData(name)
+	if err != nil {
+		t.Fatalf("Reporter.Report() error = %v, wantErr %v", err, false)
+		return nil
+	}
+	if len(d) != 1 {
+		t.Fatalf("Reporter.Report() len(d) %v, want %v", len(d), 1)
```
vagababov:
```suggestion
		t.Fatalf("Reporter.Report() len(d) = %d, want 1", len(d))
```

metrics/config.go
yanweiguo: I don't see any case it could return an error.

metrics/record.go
yanweiguo: Which two lines?

I just want to send this log for debugging purpose in case this function is called before the exporter is initialized..

metrics/record.go
yanweiguo: I can't. It checks the join result.

metrics/config.go
vagababov: still see the old formatting.

metrics/config.go
vagababov: ditto

metrics/config.go
vagababov: well you log an error. If it's not an error, then logging it as an error, also doesn't make sense?

metrics/record.go
vagababov: `logging.FromContext(ctx).Error(....)`.
And what I mean is: is it useful to log this at all?

metrics/config.go
yanweiguo: The error is returned from `newMetricsExporter`. I think you were asking whether `isNewExporterRequired` should return an error?

metrics/record.go
yanweiguo: I think it is useful in case we don't see the expected metrics. If we pass sufficient information in `ctx`, the logs should include extra information. I changed this log from `Error` to `Warn`.

metrics/config.go
yanweiguo: You need refresh the page?

grantr:
/assign mattmoor

I'm afraid I don't know whether this can be approved.

yanweiguo:
> I want to clarify if "metrics" here is a shorthand for "observation", which could feed into multiple paths including log records and tracing, or if this is specifically intended to only cover realtime aggregated data.
> 
> More review forthcoming.

"metrics" here only covers realtime aggregated data reported by knative components, i.e. autoscaler of knative serving, webhook of knative build. Metric example: `knative.dev/serving/activator/request_count`

PR #297, opened 2019-02-27
Use FormatLogger instead of func parameter for zipoking tracing calls

srinivashegde86:


dushyanthsc:
/lgtm 

Add some description.

PR #300, opened 2019-02-28
Fix the sugared logger usage

vagababov:
Without `w` the logged messages are quite useless and annoying.

/lint
/cc @evankanderson


controller/controller.go
```diff
@@ -270,20 +270,20 @@ func (c *Impl) processNextWorkItem() bool {
 	// resource to be synced.
 	if err = c.Reconciler.Reconcile(ctx, key); err != nil {
 		c.handleErr(err, key)
-		logger.Errorf("Reconcile failed. Time taken: %v.", time.Now().Sub(startTime))
+		logger.Infof("Reconcile failed. Time taken: %v.", time.Since(startTime))
```
evankanderson: Change from Error to Info for reconcile failed?

vagababov: yep, this just logs the timing. `handleError` above, does the Error level error logging anyway.

PR #302, opened 2019-03-02
Separate common monitoring methods into monitoring package

srinivashegde86:
Part of https://github.com/knative/test-infra/issues/540

This removes some things from the zipkin package into a separate monitoring package that can be used for all monitoring components like the prometheus package as well. This will help us consolidate things for monitorng usage in our tests

### DIFF
--- a/test/monitoring/monitoring.go
+++ b/test/monitoring/monitoring.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


test/monitoring/monitoring.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+
+	"github.com/knative/pkg/test/logging"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	namespace = "istio-system"
```
vagababov: this feels more specific as "istioNamespace"

PR #314, opened 2019-03-08
Simplify init function for metrics config used in unit tests

yanweiguo:
Simplify init function for metrics config used in unit tests and move it to `metrics` package so that it can be shared.

Also move fake objects relative to unit tests to `testing.go` file.


yanweiguo:
I think I'll go in another way.

PR #315, opened 2019-03-08
Record metrics if config is nil

yanweiguo:
`metrics.Record` adds a layer around `stats.Record` to guide what metrics can be record based on package level config.

Expected:
If the config is not set, do record. User should make sure the config is set before using `metrics.Record` otherwise it just proxies to `stats.Record` based the assumption that users expect the metrics to be recorded when they call this function.

Actual:
If the config is not set, don't record.

### DIFF
--- a/metrics/record.go
+++ b/metrics/record.go
@@ -26,10 +26,10 @@ import (
 
 // Record decides whether to record one measurement via OpenCensus based on the
 // following conditions:
-//   1) No package level metrics config. Users must ensure metrics config are set
-//      to get expected behavior. Otherwise it just proxies to OpenCensus based
-//      on the assumption that users intend to record metric when they call this
-//      function.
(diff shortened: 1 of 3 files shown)


metrics/record.go
```diff
@@ -20,27 +20,34 @@ import (
...
 		return
 	}
-	// Should record if one of the following conditions satisfies:
-	// 1) the backend is not Stackdriver
-	// 2) the backend is Stackdriver and it is allowed to use custom metrics
-	// 3) the backend is Stackdriver and the metric is "knative_revison" built-in metric.
+
 	if !mc.isStackdriverBackend || mc.allowStackdriverCustomMetrics {
+		// Condition 2) and 3)
 		stats.Record(ctx, ms)
```
vagababov: I'd return here and unindent the rest.

Issue #318, opened 2019-03-15
Increase Zipkin timeout before fetching the traces


yt3liu:
## Expected Behavior

See zipkin logs for the trace. 

## Actual Behavior

Currently, the zipkin library runs logZipkinTrace immediately after the response is created. Thus, it has a static timeout to wait for 5 seconds before fetching the traces from the backend. When testing for the zipkin logging, the 5 seconds timeout is very flaky as half of the time, I do not see the zipkin logging.

PR #319, opened 2019-03-15
Increase zipkin wait time before fetching the trace (#318)

yt3liu:
Increase zipkin wait time before fetching the trace. 

Fixes: #318

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -274,7 +274,7 @@ func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
 
 	zipkinTraceEndpoint := zipkin.ZipkinTraceEndpoint + traceID
 	// Sleep to ensure all traces are correctly pushed on the backend.
-	time.Sleep(5 * time.Second)
+	time.Sleep(30 * time.Second)
 	resp, err := http.Get(zipkinTraceEndpoint)
 	if err != nil {
 		sc.logf("Error retrieving Zipkin trace: %v", err)


srinivashegde86:
/hold 
I am ok with this change, but I will wait on Dusyanth to comment on this.

test/spoof/spoof.go
```diff
@@ -274,7 +274,7 @@ func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
 
 	zipkinTraceEndpoint := zipkin.ZipkinTraceEndpoint + traceID
 	// Sleep to ensure all traces are correctly pushed on the backend.
-	time.Sleep(5 * time.Second)
+	time.Sleep(30 * time.Second)
```
dushyanthsc: 30 seconds is very high. We debated a lot when we set it to 5 secs because it adds latency. So if I understand it correctly you are saying that after 5 secs you are not seeing traces and a higher wait time causes you see traces?

yt3liu: Yes, that is correct. For 5 seconds, I see traces half of the time. I can try to run a few more tests to see if we can bring down 30 second sleep time. 

Note that I was mainly testing if zipkin logs are printed for success responses by commenting out the server error check. As I could not figure out a nice way to always force a 500/600 response. However, for cases where it did fail with 504 error, I noticed the log to be much shorter than success. It could be the 5 seconds is not enough for success, but may be enough for server error. Please let me know if there's a good way to test this hunch and confirm this PR is not required.

dushyanthsc: Ya I feel that success based retrieval is overwhelming the system. Is the zipkin setup in-memory or elastic search backed?

yt3liu: It is elastic search backed.

yt3liu:
Per offline discussion with dushyanthsc@. The 30 seconds wait is too long and it is highly likely due to success content overwhelming the zipkin system. Discarding this change as the current zipkin setup is only meant to log server error responses.

Issue #26, opened 2018-07-28
Common e2e/conformance infrastructure


mattmoor:
This repo houses shared Knative libraries.  The original intent was to share common controller libraries and whatnot, but I think that there's a clear and growing need to share common test and release logic as well.

Two clear candidates off the top of my head (the common bits):
1. `github.com/knative/serving/pkg/controller/testing`
1. `github.com/knative/serving/test`

cc @bobcatfish (emeritus)

srinivashegde86:
All the common things have been moved to pkg/.

bbrowning:
Looking in each repo's test folder, we can probably still reduce copy/paste across e2e test setups in every repo. The files under the `test/` folders like `cleanup.go`, `clients.go`, `crd.go`, `crd_checks.go`, `states.go`, and perhaps more. They aren't identical in every repo but there is a good bit of overlap.

srinivashegde86:
The PR you just linked has some commonalities between eventing and serving repos. We did not want to keep serving specific methods in pkg/test/.

That said, I can see a few methods that can be moved to pkg that talk to k8s or istio

srinivashegde86:
This was originally used for creating a common pkg/test package, which is done. If we need new ones, lets open new issues to track them. 

/close

PR #321, opened 2019-03-15
Add some more common test componets in pkg/test/

srinivashegde86:
This PR adds:

1. Export GetConfigMap method
1. Create methods for working with k8s pods
1. Create a new file `crd.go` that adds few methods to deal with crd's. 
1. Add `DOCKER_REPO` and `TAG` flags here as we need it in other repos.
1. Add a few wait cheks for k8s pods

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -19,8 +19,12 @@ limitations under the License.
 package test
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/knative/pkg/test/logging"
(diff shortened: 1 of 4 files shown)


srinivashegde86:
https://github.com/knative/eventing-sources/pull/178/ was adding a few k8s specific details to the eventing-sources repo. These should be in the pkg/test repo instead as these are generic to k8s and can be re-used by others

srinivashegde86:
Part of #26

Issue #322, opened 2019-03-15
Allow pkg/webhook to work with duck types


mattmoor:
This code [here](https://github.com/knative/pkg/blob/cd278f2d3394c865fda66bca12459e879e0279b8/webhook/webhook.go#L538-L544) can simply be guarded by:

```
if _, ok := newObj.(duck.Populatable); !ok {
   ... // existing code.
}
```

PR #323, opened 2019-03-16
Modify the webhook to allow the use of duck types.

mattmoor:
This change enables us to define a duck type that applies to a whole class of GroupVersionKinds and leverage it to perform generic validation, defaulting, etc.

Fixes: #322

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -535,13 +535,18 @@ func (ac *AdmissionController) mutate(ctx context.Context, req *admissionv1beta1
 	}
 	var patches duck.JSONPatch
 
-	// Add these before defaulting fields, otherwise defaulting may cause an illegal patch because
-	// it expects the round tripped through Golang fields to be present already.
-	rtp, err := roundTripPatch(newBytes, newObj)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


mattmoor:
An example of using this to synthesize patches over a duck type (in this case "pod speccable" types) can be found [here](https://github.com/mattmoor/boo-maps/blob/4aab00cad1017c834cb8fa7974ea8d9b0fdd3414/cmd/webhook/main.go#L122-L126)

webhook/webhook.go
```diff
@@ -535,13 +535,18 @@ func (ac *AdmissionController) mutate(ctx context.Context, req *admissionv1beta1
 	}
 	var patches duck.JSONPatch
 
-	// Add these before defaulting fields, otherwise defaulting may cause an illegal patch because
-	// it expects the round tripped through Golang fields to be present already.
-	rtp, err := roundTripPatch(newBytes, newObj)
-	if err != nil {
-		return nil, fmt.Errorf("cannot create patch for round tripped newBytes: %v", err)
+	var err error
+	// Skip this step if the type we're dealing with is a duck type, simce it is inherently
```
vagababov:
```suggestion
	// Skip this step if the type we're dealing with is a duck type, since it is inherently
```

PR #328, opened 2019-03-18
Need to use patch merge for conditions.

n3wscott:
The new conditions does not tell k8s client how to patch merge. Adding those annotations.


evankanderson:
But you have test failures...

n3wscott:
@evankanderson sorry, copied too much codez... Fixed.

PR #331, opened 2019-03-20
Use latest tag as default to get test images

srinivashegde86:


adrcunha:
I am not sure if this is the right move due to the comment in https://github.com/knative/serving/blob/master/test/README.md#using-a-docker-tag

srinivashegde86:
We were always using latest as the default tag value for serving. My previous change made the default to e2e causing issues with image pulls and this PR updates it to the original state.

adrcunha:
So that comment is outdated and should be removed then?

Issue #306, opened 2019-03-06
Configmap for defaults


josephburnett:
## Expected Behavior

I would like to be able to use a config map when setting defaults in the webhook.

## Actual Behavior

However, because the webhook is common and abstracted, I cannot construct one and get it through to the defaulting function.  E.g. I would like to pass in a config map watcher [here](https://github.com/knative/serving/blob/4c08227dbdb5145622ad4d246cfcacae86c5bdf9/cmd/webhook/main.go#L106) and have that config map passed to the `SetDefaults` function [here](https://github.com/knative/serving/blob/4c08227dbdb5145622ad4d246cfcacae86c5bdf9/pkg/apis/autoscaling/v1alpha1/pa_defaults.go#L24).

## Additional Info

I used a package scoped var to work around this limitation in https://github.com/knative/serving/pull/3103.

mattmoor:
I think what I'd like to see here is changing these interfaces to pass a `context.Context` argument, and using the work by @dprotaso to attach some of these configs to it.

PR #332, opened 2019-03-20
This threads a context.Context through the webhook interfaces we expose.

mattmoor:
Related to: https://github.com/knative/pkg/issues/306

/hold

I want to stage the PR to serving (and @n3wscott to Eventing) before I remove the hold

### DIFF
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -17,6 +17,8 @@ limitations under the License.
 package apis
 
 import (
+	"context"
+
 	authenticationv1 "k8s.io/api/authentication/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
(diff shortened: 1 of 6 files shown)


apis/interfaces.go
```diff
@@ -17,28 +17,30 @@ limitations under the License.
...
+
 	authenticationv1 "k8s.io/api/authentication/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // Defaultable defines an interface for setting the defaults for the
 // uninitialized fields of this instance.
 type Defaultable interface {
-	SetDefaults()
+	SetDefaults(context.Context)
```
grantr: What is this context argument expected to be used for? Do we have an example?

vagababov: For example we have a request to make default revision timeout configurable by the operator.
ConfigMaps, attached to the context can be used to get the operator desired default.

Issue #185, opened 2018-12-02
Find a way to track queue time


mattmoor:
It would be good to be able to assess when things enter the controller's workqueue, so that we can surface a metric for queue time.

cc @mdemirhan

PR #333, opened 2019-03-20
stats: track time waiting in reconcile queue

srvaroa:
Add a timer that tracks time that each item has spent in the WorkQueue.
This is implemented by keeping an `entryTimes` map for each entry that
gets removed when the work item is removed from the WorkQueue.

Fixes: #185


controller/controller.go
```diff
@@ -84,6 +84,10 @@ type Impl struct {
 	// simultaneously in two different workers.
 	WorkQueue workqueue.RateLimitingInterface
 
+	// entryTimes keeps the entry time of a given item in the
+	// WorkQueue, which is used on removal to calculate time spent queuing
```
vagababov: The `sync.Map` docs suggest that it should be used in some specific cases (this isn't that case) and instead recommends to use mutex+map.

srvaroa: will change

srvaroa:
@vagababov thanks for the pointer. I considered using those although my reading was that extracting the individual latency metric (which [exists](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go#L81)) the right way of reusing it would be to write a [MetricsProvider](https://github.com/kubernetes/client-go/blob/master/util/workqueue/metrics.go#L195) for opencensus. I undestand that this means rewriting the stats_reporter altogether.

I'm happy to go that route if you guys think it's best. Or let me know if I'm overcomplicating the issue, it will likely be the case :)

PR #334, opened 2019-03-22
Factor out common logic on obtaining the ingress' endpoint.

markusthoemmes:
These methods are useful outside of the spoof client and for different protocols than those the spoof client supports.


test/spoof/spoof.go
```diff
@@ -115,9 +110,9 @@ func New(kubeClientset *kubernetes.Clientset, logf logging.FormatLogger, domain
 		if endpointOverride == "" {
 			var err error
 			// If the domain that the Route controller is configured to assign to Route.Status.Domain
-			// (the domainSuffix) is not resolvable, we need to retrieve the IP of the endpoint and
-			// spoof the Host in our requests.
-			e, err = GetServiceEndpoint(kubeClientset)
+			// (the domainSuffix) is not resolvable, we need to retrieve the the endpoint and spoof
```
adrcunha:
```suggestion
			// (the domainSuffix) is not resolvable, we need to retrieve the endpoint and spoof
```

PR #335, opened 2019-03-23
Add some helpers to reduce boilerplate in controller binaries.

mattmoor:
You can see the reduction in [this PR](https://github.com/knative/serving/pull/3503) taking advantage of these helpers.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -39,6 +39,10 @@ const (
 	trueString  = "true"
 )
 
+var (
+	DefaultThreadsPerController = 2
+)
+
 // Reconciler is the interface that controller implementations are expected
 // to implement, so that the shared controller.Impl can drive work through it.
 type Reconciler interface {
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -39,6 +39,10 @@ const (
 	trueString  = "true"
 )
 
+var (
```
vagababov: This is a const.

controller/controller_test.go
```diff
@@ -625,3 +630,67 @@ func checkStats(t *testing.T, r *FakeStatsReporter, reportCount, lastQueueDepth,
 		t.Errorf("Reconcile success = %v, wanted %v", got, want)
 	}
 }
+
+type FixedInformer struct {
```
vagababov:
```suggestion
type fixedInformer struct {
```
perhaps?

controller/controller_test.go
```diff
@@ -498,7 +500,8 @@ func TestStartAndShutdownWithErroringWork(t *testing.T) {
 
 	var eg errgroup.Group
 	eg.Go(func() error {
-		return impl.Run(1, stopCh)
+		StartAll(stopCh, impl)
```
vagababov: this (and others) can be redone with channels as the new tests, so sleeps can go away.

controller/controller_test.go
```diff
@@ -625,3 +630,67 @@ func checkStats(t *testing.T, r *FakeStatsReporter, reportCount, lastQueueDepth,
...
+		errCh <- StartInformers(stopCh, fi)
+	}()
+
+	select {
+	case err := <-errCh:
+		t.Errorf("Unexpected send on errCh: %v", err)
+	case <-time.After(1 * time.Second):
+		// Wait a brief period to ensure nothing is sent.
+	}
+
```
vagababov: Do we want a test, were sunk goes from false -> true? ⬛️

controller/controller.go
mattmoor: Added a detailed comment, I can see us changing this default at a per-binary level

controller/controller_test.go
mattmoor: Done, but it ends up being longer.

controller/controller_test.go
mattmoor: Done

controller/controller_test.go
mattmoor: Done

controller/controller_test.go
vagababov: But no sleeps! :)

PR #336, opened 2019-03-23
Add a GetCondition method to Status

mattmoor:
I wanted this when rewriting the Pipeline e2e test in Serving to use the dynamic client and checking status via this duck type.

See here: https://github.com/knative/serving/blob/b0ad29dbc869716cf5d80b7557d14472e33fc03f/test/e2e/build_pipeline_test.go#L84

### DIFF
--- a/apis/duck/v1alpha1/condition_type_test.go
+++ b/apis/duck/v1alpha1/condition_type_test.go
@@ -225,3 +225,33 @@ func TestJson(t *testing.T) {
 		})
 	}
 }
+
+func TestStatusGetCondition(t *testing.T) {
+	foo := &Condition{
+		Type:    "Foo",
+		Status:  corev1.ConditionTrue,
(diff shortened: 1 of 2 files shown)


skyn1ce:
:)

apis/duck/v1alpha1/condition_type_test.go
```diff
@@ -225,3 +225,31 @@ func TestJson(t *testing.T) {
 		})
 	}
 }
+
+func TestStatusGetCondition(t *testing.T) {
+	foo := Condition{
```
vagababov: I'd go with `foo := &Condition{}`, and then just `cmp.Equal()` or `cmp.Diff()`.
Alternatively `&foo` below and avoid `got == nil` checks.

mattmoor: Done

PR #338, opened 2019-03-25
Allow webhook implementations to optionally disallow unknown fields.

mattmoor:
Related: https://github.com/knative/serving/issues/3309

### DIFF
--- a/webhook/webhook_test.go
+++ b/webhook/webhook_test.go
@@ -135,6 +135,31 @@ func TestUnknownVersionFails(t *testing.T) {
 	expectFailsWith(t, ac.admit(TestContextWithLogger(t), req), "unhandled kind")
 }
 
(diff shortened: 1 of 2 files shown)


webhook/webhook_test.go
```diff
@@ -699,8 +699,9 @@ func setUserAnnotation(userC, userU string) jsonpatch.JsonPatchOperation {
 func NewAdmissionController(client kubernetes.Interface, options ControllerOptions,
 	logger *zap.SugaredLogger) (*AdmissionController, error) {
 	return &AdmissionController{
-		Client:  client,
-		Options: options,
+		Client:                client,
```
vagababov: do we actually want a test, where something will fail?

mattmoor: Added one.

PR #339, opened 2019-03-25
Adding Strict Validation based on struct.*Deprecated*Foo

n3wscott:
This adds Strict Validation mode that will reject creates and updates to fields that are marked `Deprecated`.

/assign @mattmoor 

This is first pass code. Let me know what you think.

I have all types supported except a slice of structs when updating a deprecated field.


webhook/strict.go
```diff
@@ -0,0 +1,297 @@
+package webhook
```
mattmoor: license

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+// Note: nil values denote absence of `old` (create) or `new` (delete) objects.
+func strictValidate(ctx context.Context, old GenericCRD, new GenericCRD) error {
+	// Check if it is a delete, if it is we don't validate.
+	if new == nil {
+		return nil
+	}
+
+	var errs *apis.FieldError
+
+	// If old is nil, then we are creating. Not not allowed to set deprecated
```
vagababov: `not not`.

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+// strictValidate validates the following rule sets:
+// - On Create, Spec and Status fields that are marked `Deprecated:` are not
+//   allowed to be set.
+// - On Update, Spec and Status fields that are marked `Deprecated:` are not
+//   allowed to be updated, unless deleting.
+//
+// Note: nil values denote absence of `old` (create) or `new` (delete) objects.
+func strictValidate(ctx context.Context, old GenericCRD, new GenericCRD) error {
+	// Check if it is a delete, if it is we don't validate.
+	if new == nil {
```
vagababov: `new` is a bad name, given it shields the builtin.

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+	case reflect.Ptr:
+		if a.IsNil() {
+			return false
+		}
+		return nonZero(a.Elem())
+
+	case reflect.Map, reflect.Slice, reflect.Array:
+		if a.IsNil() {
+			return false
+		}
```
vagababov: empty as well?

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+	if oldField.Type() != resField.Type() {
+		return &apis.FieldError{
+			Message: "failed to match filed type",
+			Paths:   []string{apis.CurrentField},
+		}
+	}
+
+	var errs *apis.FieldError
+
+	// for each field,
```
vagababov: this is partial,

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+							strictValidateUpdateReflected(o.Elem(), v.Elem()).ViaField(f.Name),
+						)
+					}
+
+				case reflect.Struct:
+					errs = errs.Also(
+						strictValidateUpdateReflected(o, v).ViaField(f.Name),
+					)
+
+				// TODO: because slices have no order, I am not sure how to test for this.
```
vagababov: sort keys?

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+
+	var errs *apis.FieldError
+
+	// for each field,
+	for i := 0; i < resField.NumField(); i++ {
+		f := resField.Type().Field(i)
+		if v := resField.Field(i); v.IsValid() {
+			if o := oldField.Field(i); o.IsValid() {
+				if strings.HasPrefix(f.Name, "Deprecated") {
+					if differ(v, o) {
```
vagababov: This doesn't sound right. Since if `new` is empty and `old` wasn't it will return error right here.

webhook/strict.go
n3wscott: there is nothing known to sort on.

webhook/strict.go
```diff
@@ -0,0 +1,313 @@
...
+
+// differ returns true if a != b
+func differ(a, b reflect.Value) bool {
+	if a.Kind() != b.Kind() {
+		return false
+	}
+
+	switch a.Kind() {
+	case reflect.Ptr:
+		if a.IsNil() || b.IsNil() {
```
n3wscott: Re: return error @vagababov, differ has a nil check

webhook/strict.go
n3wscott: This is matching the rest of webhook's impl.

webhook/strict.go
n3wscott: it is intended

webhook/strict.go
n3wscott: I am going to update this code. It is confusing....

webhook/strict.go
n3wscott: I am counting empty list as invalid. Not allowed to set the deprecated field.

validation/deprecated.go
```diff
@@ -0,0 +1,131 @@
...
+			return false
+		}
+		return true
+	}
+}
+
+// differ returns true if a != b
+func differ(a, b reflect.Value) bool {
+	if a.Kind() != b.Kind() {
+		return false
```
vagababov: if `a` and `b` are of different types, I think they _do_ differ, n'est pas?

n3wscott: HA! whoops! nice catch.

n3wscott:
This is ready.

/assign @mattmoor 
/cc @vagababov

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+			if nonZero(value) {
+				// Not allowed to set the value.
+				errs = errs.Also(ErrDisallowedFields(name))
+			}
+		}
+	}
+	return errs
+}
+
+func getPrefixedNamedFieldValues(prefix string, obj interface{}) map[string]reflect.Value {
```
mattmoor: I think this needs to handle `json:",inline"` fields.

apis/deprecated_test.go
```diff
@@ -0,0 +1,371 @@
...
+			},
+			wantErrs: []string{
+				"must not set",
+				"structPtr",
+			},
+		},
+	}
+	for n, tc := range testCases {
+		t.Run(n, func(t *testing.T) {
+			ctx := context.TODO()
```
mattmoor:
```suggestion
			ctx := context.Background()
```

webhook/webhook.go
```diff
@@ -99,6 +99,9 @@ type ControllerOptions struct {
 	// TLS Client Authentication.
 	// The default value is tls.NoClientCert.
 	ClientAuth tls.ClientAuthType
+
+	// Strict defines strict mode validation.
+	Strict bool
```
mattmoor: We already provide a way to alter the context, so this isn't necessary.

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+		return fields
+	}
+
+	for i := 0; i < objValue.NumField(); i++ {
+		tf := objValue.Type().Field(i)
+		if v := objValue.Field(i); v.IsValid() {
+			if strings.HasPrefix(tf.Name, prefix) {
+				jTag := tf.Tag.Get("json")
+				name := strings.Split(jTag, ",")[0]
+				if name == "" {
```
n3wscott: > I think this needs to handle json:",inline" fields.

if name is `""` it will use the full `DeprecatedFoo` name.

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+package apis
+
+import (
+	"context"
+	"reflect"
+	"strings"
+)
+
+const (
+	deprecated = "Deprecated"
```
mattmoor: `deprecatedPrefix`

apis/deprecated.go
n3wscott: ^^ @mattmoor

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+// CheckDeprecated checks whether the provided named deprecated fields
+// are set in a context where deprecation is disallowed.
+// This is a shallow check.
+func CheckDeprecatedUpdate(ctx context.Context, obj interface{}, org interface{}) *FieldError {
+	if IsDeprecatedAllowed(ctx) {
+		return nil
+	}
+
+	var errs *FieldError
+	newFields := getPrefixedNamedFieldValues(deprecated, obj)
```
mattmoor: `newFields` is a confusing name since `obj` isn't called `new`.  I also dislike the name `org` as a shortening of "original"(?)

apis/deprecated.go
n3wscott: the prefix refers to the golang struct field name.

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if nonZero(reflect.ValueOf(org)) {
+		orgFields := getPrefixedNamedFieldValues(deprecated, org)
+
+		for name, newValue := range newFields {
```
mattmoor: This could use a comment elaborating on why it is sufficient to walk just `newFields` and not `orgFields`

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+	// If res is not valid or a struct, don't even try to use it.
+	if !objValue.IsValid() || objValue.Kind() != reflect.Struct {
+		return fields
+	}
+
+	for i := 0; i < objValue.NumField(); i++ {
+		tf := objValue.Type().Field(i)
+		if v := objValue.Field(i); v.IsValid() {
+			if strings.HasPrefix(tf.Name, prefix) {
+				jTag := tf.Tag.Get("json")
```
mattmoor: cc @dgerd (we were talking about this for the diff stuff)

apis/deprecated.go
```diff
@@ -0,0 +1,141 @@
...
+			return false
+		}
+		return true
+
+	// This is a nil interface{} type.
+	case reflect.Invalid:
+		return false
+
+	default:
+		if reflect.DeepEqual(a.Interface(), reflect.Zero(a.Type()).Interface()) {
```
mattmoor: my understanding was that this was enough for any type?  see SO link in my function.

n3wscott: it only works for structs in reality. It ends up having the same problem as `ptr == ptr` rather than `*ptr == *ptr` and the case of maps and slices is complicated.

n3wscott: also, that does not work with nil types. /tableflip

apis/deprecated.go
n3wscott: done.

apis/deprecated.go
```diff
@@ -36,60 +36,99 @@ func CheckDeprecated(ctx context.Context, obj interface{}) *FieldError {
...
 					// Not allowed to update the value.
 					errs = errs.Also(ErrDisallowedUpdateDeprecatedFields(name))
 				}
 			}
 		}
+		// Look for deprecated inlined updates.
+		if len(objInlined) > 0 {
+			for name, value := range objInlined {
+				errs = errs.Also(CheckDeprecatedUpdate(ctx, value, originalInlined[name]))
+			}
```
mattmoor: Add the test we talked about.

apis/deprecated.go
```diff
@@ -36,60 +36,99 @@ func CheckDeprecated(ctx context.Context, obj interface{}) *FieldError {
 // CheckDeprecated checks whether the provided named deprecated fields
 // are set in a context where deprecation is disallowed.
 // This is a shallow check.
```
mattmoor: Update to indicate that it is JSON-shallow

PR #342, opened 2019-03-27
This creates a way for clients of the webhook to decorate the request context.

mattmoor:
Clients of webhook can now decorate the request context with additional metadata via:
```
  ac := &AdmissionController{
    ... // As before
    WithContext: func(ctx context.Context) context.Context {
      // logic to attach stuff to ctx
    }
  }
```

This metadata can then be accessed off of the context in methods like
`SetDefaults` and `Validate` on types registered as webhook handlers.

Fixes: https://github.com/knative/pkg/issues/306

### DIFF
--- a/testing/resource.go
+++ b/testing/resource.go
@@ -55,6 +55,7 @@ var _ apis.Listable = (*Resource)(nil)
 // ResourceSpec represents test resource spec.
 type ResourceSpec struct {
 	FieldWithDefault               string `json:"fieldWithDefault,omitempty"`
+	FieldWithContextDefault        string `json:"fieldWithContextDefault,omitempty"`
 	FieldWithValidation            string `json:"fieldWithValidation,omitempty"`
 	FieldThatsImmutable            string `json:"fieldThatsImmutable,omitempty"`
 	FieldThatsImmutableWithDefault string `json:"fieldThatsImmutableWithDefault,omitempty"`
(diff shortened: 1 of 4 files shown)


testing/resource.go
```diff
@@ -65,11 +66,34 @@ func (c *Resource) SetDefaults(ctx context.Context) {
 	c.Spec.SetDefaults(ctx)
 }
 
+type onContextKey struct{}
+
+// WithContext is the type of a context decorator.
+type WithContext func(context.Context) context.Context
```
n3wscott: I would call this `MutateContext`

webhook/webhook.go
```diff
@@ -121,9 +121,14 @@ type AdmissionController struct {
 	Handlers map[schema.GroupVersionKind]GenericCRD
 	Logger   *zap.SugaredLogger
 
+	WithContext           func(context.Context) context.Context
```
n3wscott: And this `MutateContext`

testing/resource.go
```diff
@@ -65,11 +66,34 @@ func (c *Resource) SetDefaults(ctx context.Context) {
...
+
+// WithValue returns a WithContext for attaching an OnContext with the given value.
+func WithValue(val string) WithContext {
```
n3wscott: I have found it useful to have methods that do:

```
ctx := foo.WithBar(ctx, barbarbar)
```

The usage is inline and then it is passed down the the call chain.

webhook/webhook.go
```diff
@@ -456,7 +461,15 @@ func (ac *AdmissionController) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		zap.String(logkey.Resource, fmt.Sprint(review.Request.Resource)),
 		zap.String(logkey.SubResource, fmt.Sprint(review.Request.SubResource)),
 		zap.String(logkey.UserInfo, fmt.Sprint(review.Request.UserInfo)))
-	reviewResponse := ac.admit(logging.WithLogger(r.Context(), logger), review.Request)
+	ctx := logging.WithLogger(r.Context(), logger)
+
+	wc := ac.WithContext
```
n3wscott: why not

```
	if ac.WithContext != nil {
		ctx = ac.WithContext(ctx)
	}
```

testing/resource.go
mattmoor: Done

testing/resource.go
mattmoor: I removed this based on the below comment.

webhook/webhook.go
mattmoor: done

webhook/webhook.go
mattmoor: I removed the type alias

testing/resource.go
```diff
@@ -65,11 +66,29 @@ func (c *Resource) SetDefaults(ctx context.Context) {
 	c.Spec.SetDefaults(ctx)
 }
 
+type onContextKey struct{}
+
+// WithValue returns a WithContext for attaching an OnContext with the given value.
+func WithValue(ctx context.Context, val string) context.Context {
+	return context.WithValue(ctx, onContextKey{}, &OnContext{Value: val})
```
n3wscott: double check this, but if you look at the docs for context.WithValue they imply that you should use a global `var onKey = onContextKey{}` and then use it `return context.WithValue(ctx, onKey, &OnContext{Value: val})`

I don't know if their internal impl cares.

mattmoor: it doesn't, I copied this from our config store logic

n3wscott: it might not matter.

PR #343, opened 2019-03-27
Retry DNS errors in spoof client.

tcnghia:
Adding retries for DNS error to prepare for using `.xip.io` or `.nip.io` in our tests.

### DIFF
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -1,0 +1,36 @@
+package spoof
+
+import (
+	"net"
+	"net/url"
+	"strings"
+)
+
+func isTCPTimeout(err error) bool {
+	if err, ok := err.(net.Error); ok && err.Timeout() {
+		return true
+	}
+	return false
+}
+
+func isDNSError(err error) bool {
+	if err, ok := err.(*url.Error); err != nil && ok {
+		if err, ok := err.Err.(*net.OpError); err != nil && ok {
+			if err, ok := err.Err.(*net.DNSError); err != nil && ok {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isTCPConnectRefuse(err error) bool {
+	// The alternative for the string check is:
+	// 	errNo := (((err.(*url.Error)).Err.(*net.OpError)).Err.(*os.SyscallError).Err).(syscall.Errno)
+	// if errNo == syscall.Errno(0x6f) {...}
+	// But with assertions, of course.
+	if strings.Contains(err.Error(), "connect: connection refused") {
+		return true
+	}
+	return false
+}
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -6,11 +6,9 @@ import (
 	"strings"
 )
 
-func isTCPTimeout(err error) bool {
-	if err, ok := err.(net.Error); ok && err.Timeout() {
-		return true
-	}
-	return false
+func isTCPTimeout(e error) bool {
+	err, ok := e.(net.Error)
+	return err != nil && ok && err.Timeout()
 }
 
 func isDNSError(err error) bool {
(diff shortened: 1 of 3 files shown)


tcnghia:
/hold
I believe this isn't correct.

tcnghia:
@vagababov RFAL, sorry about the false start.

test/spoof/error_checks.go
```diff
@@ -0,0 +1,36 @@
...
+		if err, ok := err.Err.(*net.OpError); err != nil && ok {
+			if err, ok := err.Err.(*net.DNSError); err != nil && ok {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isTCPConnectRefuse(err error) bool {
```
vagababov: this is easy to test, by connecting to a port/target that won't accept.

test/spoof/error_checks.go
```diff
@@ -0,0 +1,36 @@
+package spoof
+
+import (
+	"net"
+	"net/url"
+	"strings"
+)
+
+func isTCPTimeout(err error) bool {
+	if err, ok := err.(net.Error); ok && err.Timeout() {
```
vagababov: make this simpler
```
err, ok := err.(net.Error)
return ok && err.Timeout()
```

test/spoof/error_checks.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	return false
+}
+
+func isTCPConnectRefuse(err error) bool {
+	// The alternative for the string check is:
+	// 	errNo := (((err.(*url.Error)).Err.(*net.OpError)).Err.(*os.SyscallError).Err).(syscall.Errno)
+	// if errNo == syscall.Errno(0x6f) {...}
+	// But with assertions, of course.
+	if strings.Contains(err.Error(), "connect: connection refused") {
```
vagababov: If you're going through the motions, you can do this as above, then (just the error chain will be deeper :disappointed: ).

vagababov: But I am fine as is, though change to `return strings.Contains...`.

test/spoof/error_checks.go
tcnghia: Done

test/spoof/error_checks_test.go
```diff
@@ -0,0 +1,101 @@
...
+				t.Errorf("Expected tcpRefuse=%v, got %v", tt.tcpRefuse, tcpRefuse)
+			}
+		})
+	}
+}
+
+func TestTCPTimeout(t *testing.T) {
+	client := &http.Client{}
+
+	// We have no postive test for TCP timeout, but we do have a few negative tests.
```
mattmoor-sockpuppet:
```suggestion
	// We have no positive test for TCP timeout, but we do have a few negative tests.
```

test/spoof/error_checks.go
tcnghia: Done

test/spoof/error_checks_test.go
tcnghia: Done

test/spoof/error_checks.go
tcnghia: I think I prefer the string check than those nested casts.

test/spoof/error_checks.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"net"
+	"net/url"
+	"strings"
+)
+
+func isTCPTimeout(e error) bool {
+	err, ok := e.(net.Error)
+	return err != nil && ok && err.Timeout()
```
vagababov: being internal we can precondition `err != nil`, but fine as is.

tcnghia:
/assign @srinivashegde86 

for test/OWNERS

test/spoof/error_checks.go
```diff
@@ -0,0 +1,34 @@
+package spoof
```
srinivashegde86: missing doc header

test/spoof/error_checks_test.go
```diff
@@ -0,0 +1,101 @@
+package spoof
```
srinivashegde86: doc header

srinivashegde86:
Can you add the headers? I Can approve once that is done.

test/spoof/error_checks.go
tcnghia: Done

test/spoof/error_checks_test.go
tcnghia: Done

tcnghia:
@srinivashegde86 I added the headers, can you please also re /lgtm when you /approve? thanks

Issue #344, opened 2019-03-27
use apps.deployment instead of extensions.deployment


vaikas-google:
## Expected Behavior
use the latest version of deployment instead of extensions v1beta1

## Actual Behavior
uses v1beta1

## Steps to Reproduce the Problem

1. In webhook/webhook.go we use v1beta1 extensions
2. don't. use apps instead which is v1.

PR #345, opened 2019-03-27
Use apps.Deployment instead of extensions.Deployment

vaikas-google:
Fixes: #344


webhook/webhook.go
```diff
@@ -42,9 +42,9 @@ import (
 	"github.com/mattbaird/jsonpatch"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	v1 "k8s.io/api/apps/v1"
```
mattmoor: nit: `appsv1`

vaikas-google: done

PR #351, opened 2019-03-28
Address some lint errors

trshafer:
Addresses some lint errors in the go report card.

### DIFF
--- a/apis/duck/v1alpha1/conditions_types.go
+++ b/apis/duck/v1alpha1/conditions_types.go
@@ -159,7 +159,7 @@ var _ duck.Populatable = (*KResource)(nil)
 var _ apis.Listable = (*KResource)(nil)
 
 // GetFullType implements duck.Implementable
-func (_ *Conditions) GetFullType() duck.Populatable {
+func (*Conditions) GetFullType() duck.Populatable {
 	return &KResource{}
 }
 
--- a/apis/duck/v1alpha1/conditions_types.go
+++ b/apis/duck/v1alpha1/conditions_types.go
@@ -187,7 +187,7 @@ func (t *KResource) Populate() {
 }
 
 // GetListType implements apis.Listable
-func (r *KResource) GetListType() runtime.Object {
+func (t *KResource) GetListType() runtime.Object {
 	return &KResourceList{}
 }
 
--- a/apis/duck/v1alpha1/conditions_types.go
+++ b/apis/duck/v1alpha1/conditions_types.go
@@ -187,7 +187,7 @@ func (t *KResource) Populate() {
 }
 
 // GetListType implements apis.Listable
-func (t *KResource) GetListType() runtime.Object {
+func (*KResource) GetListType() runtime.Object {
 	return &KResourceList{}
 }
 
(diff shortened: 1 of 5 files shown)


apis/duck/v1alpha1/conditions_types.go
```diff
@@ -187,7 +187,7 @@ func (t *KResource) Populate() {
 }
 
 // GetListType implements apis.Listable
-func (r *KResource) GetListType() runtime.Object {
+func (t *KResource) GetListType() runtime.Object {
```
evankanderson: Why does this need `t`?

trshafer: There is a lint test that errors when a receiver has a different name within the same file. The receiver is named `t` on line 177: https://github.com/knative/pkg/pull/351/files/f35a510212e1636b012c3bee893f6043c18bde95#diff-04abbb75ad1eb1e871f98943273acd40L177

mattmoor: I think the question is why we need it at all?

apis/duck/v1alpha1/legacy_targetable_types.go
```diff
@@ -80,7 +80,7 @@ func (t *LegacyTarget) Populate() {
 }
 
 // GetListType implements apis.Listable
-func (r *LegacyTarget) GetListType() runtime.Object {
+func (t *LegacyTarget) GetListType() runtime.Object {
```
mattmoor: delete?

apis/duck/v1alpha1/retired_targetable_types.go
```diff
@@ -82,7 +82,7 @@ func (t *Target) Populate() {
 }
 
 // GetListType implements apis.Listable
-func (r *Target) GetListType() runtime.Object {
+func (t *Target) GetListType() runtime.Object {
```
mattmoor: delete?

apis/duck/v1alpha1/conditions_types.go
trshafer: Ah thanks for the clarification. My understanding is that this implements the Listable interface: https://github.com/trshafer/pkg/blob/f35a510212e1636b012c3bee893f6043c18bde95/apis/interfaces.go#L48. So the objects can be used here: https://github.com/trshafer/pkg/blob/f35a510212e1636b012c3bee893f6043c18bde95/apis/duck/typed.go#L49

mattmoor: To further clarify, why not this?
```suggestion
func (*KResource) GetListType() runtime.Object {
```

apis/duck/v1alpha1/legacy_targetable_types.go
mattmoor:
```suggestion
func (*LegacyTarget) GetListType() runtime.Object {
```

apis/duck/v1alpha1/retired_targetable_types.go
mattmoor:
```suggestion
func (*Target) GetListType() runtime.Object {
```

apis/duck/v1alpha1/conditions_types.go
trshafer: Great suggestion. Thanks.

PR #352, opened 2019-03-28
Fix lint error: Address struct field naming

trshafer:
Fixed a few lint errors in the go report card.
Ran ./hack/update-codegen.sh


mattmoor:
Actually eventing uses this too, so eventing too.

trshafer:
Serving: https://github.com/knative/serving/pull/3590
Eventing: https://github.com/knative/eventing/pull/1006

These builds fail as the code from pkg has not been checked in yet and vendored.

Issue #357, opened 2019-03-30
InitializeConditions should actually reset the Conditions.


mattmoor:
Today `InitializeCondition` checks whether a Condition exists before resetting it to its initial state: https://github.com/knative/pkg/blob/916205998db9d085aaff357221b208bc89290ea5/apis/duck/v1alpha1/condition_set.go#L334-L342

This is called at the beginning of `Reconcile()` calls, and at the end we update `Status` if things have changed.  Early on, we struggled with status updates "touching" the resource and triggering another reconciliation that "touches" the resource (and round and round), but `apis.VolatileTime` largely mitigated that.

In thinking about how we might change the space of `Condition` types we expose in a non-additive way, I got to wondering whether we still needed to be so cautious.  In principle, reconciliation should produce a stable set of conditions and our `apis.VolatileTime` will keep us from updating when all that changes is the timestamps.

To experiment with this, I changed the above to unconditionally set the conditions during `InitializeConditions()` and cranked up the resync period on our resources to `10 * time.Second`, and the Serving resources have remained stable.

This is too risky for 0.5, but we should land this early in 0.6 across the various repos so that it has time to bake.

cc @n3wscott

mattmoor:
There are several imperative uses of `lastTransitionTime` that we will need to eliminate before we can make this change:
1. [Here](https://github.com/knative/serving/blob/218f3976169a1fda7f64f30b644b5b5c5dc86e56/pkg/reconciler/v1alpha1/revision/reconcile_resources.go#L232) we use the last transition time to time out deployments.
1. [Here](https://github.com/knative/serving/blob/0024f1c106442b6e99251b5ff78b4947241e6173/pkg/apis/autoscaling/v1alpha1/pa_lifecycle.go#L136) we use the last transition time to sleep before scaling Revisions to zero.

There may be others as well.

mattmoor:
For the former case, we should consider using the `creationTimestamp` on the resource instead.

PR #358, opened 2019-03-31
Stop initializing conditions imperatively.

mattmoor:
Reconcilers should consistently produce a set of conditions, so when we initialize we reset the conditions and reconciliation should bring us back to a consistent state.  The status update is already careful about spinning when only timestamps change during this process, so this *should* be safe for all Reconcilers.

Fixes: https://github.com/knative/pkg/issues/357

/hold
Not before 0.5 cuts


mattmoor:
This is going to take a bunch of updating in knative/serving, so closing until I can reopen with https://github.com/knative/pkg/pull/361 covered and fixes to other repos stages.

PR #359, opened 2019-03-31
[WIP] Move the Condition definitions out of duck and into apis.

mattmoor:
The conditions implementation is independent of duck typing, so this moves
the bulk of it out of `pkg/apis/duck/v1alpha1` and into `pkg/apis`.  The part
that remains is our duck-typed `Status` definition.

At this point, I'm really still just exploring this (staging it to see how disruptive it is to Serving), but thoughts on this are welcome.


apis/duck/v1alpha1/status_types.go
```diff
@@ -0,0 +1,117 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

mattmoor:
I think that practically we will need to keep the existing implementation in parallel until we can migrate everything to the new location. 🤔

mattmoor:
I think I have a plan to do this.  Will reopen this when it is ready.

mattmoor:
@vdemeester Yeah, I was staging this to gauge disruption and mitigate it in `./vendor` (ordinarily, I'd just prototype things there, but I needed to `./hack/update-codegen.sh`)

PR #360, opened 2019-04-01
Add CredentialName into Gateway

ZhiminXiang:
Add `CredentialName` into `Gateway` for TLS configuration.
Note that this feature depends on Istio 1.1. So don't check it in until Istio 1.1 is landed within Knative.


ZhiminXiang:
As Istio 1.1 was landed, we can check in this PR now.

PR #363, opened 2019-04-04
refactor mkdir into a common function

Fredy-Z:
Refactor #[656](https://github.com/knative/test-infra/issues/656)

### DIFF
--- a/test/helpers/utils.go
+++ b/test/helpers/utils.go
@@ -21,10 +21,12 @@ import (
 	"os"
 )
 
+const allUsersFullPermission = 0777
+
 // CreateDir creates dir if does not exist.
 // The created dir will have the permission bits as 0777, which means everyone can read/write/execute it.
 func CreateDir(dirPath string) error {
-	return CreateDirWithFileMode(dirPath, 0777)
+	return CreateDirWithFileMode(dirPath, allUsersFullPermission)
 }
 
 // CreateDirWithFileMode creates dir if does not exist.
--- a/test/helpers/utils.go
+++ b/test/helpers/utils.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


test/helpers/utils.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+// CreateDir creates dir if does not exist.
+// The created dir will have the permission bits as 0777, which means everyone can read/write/execute it.
+func CreateDir(dirPath string) error {
+	return CreateDirWithFileMode(dirPath, 0777)
```
srinivashegde86: 0777 should be a Const

test/helpers/utils.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// CreateDirWithFileMode creates dir if does not exist.
+// The created dir will have the permission bits as perm, which is the standard Unix rwxrwxrwx permissions.
+func CreateDirWithFileMode(dirPath string, perm os.FileMode) error {
```
srinivashegde86: Can we keep these in an existing file instead of creating a new one?

Fredy-Z:
> Also, for completeness sake, don't just add a new function. Refactor existing code to use the new function and thus make the codebase better. An example of a piece of code that would benefit from this:
> 
> https://github.com/knative/test-infra/blob/master/shared/testgrid/testgrid.go#L70

I have done it locally but haven't created the PR, since it will have build failure before this PR is merged.

test/helpers/utils.go
Fredy-Z: Done.

test/helpers/utils.go
Fredy-Z: I also prefer not to create a new file, but seems there is no appropriate file to put it in.
Maybe we can rename the `data.go` file to `utils.go`? Or any other suggestions?

adrcunha: Based on your other reply, I think this is the wrong repo for such helper function.

Fredy-Z: Are you suggesting to put it under `test-infra/shared`? Seems we'll also need to create a new file if doing so, since all existing folders are individual components. Is it fine if I create a file `test-infra/shared/common/common.go` for holding this function?

adrcunha: I think so. @srinivashegde86 objections?

srinivashegde86: nope. sg.

PR #368, opened 2019-04-06
Deprecate apis.Immutable and apis.Annotatable.

mattmoor:
Deprecate apis.Immutable and apis.Annotatable.

This deprecates the `apis.Immutable` and `apis.Annotatable` interfaces,
which were both awkward niche extensions of `apis.Validatable` and
`apis.SetDefaults` for specific contexts that the former set didn't
cover well.

With this change, the expectation is that types that want to check
for immutability will instead access the "baseline" object via the
context from within updates.  For example:

```
func (new *Type) Validate(ctx context.Context) *apis.FieldError {
  if apis.IsInUpdate(ctx) {
    old := apis.GetBaseline(ctx).(*Type)
    // Update specific validation based on new and old.
  }
}
```

For applying user annotations, the type writer can write:

```
func (new *Type) SetDefaults(ctx context.Context) {
  if apis.IsInCreate(ctx) {
    ui := apis.GetUserInfo(ctx)
    // Set creator annotation from ui
  }

  if apis.IsInUpdate(ctx) {
    ui := apis.GetUserInfo(ctx)
    old := apis.GetBaseline(ctx).(*Type)
    // Compare old.Spec vs. new.Spec and on changes
    // update the "updater" annotation from ui.
  }
}
```

One of the key motivations for this refactoring was to enable us
to do more powerful validation in `apis.Validate` beyond the niche
of immutability checking (and without introducing yet-another
one-off niche interface).  In the BYO Revision name PoC I abused
`apis.Immutable` to do more arbitrary before/after validation,
which with this can simply be a part of `apis.Validatable`.

See: knative/serving#3562

## BREAKING CHANGES

The general stance on deprecating interfaces such as these will be
to deprecate them in a non-breaking way (via a comment for now). They
will be hollowed out when the functionality is removed from the webhook,
but left in because of diamond dependency problems.  In this change
we remove the `apis.Annotatable` functionality and deprecate the
`apis.Immutable` functionality.

### DIFF
--- a/apis/contexts.go
+++ b/apis/contexts.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apis
+
+import (
+	"context"
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+)
+
+// This is attached to contexts passed to webhook interfaces when
+// the receiver being validated is being created.
+type inCreateKey struct{}
+
+// WithinCreate is used to note that the webhook is calling within
+// the context of a Create operation.
+func WithinCreate(ctx context.Context) context.Context {
+	return context.WithValue(ctx, inCreateKey{}, struct{}{})
+}
+
+// IsInCreate checks whether the context is a Create.
+func IsInCreate(ctx context.Context) bool {
+	return ctx.Value(inCreateKey{}) != nil
+}
+
+// This is attached to contexts passed to webhook interfaces when
+// the receiver being validated is being updated.
+type inUpdateKey struct{}
+
+// WithinUpdate is used to note that the webhook is calling within
+// the context of a Update operation.
+func WithinUpdate(ctx context.Context, base interface{}) context.Context {
+	return context.WithValue(ctx, inUpdateKey{}, base)
+}
+
+// IsInUpdate checks whether the context is an Update.
+func IsInUpdate(ctx context.Context) bool {
+	return ctx.Value(inUpdateKey{}) != nil
+}
+
+// GetBaseline returns the baseline of the update, or nil when we
+// are not within an update context.
+func GetBaseline(ctx context.Context) interface{} {
+	return ctx.Value(inUpdateKey{})
+}
+
+// This is attached to contexts passed to webhook interfaces when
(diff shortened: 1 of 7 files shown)


dprotaso:
cc @scothis as I know there was another breaking change in this repo that affected you

mattmoor:
It looks like Build and Tekton aren't using this.  Eventing is, but the only immediate change they'd need is `apis.Immutable` => `apis.DeprecatedImmutable`.

n3wscott:
Eventing Staged: https://github.com/knative/eventing/pull/1037

mattmoor:
I'm going to revert the breaking portion of this change.  At one level, it's just the right amount of breaking, but with the diamond problem, it's a glorious PITA.

vdemeester:
> It looks like Build and Tekton aren't using this. Eventing is, but the only immediate change they'd need is `apis.Immutable` => `apis.DeprecatedImmutable`.

I'll take a look at it. That said, why not *just* marking them as `Deprecated` (the comment) and wait for a given amount of time, before removing the deprecated struct ?

mattmoor:
@vdemeester that's what I'm about to change this to.

mattmoor:
I'm ripping out the `apis.Annotatable` now though.

scothis:
Breaking changes to pkg may seem harmless, but it creates downstream pain when attempting to consume different versions of Knative components. For example, with riff we wanted to upgrade Serving and Build to 0.5.0 while leaving Eventing at 0.4.1. A breaking change in pkg meant we had to leave all of our types at 0.4.x. Not horrible, but also annoying.

mattmoor:
@scothis see the updated commit message.  We're going to leave things as-is when deprecating, and remove the interface when we remove the functionality from the webhook.

apis/contexts.go
```diff
@@ -0,0 +1,77 @@
...
+	return ctx.Value(inCreateKey{}) != nil
+}
+
+// This is attached to contexts passed to webhook interfaces when
+// the receiver being validated is being updated.
+type inUpdateKey struct{}
+
+// WithinUpdate is used to note that the webhook is calling within
+// the context of a Update operation.
+func WithinUpdate(ctx context.Context, base interface{}) context.Context {
```
n3wscott: `WithInUpdate`

apis/contexts.go
```diff
@@ -0,0 +1,77 @@
...
+	return context.WithValue(ctx, inUpdateKey{}, base)
+}
+
+// IsInUpdate checks whether the context is an Update.
+func IsInUpdate(ctx context.Context) bool {
```
n3wscott: I would be fine with `InUpdate` as the method name

apis/contexts.go
mattmoor: `WithinUpdate` expresses what I want to express, I worry `WithInUpdate` exposes implementation details (e.g. `inUpdate`) IMO.

n3wscott: ok

mattmoor:
/hold cancel

No longer breaking for anything but serving.

mattmoor:
🤔 We can't even remove the interface without the diamond problem between serving / eventing.

I'll leave both interface for now, but Serving is about to drop both entirely.

PR #369, opened 2019-04-07
Add the Convertible interface.

mattmoor:
This sketches the "Convertible" interface that @vagababov outlined in
last week's Serving API WG meeting.

See also: https://docs.google.com/document/d/1UFUfwIF4q4r6JVigom0jQivbcI8mFycWPc9z48yjSIc/edit#heading=h.9np2by1b65b

/lint

### DIFF
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -35,6 +35,16 @@ type Validatable interface {
 	Validate(context.Context) *FieldError
 }
 
+// Convertible indicates that a particular type supports conversions to/from
+// "higher" versions of the same type.
+type Convertible interface {
+	// ConvertUp up-converts the receiver into `to`.
+	ConvertUp(to Convertible) error
+
+	// ConvertDown down-converts from `from` into the receiver.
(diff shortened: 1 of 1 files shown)


apis/interfaces.go
```diff
@@ -35,6 +35,16 @@ type Validatable interface {
 	Validate(context.Context) *FieldError
 }
 
+// Convertible indicates that a particular type supports conversions to/from
+// "higher" versions of the same type.
+type Convertible interface {
```
vagababov: Do we care to specify whether we're doing staged a⇒b⇒c or a⇒c, or note that how a is converted to c is not prescribed and is an implementation detail?

mattmoor: I think that beyond the direction it's an implementation detail for you get from `a` to `c`.

vagababov: Well, then it would make sense to say, that any version above/below `x` is acceptable

vagababov:
/lgtm
/hold
just in case anyone else has comments.

PR #375, opened 2019-04-09
Add a method to check if a single pod running

Fredy-Z:
Sometimes we may not want to check if all pods are running but just one single pod.

So I add one function `WaitForPodRunning` to fulfill this need.

### DIFF
--- a/test/kube_checks.go
+++ b/test/kube_checks.go
@@ -105,9 +105,14 @@ func WaitForAllPodsRunning(client *KubeClient, namespace string) error {
 }
 
 // WaitForPodRunning waits for the given pod to be in running state
-func WaitForPodRunning(client *KubeClient, pod *corev1.Pod, namespace string) error {
+func WaitForPodRunning(client *KubeClient, name string, namespace string) error {
+	p := client.Kube.CoreV1().Pods(namespace)
 	return wait.PollImmediate(interval, podTimeout, func() (bool, error) {
-		return PodRunning(pod)
+		p, err := p.Get(name, metav1.GetOptions{})
+		if err != nil {
+			return true, err
+		}
+		return PodRunning(p)
 	})
 }
 
--- a/test/kube_checks.go
+++ b/test/kube_checks.go
@@ -112,14 +112,14 @@ func WaitForPodRunning(client *KubeClient, name string, namespace string) error 
 		if err != nil {
 			return true, err
 		}
-		return PodRunning(p)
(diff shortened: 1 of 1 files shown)


test/kube_checks.go
```diff
@@ -104,12 +104,32 @@ func WaitForAllPodsRunning(client *KubeClient, namespace string) error {
...
 	for _, pod := range podList.Items {
-		if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodSucceeded {
+		if isRunning, _ := PodRunning(&pod); !isRunning {
 			return false, nil
 		}
 	}
 	return true, nil
 }
+
+// PodRunning will check the status conditions of the pod and return true if it's Running
```
vagababov: Is this going to be used outside of this file?

test/kube_checks.go
```diff
@@ -104,12 +104,32 @@ func WaitForAllPodsRunning(client *KubeClient, namespace string) error {
...
+func PodRunning(pod *corev1.Pod) (bool, error) {
+	if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodSucceeded {
```
vagababov: this function never returns error. So just bool.

test/kube_checks.go
vagababov: Hm... ignore this one :)

test/kube_checks.go
Fredy-Z: Done.

test/kube_checks.go
```diff
@@ -104,12 +104,32 @@ func WaitForAllPodsRunning(client *KubeClient, namespace string) error {
...
+		if isRunning := PodRunning(&pod); !isRunning {
 			return false, nil
 		}
 	}
 	return true, nil
 }
+
+// PodRunning will check the status conditions of the pod and return true if it's Running
+func PodRunning(pod *corev1.Pod) bool {
+	if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodSucceeded {
```
vagababov: and now we can inline this into: 
`return pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodSucceeded` :-)

Fredy-Z: Done. Thanks!

/meow

vagababov:
That's a huge cat.

PR #376, opened 2019-04-09
Use string match to classify DNS error.

tcnghia:
It seems that typed-check aren't robust enough, so we are switching to
checking the error message here.

### DIFF
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -20,7 +20,6 @@ package spoof
 
 import (
 	"net"
-	"net/url"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


test/spoof/error_checks.go
```diff
@@ -30,14 +29,13 @@ func isTCPTimeout(e error) bool {
...
-			}
-		}
+	if err == nil {
+		return false
 	}
-	return false
+	// Checking by casting to url.Error and casting the nested error
+	// seems to be not as robust as string check.
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "no such host") || strings.Contains(msg, ":53")
```
srinivashegde86: can we add a sample err string in the comment? I would like to understand what :53 means here

tcnghia: Done.  :53 is the default port for DNS

srinivashegde86: thanks :+1:

PR #377, opened 2019-04-09
Do not hard code namespace when checking pod logs

Fredy-Z:
I think it's better to let user specify the `namespace`, rather than hard code it with the one passed as a flag, since it's not mandatory for the user to pass the `namespace` flag. 

Other functions in the `kube_checks.go` file are actually either allowing passing `namespace` or directly using the `pod` or `deployment` reference.


srinivashegde86:
why hold? eventing has vendored it in and will get the update when pkg is updated.

/lgtm
/approve

PR #380, opened 2019-04-10
Adding string pointer.

n3wscott:


n3wscott:
> we could have just pulled in proto2 go package as a dependency :)

not invented here!

vagababov:
> we could have just pulled in proto2 go package as a dependency :)

i.e.: https://github.com/golang/protobuf/blob/8d0c54c1246661d9a51ca0ba455d22116d485eaa/proto/lib.go#L459

PR #381, opened 2019-04-10
Add a version of GlobalResync that calls an event handler.

tcnghia:
This allows impl to apply filtering and queuing logic of their own,
instead of calling Enqueue() for everything.

Currently ConfigMap update event triggers GlobalResync which will evade the filtering logic like ClusterIngress class annotation (or KPA class annotation).  Once this is checked in we can switch ConfigMap update to use this version instead.

### DIFF
--- a/controller/helper.go
+++ b/controller/helper.go
@@ -53,7 +53,7 @@ func EnsureTypeMeta(f Callback, gvk schema.GroupVersionKind) Callback {
 }
 
 // SendGlobalUpdates triggers an update event for all objects from the
-// passed ShareInformer.
+// passed SharedInformer.
 //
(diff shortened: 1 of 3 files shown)


controller/helper.go
```diff
@@ -50,3 +51,17 @@ func EnsureTypeMeta(f Callback, gvk schema.GroupVersionKind) Callback {
 		f(copy)
 	}
 }
+
+// SendGlobalUpdates triggers an update event for all objects from the
+// passed ShareInformer.
```
vagababov:
```suggestion
// passed SharedInformer.
```

tcnghia: Done

PR #385, opened 2019-04-15
Add support for the Istio 1.1 per-route Headers.

mattmoor:
/hold

I want to have the Serving change ready.


mattmoor:
The serving change is here btw: https://github.com/knative/serving/pull/3758

PR #400, opened 2019-05-01
Add support for a structured URL type.

mattmoor:
This type can be used to accept `url: http://asdf.com` where in code we
get a `url.URL` to interact with.

This is based on: https://github.com/knative/pkg/pull/362


apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+// ParseURL attempts to parse the given string as a URI-Reference.
+func ParseURL(u string) *URL {
+	if u == "" {
+		return nil
+	}
+	pu, err := url.Parse(u)
+	if err != nil {
+		return nil
+	}
+	url := URL(*pu)
```
vagababov: why do you create a copy here?

apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+// String returns the full string representation of the URI-Reference.
+func (u *URL) String() string {
+	if u == nil {
+		return ""
+	}
+	uu := url.URL(*u)
+	return uu.String()
```
vagababov: probably joinable in a single line.

apis/url.go
mattmoor: It's a cast, so we can take the address.  If you prefer I can cast the pointer, but that felt gross.

apis/url.go
mattmoor: No, String is a pointer method.

apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+
+package apis
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+)
+
+// URL is a wrapper to url.URL.
```
akashrv: Update comment. It is an alias now

apis/url.go
vagababov: `return (*URL)(pu)` ?
Not the prettiest, but, IMO, better than a copy of the object?

apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+		return err
+	}
+	r := ParseURL(ref)
+	if r != nil {
+		*u = *r
+	}
+	return nil
+}
+
+// String returns the full string representation of the URI-Reference.
```
akashrv: URL - nit - fix the comments to refer to URL alias rather than URI-Reference

apis/url.go
vagababov: So why did you dereference it to a value? :)

apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+type URL url.URL
+
+// ParseURL attempts to parse the given string as a URI-Reference.
+func ParseURL(u string) *URL {
+	if u == "" {
+		return nil
+	}
+	pu, err := url.Parse(u)
+	if err != nil {
+		return nil
```
akashrv: return *URL, error?

apis/url.go
```diff
@@ -0,0 +1,73 @@
...
+
+// UnmarshalJSON implements the json unmarshal method used when this type is
+// unmarsheled using json.Unmarshal.
+// json.Unmarshaler impl
+func (u *URL) UnmarshalJSON(b []byte) error {
+	var ref string
+	if err := json.Unmarshal(b, &ref); err != nil {
+		return err
+	}
+	r := ParseURL(ref)
```
akashrv: If ParseURL returns error based on the other comment, then it could be handled here and returned in case of err!=nil

apis/url.go
mattmoor: Done

apis/url.go
mattmoor: Done

apis/url.go
mattmoor: Done

apis/url.go
mattmoor: Done

apis/url.go
mattmoor: Done

Issue #395, opened 2019-04-23
Extend Addressable


akashrv:
Today, Addressable struct is defined as 
```
type Addressable struct {
	Hostname string `json:"hostname,omitempty"`		
}
```
[Addressable interface](https://github.com/knative/eventing/blob/master/docs/spec/interfaces.md) is defined as 

> An Addressable resource receives events over a network transport (currently only HTTP is supported). The Addressable returns success when it has successfully handled the event (for example, by committing it to stable storage). When used as an Addressable, only the acknowledgement or return code is used to determine whether the event was handled successfully. One example of an Addressable is a Channel.

In its current definition, Addressable imposes certain limitations.

1. http scheme is implied and not other scheme is supported.
2. Even for http scheme, detailed http URL address cannot be specified such as ports, path etc.
3. Only one address could be specified by an addressable.

These limitations have been noticed in multiple use cases.
1. Channel dispatchers would like requests from different Channels to be routed to same dispatcher service. However, the dispatcher service needs to disambiguate requests coming from different channels. Doing this with just HostHeader requires dispatchers to create ExternalName services or multiple K8s services and then maintain a map of which service points to which channel. This could be simplified by using 'url path' as is used by Triggers.
2. Serving has a a type that requires an internal and an external address. @mattmoor  to add more details about this one.

Related issues: 
- [Add support for a urlPath to Sinks](https://github.com/knative/eventing-sources/issues/325)
- [Extend addressable with URL](https://github.com/knative/pkg/pull/362)

Proposal: Just based on above limitations here is a probable proposal
```
type Addressable struct {
         addresses map[string]*apis.URLRef `json:"url,omitempty"`
}
```
- Addressable would have list of named addresses, where
  - key to the map is the name of the address
  - value is a URLRef. URLRef will be new type that will enable URLs to be represented as strings on wire but in code it will be parsed into a url.URL. Refer to https://github.com/knative/pkg/pull/362/files for more details.
- a URL (scheme:[//authority]path[?query][#fragment]), where authority = [userinfo@]host[:port] keeps the design open-ended to support different schemes in Knative in future.
- We can have the "default" key reserved as the default URL or can add another field that will be a mandatory default URL, while the map is optional.

Above list of limitation, use cases and the proposal is just to kick off the discussion. The goal of this proposal is to capture all such requirements and come-up with a proposal that could be implemented with 'beta' Knative apis.

Please feel free to add your thoughts or extend the list of limitation and requirements in comments.

[...]

PR #401, opened 2019-05-02
Implement Addressable v1beta1 to use URL.

mattmoor:
Fixes: https://github.com/knative/pkg/issues/395

/hold


apis/duck/v1beta1/addressable_types.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Status AddressStatus `json:"status"`
+}
+
+// AddressStatus shows how we expect folks to embed Addressable in
+// their Status field.
+type AddressStatus struct {
+	Address *Addressable `json:"address,omitempty"`
```
vagababov: Not inlined?

mattmoor: This is how it works in `v1alpha1`, which we aren't changing.

mattmoor:
/hold cancel

No real opposition on the issue, so removing the hold\ here

PR #404, opened 2019-05-06
Drop error return type from controller.StartAll.

markusthoemmes:
This confused me in terms of the error modes available here. Went down a rabbithole propagating errors from here through StartAll only to notice this can never return an error in the first place.


markusthoemmes:
@dprotaso Indeed some downstream controllers check the error of this. The diff to fix that should be relatively straightforward though.

mattmoor:
/lgtm
/approve
/hold

Can you stage the change for the other repos?

markusthoemmes:
@mattmoor looking at it, I think there's nothing to stage actually. I do wonder however if we should punt on this change. The Run method of the admissioncontroller surfaces errors for instance. I'm a bit on the fence wrt. the usefulness of this.

mattmoor:
I'm fine with leaving the door open to error checking later.

PR #405, opened 2019-05-08
add trace id

lvjing2:
add trace id for reconcile, cause the knative/key is not enough yet if it reconcile for the same key for many times.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -289,7 +289,7 @@ func (c *Impl) processNextWorkItem() bool {
 
 	// Embed the key into the logger and attach that to the context we pass
 	// to the Reconciler.
-	logger := c.logger.With(zap.String(logkey.TraceId, uuid.New().String())).With(zap.String(logkey.Key, key))
+	logger := c.logger.With(zap.String(logkey.TraceId, uuid.New().String()), zap.String(logkey.Key, key))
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -277,7 +278,7 @@ func (c *Impl) processNextWorkItem() bool {
 
 	// Embed the key into the logger and attach that to the context we pass
 	// to the Reconciler.
-	logger := c.logger.With(zap.String(logkey.Key, key))
+	logger := c.logger.With(zap.String(logkey.TraceId, uuid.New().String())).With(zap.String(logkey.Key, key))
```
vagababov: `With` takes a list.

lvjing2: oh, nice tip

lvjing2: done

logging/logkey/constants.go
```diff
@@ -23,6 +23,8 @@ const (
 	// Key is the key (namespace/name) being reconciled.
 	Key = "knative.dev/key"
 
+	TraceId = "knative.dev/traceid"
```
mdemirhan: Can you please add the documentation for this? Something like:

// TraceId is used to track an asynchronous or long running operation within Knative.

lvjing2: ok, done

PR #407, opened 2019-05-08
Move serving/pkg/tracing to pkg/tracing, so that eventing can start using it

akashrv:
Moving serving/pkg/tracing to pkg/tracing, so that eventing can start using it. Moving the files as it is with no changes except for imports, as the package is independent and doesn't use anything that is specific to serving.
Will create an issue in serving repo to consume tracing from pkg, as soon as this PR is merged.


Harwayne:
Concretely, this helps with knative/eventing#704.

Issue #412, opened 2019-05-13
ConfigMap Watcher with Defaults


Harwayne:
## Expected Behavior

I should be able to create a [configmap.Watcher](https://github.com/knative/pkg/blob/9e0db8f0a7f47def18cf65e0859066d5e0377304/configmap/watcher.go#L30) that tolerates missing ConfigMaps by using a code supplied default.

I think it would be something like:

```golang
type DefaultingWatcher interface {
	// Watch is called to register a callback to be notified when a named ConfigMap changes.
	Watch(cmName string, o Observer)

	// WatchWithDefault is called to register a callback to be notified when a named 
	// ConfigMap changes. If no ConfigMap with that name exists or is deleted, then
	// the default value is used.
	WatchWithDefault(cmName string, def *v1.ConfigMap, o Observer)

	// Start is called to initiate the watches and provide a channel to signal when we should
	// stop watching.  When Start returns, all registered Observers will be called with the
	// initial state of the ConfigMaps they are watching.
	Start(<-chan struct{}) error
}
```

## Actual Behavior

If a watched ConfigMap is missing, configmap.Watcher fails to start leading the binary to fail to start (everywhere I have seen it started is in a `main`).

## Additional Info

This has come up for dynamic tracing support in knative/eventing. We want to allow per namespace tracing, so we can't simply precreate the `config-tracing` ConfigMap while installing eventing. So we want to use that ConfigMap if present, or a code supplied default if not.

PR #415, opened 2019-05-15
Defaulting watcher

Harwayne:
Fixes: #412 

## Proposed Changes

- Extend the existing `configmap.InformedWatcher` to use code provided defaults if the real ConfigMap does not exist.
    - Existing usage will not require any changes. All new functionality is strictly opt-in.

### DIFF
--- a/configmap/informed_watcher.go
+++ b/configmap/informed_watcher.go
@@ -62,8 +62,8 @@ type InformedWatcher struct {
 	informer corev1informers.ConfigMapInformer
 	started  bool
 
-	// cfgs are the default ConfigMaps to use if the real ones do not exist or are deleted.
-	cfgs map[string]*corev1.ConfigMap
+	// defaults are the default ConfigMaps to use if the real ones do not exist or are deleted.
+	defaults map[string]*corev1.ConfigMap
 
 	// Embedding this struct allows us to reuse the logic
 	// of registering and notifying observers. This simplifies the
@@ -79,10 +79,10 @@ var _ DefaultingWatcher = (*InformedWatcher)(nil)
 
 // WatchWithDefault implements DefaultingWatcher.
 func (i *InformedWatcher) WatchWithDefault(cm corev1.ConfigMap, o Observer) {
-	if i.cfgs == nil {
-		i.cfgs = map[string]*corev1.ConfigMap{}
+	if i.defaults == nil {
+		i.defaults = map[string]*corev1.ConfigMap{}
 	}
-	i.cfgs[cm.Name] = &cm
+	i.defaults[cm.Name] = &cm
 
 	i.m.Lock()
 	started := i.started
@@ -105,8 +105,8 @@ func (i *InformedWatcher) Start(stopCh <-chan struct{}) error {
 	// the informer to ensure that if a defaulted ConfigMap does exist, then the real value is
 	// processed after the default one.
 	for k := range i.observers {
-		if cfg, ok := i.cfgs[k]; ok {
-			i.addConfigMapEvent(cfg)
+		if def, ok := i.defaults[k]; ok {
+			i.addConfigMapEvent(def)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


configmap/informed_watcher.go
```diff
@@ -61,24 +62,61 @@ type InformedWatcher struct {
...
 		return err
 	}
 
 	// Wait until it has been synced (WITHOUT holing the mutex, so callbacks happen)
 	if ok := cache.WaitForCacheSync(stopCh, i.informer.Informer().HasSynced); !ok {
-		return errors.New("Error waiting for ConfigMap informer to sync.")
+		return errors.New("error waiting for ConfigMap informer to sync")
 	}
 
 	return i.checkObservedResourcesExist()
```
mattmoor: The point of this is to block until we've seen each thing at least once.  I am curious in which scenarios that might be inadequate?

Harwayne: This doesn't block, it returns an error if a ConfigMap doesn't. 

As for why, from #412:

This has come up for dynamic tracing support in knative/eventing. We want to allow per namespace tracing, so we can't simply precreate the config-tracing ConfigMap while installing eventing. 

We expect the ConfigMap to be absent most of them time. Only when the user has decided they want tracing in this namespace, will they create it. At which point, we want the new config picked up without requiring a restart of all the Pod.

Our expectation is that we have tracing either disabled or sampling set to zero in a code default. Have each Pod we control run the configmap.InformedWatcher so that when the ConfigMap is created/updated/deleted the Pod immediately notices and starts using the updated config.

configmap/informed_watcher.go
```diff
@@ -61,24 +62,61 @@ type InformedWatcher struct {
 	informer corev1informers.ConfigMapInformer
 	started  bool
 
+	// cfgs are the default ConfigMaps to use if the real ones do not exist or are deleted.
+	cfgs map[string]*corev1.ConfigMap
```
vagababov: Perhaps
```suggestion
	defaults map[string]*corev1.ConfigMap
```
?

configmap/informed_watcher.go
```diff
@@ -61,24 +62,61 @@ type InformedWatcher struct {
...
 // Asserts that InformedWatcher implements Watcher.
 var _ Watcher = (*InformedWatcher)(nil)
 
-// Start implements Watcher
+// Asserts that InformedWatcher implements DefaultingWatcher.
+var _ DefaultingWatcher = (*InformedWatcher)(nil)
+
+// WatchWithDefault implements DefaultingWatcher.
+func (i *InformedWatcher) WatchWithDefault(cm corev1.ConfigMap, o Observer) {
+	if i.cfgs == nil {
```
vagababov: Move this to `NewInformedWatcher`?

Harwayne: I decided against that because people can (but probably shouldn't) make an `InformedWatcher` directly, without going through `NewInformedWatcher`, so I would want this check here regardless.

configmap/informed_watcher.go
Harwayne: Done.

configmap/informed_watcher.go
vagababov: How? `sif` and `informer` are private and cannot be set outside of the provided constructors.

Harwayne: You're right, people can do it, but it will panic as soon as someone calls `Start()` because `informer` is `nil`. So panicking in a slightly different spot isn't worse.

Moved.

configmap/informed_watcher.go
```diff
@@ -61,24 +63,58 @@ type InformedWatcher struct {
...
+	i.m.Lock()
+	started := i.started
+	i.m.Unlock()
+	if started {
+		// TODO make both Watch and WatchWithDefault work after the InformedWatcher has started.
+		// This likely entails changing this to `o(&cm)` and having Watch check started, if it has
+		// started, then ensuring i.informer.Lister().ConfigMaps(i.Namespace).Get(cmName) exists and
+		// calling this observer on it. It may require changing Watch and WatchWithDefault to return
+		// an error.
+		panic(errors.New("cannot WatchWithDefault after the InformedWatcher has started"))
```
vagababov:
```suggestion
		panic("cannot WatchWithDefault after the InformedWatcher has started")
```

Harwayne: Done.

Issue #417, opened 2019-05-16
Risk condition: Failed calling admission webhook "webhook.eventing.knative.dev"


zxDiscovery:
## Expected Behavior

Deploy the knative eventing normolly.

## Actual Behavior

Deloy the knative eventing failed, error as follows:
Error: release knative failed: Internal error occurred: failed calling admission webhook "webhook.eventing.knative.dev": Post https://webhook.knative-eventing.svc:443/?timeout=30s: dial tcp 10.0.25.12:443: connect: connection refused

## Steps to Reproduce the Problem

This issue is led by https://github.com/knative/pkg/blob/5e4512dcb2cafff2779e6b1ba5091c642b4df2e1/webhook/webhook.go#L290
`RegistrationDelay` can't make sure the webhook create order correctly. 

We need the [server](https://github.com/knative/pkg/blob/5e4512dcb2cafff2779e6b1ba5091c642b4df2e1/webhook/webhook.go#L303) creates ready before the [`mutatingwebhookconfiguration`](https://github.com/knative/pkg/blob/5e4512dcb2cafff2779e6b1ba5091c642b4df2e1/webhook/webhook.go#L380) create

## Additional Info

https://github.com/knative/eventing/blob/master/config/500-webhook.yaml
https://github.com/knative/eventing/blob/master/config/400-webhook-service.yaml

evankanderson:
Will this fix need to be backported to 0.6.0?

Do we know when it was introduced?

vagababov:
> Will this fix need to be backported to 0.6.0?
> 
> Do we know when it was introduced?

My guess it has been there _forever_.

PR #418, opened 2019-05-16
Risk condition for server listen

zxDiscovery:
Fix https://github.com/knative/pkg/issues/417

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -267,6 +268,7 @@ func configureCerts(ctx context.Context, client kubernetes.Interface, options *C
 
 // Run implements the admission controller run loop.
 func (ac *AdmissionController) Run(stop <-chan struct{}) error {
+	var wg sync.WaitGroup
 	logger := ac.Logger
 	ctx := logging.WithLogger(context.TODO(), logger)
 	tlsConfig, caCert, err := configureCerts(ctx, ac.Client, &ac.Options)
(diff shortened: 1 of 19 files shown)


webhook/webhook.go
```diff
@@ -281,6 +283,19 @@ func (ac *AdmissionController) Run(stop <-chan struct{}) error {
...
+	go func() {
+		if err := server.ListenAndServeTLS("", ""); err != nil {
+			logger.Errorw("ListenAndServeTLS for admission webhook returned error", zap.Error(err))
+			close(serverBootstrapErrCh)
+		}
+		wg.Done()
+		logger.Info("server ready to listen")
+	}()
+
+	wg.Wait()
```
mattmoor: This is equivalent to not doing this in a go-routine, which I believe is incorrect given the examples:
https://golang.org/pkg/net/http/#ListenAndServe

webhook/webhook.go
```diff
@@ -281,6 +283,19 @@ func (ac *AdmissionController) Run(stop <-chan struct{}) error {
...
 	}
 
+	serverBootstrapErrCh := make(chan struct{})
+	wg.Add(1)
+	go func() {
+		if err := server.ListenAndServeTLS("", ""); err != nil {
+			logger.Errorw("ListenAndServeTLS for admission webhook returned error", zap.Error(err))
+			close(serverBootstrapErrCh)
+		}
+		wg.Done()
```
vagababov: always (unless strong reason not to exists) do `defer wg.Done()`

webhook/webhook.go
vagababov: +1 the way it's written this `wg` won't be ready until after the server starts. You need one more goroutine to probe the server, I guess, and in that goroutine trigger the `wg.Done`.

zxDiscovery: Thanks for your review. @mattmoor @vagababov 
Now I try to find a method to make sure the server is ready to create. Can I use the `http.Get()` method to access the server to check the server is ready?

zxDiscovery:
The CI test pass. Please help to review, thanks! @mattmoor @vagababov @grantr

webhook/webhook.go
```diff
@@ -265,8 +277,113 @@ func configureCerts(ctx context.Context, client kubernetes.Interface, options *C
...
+			case endpointCheckShutdown:
+				return true
+			case endpointCheckReady:
+				return false
+			case endpointCheckNotReady:
+				// continue waiting for endpoint to be ready
+			}
+		}
+	}
+}
```
mattmoor: I'm concerned about the amount of code we are using to solve this (without unit testing).

The initial implementation of this had a deadlock, and this basically uses an Endpoints informer to perform a readiness check, which itself has two problems:
1) Even Endpoints propagation is eventually consistent and no guarantee that the API Server will reach our pod.
2) Our webhooks don't currently define readiness probes, so they will report as ready instantly (before the http endpoint is available).

Let's discuss options more in the issue first?

mattmoor:
Going to close this, as I'm not sure we're in agreement on an approach here.  Let's discuss options in an issue first before jumping into the code.

PR #420, opened 2019-05-18
Add a new method to wrap the signal channel in a context.

mattmoor:


mattmoor:
I want things to be able to use the more idiomatic `<-ctx.Done()`, and passing a `ctx` around instead of threading a `stopCh` everywhere.

This is broken off of a larger experiment to explore a potential pattern for injection and boilerplate reduction around `ctx` during controller setup.

cc @n3wscott

vagababov:
But wouldn't the context be based off request?

signals/signal.go
```diff
@@ -41,3 +43,33 @@ func SetupSignalHandler() (stopCh <-chan struct{}) {
...
+	return &signalContext{stopCh: SetupSignalHandler()}
+}
+
+type signalContext struct {
+	stopCh <-chan struct{}
+}
+
+// Deadline implements context.Context
+func (scc *signalContext) Deadline() (deadline time.Time, ok bool) {
+	return time.Now(), false
```
dprotaso: I don't think we need to return `time.Now()`

```suggestion
	return
```

signals/signal.go
```diff
@@ -41,3 +43,33 @@ func SetupSignalHandler() (stopCh <-chan struct{}) {
...
+}
+
+// Done implements context.Context
+func (scc *signalContext) Done() <-chan struct{} {
+	return scc.stopCh
+}
+
+// Err implements context.Context
+func (scc *signalContext) Err() error {
```
dprotaso: Probably should follow the semantics of `Err()`as described in the `Context` [godoc](https://golang.org/pkg/context/#Context)

```
If Done is closed, Err returns a non-nil error explaining why:
```

dprotaso:
> This is broken off of a larger experiment to explore a potential pattern for injection and boilerplate reduction around ctx during controller setup.

Do you have more context on this? ie. are you looking at https://github.com/google/wire ?

vagababov:
Guice in Go.

dprotaso:
I thought folks moved to dagger 2

signals/signal.go
mattmoor: Done

signals/signal.go
mattmoor: Done.

mattmoor:
@dprotaso @vagababov 

More context: https://github.com/knative/serving/compare/master...mattmoor:safe-injection-site?expand=1

I want to reach the point where we can bootstrap controller binaries by effectively:
1. Link a shared "main"
2. Link a set of "reconcilers" (that register `NewController` methods)

PR #422, opened 2019-05-22
Update OWNERS to use OWNERS_ALIASES

adrcunha:
Part of knative/test-infra#579


adrcunha:
> This only does this for two groups:

Thanks, I forgot to push the remaining changes. PTAL.

PR #423, opened 2019-05-24
Create a library for Reconciler dependency injection.

mattmoor:
This creates a base library that enables us to leverage the dependency
graph to convey injection dependencies.  Our controllers will start
to get their informers off of a `context.Context` via `fooinformers.Get(ctx)`.

By simply linking `fooinformers` to access the informer in this way, we trigger
a cascade of `init`-time registrations for the informer, and its transitive
dependencies (e.g. shared informer factory, client).

The `ctx` is infused with these informers in `package main` by linking the
appropriate reconcilers (transitively registering everything) and then calling
`ctx, informers := injection.SetupInformers(ctx, cfg, resyncPeriod)`.

`informers` contains the unstarted informers, which should be passed to
`controller.StartInformers` after the controllers have been initialized.

/hold

### DIFF
--- a/injection/clients.go
+++ b/injection/clients.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package injection
+
+import (
+	"context"
+
+	"k8s.io/client-go/rest"
+)
+
+// ClientInjector holds the type of a callback that attaches a particular
+// client type to a context.
+type ClientInjector func(context.Context, *rest.Config) context.Context
+
+func (i *impl) RegisterClient(ci ClientInjector) {
+	i.m.Lock()
+	defer i.m.Unlock()
+
+	i.clients = append(i.clients, ci)
+}
+
+func (i *impl) GetClients() []ClientInjector {
+	i.m.RLock()
+	defer i.m.RUnlock()
+
+	// Copy the slice before returning.
+	return append(i.clients[:0:0], i.clients...)
+}
--- a/injection/clients/dynamicclient/dynamicclient.go
+++ b/injection/clients/dynamicclient/dynamicclient.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dynamicclient
+
+import (
+	"context"
+
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+
+	"github.com/knative/pkg/injection"
+)
+
+func init() {
+	injection.Default.RegisterClient(withClient)
+}
+
+// Key is used as the key for associating information
+// with a context.Context.
+type Key struct{}
+
+func withClient(ctx context.Context, cfg *rest.Config) context.Context {
+	return context.WithValue(ctx, Key{}, dynamic.NewForConfigOrDie(cfg))
+}
+
+// Get extracts the Dynamic client from the context.
+func Get(ctx context.Context) dynamic.Interface {
+	untyped := ctx.Value(Key{})
+	if untyped == nil {
+		return nil
+	}
+	return untyped.(dynamic.Interface)
+}
--- a/injection/clients/dynamicclient/dynamicclient_test.go
+++ b/injection/clients/dynamicclient/dynamicclient_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dynamicclient
+
+import (
+	"context"
+	"testing"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/knative/pkg/controller"
+	"github.com/knative/pkg/injection"
+)
+
+func TestRegistration(t *testing.T) {
+	ctx := context.Background()
+
+	// Get before registration
+	if empty := Get(ctx); empty != nil {
+		t.Errorf("Unexpected informer: %v", empty)
+	}
+
+	// Check how many informers have registered.
+	inffs := injection.Default.GetClients()
+	if want, got := 1, len(inffs); want != got {
+		t.Errorf("GetClients() = %d, wanted %d", want, got)
+	}
+
+	// Setup the informers.
+	var infs []controller.Informer
+	ctx, infs = injection.Default.SetupInformers(ctx, &rest.Config{})
+
+	// We should see that a single informer was set up.
+	if want, got := 0, len(infs); want != got {
+		t.Errorf("SetupInformers() = %d, wanted %d", want, got)
+	}
+
+	// Get our informer from the context.
+	if inf := Get(ctx); inf == nil {
+		t.Error("Get() = nil, wanted non-nil")
+	}
+}
--- a/injection/clients/dynamicclient/fake/fake.go
+++ b/injection/clients/dynamicclient/fake/fake.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fake
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/rest"
+
+	"github.com/knative/pkg/injection"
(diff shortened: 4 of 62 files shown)


mattmoor:
cc @dprotaso

injection/informers.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Based on the reconcilers we have linked, build up a set of informer factories
+	// and inject them onto the context.
+	for _, ifi := range GetInformerFactories() {
+		ctx = ifi(ctx, resyncPeriod)
+	}
+
+	// Based on the reconcilers we have linked, build up a set of informers
+	// and inject them onto the context.
+	var inf controller.Informer
```
vagababov: What if two reconcilers ask for the same informer? Will there be duplicates?

injection/sharedmain/main.go
```diff
@@ -0,0 +1,136 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

injection/clients.go
```diff
@@ -0,0 +1,47 @@
...
+
+func RegisterClient(ci ClientInjector) {
+	clientMutex.Lock()
+	defer clientMutex.Unlock()
+
+	clients = append(clients, ci)
+}
+
+func GetClients() []ClientInjector {
+	clientMutex.Lock()
```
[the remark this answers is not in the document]
mattmoor: Done

injection/informers.go
mattmoor: `init()` is run once per package according to the dependency graph, so it should work exactly as intended.

injection/clients/dynamicclient/dynamicclient.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/knative/pkg/injection"
+)
+
+func init() {
+	injection.RegisterClient(withDynamicClient)
+}
+
+// dynamicClientKey is used as the key for associating information
+// with a context.Context.
+type dynamicClientKey struct{}
```
dprotaso: If we go with the typed struct we'd need the ability to set clients/informers etc. Thus either this key would need to be public or we provide a `Set` func

injection/clients/dynamicclient/dynamicclient_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Get before registration
+	if empty := Get(ctx); empty != nil {
+		t.Errorf("Unexpected informer: %v", empty)
+	}
+
+	// Check how many informers have registered.
+	inffs := injection.GetClients()
+	if want, got := 1, len(inffs); want != got {
+		t.Errorf("GetInformers() = %d, wanted %d", want, got)
```
dprotaso: This block is fetching `clients` but the comment & error message say `informers`

injection/clients/kubeclient/kubeclient_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Get before registration
+	if empty := Get(ctx); empty != nil {
+		t.Errorf("Unexpected informer: %v", empty)
+	}
+
+	// Check how many informers have registered.
+	inffs := injection.GetClients()
+	if want, got := 1, len(inffs); want != got {
+		t.Errorf("GetInformers() = %d, wanted %d", want, got)
```
dprotaso: same comment/method call mismatch

injection/doc.go
```diff
@@ -0,0 +1,65 @@
...
+//   package main
+//
+//   import (
+//   	// The set of controllers this controller process runs.
+//      // Linking these will register the controllers and their transitive
+//      // dependencies, after which the shared main can set up the rest.
+//   	_ "github.com/knative/foo/pkg/reconciler/matt"
+//   	_ "github.com/knative/foo/pkg/reconciler/scott"
+//   	_ "github.com/knative/foo/pkg/reconciler/ville"
+//   	_ "github.com/knative/foo/pkg/reconciler/dave"
```
dprotaso: lol

injection/clients.go
```diff
@@ -0,0 +1,47 @@
...
+	defer clientMutex.Unlock()
+
+	return clients[:]
```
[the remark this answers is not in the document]
mattmoor: My understanding was that this was the simplest way to copy it so that it couldn't be modified by an `append()`.  That may be incorrect (and some simple Googling doesn't suggest I am right).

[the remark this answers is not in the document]
vagababov: It will be a copy, yes.

injection/doc.go
vagababov: Something's wrong. No `wesley/hutchinson`.

mattmoor: :)

injection/clients.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// ClientInjector holds the type of a callback that attaches a particular
+// client type to a context.
+type ClientInjector func(context.Context, *rest.Config) context.Context
+
+var (
+	clientMutex sync.Mutex
+	clients     []ClientInjector
+)
```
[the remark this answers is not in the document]
mattmoor: cc @n3wscott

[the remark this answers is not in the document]
mattmoor: tl;dr I think I like this idea, I'll play around some more (hopefully) tomorrow

Thanks, @dprotaso

injection/clients/dynamicclient/dynamicclient_test.go
mattmoor: good catch

injection/clients/kubeclient/kubeclient_test.go
mattmoor: Good catch

injection/informers.go
```diff
@@ -0,0 +1,74 @@
...
+	// Based on the reconcilers we have linked, build up a set of clients and inject
+	// them onto the context.
+	for _, ci := range GetClients() {
+		ctx = ci(ctx, cfg)
+	}
+
+	// Based on the reconcilers we have linked, build up a set of informer factories
+	// and inject them onto the context.
+	for _, ifi := range GetInformerFactories() {
+		ctx = ifi(ctx, resyncPeriod)
```
mattmoor: I think we want `resyncPeriod` on `context.Context`, so that the tracker can key off of the same value.

injection/clients.go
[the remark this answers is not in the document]
dprotaso: Slices are value types that point to a backing array on a heap.

Returning `clients` is already returning a copy ;)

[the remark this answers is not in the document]
dprotaso: If the intent is to prevent side-effects (ie. replacing an element at some index affects all arrays) then you need to use `make` and copy the elements into the new slice.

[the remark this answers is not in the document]
vagababov: I guess, we're looking for `append(a[:0:0], a...) ` :-)

injection/clients/dynamicclient/dynamicclient.go
mattmoor: 🤔 I think we'll only need this to be public if we *don't* do that.  Still we may want it to be public for more sophisticated testing.

injection/clients.go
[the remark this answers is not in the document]
mattmoor: I added `injection.Interface` and `injection.Default` / `injection.Fake` (instances of an implementation of `injection.Interface`).

injection/clients/dynamicclient/fake/fake.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/rest"
+
+	"github.com/knative/pkg/injection"
+	"github.com/knative/pkg/injection/clients/dynamicclient"
+)
+
+var Get = dynamicclient.Get
```
mattmoor: Rather than aliasing here, this should reimplement Get to return the client with the fake type.

dprotaso:
Looking at some old exploration I did what could be useful here is that SharedInformerFactory will manage and track child informers for your.

This means you can look up [informers using GroupVersionKind](https://github.com/dprotaso/serving/blob/route-reconciler-wip/pkg/reconciler/dependencies.go#L121)

And then `Start`'ing your SharedInformerFactory will start all the managed informers ([example](https://github.com/dprotaso/serving/blob/route-reconciler-wip/pkg/reconciler/dependencies.go#L89-L115)).

injection/sharedmain/main.go
```diff
@@ -0,0 +1,115 @@
...
+const (
+	component = "controller"
+)
+
+var (
+	masterURL  = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	kubeconfig = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+)
+
+func Main() {
```
dprotaso: Can we add some flexibility here wrt. to enabling customization.

For example - I may want to expose some runtime stats using https://golang.org/pkg/expvar - but I need the context (from signals) to help coordinate shutdown (or I pass one in) etc...

mattmoor:
@dprotaso we used to use `Start` on the shared informer factories, but IIRC there is a bug where if nothing registers an event it never successfully starts (it just hangs).  Happy to discuss further, but I'm curious what you are trying to simplify?  We'd either have to gather the list of factories or the list of informers to start regardless, right?

injection/sharedmain/main.go
mattmoor: Yeah, we can finesse this a bit to improve it's flexibility, this was just a starting point.

mattmoor: I added a few variations to let folks customize aspects of this while sharing the meat of it.

dprotaso:
> We'd either have to gather the list of factories or the list of informers to start regardless, right?

Yeah - there's just less factories to manage (ie. 5) vs many more informers. I figured the less boilerplate the better.

dprotaso:
> but IIRC there is a bug where if nothing registers an event it never successfully starts (it just hangs).

Do you mean an event handler?

mattmoor:
@dprotaso Not sure what boilerplate you are referring to?  Done generically it seems like the same LoC, right?  Ultimately all of the boilerplate should end up hidden regardless.

dprotaso:
>  Not sure what boilerplate you are referring to?

If that `SharedInformerFactory` bug is resolved you could drop the logic of registering informers.

ie. 
```go
func init() {
	injection.Default.RegisterInformer(withInformer)
}

// Key is used as the key for associating information
// with a context.Context.
type Key struct{}

func withInformer(ctx context.Context) (context.Context, controller.Informer) {
	f := factory.Get(ctx)
	inf := f.Autoscaling().V1().HorizontalPodAutoscalers()
	return context.WithValue(ctx, Key{}, inf), inf.Informer()
}

// Get extracts the Kubernetes Hpa informer from the context.
func Get(ctx context.Context) autoscalingv1.HorizontalPodAutoscalerInformer {
	untyped := ctx.Value(Key{})
	if untyped == nil {
		return nil
	}
	return untyped.(autoscalingv1.HorizontalPodAutoscalerInformer)
}
```

becomes
```go
// Get extracts the Kubernetes Hpa informer from the context.
func Get(ctx context.Context) autoscalingv1.HorizontalPodAutoscalerInformer {
	f := factory.Get(ctx)
        return f.Autoscaling().V1().HorizontalPodAutoscalers()
}
```

mattmoor:
@dprotaso Ah, I see what you mean.  Yes, but my point was that as far as the end-user is concerned this is all codegen's problem.

mattmoor:
I have started to sketch some thoughts [here](https://docs.google.com/document/d/1v_aCyF8gwR5WHPFwsEd8OFSFi-LMERtPWluiwabAhws/edit?hl=en#) (on team drive)

mattmoor:
I added a README.md with a mini-HOWTO and roughly describing how things work.  I'll fill out the code-gen section in the PR that implements the code-gen stacked on top of this.

PR #425, opened 2019-05-26
Share the configmap name logic for Logging and Metrics.

mattmoor:
This also adds a way to fetch the metrics "domain" following a similar pattern to `system.Namespace()`.

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -29,6 +29,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	DomainEnv        = "METRICS_DOMAIN"
+	ConfigMapNameEnv = "CONFIG_OBSERVABILITY_NAME"
+)
+
 // metricsBackend specifies the backend to use for metrics
 type metricsBackend string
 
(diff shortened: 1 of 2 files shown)


metrics/config.go
```diff
@@ -201,8 +206,8 @@ func getMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsC
 
 // UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
 // when a config map is updated.
-// DEPRECATED. Use UpdateExporter instead.
-func UpdateExporterFromConfigMap(domain string, component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
```
yanweiguo: Do you intend to use this function in serving repo? I didn't find this is being used in https://github.com/knative/serving/pull/4163.

mattmoor: It is, but indirectly.  This enables the entire controller `main()` to be defined in `knative/pkg` and simply called from Serving (or Eventing).

mattmoor: At whoops, not sure if the PR you linked does it quite yet, but https://github.com/knative/pkg/pull/423 and https://github.com/knative/serving/pull/4162 do what I mentioned.

PR #427, opened 2019-05-27
Simplify the default controller.Impl constructor.

mattmoor:
A while back we added a "StatsReporter" argument to `controller.NewImpl`,
but in serving every callsite of this is passing:
```
controller.NewImpl(r, logger, "my-string", MustNewStatsReporter("my-string", logger))
```

Where `MustNewStatsReporter` is just a form of knative/pkg's `controller.NewStatsReporter`
that logs fatally when an error is returned.  It is notable that Serving's logic has been
duplicated to both Build and Eventing.

There are a handful of changes here:
1. Move MustNewStatsReporter into knative/pkg
2. Expose the current interface as NewImplWithStats
3. Drop the StatsReporter from NewImpl and default to `MustNewStatsReporter()`

This is a breaking change for downstream repositories, but should make their callsites universally simpler.

/hold

I am going to stage this change on the other repositories, but please voice any objections here in the meantime.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -123,7 +123,11 @@ type Impl struct {
 
 // NewImpl instantiates an instance of our controller that will feed work to the
 // provided Reconciler as it is enqueued.
-func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter) *Impl {
+func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string) *Impl {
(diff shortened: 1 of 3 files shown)


mattmoor:
cc @ImJasonH @n3wscott FYI

mattmoor:
I've staged PRs against Serving, Build and Eventing with this change.

mattmoor:
As some further context, I was starting to put together a `sample-controller` and found myself needing to copy/paste this yet again and so I wanted to address this at its root.

dprotaso:
FYI - just made an issue to capture opinions of versioning (ie. accepting breaking changes etc.) - see referenced issue above

PR #428, opened 2019-05-27
This upstreams a useful helper from serving.

mattmoor:
In serving we have `reconciler.Handler` that wraps a handler function
(e.g. `Enqueue`) in the `cache.ResourceEventHandlerFuncs`.  This pattern
was becoming pervasive, so this simpler handler dramatically reduced our
boilerplate.


mattmoor:
LOL, it seems eventing has cloned this as well 🤣

PR #429, opened 2019-05-28
Add a testing package to setup metrics.Domain()

mattmoor:
This mirrors what we do for `system.Namespace()` today.


dprotaso:
/lgtm

so many `init()`'s lately ;)

PR #432, opened 2019-05-28
Remove OWNERS/tests under /vendor in update-deps.sh

adrcunha:
These are unnecessary.


adrcunha:
Unrelated unit test failure.

/retest

PR #433, opened 2019-05-28
change metav1.Object to runtime.Object so we can test moar objects

vaikas-google:
Without this fix you can't expect unstructured.Unstructured object creations which are needed for some duck related stuff.


dprotaso:
🦆

PR #436, opened 2019-05-30
Logging config parses all components, rather than just a list passed in.

Harwayne:
Logging config parses all components, rather than just a list passed in.

The current API exposed before this change is easy to get wrong. For example, knative/serving's cmd/controller/main.go doesn't pass in the components list, so dynamic updates never occur.

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -142,7 +142,7 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 
 	lc.LoggingLevel = make(map[string]zapcore.Level)
 	for k, v := range data {
-		if component := strings.TrimPrefix(k, "loglevel."); component != k {
+		if component := strings.TrimPrefix(k, "loglevel."); component != k && component != "" {
 			if len(v) > 0 {
 				level, err := levelFromString(v)
 				if err != nil {
--- a/logging/config.go
+++ b/logging/config.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -131,7 +132,7 @@ const defaultZLC = `{
 
 // NewConfigFromMap creates a LoggingConfig from the supplied map,
 // expecting the given list of components.
-func NewConfigFromMap(data map[string]string, components ...string) (*Config, error) {
+func NewConfigFromMap(data map[string]string) (*Config, error) {
 	lc := &Config{}
 	if zlc, ok := data["zap-logger-config"]; ok {
 		lc.LoggingConfig = zlc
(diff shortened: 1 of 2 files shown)


logging/config.go
```diff
@@ -140,25 +141,25 @@ func NewConfigFromMap(data map[string]string, components ...string) (*Config, er
...
 
 	lc.LoggingLevel = make(map[string]zapcore.Level)
-	for _, component := range components {
-		if ll := data["loglevel."+component]; len(ll) > 0 {
-			level, err := levelFromString(ll)
-			if err != nil {
-				return nil, err
+	for k, v := range data {
+		if strings.HasPrefix(k, "loglevel.") {
+			component := k[len("loglevel."):]
```
vagababov: 1. const "loglevel."?
2. Move `component :=` into, it seems unused outside of it.

logging/config_test.go
```diff
@@ -131,7 +131,7 @@ func TestNewConfigNoEntry(t *testing.T) {
 
 func TestNewConfig(t *testing.T) {
 	wantCfg := "{\"level\": \"error\",\n\"outputPaths\": [\"stdout\"],\n\"errorOutputPaths\": [\"stderr\"],\n\"encoding\": \"json\"}"
-	wantLevel := zapcore.InfoLevel
+	wantLevel := zapcore.ErrorLevel
```
vagababov: any reason for this change?

Harwayne: I want to show that the ConfigMap was actually parsed correctly. InfoLevel is the zero value, so will be returned if that key is not set for some reason.

logging/config.go
Harwayne: I can't simply move things around as I don't want to index past the end of the string. But I can do:

```
if component := strings.TrimPrefix(k, "loglevel."); component != k {}
```

Would you like that?

logging/config.go
```diff
@@ -140,25 +141,24 @@ func NewConfigFromMap(data map[string]string, components ...string) (*Config, er
 	}
 
 	lc.LoggingLevel = make(map[string]zapcore.Level)
-	for _, component := range components {
-		if ll := data["loglevel."+component]; len(ll) > 0 {
-			level, err := levelFromString(ll)
-			if err != nil {
-				return nil, err
+	for k, v := range data {
+		if component := strings.TrimPrefix(k, "loglevel."); component != k {
```
vagababov: `&& component != ""` ?

Harwayne: TrimPrefix will return the original string unaltered, if the prefix is not present. So this line is removing the prefix if it is there, then seeing if anything changed (showing that the prefix was present).

vagababov: I know, but key be "loglevel.", then removing prefix would be "".

Harwayne: Done.

PR #447, opened 2019-06-07
Eliminate controller registration in favor of passing a list to the shared main.

mattmoor:


injection/doc.go
```diff
@@ -25,7 +25,7 @@ limitations under the License.
 //   import (
 //     // Simply linking this triggers the injection of the informer, which links
 //     // the factory triggering its injection, and which links the client,
-//     // triggering its injection.
+//     // triggering its injection.  All you need to know is that it works :)
```
n3wscott: lol

dprotaso:
The doc is out of date now - not sure if you want to update it

https://docs.google.com/document/d/1v_aCyF8gwR5WHPFwsEd8OFSFi-LMERtPWluiwabAhws/edit?hl=en#


Secondly, can you comment here why we went the 'other' way in this scenario. I have guess.

PR #455, opened 2019-06-10
Remove GCP default auth.

n3wscott:
Remote auth for GCP was on by default. Removed the import.


injection/sharedmain/main.go
```diff
@@ -25,7 +25,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	// Uncomment the following line to load the gcp plugin (only required to authenticate against GKE clusters).
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	//_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
mattmoor: We should just remove this completely.  If folks want this, they can put this line into `./cmd/controller/main.go`

PR #463, opened 2019-06-14
upstreaming channel ducktypes and the addressable informer.

n3wscott:
/assign @mattmoor


apis/duck/addressable_resolver_test.go
```diff
@@ -0,0 +1,284 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

system/cluster.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

system/cluster_test.go
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

n3wscott:
I think this will have breaking downstream changes. I need to test integrating it with a few deps.

apis/duck/v1alpha1/channelable_types.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// ChannelableSpec contains Spec of the Channelable object
+type ChannelableSpec struct {
+	SubscribableTypeSpec `json:",inline"`
+}
+
+// ChannelableStatus contains the Status of a Channelable object.
+type ChannelableStatus struct {
+	// AddressStatus is the part where the Channelable fulfills the Addressable contract.
```
vaikas-google: FYI, since you're upstreaming, just curious about your thoughts on this?
https://github.com/knative/eventing/issues/1375

I was going to take a crack at this on Monday, but if you're doing it here, consider adding the duckv1beta1.Status field here as well? Currently it's not easy to figure out what the Status of Channelable is without knowing what type it is to get to the actual Status.

n3wscott: I think I made a mess of this PR and it needs to be scoped down...

n3wscott: I will put a pin in this one and then look at moving the types up one at a time.

`apis/duck/addressable_resolver.go` was the thing that make it hard.

n3wscott:
/hold

I will break this apart.

PR #464, opened 2019-06-14
Add a doc on releasing knative/pkg.

mattmoor:


RELEASING.md
```diff
@@ -0,0 +1,79 @@
...
+If problems are found, they are addressed and the update is merged, and this
+process repeats until knative/pkg can be cleanly updated without any changes.
+
+> If by mid-week, we do not have a clean PR in each repository, the person
+> driving the update should escalate to avoid delaying the release.
+
+
+### Step #2: Friday the week prior to each release.
+
+A release branch is snapped on knative/pkg with the form `release-0.X` where `X`
```
n3wscott: how does one "snap" a branch? it is automated?

mattmoor: I usually do it through the github UI :)

RELEASING.md
```diff
@@ -0,0 +1,79 @@
+# Releasing knative/pkg
+
+We release the components of Knative ever 6 weeks.  All of these components must
```
vaikas-google: nit: every 6 weeks.

RELEASING.md
```diff
@@ -0,0 +1,79 @@
...
+be moved to the latest "release" of the knative/pkg shared library prior to each
+release, but likely this should happen incrementally over each milestone.
+
+## Release Process.
+
+### Step #1: Monday the week prior to each release.
+
+Early the week prior to the Knative release, each of the downstream repositories
```
vaikas-google: Title says Monday, this says early. Either remove the Monday from the title or make them match?

RELEASING.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+Then the following is run:
+
+```shell
+dep ensure -update github.com/knative/pkg
+```
+
+If problems are found, they are addressed and the update is merged, and this
+process repeats until knative/pkg can be cleanly updated without any changes.
```
vaikas-google: There should probably be something about the repos need to keep checking (or can some sort of notification be set up?). I'm thinking of the scenario where, say eventing repo does this on Monday morning, everything looks great and then some other repo has an issue, requiring an update to pkg, so now every repo needs to grab the newer version. Saying you have to manually check the pkg repo is fine, but point being is that there's some on-going work to keep checking and re-testing.

RELEASING.md
n3wscott: "Each downstream" can we list them here and update when we find more?

RELEASING.md
```diff
@@ -0,0 +1,79 @@
...
+```toml
+[[override]]
+  name = "github.com/knative/pkg"
+  branch = "master"
+```
+
+Then the following is run:
+
+```shell
+dep ensure -update github.com/knative/pkg
```
n3wscott: This brings in all the owner files from pkg.

n3wscott: they also need to run `./hack/update-deps.sh`

RELEASING.md
mattmoor: We now have automation that updates the list of repos above to the latest commit in master daily like the rest, so the above basically devolves to making sure that is working, but it's written as if it's the only game in town (because I wrote the automation after putting this together to try and make this easier!).

RELEASING.md
```diff
@@ -0,0 +1,91 @@
...
+./hack/update-deps.sh
+```
```
n3wscott: Change to: `./hack/update-codegen.sh`

RELEASING.md
```diff
@@ -0,0 +1,91 @@
...
+### Step #1: Monday the week prior to each release.
+
+On Monday of the week prior to the Knative release, each of the downstream
+repositories should stage a `[WIP]` Pull Request that advances the knative/pkg
+dependency to the latest commit.
+
+At present, these downstream repositories include:
+
+1. knative/serving
+1. knative/eventing
```
n3wscott: please add `knative/eventing-contrib`

RELEASING.md
mattmoor: done

RELEASING.md
mattmoor: done

PR #467, opened 2019-06-17
#457 Duck type user annotation logic

aliok:
Fixes https://github.com/knative/pkg/issues/457

Moving the user annotation logic done in Knative eventing in https://github.com/knative/eventing/pull/1293 to Knative `pkg`.

Once these changes are merged, the only thing to do at Knative modules is making the resources implement `HasSpec` interface to get them annotated with `abc.knative.dev/creator` and `abc.knative.dev/updater` annotations. See the reason for `HasSpec` interface here: https://github.com/knative/pkg/pull/467#pullrequestreview-250415818

~~I will adapt the tests after I get a quick feedback on the direction.~~
Tests are also adapted.

### DIFF
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -70,6 +70,6 @@ type Annotatable interface{}
 // HasSpec indicates that a particular type has a specification information
 // and that information is retrievable.
 type HasSpec interface {
-	// GetSpec returns the spec of the resource.
-	GetSpec() interface{}
+	// GetUntypedSpec returns the spec of the resource.
+	GetUntypedSpec() interface{}
 }
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -66,3 +66,10 @@ type Listable interface {
 // The webhook functionality for this has been turned down, which is why this
 // interface is empty.
 type Annotatable interface{}
+
+// HasSpec indicates that a particular type has a specification information
+// and that information is retrievable.
+type HasSpec interface {
+	// GetSpec returns the spec of the resource.
+	GetSpec() interface{}
+}
--- a/webhook/user_info.go
+++ b/webhook/user_info.go
@@ -14,10 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package apis
+package webhook
 
 import (
 	"context"
+	"github.com/knative/pkg/apis"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -27,14 +28,14 @@ const (
 	// the user that created the resource.
 	CreatorAnnotationSuffix = "/creator"
 
-	// UpdaterAnnotationSuffix is the suffix of the annotation key to describe
-	// the user that last updated the resource.
-	UpdaterAnnotationSuffix = "/lastModifier"
+	// LastModifierAnnotationSuffix is the suffix of the annotation key to describe
+	// the user who last modified the resource.
+	LastModifierAnnotationSuffix = "/lastModifier"
 )
 
-// SetUserInfoAnnotations sets creator and updater annotations on a resource.
-func SetUserInfoAnnotations(resource HasSpec, ctx context.Context, groupName string) {
-	if ui := GetUserInfo(ctx); ui != nil {
+// SetUserInfoAnnotations sets creator and lastModifier annotations on a resource.
+func SetUserInfoAnnotations(resource apis.HasSpec, ctx context.Context, groupName string) {
+	if ui := apis.GetUserInfo(ctx); ui != nil {
 		objectMetaAccessor, ok := resource.(metav1.ObjectMetaAccessor)
 		if !ok {
 			return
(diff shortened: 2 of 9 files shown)


apis/interfaces.go
```diff
@@ -66,3 +66,10 @@ type Listable interface {
 // The webhook functionality for this has been turned down, which is why this
 // interface is empty.
 type Annotatable interface{}
+
+// HasSpec indicates that a particular type has a specification information
+// and that information is retrievable.
+type HasSpec interface {
```
aliok: Quoting from https://github.com/knative/eventing/pull/1293

> * The reason for `HasSpec` interface is, in the CRD using the `spec` field for specification of a resource is a convention but following also would work if the operator is designed to read the specs from `LeSpecification` field:

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
spec:
  ...
  validation:
    openAPIV3Schema:
      properties:
        LeSpecification:
          properties:
            ...
```

aliok:
I will add some tests now

webhook/user_info.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+const (
+	// CreatorAnnotationSuffix is the suffix of the annotation key to describe
+	// the user that created the resource.
+	CreatorAnnotationSuffix = "/creator"
+
+	// LastModifierAnnotationSuffix is the suffix of the annotation key to describe
+	// the user who last modified the resource.
+	LastModifierAnnotationSuffix = "/lastModifier"
```
aliok: This annotation is changed from `xyz.knative.dev/updater` to `xyz.knative.dev/lastModifier`.

It was that way in eventing module and made more sense to me, thus I used it that way.

Happy to revert to `updater` if there are any objections.

aliok:
tests are added/adapted now. ready for final review.

webhook/user_info.go
mattmoor: I believe this is used by client tooling, so it's a breaking change

apis/interfaces.go
mattmoor: This should be `GetUntypedSpec` or something.  Otherwise, implementations cannot use `GetSpec` to return their typed Spec object, which I was literally just recommending for the Ingress work in Serving.

webhook/user_info_test.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+func TestSetUserInfoAnnotationsWhenWithinCreate(t *testing.T) {
+	tests := []struct {
+		name                string
+		configureContext    func(context.Context) context.Context
+		setup               func(context.Context, *Resource)
+		expectedAnnotations map[string]string
+	}{
+		{
```
mattmoor: nit: use `{{` and `}, {` to reduce indentation

webhook/user_info_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func TestSetUserInfoAnnotationsWhenWithinUpdate(t *testing.T) {
+	tests := []struct {
+		name                string
+		configureContext    func(context.Context, *Resource) context.Context
+		setup               func(context.Context, *Resource)
+		expectedAnnotations map[string]string
+	}{
+		{
```
mattmoor: As above `{{` and `}, {`

webhook/webhook.go
```diff
@@ -99,6 +99,9 @@ type ControllerOptions struct {
 	// TLS Client Authentication.
 	// The default value is tls.NoClientCert.
 	ClientAuth tls.ClientAuthType
+
+	// GroupName is the name of the group that is using this webhook.
+	GroupName string
```
mattmoor: For serving, this is not a singular group.  Can this be extracted from the resource TypeMeta instead?

mattmoor: e.g. we have `kmeta.OwnerRefable`, which most things should implement already, which should maybe be incorporated into either `HasSpec` or `GenericCRD`.

webhook/user_info.go
vagababov: Yes, please don't change the label. It might actually be part of the spec.

apis/interfaces.go
aliok: +1

webhook/user_info.go
aliok: Ok, going to change it back to `/updater`

webhook/webhook.go
aliok: Thinking about it, how about getting the group name from the `AdmissionRequest`?https://github.com/aliok/pkg/blob/139162bdb52ec4e177616188f27d1254d15b9bfb/webhook/webhook.go#L493 

I will give this a try.

`req.Resource.Group` is there.

aliok:
Pushed new commits. Now the group name is read from the request that's made to admission hook.

Tested the changes locally with adapting the eventing module for the changes.

webhook/user_info.go
aliok: Reverted to `/updater` with commits later.

aliok:
@vagababov this is all green now.
There was some unrelated random e2e test failures which I guess is caused by the underlying infrastructure.

PR #480, opened 2019-06-22
Enrich the logs streamed in tests with useful information.

markusthoemmes:
- Level is very useful to spot errors at a glance.
- Time is very helpful to be able to order the statements and notice time gaps.
- Key helps identifying which ressource is being reconciled (could be a child resource).

/assign @mattmoor 

Let's discuss here if this is what we want/need for now. Just printing time without ordering things is not ideal, but maybe a good first step?

### DIFF
--- a/test/logstream/kubelogs.go
+++ b/test/logstream/kubelogs.go
@@ -48,7 +48,7 @@ type logger func(string, ...interface{})
 var _ streamer = (*kubelogs)(nil)
 
 // timeFormat defines a simple timestamp with millisecond granularity
-const timeFormat = "2006-01-02 15:04:05.000"
+const timeFormat = "15:04:05.000"
 
 func (k *kubelogs) init(t *testing.T) {
 	k.keys = make(map[string]logger)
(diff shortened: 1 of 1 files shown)


test/logstream/kubelogs.go
```diff
@@ -47,6 +47,9 @@ type logger func(string, ...interface{})
 
 var _ streamer = (*kubelogs)(nil)
 
+// timeFormat defines a simple timestamp with millisecond granularity
+const timeFormat = "2006-01-02 15:04:05.000"
```
vagababov: Do we want the date, in reality? The test run times are usually bound to the day, but this increases the line length and adds noise.
I like the way glog does formatting
[FirstCharOfLevel][DDMM][HHMMSS.SSS] -> [0201] 150405.000.
Here of course we can just keep `[FirstCharOfLevel]150405.000`.

markusthoemmes: Points taken, I adjusted the time format and reduced the log level to only the first char. I kept `:` in the time. If people feel strongly, I can remove those as well, but I feel like they add to readability a lot.

PR #489, opened 2019-06-25
Migrate pkg to use the knative.dev/pkg import path

mattmoor:
I have broken this into two commits:
1. The manual changes I made.
2. The scripted changes I made.

### This is a BREAKING change for all contributors

After this has merged, the following changes are needed:
1. Move or recreate your `knative/serving` directory to `${GOPATH}/src/knative.dev/pkg`
2. Run the following sed command pkg any outstanding changes:

```
sed -i 's@github.com/knative/pkg@knative.dev/pkg@g' $(find -name '*.go' | xargs grep github.com/knative/pkg | cut -d':' -f 1 | uniq)
```

In order to merge this, we are going to have to effectively take a lock on Prow for a few hours because we will need to change the repository configuration and land this change (which also won't pass CI until that change is made).

Related to: https://github.com/knative/serving/issues/971

/hold

Until after we cut 0.7 (even though this is cut for pkg, I don't want to risk the other cuts).


Harwayne:
> 1. Move or recreate your `knative/serving` directory to `${GOPATH}/src/knative.dev/pkg`

This should be `knative/pkg`, right?

jchesterpivotal:
/lgtm 

for same reasons as given in knative/serving#4521.

PR #502, opened 2019-07-02
Auto-update dependencies

mattmoor-sockpuppet:
Produced via:
  `dep ensure -update github.com/knative/test-infra`
/assign @mattmoor


evankanderson:
/lgtm
/approve

I'm still confused as to why we think these are doing anything.

PR #503, opened 2019-07-02
Add metrics to webhook package

anniefu:
Metrics added:

- "request_count": the number of AdmissionReview requests the webhook received and responded to (invalid requests are ignored)
- "request_latencies": the time in ms it took from receiving the request to sending the response

Other:

- Refactor reused metrics testing code into "metricstest" package that can be shared

See https://github.com/knative/serving/issues/1279

### DIFF
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -51,7 +51,7 @@ func CheckStatsNotReported(t *testing.T, names ...string) {
 
 // CheckCountData checks the view with a name matching string name to verify that the CountData stats
 // reported are tagged with the tags in wantTags and that wantValue matches reported count.
-func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantValue int) {
+func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantValue int64) {
 	t.Helper()
 	if row := checkExactlyOneRow(t, name); row != nil {
 		for _, got := range row.Tags {
@@ -64,8 +64,8 @@ func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantV
 		}
 
 		if s, ok := row.Data.(*view.CountData); !ok {
-			t.Error("Reporter expected a SumData type")
-		} else if s.Value != int64(wantValue) {
+			t.Error("Reporter expected a CountData type")
+		} else if s.Value != wantValue {
 			t.Errorf("For %s value = %v, want: %d", name, s.Value, wantValue)
 		}
 	}
@@ -74,7 +74,7 @@ func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantV
 // CheckDistributionData checks the view with a name matching string name to verify that the DistributionData stats reported
 // are tagged with the tags in wantTags and that expectedCount number of records were reported.
 // It also checks that expectedMin and expectedMax match the minimum and maximum reported values, respectively.
-func CheckDistributionData(t *testing.T, name string, wantTags map[string]string, expectedCount int, expectedMin float64, expectedMax float64) {
+func CheckDistributionData(t *testing.T, name string, wantTags map[string]string, expectedCount int64, expectedMin float64, expectedMax float64) {
 	t.Helper()
 	if row := checkExactlyOneRow(t, name); row != nil {
 		for _, got := range row.Tags {
@@ -89,7 +89,7 @@ func CheckDistributionData(t *testing.T, name string, wantTags map[string]string
 		if s, ok := row.Data.(*view.DistributionData); !ok {
 			t.Error("Reporter expected a DistributionData type")
 		} else {
-			if s.Count != int64(expectedCount) {
+			if s.Count != expectedCount {
 				t.Errorf("For metric %s: reporter count = %d, want = %d", name, s.Count, expectedCount)
 			}
 			if s.Min != expectedMin {
@@ -117,7 +117,7 @@ func CheckLastValueData(t *testing.T, name string, wantValue float64) {
 
 // CheckSumData checks the view with a name matching string name to verify that the SumData stats
 // reported are tagged with the tags in wantTags and that wantValue matches the reported sum.
-func CheckSumData(t *testing.T, name string, wantTags map[string]string, wantValue int) {
+func CheckSumData(t *testing.T, name string, wantTags map[string]string, wantValue float64) {
 	t.Helper()
 	if row := checkExactlyOneRow(t, name); row != nil {
 		for _, got := range row.Tags {
@@ -131,8 +131,8 @@ func CheckSumData(t *testing.T, name string, wantTags map[string]string, wantVal
 
 		if s, ok := row.Data.(*view.SumData); !ok {
 			t.Error("Reporter expected a SumData type")
-		} else if s.Value != float64(wantValue) {
-			t.Errorf("For %s value = %v, want: %d", name, s.Value, wantValue)
+		} else if s.Value != wantValue {
+			t.Errorf("For %s value = %v, want: %v", name, s.Value, wantValue)
 		}
 	}
 }
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -118,6 +118,21 @@ func CheckSumData(t *testing.T, name string, wantTags map[string]string, wantVal
 	}
 }
 
+// Unregister unregisters the metrics that were registered.
+// This is useful for testing since golang execute test iterations within the same process and
+// opencensus views maintain global state. At the beginning of each test, tests should
+// unregister for all metrics and then re-register for the same metrics. This effectively clears
+// out any existing data and avoids a panic due to re-registering a metric.
+//
+// In normal process shutdown, metrics do not need to be unregistered.
+func Unregister(names ...string) {
+	for _, n := range names {
+		if v := view.Find(n); v != nil {
+			view.Unregister(v)
+		}
+	}
+}
+
 func checkExactlyOneRow(t *testing.T, name string, wantTags map[string]string) *view.Row {
 	t.Helper()
 	d, err := view.RetrieveData(name)
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -31,7 +31,7 @@ func CheckStatsReported(t *testing.T, names ...string) {
 			t.Errorf("Reporter.Report() error = %v, wantErr %v", err, false)
 		}
 		if len(d) < 1 {
-			t.Errorf("No data reported when data was expected. len(d)=%v", len(d))
+			t.Error("No data reported when data was expected, view data is empty.")
 		}
 	}
 }
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -1,0 +1,149 @@
+/*
+Copyright 2018 The Knative Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metricstest
+
+import (
+	"testing"
+
+	"go.opencensus.io/stats/view"
+)
+
+// CheckStatsReported checks that there is a view registered with the given name for each string in names,
+// and that each view has at least one record.
+func CheckStatsReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		d, err := view.RetrieveData(name)
+		if err != nil {
+			t.Errorf("Reporter.Report() error = %v, wantErr %v", err, false)
+		}
+		if len(d) < 1 {
+			t.Errorf("No data reported when data was expected. len(d)=%v", len(d))
+		}
+	}
+}
+
+// CheckStatsNotReported checks that there are no records for any views that a name matching a string in names.
+// Names that do not match registered views are considered not reported.
+func CheckStatsNotReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		if d, err := view.RetrieveData(name); err != nil {
+			if len(d) > 0 {
+				t.Errorf("Unexpected data reported when no data was expected. Reporter len(d) = %d", len(d))
+			}
+		}
+	}
+}
+
+// CheckCountData checks the view with a name matching string name to verify that the CountData stats
+// reported are tagged with the tags in wantTags and that wantValue matches reported count.
+func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantValue int) {
+	t.Helper()
+	if row := checkExactlyOneRow(t, name); row != nil {
+		for _, got := range row.Tags {
+			n := got.Key.Name()
+			if want, ok := wantTags[n]; !ok {
+				t.Errorf("Reporter got an extra tag %v: %v", n, got.Value)
+			} else if got.Value != want {
+				t.Errorf("Reporter expected a different tag value for key: %s, got: %s, want: %s", n, got.Value, want)
+			}
+		}
+
+		if s, ok := row.Data.(*view.CountData); !ok {
+			t.Error("Reporter expected a SumData type")
+		} else if s.Value != int64(wantValue) {
+			t.Errorf("For %s value = %v, want: %d", name, s.Value, wantValue)
+		}
+	}
+}
+
+// CheckDistributionData checks the view with a name matching string name to verify that the DistributionData stats reported
+// are tagged with the tags in wantTags and that expectedCount number of records were reported.
+// It also checks that expectedMin and expectedMax match the minimum and maximum reported values, respectively.
+func CheckDistributionData(t *testing.T, name string, wantTags map[string]string, expectedCount int, expectedMin float64, expectedMax float64) {
+	t.Helper()
+	if row := checkExactlyOneRow(t, name); row != nil {
+		for _, got := range row.Tags {
+			n := got.Key.Name()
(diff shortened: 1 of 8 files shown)


metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,149 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

webhook/stats_reporter.go
```diff
@@ -0,0 +1,150 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

webhook/stats_reporter_test.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

webhook/stats_reporter.go
```diff
@@ -0,0 +1,150 @@
...
+	requestCountM = stats.Int64(
+		requestCountName,
+		"The number of requests that are routed to webhook",
+		stats.UnitDimensionless)
+	responseTimeInMsecM = stats.Float64(
+		"request_latencies",
+		"The response time in milliseconds",
+		stats.UnitMilliseconds)
+
+	defaultLatencyDistribution = view.Distribution(0, 5, 10, 20, 40, 60, 80, 100, 150, 200, 250, 300, 350, 400, 450, 500, 600, 700, 800, 900, 1000, 2000, 5000, 10000, 20000, 50000, 100000)
```
vagababov: I think you should remove bucket "0".

webhook/stats_reporter.go
```diff
@@ -0,0 +1,150 @@
...
+		kindVersionKey,
+		kindKindKey,
+		resourceGroupKey,
+		resourceVersionKey,
+		resourceResourceKey,
+		resourceNamespaceKey,
+		resourceNameKey,
+		admissionAllowedKey}
+
+	err := view.Register(
```
vagababov: join the statements into one:
```
if err := f(...
...
); err != nil {
}
```

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+			t.Error("Reporter.Report() expected a LastValueData type")
+		} else if s.Value != wantValue {
+			t.Errorf("Reporter.Report() expected %v got %v. metric: %v", s.Value, wantValue, name)
+		}
+	}
+}
+
+// CheckSumData checks the view with a name matching string name to verify that the SumData stats
+// reported are tagged with the tags in wantTags and that wantValue matches the reported sum.
+func CheckSumData(t *testing.T, name string, wantTags map[string]string, wantValue int) {
```
vagababov: same here, just pass in float64 from the getgo.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+			if len(d) > 0 {
+				t.Errorf("Unexpected data reported when no data was expected. Reporter len(d) = %d", len(d))
+			}
+		}
+	}
+}
+
+// CheckCountData checks the view with a name matching string name to verify that the CountData stats
+// reported are tagged with the tags in wantTags and that wantValue matches reported count.
+func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantValue int) {
```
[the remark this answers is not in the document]
anniefu: Most of the calling sites will have an int, so either the caller casts it or the helper function casts it. CheckDistributionData (which I cut from a different file) already set the precedent of the helper function casting it.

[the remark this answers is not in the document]
vagababov: For example `metricstest.CheckCountData(t, requestCountName, expectedTags, 2)` will work just fine, I think if you change param to `int64`. Same with `CheckDistributionData`.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+// and that each view has at least one record.
+func CheckStatsReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		d, err := view.RetrieveData(name)
+		if err != nil {
+			t.Errorf("Reporter.Report() error = %v, wantErr %v", err, false)
+		}
+		if len(d) < 1 {
+			t.Errorf("No data reported when data was expected. len(d)=%v", len(d))
```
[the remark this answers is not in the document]
anniefu: I would rather leave it in because on failure it gives more info on what "No data reported" means. There's little harm in a little over-logging for test failures :)

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+
+import (
+	"testing"
+
+	"go.opencensus.io/stats/view"
+)
+
+// CheckStatsReported checks that there is a view registered with the given name for each string in names,
+// and that each view has at least one record.
+func CheckStatsReported(t *testing.T, names ...string) {
```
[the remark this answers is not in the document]
anniefu: For tests, it does save a couple of lines of code re-writing for each call to just error. 

In this case, it would be wonky to have this function's behavior differ from the rest of the helpers, so I'll leave it be.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+}
+
+// CheckStatsNotReported checks that there are no records for any views that a name matching a string in names.
+// Names that do not match registered views are considered not reported.
+func CheckStatsNotReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		if d, err := view.RetrieveData(name); err != nil {
```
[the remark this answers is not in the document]
anniefu: I think so, erroring here implies that stat does not exist and hence was not reported, so the check is passed at that point..

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+			n := got.Key.Name()
+			if want, ok := wantTags[n]; !ok {
+				t.Errorf("Reporter got an extra tag %v: %v", n, got.Value)
+			} else if got.Value != want {
+				t.Errorf("Reporter expected a different tag value for key: %s, got: %s, want: %s", n, got.Value, want)
+			}
+		}
+
+		if s, ok := row.Data.(*view.CountData); !ok {
+			t.Error("Reporter expected a SumData type")
```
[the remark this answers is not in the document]
anniefu: Good catch.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+			n := got.Key.Name()
+			if want, ok := wantTags[n]; !ok {
+				t.Errorf("Reporter got an extra tag %v: %v", n, got.Value)
+			} else if got.Value != want {
+				t.Errorf("Reporter expected a different tag value for key: %s, got: %s, want: %s", n, got.Value, want)
+			}
+		}
+
+		if s, ok := row.Data.(*view.DistributionData); !ok {
+			t.Error("Reporter expected a DistributionData type")
```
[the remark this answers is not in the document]
anniefu: Yeah, I guess a lot of the errors in these helpers should have been fatal. I'll do a pass through and change some of them.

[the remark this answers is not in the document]
anniefu: Actually, on second glance. The Error vs Fatal won't make a difference in what gets executed and leaving them as Errors allows further checks to continue if they're added in the future. So I'm going to keep it as an Error in the interest of getting as many test failures out in one test run as possible.

metrics/metricstest/metricstest.go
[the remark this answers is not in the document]
vagababov: Perhaps the text should be different then? :) Printing always 0 does not make sense. Even if you want to print it: just put it in the error text, no need to use formatting for that.

metrics/metricstest/metricstest.go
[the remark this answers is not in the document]
vagababov: It does for the code readability. Your whole `else{...}` indentation goes away.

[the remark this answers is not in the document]
anniefu: That's true, but using fatal also adds the restriction that the DistributionData check must always be the last check in the function if the show-as-many-errors-as-possible functionality is to be preserved.

For example, I would not be able to arbitrarily move the tag-check for-loop down.

I think it comes down to personal preference, but I usually err on letting the original author's format stay in that case (these functions are mostly transplants from previous tests).

anniefu:
@vagababov thanks for taking the time to review!

metrics/metricstest/metricstest.go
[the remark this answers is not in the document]
vagababov: Eh, unit test, fast iterations, etc :-)
Up to you. I am not bent on this.

webhook/stats_reporter_test.go
```diff
@@ -0,0 +1,69 @@
...
+	expectedTags := createExpectedMetricTags(admReq, resp)
+
+	r.ReportRequest(admReq, resp, time.Duration(shortTime)*time.Millisecond)
+	r.ReportRequest(admReq, resp, time.Duration(longTime)*time.Millisecond)
+
+	metricstest.CheckCountData(t, requestCountName, expectedTags, 2)
+	metricstest.CheckDistributionData(t, requestLatenciesName, expectedTags, 2, shortTime, longTime)
+}
+
+func createExpectedMetricTags(req *admissionv1beta1.AdmissionRequest, resp *admissionv1beta1.AdmissionResponse) map[string]string {
```
vagababov: Why not just
`var = map[]`... ?

anniefu: At the time, I expected to write more than one test, but it did end up being unnecessary. I'll just go with the one variable instead of the function.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+}
+
+// CheckStatsNotReported checks that there are no records for any views that a name matching a string in names.
+// Names that do not match registered views are considered not reported.
+func CheckStatsNotReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		if d, err := view.RetrieveData(name); err != nil {
```
evankanderson: Is this check backwards? We're checking that `err != nil`, but then proceeding to use `d` when `err` was **non-`nil`**.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+// CheckStatsReported checks that there is a view registered with the given name for each string in names,
+// and that each view has at least one record.
+func CheckStatsReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		d, err := view.RetrieveData(name)
+		if err != nil {
+			t.Errorf("Reporter.Report() error = %v, wantErr %v", err, false)
```
evankanderson: Why `wantErr false` in this output?

evankanderson: Also, should this and the next error include `name`?

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// CheckStatsNotReported checks that there are no records for any views that a name matching a string in names.
+// Names that do not match registered views are considered not reported.
+func CheckStatsNotReported(t *testing.T, names ...string) {
+	t.Helper()
+	for _, name := range names {
+		if d, err := view.RetrieveData(name); err != nil {
+			if len(d) > 0 {
+				t.Errorf("Unexpected data reported when no data was expected. Reporter len(d) = %d", len(d))
```
evankanderson: Do you want to include `name` in the `Errorf`?

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+	t.Helper()
+	if row := checkExactlyOneRow(t, name); row != nil {
+		for _, got := range row.Tags {
+			n := got.Key.Name()
+			if want, ok := wantTags[n]; !ok {
+				t.Errorf("Reporter got an extra tag %v: %v", n, got.Value)
+			} else if got.Value != want {
+				t.Errorf("Reporter expected a different tag value for key: %s, got: %s, want: %s", n, got.Value, want)
+			}
+		}
```
evankanderson: You might get a clearer output if you convert row to a map[string]string, and then use [`cmp.Diff`](https://godoc.org/github.com/google/go-cmp/cmp#Diff) with an [`cmpopts.IgnoreMapEntries`](https://godoc.org/github.com/google/go-cmp/cmp/cmpopts#IgnoreMapEntries) argument.

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+
+		if s, ok := row.Data.(*view.CountData); !ok {
+			t.Error("Reporter expected a CountData type")
```
evankanderson: Include `name` here?

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+
+// CheckCountData checks the view with a name matching string name to verify that the CountData stats
+// reported are tagged with the tags in wantTags and that wantValue matches reported count.
+func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantValue int64) {
+	t.Helper()
+	if row := checkExactlyOneRow(t, name); row != nil {
+		for _, got := range row.Tags {
+			n := got.Key.Name()
+			if want, ok := wantTags[n]; !ok {
+				t.Errorf("Reporter got an extra tag %v: %v", n, got.Value)
```
evankanderson: Include `name` for these?

metrics/metricstest/metricstest.go
```diff
@@ -0,0 +1,151 @@
...
+			t.Errorf("For %s value = %v, want: %v", name, s.Value, wantValue)
+		}
+	}
+}
+
+func checkExactlyOneRow(t *testing.T, name string) *view.Row {
+	t.Helper()
+	d, err := view.RetrieveData(name)
+	if err != nil {
+		t.Errorf("Reporter.Report() error = %v, wantErr %v", err, false)
```
evankanderson: Include `name` here?

webhook/stats_reporter.go
```diff
@@ -0,0 +1,149 @@
...
+	requestCountM = stats.Int64(
+		requestCountName,
+		"The number of requests that are routed to webhook",
+		stats.UnitDimensionless)
+	responseTimeInMsecM = stats.Float64(
+		"request_latencies",
+		"The response time in milliseconds",
+		stats.UnitMilliseconds)
+
+	defaultLatencyDistribution = view.Distribution(5, 10, 20, 40, 60, 80, 100, 150, 200, 250, 300, 350, 400, 450, 500, 600, 700, 800, 900, 1000, 2000, 5000, 10000, 20000, 50000, 100000)
```
evankanderson: I think @grantr created a library for building distributions with a nice bucketing. Not sure where it's living right now, though.

metrics/metricstest/metricstest.go
anniefu: Its confusing to look at, but it should be ok.

We want this function to pass in two cases:
1. err == nil (implies no component registered a stat with "name", so it's not possible that it's reported)
2. if err != nil && len(d) == 0 (there is a stat registered with "name", but nothing actually recorded a measurement for that stat)

metrics/metricstest/metricstest.go
anniefu: Good suggestion, I'll add name where applicable.

metrics/metricstest/metricstest.go
anniefu: Mmmmhh yeah it's not really sensible. I think I copy-pasted the error message from somewhere. I'll remove it and add something about the stat not existing.

metrics/metricstest/metricstest.go
anniefu: Interesting, I will give it a try.

anniefu: I'm going to leave it as-is because the error messages here are clear enough. Using cmp.Diff with cmpopts.IgnoreMapEntries is a little bit more complex code-wise and more inefficient algorithmically.

webhook/stats_reporter.go
anniefu: Mmmh, I don't see it anywhere in this repo. I've reached out to @grantr offline. In the meanwhile, I'll stick with this (which matches the activator's latency distribution).

anniefu: Grant got back to me, I've copied the bucketing function into the pkg/metrics package so that other components can use it in the future.

grantr: @evankanderson actually wrote the code in https://github.com/knative/eventing/pull/1177#discussion_r292208873, I just copied it :smile:

metrics/metricstest/metricstest.go
evankanderson: I still think this is backwards. Let me rewrite, with comments:

```golang
d, err := view.RetrieveData(name)
// At this point, we have two cases:
// 1. d has data, and err is nil
// 2. d is nil, and err is non-nil
if err != nil { // We are in case 2
  if len(d) >0 {
    ....
  }
}

**Comment by n3wscott** [comment]
This is causing NPE in eventing.

File: webhook/webhook.go [hunk 22]
```diff
@@ -452,6 +454,9 @@ func (ac *AdmissionController) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		http.Error(w, fmt.Sprintf("could encode response: %v", err), http.StatusInternalServerError)
 		return
 	}
+
+	// Only report valid requests
+	ac.StatsReporter.ReportRequest(review.Request, response.Response, time.Since(ttStart))
```
n3wscott: This causes a NPE. StatsReporter is new and nil in all repos.

anniefu: Ah yes, that makes sense. I was pulling this in and making the changes in the serving repo to initialize this correctly, but I forgot this was a common package :/

I'll push a fix out in a couple of hours or roll it back.

Issue #441, opened 2018-07-19
Log output from logging library should be optional


bobcatfish:
/area API
/area test-and-release

/kind cleanup

## Expected Behavior
In general libraries should make logging optional or not log at all, so that the user of the library can control when the libraries log.

Specifically, [the logging library in serving](https://github.com/knative/serving/tree/master/pkg/logging) should allow users to control whether or not it outputs logs, or not log at all.

## Actual Behavior
If a new logger is instantiated with knative/serving/blob/master/pkg/logging/config.go, logs are output at `Info` and `Error` levels. This can be seen when running the e2e tests (see below), the first line of output is:

```
info	logging/config.go:90	Successfully created the logger.	{"knative.dev/jsonconfig": "{\n\t  \"level\": \"info\",\n\t  \"encoding\": \"console\",\n\t  \"outputPaths\": [\"stdout\"],\n\t  \"errorOutputPaths\": [\"stderr\"],\n\t  \"encoderConfig\": {\n\t    \"messageKey\": \"message\",\n\t\t\t\"levelKey\": \"level\",\n\t\t\t\"nameKey\": \"logger\",\n\t\t\t\"callerKey\": \"caller\",\n\t\t\t\"messageKey\": \"msg\",\n      \"stacktraceKey\": \"stacktrace\",\n      \"lineEnding\": \"\",\n      \"levelEncoder\": \"\",\n      \"timeEncoder\": \"\",\n      \"durationEncoder\": \"\",\n      \"callerEncoder\": \"\"\n\t  }\n\t}"}
```

This should be optional so the tests can suppress it. (Bonus, if we DO want output like this, it could be human readable.)

## Steps to Reproduce the Problem

You can reproduce this by running the tests with `go test -v -tags=e2e -count=1 ./test/conformance`:

```bash
➜  serving git:(master) ✗ go test -v -tags=e2e -count=1 ./test/conformance
[...]

PR #506, opened 2019-07-03
Reducing the log level for github commit to info

### LINKED ISSUE #505: GitHub commit id missing from kodata should not be a warning
## Expected Behavior
No spurious warning messages are generated when instantiating the logger:
```
I0626 23:56:51.574] [gopath-test] [step-test-build] {"level":"warn","ts":1561593313.3031838,"logger":"fallback-logger","caller":"logging/config.go:69","msg":"Fetch GitHub commit ID from kodata failed: \"KO_DATA_PATH\" does not exist or is empty"}
```

This happens all the time when starting application that use the [logger from knative/pkg](https://github.com/knative/pkg/blob/222dd25986d91c12d54f6ac7758251f005053b8f/logging/config.go#L69).

The end result is that a lot of warnings end-up in [CI logs](As far as I understand, setting the commit ID in kodata is not mandatory, so this should not be logged at "warn" level, which generates a lot of noise when troubleshooting actual issues.), which generates noise.

The log level should be tuned down to info or debug.

## Actual Behavior

The logger emits a warning if it cannot find 

1. Use the logger in an application
2. Do not include a github commit id in kodata
3. Build the app (with ko)
4.  Start the app, check the logs

afrittoli:
The logger by default attempts to fetch the commit id from kodata.
When that fails log a message with info rather than warning, to
avoid creating spurious warning message when starting applications.

Fixes #505

Signed-off-by: Andrea Frittoli <andrea.frittoli@uk.ibm.com>

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -66,7 +66,7 @@ func enrichLoggerWithCommitID(logger *zap.SugaredLogger) *zap.SugaredLogger {
 		return logger.With(zap.String(logkey.GitHubCommitID, commmitID))
 	}
 
-	logger.Warnf("Fetch GitHub commit ID from kodata failed: %v", err)
+	logger.Infof("Fetch GitHub commit ID from kodata failed: %v", err)
 	return logger
 }
 


bobcatfish:
I think this is a good first step! That should not be a warning for sure.

In the long run I hope someone will address https://github.com/knative/pkg/issues/441 - libs should opt into logging instead of always doing it and not letting the user control when it happens.

PR #518, opened 2019-07-09
Add check for nil StatsReporter in webhook package

### LINKED ISSUE #517: Stats Reporter causes a NPE. StatsReporter is new and nil in all repos.
This causes a NPE. StatsReporter is new and nil in all repos.

_Originally posted by @n3wscott in https://github.com/knative/pkg/pull/503/files#diff-1034b06755820a5860e30a206b72a333R459_

anniefu:
Fixes #517 

Also add constructors for consumers to use.

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -150,7 +150,7 @@ func NewAdmissionController(
 		panic(err)
 	}
 
-	return NewAdmissionControllerWithStats(client, opts, handlers, logger, reporter, ctx, disallowUnknownFields)
+	return NewAdmissionControllerWithStats(client, opts, handlers, logger, &reporter, ctx, disallowUnknownFields)
 }
 
 // NewAdmissionControllerWithStats constructs an AdmissionController where the caller provides the StatsReporter
@@ -159,7 +159,7 @@ func NewAdmissionControllerWithStats(
 	opts ControllerOptions,
 	handlers map[schema.GroupVersionKind]GenericCRD,
 	logger *zap.SugaredLogger,
-	reporter StatsReporter,
+	reporter *StatsReporter,
 	ctx func(context.Context) context.Context,
 	disallowUnknownFields bool) *AdmissionController {
 	return &AdmissionController{
@@ -167,7 +167,7 @@ func NewAdmissionControllerWithStats(
 		Options:               opts,
 		Handlers:              handlers,
 		Logger:                logger,
-		StatsReporter:         reporter,
+		StatsReporter:         *reporter,
 		WithContext:           ctx,
 		DisallowUnknownFields: disallowUnknownFields,
 	}
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -136,6 +136,43 @@ type GenericCRD interface {
 	runtime.Object
 }
 
+// NewAdmissionController constructs an AdmissionController
+func NewAdmissionController(
+	client kubernetes.Interface,
+	opts ControllerOptions,
+	handlers map[schema.GroupVersionKind]GenericCRD,
+	logger *zap.SugaredLogger,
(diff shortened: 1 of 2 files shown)


anniefu:
I validated that the fix passes the e2e tests for knative/serving master branch after updating the knative.dev/pkg version to my personal fork with the fix.

webhook/webhook.go
```diff
@@ -136,6 +136,43 @@ type GenericCRD interface {
...
+
+	reporter, err := NewStatsReporter()
+	if err != nil {
+		panic(err)
+	}
+
+	return NewAdmissionControllerWithStats(client, opts, handlers, logger, &reporter, ctx, disallowUnknownFields)
+}
+
+// NewAdmissionControllerWithStats constructs an AdmissionController where the caller provides the StatsReporter
```
vagababov: Is this used anywhere outside this file?

webhook/webhook.go
```diff
@@ -136,6 +136,43 @@ type GenericCRD interface {
...
+	client kubernetes.Interface,
+	opts ControllerOptions,
+	handlers map[schema.GroupVersionKind]GenericCRD,
+	logger *zap.SugaredLogger,
+	ctx func(context.Context) context.Context,
+	disallowUnknownFields bool) *AdmissionController {
+
+	reporter, err := NewStatsReporter()
+	if err != nil {
+		panic(err)
```
vagababov: Don't Panic (c). Can we log fatal, perhaps?

webhook/webhook.go
anniefu: In the tests now and consuming packages can choose to use these instead of initializing the struct themselves.

I plan on having knative/serving use this constructor when I pull in the changes.

vagababov: Sounds good.

webhook/webhook.go
evankanderson: Or return `*AdmissionController, error`

vagababov:
> /approve
> 
> /hold
> In case you want to change the return type of `NewAdmissionController`

I didn't suggest that, since it's a NBC change and requires changes to the downstream consumers, I presume.

anniefu:
> > /approve
> > /hold
> > In case you want to change the return type of `NewAdmissionController`
> 
> I didn't suggest that, since it's a NBC change and requires changes to the downstream consumers, I presume.

What does NBC stand for? The constructor is introduced in this PR, so there are no downstream consumers that need to change.

vagababov:
not backwards compatible :-)

but if it's new then it doesn't matter -- returning error is a preferred method.

webhook/webhook.go
```diff
@@ -136,6 +136,43 @@ type GenericCRD interface {
...
+
+// NewAdmissionControllerWithStats constructs an AdmissionController where the caller provides the StatsReporter
+func NewAdmissionControllerWithStats(
+	client kubernetes.Interface,
+	opts ControllerOptions,
+	handlers map[schema.GroupVersionKind]GenericCRD,
+	logger *zap.SugaredLogger,
+	reporter *StatsReporter,
+	ctx func(context.Context) context.Context,
+	disallowUnknownFields bool) *AdmissionController {
```
evankanderson: Looking at this, I'm realizing that you're repeating the "optional parameters" pattern here: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

You might consider putting `*StatsReporter` in ControllerOptions, and then initializing it if nil in `NewAdmissionController`. This would mean that you only need to export one function, rather than `Foo` and `FooWithStats`.

anniefu: It's difficult to go as far as that article with the functional options because of existing usage, but I went with putting StatsReporter in ControllerOptions and consolidating the constructors.

PR #527, opened 2019-07-14
Expose the RegistrationDelay for webhook deployment

zxDiscovery:
Expose the RegistrationDelay so that user can set this env `REG_DELAY_TIME` in [500-webhook.yaml](https://github.com/knative/eventing/blob/master/config/500-webhook.yaml).
Workaround for https://github.com/knative/pkg/issues/417

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -25,7 +25,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


zxDiscovery:
This change is only for knative-eventing,  it shouldn't in the `pkg`. 
So close this issue.

PR #528, opened 2019-07-14
Improve kmp.ShortDiff when one object is invalid

fatkun:
Fixed knative/serving#4655

* Both are invalid throw an error

* If x is invalid and y is valid, add '+' prefix and no '-'

* If x is valid and y is invalid, add '-' prefix and not '+'

### DIFF
--- a/kmp/reporters.go
+++ b/kmp/reporters.go
@@ -111,12 +111,24 @@ func (r *ShortDiffReporter) Report(rs cmp.Result) {
 	t := cur.Type()
 	var diff string
 	// Prefix struct values with the types to add clarity in output
-	if !vx.IsValid() || !vy.IsValid() {
-		r.err = fmt.Errorf("Unable to diff %+v and %+v on path %#v", vx, vy, r.path)
+	if !vx.IsValid() && !vy.IsValid() {
+		r.err = fmt.Errorf("Both are invalid on path %#v", r.path)
 	} else if t.Kind() == reflect.Struct {
-		diff = fmt.Sprintf("%#v:\n\t-: %+v: \"%+v\"\n\t+: %+v: \"%+v\"\n", r.path, t, vx, t, vy)
+		diff = fmt.Sprintf("%#v:\n", r.path)
+		if vx.IsValid() {
+			diff += fmt.Sprintf("\t-: %+v: \"%+v\"\n", t, vx)
+		}
+		if vy.IsValid() {
+			diff += fmt.Sprintf("\t+: %+v: \"%+v\"\n", t, vy)
+		}
 	} else {
-		diff = fmt.Sprintf("%#v:\n\t-: \"%+v\"\n\t+: \"%+v\"\n", r.path, vx, vy)
+		diff = fmt.Sprintf("%#v:\n", r.path)
+		if vx.IsValid() {
+			diff += fmt.Sprintf("\t-: \"%+v\"\n", vx)
+		}
+		if vy.IsValid() {
+			diff += fmt.Sprintf("\t+: \"%+v\"\n", vy)
+		}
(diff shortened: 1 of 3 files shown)


fatkun:
@dgerd Sorry for the delay, do you have time to review the code?

dgerd:
No worries. Thanks for the PR!! I will take a look Monday morning.

kmp/reporters.go
```diff
@@ -111,12 +111,24 @@ func (r *ShortDiffReporter) Report(rs cmp.Result) {
 	t := cur.Type()
 	var diff string
 	// Prefix struct values with the types to add clarity in output
-	if !vx.IsValid() || !vy.IsValid() {
-		r.err = fmt.Errorf("Unable to diff %+v and %+v on path %#v", vx, vy, r.path)
+	if !vx.IsValid() && !vy.IsValid() {
+		r.err = fmt.Errorf("Both are invalid on path %#v", r.path)
```
dgerd: I would keep the original error message just for clarity.

kmp/reporters.go
```diff
@@ -111,12 +111,24 @@ func (r *ShortDiffReporter) Report(rs cmp.Result) {
...
 	} else if t.Kind() == reflect.Struct {
```
dgerd: It probably makes sense now to reverse the if() conditions to reduce the number of duplicated lines.

i.e.
```
diff = fmt.Sprintf("%#v:\n", r.path)
if vx.IsValid() {
    if t.Kind() == reflect.Struct {
          diff += ...
    } else {
          diff += ...
   }
}

if vy.IsValid() {
    if t.Kind() == reflect.Struct {
          diff += ...
    } else {
          diff += ...
   }
}
```

kmp/reporters_test.go
```diff
@@ -217,7 +217,9 @@ func TestImmutableReporter(t *testing.T) {
 		x: testStruct{
 			StringField: "foo",
 		},
-		expectErr: true,
+		want: `root:
+	-: "{A: StringField:foo IntField:0 StructField:{ChildString: ChildInt:0} Omit: Ignore: Dash: MultiComma:}"
+`,
 	}, {
 		name: "Single character field name",
```
dgerd: Can you add a test case for a map with a missing key since that is where you are seeing the bug?

kmp/reporters.go
fatkun: done

kmp/reporters.go
fatkun: I extract this block into method now, please review again.
```
    if t.Kind() == reflect.Struct {
          diff += ...
    } else {
          diff += ...
   }
```

kmp/reporters_test.go
fatkun: done

kmp/reporters.go
```diff
@@ -111,16 +111,30 @@ func (r *ShortDiffReporter) Report(rs cmp.Result) {
...
+			diff += r.diffString("+", t, vy)
+		}
 	}
 	r.diffs = append(r.diffs, diff)
 }
 
+func (r *ShortDiffReporter) diffString(diffType string, t reflect.Type, v reflect.Value) string {
+	var diff string
+	if t.Kind() == reflect.Struct {
+		diff = fmt.Sprintf("\t%s: %+v: \"%+v\"\n", diffType, t, v)
```
dgerd: nit: this can just be return on both this line and 133.

fatkun: done

dgerd:
/assign @mattmoor 
for approval

dgerd:
@fatkun Can you remove the "[WIP]"?

PR #538, opened 2019-07-21
Implement ClearCondition for ConditionManager

johnugeorge:
This implements ClearCondition to remove a non terminal condition that matches
the condition type. 

Related: https://github.com/kubeflow/kfserving/pull/253

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -60,6 +60,9 @@ type ConditionManager interface {
 	// If there is an update, Conditions are stored back sorted.
 	SetCondition(new Condition)
 
+	// RemoveCondition removes the condition that matches the ConditionType
+	RemoveCondition(t ConditionType)
+
 	// MarkTrue sets the status of t to true, and then marks the happy condition to
 	// true if all dependents are true.
 	MarkTrue(t ConditionType)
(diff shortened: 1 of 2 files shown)


apis/condition_set.go
```diff
@@ -207,6 +210,78 @@ func (r conditionsImpl) severity(t ConditionType) ConditionSeverity {
 	return ConditionSeverityInfo
 }
 
+// RemoveCondition removes the condition that matches the ConditionType
+func (r conditionsImpl) RemoveCondition(t ConditionType) {
```
n3wscott: If you are adding `RemoveCondition` I would expect you would also add `AddCondition` to get back to the orignal setup.

apis/condition_set.go
```diff
@@ -207,6 +210,78 @@ func (r conditionsImpl) severity(t ConditionType) ConditionSeverity {
...
+		}
+	}
+
+	// Sorted for convenience of the consumer, i.e. kubectl.
+	sort.Slice(conditions, func(i, j int) bool { return conditions[i].Type < conditions[j].Type })
+	r.accessor.SetConditions(conditions)
+
+	happy := r.GetCondition(r.happy)
+	if r.isTerminal(t) {
+		if cond.IsFalse() && happy.IsFalse() {
```
n3wscott: I think all of this logic can be reduced down to resetting all of the dependents to their current state with the mark methods without regard to the happy condition with testing that there are dependents. If you removed the only dependent, I would assume happy goes from false to unknown, or stays the same? 

So: ignore the happy condition if there are deps, for each dep, set it again using the `Mark` methods. This will cause `happy` updates to be propagated.

apis/condition_set_impl_test.go
```diff
@@ -840,3 +840,94 @@ func TestTerminalInitialization(t *testing.T) {
...
+	if got == nil {
+		t.Errorf("GetCondition = nil, wanted %s", want)
+	}
+	if status := got.Status; status != want {
+		t.Errorf("GetCondition = %s, wanted %s", status, want)
+	}
+
+}
+func TestRemoveTerminalConditions(t *testing.T) {
+	set := NewLivingConditionSet("Foo", "Bar")
```
n3wscott: can you also make a test that just has a condition set of "Foo" and then remove "Foo". So then happy has no dependents....

johnugeorge: @n3wscott what is the expected result here? When all dependent conditions are removed, does happy condition also need to be removed(like the initial case where there are no conditions)?

apis/condition_set.go
johnugeorge: Won't AddCondition be a confusing one? All `Mark` functions add condition if they are not added earlier. RemoveCondition was added as this functionality was missing earlier.

apis/condition_set_impl_test.go
n3wscott: A resource needs to have a happy condition. Ready in most cases. There are resources that do not have dependents, and using this method you could simulate what k8s resources do and when you are fully ready, you delete all deps and Ready remains true, and then when a depdent changes, you add it back to the deps list and set that dep to the new state (unknown or false). This would result in ready unknown or false and just the unknown or false dep being displayed in status.

apis/condition_set.go
```diff
@@ -221,7 +262,7 @@ func (r conditionsImpl) MarkTrue(t ConditionType) {
 	for _, cond := range r.dependents {
 		c := r.GetCondition(cond)
 		// Failed or Unknown conditions trump true conditions
-		if !c.IsTrue() {
+		if c != nil && !c.IsTrue() {
```
n3wscott: this change is likely an api breaking change for knative. If the condition is nil, it means it is unknown. The expected behavior is to propagate unknown to Ready (happy)

johnugeorge: done

apis/condition_set_impl_test.go
evankanderson: FWIW, feedback from client UX indicated that in the typical case, customers expect to see a consistent set of Conditions, even if they go "green" very quickly, rather than having some Conditions disappear after they are ready.

That's not to say that this functionality shouldn't exist, just a warning on UX.

n3wscott: Point taken, this feature is for FYI conditions that do not apply to modes that the resource can transition into. And in that case, the condition as it is set does not make sense for the updated mode. And this give hooks to allow the deletion of that condition so the user does not see it and get confused.

PR #539, opened 2019-07-22
Add validation for creator and lastModifier annotation

savitaashture:
Fixes https://github.com/knative/serving/issues/4648

`ValidateCreatorAndModifier` function will be used by all the CRD's which will be created by user like
Service, Configuration, Route to validate `/creator` and `/lastModifier` annotation

/assign @markusthoemmes @dgerd

### DIFF
--- a/apis/metadata_validation.go
+++ b/apis/metadata_validation.go
@@ -19,8 +19,19 @@ package apis
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// CreatorAnnotationSuffix is the suffix of the annotation key to describe
+	// the user that created the resource.
+	CreatorAnnotationSuffix = "/creator"
+
+	// UpdaterAnnotationSuffix is the suffix of the annotation key to describe
+	// the user who last modified the resource.
+	UpdaterAnnotationSuffix = "/lastModifier"
 )
 
 // ValidateObjectMetadata validates that `metadata` stanza of the
@@ -59,4 +70,19 @@ func ValidateObjectMetadata(meta metav1.Object) *FieldError {
 	}
 
 	return nil
+}
+
+// ValidateCreatorAndModifier validates `metadata.annotation`
+func ValidateCreatorAndModifier(oldSpec, newSpec interface{}, oldAnnotation, newAnnotation map[string]string, groupName string) *FieldError {
+	var errs *FieldError
+	if oldAnnotation[groupName+CreatorAnnotationSuffix] != newAnnotation[groupName+CreatorAnnotationSuffix] {
+		errs = errs.Also(&FieldError{
+			Message: "annotation value is immutable",
+			Paths:   []string{groupName + CreatorAnnotationSuffix},
+		})
+	}
+	if equality.Semantic.DeepEqual(oldSpec, newSpec) && oldAnnotation[groupName+UpdaterAnnotationSuffix] != newAnnotation[groupName+UpdaterAnnotationSuffix] {
+		errs = errs.Also(ErrInvalidValue(newAnnotation[groupName+UpdaterAnnotationSuffix], groupName+UpdaterAnnotationSuffix))
+	}
+	return errs
 }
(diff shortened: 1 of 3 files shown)


apis/metadata_validation.go
```diff
@@ -60,3 +71,18 @@ func ValidateObjectMetadata(meta metav1.Object) *FieldError {
...
+// ValidateCreatorAndModifier validates `metadata.annotation`
+func ValidateCreatorAndModifier(oldSpec, newSpec interface{}, oldAnnotation, newAnnotation map[string]string) *FieldError {
+	var errs *FieldError
+	if oldAnnotation[CreatorAnnotation] != newAnnotation[CreatorAnnotation] {
+		errs = errs.Also(&FieldError{
+			Message: "annotation value is immutable",
+			Paths:   []string{CreatorAnnotation},
+		})
+	}
+	if equality.Semantic.DeepEqual(oldSpec, newSpec) && oldAnnotation[UpdaterAnnotation] != newAnnotation[UpdaterAnnotation] {
```
vagababov: do we care to update, if labels/annotations changed or only spec?
(I think spec only is fine, but just to make sure -- since label/annotation change will trigger a reconcile of child resources).

savitaashture: In update case
1. spec change or not change but if `/creator` annotation changes it should give validation error
2. `NO spec` change but only changes to `/lastmodifier` annotation than it should give error
3. Changes to spec and `/lastModifer` will handle automatically and `/lastModifier` get updated if the spec change is done by different user.

So the 2 point will handle
```
if equality.Semantic.DeepEqual(oldSpec, newSpec) && oldAnnotation[UpdaterAnnotation] != newAnnotation[UpdaterAnnotation] 
```

apis/metadata_validation.go
```diff
@@ -19,10 +19,21 @@ package apis
...
 )
 
+const (
+	groupName = "serving.knative.dev"
+	// CreatorAnnotation is the annotation key to describe the user that
+	// created the resource.
+	CreatorAnnotation = groupName + "/creator"
+	// UpdaterAnnotation is the annotation key to describe the user that
+	// last updated the resource.
+	UpdaterAnnotation = groupName + "/lastModifier"
```
markusthoemmes: Aren't these constants defined somewhere already?

apis/metadata_validation.go
```diff
@@ -19,10 +19,21 @@ package apis
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	groupName = "serving.knative.dev"
```
markusthoemmes: I don't think this should be hardcoded here.

apis/metadata_validation.go
savitaashture: yes they are defined in
https://github.com/knative/serving/blob/master/pkg/apis/serving/register.go#L20
But in knative.dev/pkg if i want to use those constants i need to import `"knative.dev/serving/pkg/apis/serving"` which leads to cyclic dependency.

markusthoemmes: They are also defined here:
https://github.com/knative/pkg/blob/74c5d67ea097dd45bc66a6e2212ff70058efe005/webhook/user_info.go#L27-L35

We should probably move them to /apis though and reference them from both points.

apis/metadata_validation.go
markusthoemmes: See 

https://github.com/knative/pkg/blob/74c5d67ea097dd45bc66a6e2212ff70058efe005/webhook/user_info.go#L38-L62

The caller should provide the respective groupName it seems.

apis/metadata_validation.go
savitaashture: Modified

Sorry again i squashed commits next time i will take care.

apis/metadata_validation_test.go
```diff
@@ -106,3 +106,50 @@ func TestValidateObjectMetadata(t *testing.T) {
...
+	tests := []struct {
+		name          string
+		old           interface{}
+		new           interface{}
+		oldAnnotation map[string]string
+		newAnnotation map[string]string
+		want          *FieldError
+	}{{
+		name: "update creator annotation",
+		old:  nil,
```
dgerd: All of the test cases have `old` and `new` as nil. What about where there are spec changes and the lastModifier should be updated. Can we add that test case as well?

savitaashture: Added test case related to spec and /lastModifier changes.

PR #541, opened 2019-07-22
Fix the issues pointed out by staticcheck

vagababov:
executed `staticcheck ./...` and fixed issues pointed out there.

/assign @vaikas-google

### DIFF
--- a/configmap/manual_watcher.go
+++ b/configmap/manual_watcher.go
@@ -40,10 +40,10 @@ func (w *ManualWatcher) Watch(name string, o Observer) {
 	defer w.m.Unlock()
 
 	if w.observers == nil {
-		w.observers = make(map[string][]Observer)
+		w.observers = make(map[string][]Observer, 1)
 	}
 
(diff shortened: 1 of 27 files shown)


configmap/manual_watcher.go
```diff
@@ -40,10 +40,10 @@ func (w *ManualWatcher) Watch(name string, o Observer) {
 	defer w.m.Unlock()
 
 	if w.observers == nil {
-		w.observers = make(map[string][]Observer)
+		w.observers = make(map[string][]Observer, 1)
 	}
 
-	wl, _ := w.observers[name]
+	wl := w.observers[name]
```
vaikas-google: can we just delete this and do:
w.observers[name] = append(w.observers[name], o)
?

vaikas-google:
/hold
/lgtm
/approve

Just one question, remove the hold if not useful.

configmap/manual_watcher.go
vagababov: Done.

PR #550, opened 2019-07-30
add openapi-gen=true for istion struct

runzexia:
Signed-off-by: runzexia <runzexia@yunify.com>
Our open source project `kubesphere` use the struct of the `istio` part of `knative/pkg`. We are now generating a swagger document for `kubesphere`, but we need to generate `openapi` for this part of the struct.
https://github.com/kubesphere/kubesphere/pull/609
The PR is going to add "// +k8s:openapi-gen=true" comments for some knative structs. Thanks.


runzexia:
@googlebot I signed it!

evankanderson:
These are internal copies of the Istio APIs from before they started publishing their own go types, not intended for public consumption.

/close

PR #553, opened 2019-07-31
Read the logging configmap via API, not from the file system.

markusthoemmes:
This makes it possible to run our controllers locally which can be super handy for quick development cycles.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -72,15 +72,11 @@ func GetConfig(masterURL, kubeconfig string) (*rest.Config, error) {
 // or via reading a configMap from the API.
 // The context is expected to be initialized with injection.
 func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
-	loggingConfigData, err := configmap.Load("/etc/config-logging")
-	if err == nil {
-		return logging.NewConfigFromMap(loggingConfigData)
-	}
 	loggingConfigMap, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{})
-	if err == nil {
-		return logging.NewConfigFromConfigMap(loggingConfigMap)
+	if err != nil {
+		return nil, err
 	}
-	return nil, err
+	return logging.NewConfigFromConfigMap(loggingConfigMap)
 }
 
 func Main(component string, ctors ...injection.ControllerConstructor) {
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -113,7 +113,6 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	cfg.Burst = len(ctors) * rest.DefaultBurst
 
 	ctx, informers := injection.Default.SetupInformers(ctx, cfg)
-	kubeClient := kubeclient.Get(ctx)
 
 	// Set up our logger.
 	loggingConfig, err := GetLoggingConfig(ctx)
@@ -125,7 +124,7 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	ctx = logging.WithLogger(ctx, logger)
 
 	// TODO(mattmoor): This should itself take a context and be injection-based.
-	cmw := configmap.NewInformedWatcher(kubeClient, system.Namespace())
+	cmw := configmap.NewInformedWatcher(kubeclient.Get(ctx), system.Namespace())
 
 	// Based on the reconcilers we have linked, build up the set of controllers to run.
 	controllers := make([]*controller.Impl, 0, len(ctors))
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -25,6 +25,7 @@ import (
 	"os/user"
 	"path/filepath"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -55,32 +56,33 @@ func MainWithContext(ctx context.Context, component string, ctors ...injection.C
...
+	kubeClient := kubeclient.Get(ctx)
+
 	// Set up our logger.
-	loggingConfigMap, err := configmap.Load("/etc/config-logging")
+	loggingConfigMap, err := kubeClient.CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{})
 	if err != nil {
 		log.Fatal("Error loading logging configuration:", err)
 	}
-	loggingConfig, err := logging.NewConfigFromMap(loggingConfigMap)
+	loggingConfig, err := logging.NewConfigFromMap(loggingConfigMap.Data)
```
mattmoor: I think there's a variant that does this for you :)

injection/sharedmain/main.go
```diff
@@ -55,32 +56,33 @@ func MainWithContext(ctx context.Context, component string, ctors ...injection.C
...
+	// Adjust our client's rate limits based on the number of controller's we are running.
+	cfg.QPS = float32(len(ctors)) * rest.DefaultQPS
+	cfg.Burst = len(ctors) * rest.DefaultBurst
+
+	ctx, informers := injection.Default.SetupInformers(ctx, cfg)
+	kubeClient := kubeclient.Get(ctx)
+
 	// Set up our logger.
-	loggingConfigMap, err := configmap.Load("/etc/config-logging")
+	loggingConfigMap, err := kubeClient.CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{})
```
dprotaso: Could we attempt to read from the filesystem first and then fall back to the API call?

markusthoemmes:
/hold cancel

You were right! Fixed.

injection/sharedmain/main.go
markusthoemmes: Yes we can! Done that and pushed an update.

injection/sharedmain/main.go
```diff
@@ -87,30 +103,26 @@ func MainWithContext(ctx context.Context, component string, ctors ...injection.C
...
-	ctx = logging.WithLogger(ctx, logger)
-
-	logger.Infof("Registering %d clients", len(injection.Default.GetClients()))
-	logger.Infof("Registering %d informer factories", len(injection.Default.GetInformerFactories()))
-	logger.Infof("Registering %d informers", len(injection.Default.GetInformers()))
-	logger.Infof("Registering %d controllers", len(ctors))
+	log.Printf("Registering %d clients", len(injection.Default.GetClients()))
+	log.Printf("Registering %d informer factories", len(injection.Default.GetInformerFactories()))
+	log.Printf("Registering %d informers", len(injection.Default.GetInformers()))
+	log.Printf("Registering %d controllers", len(ctors))
```
dprotaso: Any reason why we don't want to use `logger` here?

markusthoemmes: We can't, logger creation now needs a client 😅

dprotaso: heh and the clients expect a logger in the context


https://github.com/knative/pkg/blob/5112595d1ff52715e10949cb7ea24d3c31e5246e/injection/clients/kubeclient/kubeclient.go#L42-L49

markusthoemmes: 😱 that's really ugly. Since we're panicing anyway though, we might aswell not use a logger at all there? @mattmoor thoughts on that?

markusthoemmes: An alternative route we could take: Don't try to read via API but fall back to the default config if you cannot read the file? Any preference?

mattmoor: I would greatly prefer to eliminate the volume mounts and simplify the YAML.  We need API access or the watchers won't work anyways, so if there's a downside I'm missing it.

markusthoemmes: @mattmoor the current issue is that the injection logic currently wants a proper "logger". We cannot create that without a client now though. We kinda have a circular dependency (hidden by the debouncing effect of going through context).

If you want to ditch the volume mount, let's go around and replace the logger dependency with a simple `panic` call first?

mattmoor: This cannot return nil: https://github.com/knative/pkg/blob/e8df94bc5a75081fe081548c261203467448857b/logging/logger.go#L56

injection/sharedmain/main.go
```diff
@@ -67,6 +68,21 @@ func GetConfig(masterURL, kubeconfig string) (*rest.Config, error) {
...
 }
 
+// GetLoggingConfig gets the logging config from either the file system if present
+// or via reading a configMap from the API.
+// The context is expected to be initialized with injection.
+func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
+	loggingConfigData, err := configmap.Load("/etc/config-logging")
+	if err == nil {
+		return logging.NewConfigFromMap(loggingConfigData)
+	}
```
mattmoor: @dprotaso  Why do we need both?  It should be safe to move from the volume to an API call, and then drop the volumes.  What am I missing? :)

dprotaso: > Why do we need both?

We don't. I realize the other configs aren't mounted anymore

PR #554, opened 2019-07-31
Derive the config for our controller setup like kubectl etc.

markusthoemmes:
Our config loading currently only tries the given flags or in-cluster config. This also adds the KUBECONFIG env variable and the user's home directoy as fallback solutions to find a valid configuration.

That helps running our controllers locally with minimal effort.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -19,7 +19,11 @@ package sharedmain
 import (
 	"context"
 	"flag"
+	"fmt"
 	"log"
+	"os"
+	"os/user"
+	"path/filepath"
 
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


dprotaso:
Seems like client-go defaults must be broken

 from `clientcmd.BuildConfigFromFlags` [godoc](https://godoc.org/k8s.io/client-go/tools/clientcmd#BuildConfigFromFlags)
> If neither masterUrl or kubeconfigPath are passed in we fallback to inClusterConfig. If inClusterConfig fails, we fallback to the default config.

markusthoemmes:
@dprotaso Nope, that sentence remains true. "default config" just isn't what we're after. This command does not look at the environment variable neither does it fall back to the user's home location. The default config is simply an empty config I think?

dprotaso:
that sucks - what you want is there though
https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L141

I guess it's not wired into `BuildConfigFromFlags`

PR #561, opened 2019-08-06
Adding apis/v1alpha1/Destination; future sink and subscriber object.

n3wscott:
Adding apis/v1alpha1/Destination; future sink and subscriber object.

A followup will be to add a destination resolver object that will assemble the destination into a full usable url.


apis/destination.go
```diff
@@ -0,0 +1,96 @@
...
+// Destination represents a target of an invocation over HTTP.
+type Destination struct {
+	// ObjectReference points to an Addressable.
+	*corev1.ObjectReference `json:",inline"`
+
+	// URI is for direct URI Designations.
+	URI *URL `json:"uri,omitempty"`
+
+	// Path is used with the resulting URL from Addressable ObjectReference or
+	// URI. Must start with `/`. Only OPatgh
```
lionelvillard: oPatgh?

apis/destination.go
```diff
@@ -0,0 +1,96 @@
...
+type Destination struct {
+	// ObjectReference points to an Addressable.
+	*corev1.ObjectReference `json:",inline"`
+
+	// URI is for direct URI Designations.
+	URI *URL `json:"uri,omitempty"`
+
+	// Path is used with the resulting URL from Addressable ObjectReference or
+	// URI. Must start with `/`.
+	Path *string `json:"path,omitempty"`
```
mattmoor: Does this override the path in the resolved URL or augment it?

apis/destination.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"context"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"net/url"
+	"strings"
+)
+
+// Destination represents a target of an invocation over HTTP.
+type Destination struct {
```
mattmoor: I'm wondering if this belongs in a versioned directory...

apis/v1alpha1/destination_test.go
```diff
@@ -0,0 +1,186 @@
+/*
+Copyright 2017 The Knative Authors
```
mattmoor-sockpuppet:
```suggestion
Copyright 2019 The Knative Authors
```

apis/destination.go
n3wscott: I added a comment that the intent is to append them but it is up to the implementation of the resolver. Which will be added to pkg in a followup.

apis/destination.go
n3wscott: not a bad idea. Moved to apis/v1alpha1 

I am not sure if we want to give it a name or just a apis with version. It felt more natural to have it `apis/v1alpha1`

apis/destination.go
```diff
@@ -0,0 +1,96 @@
...
+	if ref.Name == "" {
+		errs = errs.Also(ErrMissingField("name"))
+	}
+	if ref.APIVersion == "" {
+		errs = errs.Also(ErrMissingField("apiVersion"))
+	}
+	if ref.Kind == "" {
+		errs = errs.Also(ErrMissingField("kind"))
+	}
+
```
lionelvillard: what about checking the other fields are not set (FieldPath UID, etc..)?

n3wscott: I am not sure I want to limit them. But I do know I need these ones at least.

apis/destination.go
mattmoor: Having the semantics be context specific feels somewhat antithetical to duck typing, does it not?

apis/v1alpha1/destination.go
```diff
@@ -0,0 +1,99 @@
+/*
```
mattmoor: Generally there's a group name between `apis` and `v1alpha1`...

apis/destination.go
n3wscott: This is not a duck, it is a field implementation.

apis/v1alpha1/destination.go
n3wscott: this is a generic component, but not a duck

PR #566, opened 2019-08-09
Adding Source ducktype.

n3wscott:
We are working on `PodSpecable` sources and it will be really important to have a canonical structure of the minimum we expect from a `source` type resource.


apis/duck/v1beta1/source_types.go
```diff
@@ -0,0 +1,174 @@
...
+var (
+	// Verify Source resources meet duck contracts.
+	_ duck.Populatable = (*Source)(nil)
+	_ apis.Listable    = (*Source)(nil)
+)
+
+const (
+	// SourceConditionSinkProvided has status True when the Source
+	// has been configured with a sink target that is resolvable.
+	SourceConditionSinkProvided apis.ConditionType = "SinkProvided"
```
vaikas-google: Could / should there be a SourceConditionReady that all should use as the top level Happy condition?

apis/duck/v1beta1/source_types.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type CloudEventOverrides struct {
+	// OutputFormat describes the CloudEvent output format the source
+	// should send events in. All formats are over HTTP.
+	// Defaults to binary.
+	// +optional
+	OutputFormat OutputFormatType `json:"outputFormat,omitempty"`
+
+	// Extensions will specify what is added to an event as an extension.
```
vaikas-google: Are there any limits here? lengths, formats, etc.? Also, these are additive only? As in, you can not use them to override existing fields?

apis/duck/v1beta1/source_types.go
```diff
@@ -0,0 +1,174 @@
...
+
+type SourceSpec struct {
+	// Sink is a reference to an object that will resolve to a domain name or a
+	// URI directly to use as the sink.
+	Sink apisv1alpha1.Destination `json:"sink,omitempty"`
+
+	// CloudEventOverrides defines overrides to control the output format and
+	// modifications of the event sent to the sink.
+	// +optional
+	CloudEventOverrides CloudEventOverrides `json:"ceOverrides,omitempty"`
```
vaikas-google: Since this is optional, should this be a pointer instead?

apis/duck/v1beta1/source_types.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+type CloudEventOverrides struct {
+	// OutputFormat describes the CloudEvent output format the source
+	// should send events in. All formats are over HTTP.
+	// Defaults to binary.
+	// +optional
+	OutputFormat OutputFormatType `json:"outputFormat,omitempty"`
```
grantr: I worry that the presence of this field, even if it's optional and a SHOULD, will practically force importers to implement both HTTP output formats. Do we actually need this in the spec? Can it be an annotation until we know that this is a common need?

n3wscott: producing binary or structured is the min bar for an SDK. I think it is reasonable to assume the source impl can do this.

apis/duck/v1beta1/source_types.go
n3wscott: Added an `IsReady` method. I can't assume Ready because we will support Job based sources.

apis/duck/v1beta1/source_types.go
n3wscott: Made it a pointer.

apis/duck/v1beta1/source_types.go
n3wscott: I added some more comments, I think the intent is for the given extensions to win always.

apis/duck/v1beta1/source_types.go
grantr: That does seem reasonable! But I still don't understand why we need this in the spec.

The [HTTP transport binding spec](https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#3-http-message-mapping) implies that all receivers must be able to handle both binary and structured modes. If we expect receivers to be CloudEvents compliant (I think we do), then in what scenario would the user ever need to set this field?

n3wscott:
Output format has been removed for now. Will just use for PodSpecable sources.

apis/duck/v1beta1/source_types.go
spencer-p: Could add a `Succeeded` method

PR #573, opened 2019-08-14
clustermanager: move common helpers and add more helpers

chaodaiG:
Turns out to be `StandardExec` needs to be mocked in more than one place for unit testing, move it out of `GKECluster` struct to make it simpler. Also, add more common functions as well as unit tests

/cc @adrcunha

### DIFF
--- a/testutils/clustermanager/util_test.go
+++ b/testutils/clustermanager/util_test.go
@@ -23,10 +23,13 @@ import (
 )
 
 func TestGetResourceName(t *testing.T) {
-	buildNumStr := "12345678901234567890fakebuildnum"
-	expOut := "kpkg-e2e-cls"
-	if common.IsProw() {
-		expOut = "kpkg-e2e-cls-12345678901234567890"
+	datas := []struct {
+		isProw      bool
+		buildNumStr string
+		exp         string
+	}{
+		{true, "12345678901234567890fakebuildnum", "kpkg-e2e-cls-12345678901234567890"},
+		{false, "", "kpkg-e2e-cls"},
 	}
 
 	// mock GetOSEnv for testing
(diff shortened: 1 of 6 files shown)


testutils/clustermanager/util_test.go
```diff
@@ -17,29 +17,33 @@ limitations under the License.
...
-		args   []string
-		expOut string
-		expErr error
-	}{
-		{"bash", []string{"-c", "echo foo"}, "foo\n", nil},
-		{"cmd_not_exist", []string{"-c", "echo"}, "", fmt.Errorf("exec: \"cmd_not_exist\": executable file not found in $PATH")},
+func TestGetResourceName(t *testing.T) {
+	buildNumStr := "12345678901234567890fakebuildnum"
+	expOut := "kpkg-e2e-cls"
+	if common.IsProw() {
```
adrcunha: While your code is ready for both cases (Prow/local) you're only testing one of them. You should test both cases.

testutils/clustermanager/util_test.go
```diff
@@ -35,15 +38,20 @@ func TestGetResourceName(t *testing.T) {
...
-	}
-
-	out, err := getResourceName(ClusterResource)
-	if nil != err {
-		t.Fatalf("getting resource name for cluster, wanted: 'no error', got: '%v'", err)
-	}
-	if out != expOut {
-		t.Fatalf("getting resource name for cluster, wanted: '%s', got: '%s'", expOut, out)
+	for _, data := range datas {
+		common.GetOSEnv = oldFunc
```
adrcunha: Actually you can move this out the loop and use the mock in all cases.

/approve
/lgtm
/hold

Holding in case you want to address the comment.

adrcunha: Sorry, spoke too soon. Actually, your mock function should correctly handle the two expected keys, PROW_JOB_ID and BUILD_ID (and because it changes according to the data, it must be in the loop, but as a simple assignment).

chaodaiG: done

testutils/clustermanager/util_test.go
chaodaiG: done

chaodaiG:
@adrcunha , lgtm lost due to fixing unit test, please take a look

PR #574, opened 2019-08-14
Add a way to get commitId for master branch

srinivashegde86:
Currently, we only look at `.git/HEAD` to get commitId. But, this only provides us with the commitId when the head is in detached state. If we are working on some branch, this will give us a ref to the branch head, which we should then read to get the commitId for that branch. 

This PR adds a way to get the commitId from the master branch.


changeset/commit.go
```diff
@@ -26,18 +26,19 @@ import (
...
+	masterCommitIDFile = "master"
+	koDataPathEnvName  = "KO_DATA_PATH"
 )
 
 var (
 	commitIDRE = regexp.MustCompile(`^[a-f0-9]{40}$`)
 )
 
-// Get tries to fetch the first 7 digitals of GitHub commit ID from HEAD file in
+// get tries to fetch the first 7 digitals of GitHub commit ID from given file in
```
vagababov:
```suggestion
// get tries to fetch the first 7 hex digits of GitHub commit ID from given file in
```
Not strongly set on this, but it reads better for me. Same below.

mattmoor:
With https://github.com/google/ko/pull/73 we can `ln -r -s .git/refs ./cmd/foo/kodata/` and I have a change here that simply improves the resolution done by `Get()` for any ref :)

srinivashegde86:
https://github.com/knative/pkg/pull/576 adds this change to read from the links in .git/HEAD. CLosing this PR as we can use that.

PR #576, opened 2019-08-15
Add support for resolving refs, when linked.

mattmoor:
With https://github.com/google/ko/pull/73 we can link the Git ref information
into images with:
```
ln -r -s .git/refs ./cmd/foo/kodata
```

If the changeset package gets a `ref: <ref>` entry, it attempts to read that
file from `KO_DATA_PATH` next, which lets the changeset work much more effectively.


changeset/commit.go
```diff
@@ -42,11 +42,18 @@ func Get() (string, error) {
 		return "", err
 	}
 	commitID := strings.TrimSpace(string(data))
-	if !commitIDRE.MatchString(commitID) {
-		err := fmt.Errorf("%q is not a valid GitHub commit ID", commitID)
-		return "", err
+	if strings.HasPrefix(commitID, "ref: ") {
+		refName := commitID[len("ref: "):]
```
vagababov:
```
rid := strings.TrimPrefix(cid, "ref: ")
if len(rid) != len(cid) { cid = readRefid }
...
```

yanweiguo:
Hmm, `\hold` didn't work

PR #578, opened 2019-08-15
Resolver package for resolving addressables to URI 

spencer-p:
Eventing previously had a [sink reconciler](https://godoc.org/knative.dev/eventing/pkg/duck#SinkReconciler) used by sources/importers. As our library of resources that reference other resource's addresses expands, the `resolver` pkg will take its place as a more general spec/destination to URI resolver.


apis/v1alpha1/addressable_resolver.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"context"
+	"fmt"
+	"path"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"knative.dev/eventing/pkg/reconciler/names"
```
mattmoor: this creates a dependency cycle.  cc @n3wscott

spencer-p: Moving it to a new pkg

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
+/*
+Copyright 2018 The Knative Authors
```
n3wscott: 2019

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	apisv1alpha1 "knative.dev/pkg/apis/v1alpha1"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/tracker"
+
+	"knative.dev/pkg/injection/clients/dynamicclient"
+)
+
+// ForDestination resolves a Destination into a URI.
+// This is a helper component for resources that references other resources' addresses.
+type ForDestination struct {
```
n3wscott: I think the type should be `URIResolver` to make it more clear.

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	tracker         tracker.Interface
+	informerFactory pkgapisduck.InformerFactory
+}
+
+// NewForDestination creates and initializes a new resolver.ForDestination.
+func NewForDestination(ctx context.Context, callback func(string)) *ForDestination {
```
n3wscott: And then one of these for `NewForDestinations` and `NewForReferences`

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return ret
+}
+
+// GetURI resolves a Destination into a URI string.
+func (r *ForDestination) GetURI(dest apisv1alpha1.Destination, parent interface{}) (string, error) {
+	// Prefer resolved object reference + path, then try URI + path, honoring the Destination documentation
+	if dest.ObjectReference != nil {
+		url, err := r.resolveObjectReference(dest.ObjectReference, parent)
```
n3wscott: We need to expose resolveObjectReference as a top level option, perhaps two objects

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		return extendPath(url, dest.Path).String(), nil
+	} else if dest.URI != nil {
+		return extendPath(dest.URI, dest.Path).String(), nil
+	} else {
+		return "", fmt.Errorf("destination missing ObjectReference and URI, expected at least one")
+	}
+}
+
+func (r *ForDestination) resolveObjectReference(ref *corev1.ObjectReference, parent interface{}) (*apis.URL, error) {
```
n3wscott: This could be an impl detail of the ReferenceResolver that the DestinationResolver uses for example

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	if ref == nil {
+		return nil, fmt.Errorf("ref is nil")
+	}
+
+	if err := r.tracker.Track(*ref, parent); err != nil {
+		return nil, fmt.Errorf("failed to track %+v: %+v", ref, err)
```
n3wscott: for errs, you don't want `%+v`, for errors almost always you want to see the `.Errors()` version of them.

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	// K8s Services are special cased. They can be called, even though they do not satisfy the
+	// Callable interface.
+	if ref.APIVersion == "v1" && ref.Kind == "Service" {
```
n3wscott: One bug we have is we do not validate that the service exists. We should add a todo to ask k8s if the service does exist.

resolver/addressable_resolver.go
spencer-p: You can always

```go
uri, err := r.GetURI(apis.Destination{ObjectReference: myRefHere}, myParent)
```
but that's verbose and costs some allocation.

Perhaps it would be nice to have `ResolveDestination` and `ResolveObjectReference` instead of `GetURI`.

That changes the language, but I think I like `URIResolver.ResolveDestination` - the input/output relationship there is more clear.

resolver/addressable_resolver.go
spencer-p: Good point - this would break some of my unit tests (lol)

resolver/addressable_resolver.go
spencer-p: Is `%v` or `%s` the correct behavior?

resolver/addressable_resolver.go
spencer-p: Do we want a separate component for each? We can have one component that understands how to do both. Destination resolution is a superset of reference resolution.

resolver/addressable_resolver.go
n3wscott: `%v` with `err` or `%s` with `err.Error()`

resolver/addressable_resolver.go
n3wscott: some parts of the code will only want to deal in addressable refs

resolver/addressable_resolver.go
spencer-p: Made this change & all the unit tests still passed. I think it was doing the right thing already, but it's changed to `%v` now.

n3wscott:
/hold cancel

the cycle has been fixed.

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"context"
+	"fmt"
+	"path"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+
+	"knative.dev/eventing/pkg/reconciler/names"
```
n3wscott: This package should not be included in pkg

resolver/addressable_resolver.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, fmt.Errorf("failed to track %+v: %v", ref, err)
+	}
+
+	// K8s Services are special cased. They can be called, even though they do not satisfy the
+	// Callable interface.
+	// TODO(spencer-p,n3wscott) Verify that the service actually exists in K8s.
+	if ref.APIVersion == "v1" && ref.Kind == "Service" {
+		url := &apis.URL{
+			Scheme: "http",
+			Host:   names.ServiceHostName(ref.Name, ref.Namespace),
```
n3wscott: copy this method locally to avoid the import

resolver/addressable_resolver.go
```diff
@@ -68,17 +70,19 @@ func (r *URIResolver) URIFromDestination(dest apisv1alpha1.Destination, parent i
 			return "", err
 		}
 		return extendPath(url, dest.Path).String(), nil
-	} else if dest.URI != nil {
+	}
+
+	if dest.URI != nil {
```
mattmoor: (no change needed)

technically `} else if {` is still fine, as it ends up being shorter, whereas `} else {` is longer because of the extra trailing `}`.

One notable exception to the `return ... } else { ... }` rule is when one of the leading `if` statements define variables, which would be out of scope when you drop the `} else {`, e.g.

```
if foo, err := bar(); err != nil {
  return err
} else {
  baz(foo)
}
```

is the same length as:
```
foo, err := bar()
if err != nil {
  return err
}
baz(foo)
```

Just FYI, in case you are interested in some of the underlying motivations behind the style madness 😅

spencer-p: Good to know; I appreciate the feedback.

spencer-p:
Looks like serving does the exact same thing: https://github.com/knative/serving/blob/master/pkg/network/domain.go

All the more reason to upstream this.

resolver/cluster_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+	defaultDomainName = "cluster.local"
+)
+
+var (
+	domainName     string
+	domainNameOnce sync.Once
+)
+
+// ClusterDomainName fetches the cluster's domain name.
+func ClusterDomainName() string {
```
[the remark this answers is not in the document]
mattmoor: I'd like to unfork this.  Can you move this file to `knative.dev/pkg/network` and add an OWNERS file with `tcnghia` in it?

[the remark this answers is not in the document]
spencer-p: Done

PR #579, opened 2019-08-16
Add cluster creation logic with using GKE SDK

chaodaiG:
Add cluster creation logic using GKE SDK

/cc @adrcunha


testutils/clustermanager/gke.go
```diff
@@ -18,54 +18,140 @@ package clustermanager
...
-	if nil != project { // if project is supplied, use it and create cluster
+func (gs *GKEClient) Setup(numNodes *int64, nodeType *string, region *string, zone *string, project *string) (ClusterOperations, error) {
+	var err error
+	gc := &GKECluster{
+		Request: &GKERequest{
+			NumNodes:      DefaultGKENumNodes,
+			NodeType:      DefaultGKENodeType,
+			Region:        DefaultGKERegion,
+			Zone:          DefaultGKEZone,
+			BackupRegions: DefaultGKEBackupRegions},
```
adrcunha: We should make this overridable with an env var.

testutils/clustermanager/gke.go
```diff
@@ -18,54 +18,140 @@ package clustermanager
...
+	ctx := context.Background()
+
+	c, err := google.DefaultClient(ctx, container.CloudPlatformScope)
+	if nil != err {
+		return nil, fmt.Errorf("failed create google client: '%v'", err)
+	}
+
+	containerService, err := container.New(c)
+	if nil != err {
+		return nil, fmt.Errorf("failed create google client: '%v'", err)
```
adrcunha: container service?

testutils/clustermanager/gke.go
```diff
@@ -18,54 +18,140 @@ package clustermanager
...
+		}
+		if nil != numNodes {
+			gc.Request.NumNodes = *numNodes
+		}
+		if nil != nodeType {
+			gc.Request.NodeType = *nodeType
+		}
+		if nil != region {
+			gc.Request.Region = *region
+			gc.Request.BackupRegions = make([]string, 0)
```
adrcunha: Can we be smarter about the regions here? A simple approach would be "use the default plus default backup regions, except one that was specified by the user". So, if  I use us-east1 here, central1 and west1 would become the backup regions.

testutils/clustermanager/gke.go
```diff
@@ -74,14 +160,76 @@ func (gc *GKECluster) Provider() string {
...
+				Name:             clusterName,
+				InitialNodeCount: gc.Request.NumNodes,
+				NodeConfig: &container.NodeConfig{
+					MachineType: gc.Request.NodeType,
+				},
+			},
+			ProjectId: *gc.Project,
+		}
+		log.Printf("Creating cluster in %s", getClusterLocation(region, gc.Request.Zone))
+		_, err := gc.oprations.create(*gc.Project, getClusterLocation(region, gc.Request.Zone), rb)
```
```suggestion
		_, err := gc.operations.create(*gc.Project, getClusterLocation(region, gc.Request.Zone), rb)
```

testutils/clustermanager/gke.go
```diff
@@ -74,14 +160,76 @@ func (gc *GKECluster) Provider() string {
...
+		if nil == err {
+			// The process above doesn't seem to wait, wait for it
+			log.Printf("Waiting for cluster creation")
+			timeout := time.After(5 * time.Minute)
```
adrcunha: Make a constant. Are we sure 5 minutes is enough?

testutils/clustermanager/gke.go
```diff
@@ -74,14 +160,76 @@ func (gc *GKECluster) Provider() string {
...
+			tick := time.Tick(2 * time.Second)
+			for {
+				select {
+				// Got a timeout! fail with a timeout error
+				case <-timeout:
+					err = fmt.Errorf("timed out waiting for cluster creation")
+					break
+				case <-tick:
+					cluster, err = gc.oprations.get(*gc.Project, getClusterLocation(region, gc.Request.Zone), clusterName)
```
```suggestion
					cluster, err = gc.operations.get(*gc.Project, getClusterLocation(region, gc.Request.Zone), clusterName)
```

testutils/clustermanager/gke.go
```diff
@@ -74,14 +160,76 @@ func (gc *GKECluster) Provider() string {
...
+				if cluster.Status != "RUNNING" && cluster.Status != "PROVISIONING" {
+					err = fmt.Errorf("cluster in bad state: '%s'", cluster.Status)
+					break
+				}
+			}
+		}
+
+		if nil != err {
+			errMsg := fmt.Sprintf("error creating cluster: '%v'", err)
+			// TODO: catch specific errors as we know what the error look like for stockout etc.
```
```suggestion
			// TODO(chaodaiG): catch specific errors as we know what the error look like for stockout etc.
```

testutils/clustermanager/gke.go
```diff
@@ -97,19 +245,27 @@ func (gc *GKECluster) Delete() error {
...
-		parts := strings.Split(string(output), "_")
-		if len(parts) != 4 {
-			return fmt.Errorf("kubectl current-context is malformed: '%s'", string(output))
+		currentContext := strings.TrimSpace(string(output))
+		if strings.HasPrefix(currentContext, "gke_") {
+			log.Printf("kubeconfig isn't empty, uses this cluster for running tests: %s", currentContext)
+			// output should be in the form of gke_PROJECT_REGION_CLUSTER
+			parts := strings.Split(currentContext, "_")
+			if len(parts) != 4 {
+				return fmt.Errorf("kubectl current-context is malformed: '%s'", currentContext)
```
adrcunha: I think we should warn, then proceed with the next option (gcloud).

testutils/clustermanager/gke.go
```diff
@@ -97,19 +245,27 @@ func (gc *GKECluster) Delete() error {
...
+			}
+			gc.Project = &parts[1]
+			gc.Cluster, err = gc.oprations.get(*gc.Project, parts[2], parts[3])
```
adrcunha: Sweet!

```suggestion
			gc.Cluster, err = gc.operations.get(*gc.Project, parts[2], parts[3])
```

testutils/clustermanager/gke.go
```diff
@@ -97,19 +245,27 @@ func (gc *GKECluster) Delete() error {
...
+			if nil != err {
+				return fmt.Errorf("couldn't find cluster, does it exist: '%v'", err)
```
```suggestion
				return fmt.Errorf("couldn't find cluster %s in %s in %s, does it exist? %v", parts[3], parts[1], parts[2], err)
```

testutils/clustermanager/gke.go
adrcunha: Why? A project with a very specific region is pretty custom, so it would make sense controlling every aspect (and in this case, it just sets E2E_BACKUP_REGION to empty). And actually I'm thinking more of those cases where we want to shift the default region to some other because of stockout/performance/etc, as it happened in the past.

testutils/clustermanager/gke.go
```diff
@@ -97,19 +250,28 @@ func (gc *GKECluster) Delete() error {
...
-		// output should be in the form of gke_PROJECT_REGION_CLUSTER
-		parts := strings.Split(string(output), "_")
-		if len(parts) != 4 {
-			return fmt.Errorf("kubectl current-context is malformed: '%s'", string(output))
+		currentContext := strings.TrimSpace(string(output))
+		if strings.HasPrefix(currentContext, "gke_") {
+			// output should be in the form of gke_PROJECT_REGION_CLUSTER
+			parts := strings.Split(currentContext, "_")
+			if len(parts) != 4 { // fall through with warning
+				log.Printf("WARNING: kubectl current-context is malformed: '%s'", currentContext)
```
```suggestion
				log.Printf("WARNING: ignoring kubectl current-context since it's malformed: '%s'", currentContext)
```

testutils/clustermanager/gke.go
chaodaiG: I see the point, will update it

testutils/clustermanager/gke_test.go
```diff
@@ -162,16 +166,96 @@ func TestGKECheckEnvironment(t *testing.T) {
...
+
+	for _, data := range datas {
+		common.GetOSEnv = func(key string) string {
+			switch key {
+			case "BUILD_NUMBER":
+				return fakeBuildID
+			case "PROW_JOB_ID": // needed to mock IsProw()
+				return "jobid"
+			}
+			return oldFunc(key)
```
adrcunha: We're missing the test cases for setting the region and backup regions using env vars.

chaodaiG: testing for `setup` function is intentionally not included in this PR, as I don't want to increase the size of current one. Will create a PR for testing `setup` and another one for `delete` once this one is merged

testutils/clustermanager/gke_test.go
```diff
@@ -29,43 +33,104 @@ var (
 	fakeCluster = "d"
 )
```
adrcunha: I could swear that I saw `TestSetup()` somewhere...

testutils/clustermanager/gke_test.go
adrcunha: /lgtm
/approve

testutils/clustermanager/gke_test.go
chaodaiG: I probably should rename that method to `setupForTest`, that's for setting up each test

PR #581, opened 2019-08-20
Shuffle knative.dev/pkg/tracing code a bit

mattmoor:
This moves around the code within the tracing package in preparation for adding support for other exporter options.

As the fields changed are all private, this should be a no-op for consumers of this package.

The goal of this change is to make it so that no files mention zipkin other than zipkin.go (or tests, which I left alone).

### DIFF
--- a/tracing/zipkin.go
+++ b/tracing/zipkin.go
@@ -17,8 +17,11 @@ limitations under the License.
 package tracing
 
 import (
+	"contrib.go.opencensus.io/exporter/zipkin"
+	zipkinmodel "github.com/openzipkin/zipkin-go/model"
 	zipkinreporter "github.com/openzipkin/zipkin-go/reporter"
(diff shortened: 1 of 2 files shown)


tracing/zipkin.go
```diff
@@ -34,3 +37,38 @@ func CreateZipkinReporter(cfg *config.Config) (zipkinreporter.Reporter, error) {
...
+		}
+
+		// We know this is set because we are called with acquireGlobal lock held
+		oct := globalOct
+		if oct.exporter != nil {
+			trace.UnregisterExporter(oct.exporter)
+		}
+
+		if oct.closer != nil {
+			// TODO(greghaynes) log this error
```
greghaynes: Not a an issue with this PR so you can leave it, but we should remove this comment

PR #582, opened 2019-08-20
Expand the tracing package to support stackdriver.

mattmoor:
This makes an assortment of changes to support the use of non-Zipkin tracing backends for publishing traces from system components based on config-tracing.

The changes to the config are modeled after config-observability, which already has support for prometheus and stackdriver backends for metrics.

/hold

This is on hold until I can properly stage downstream changes to serving and eventing.

### DIFF
--- a/tracing/config/tracing.go
+++ b/tracing/config/tracing.go
@@ -19,6 +19,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"reflect"
 	"strconv"
 
 	"cloud.google.com/go/compute/metadata"
@@ -61,11 +62,7 @@ type Config struct {
 
 // Equals returns true if two Configs are identical
 func (cfg *Config) Equals(other *Config) bool {
-	return other.Backend == cfg.Backend &&
-		other.StackdriverProjectID == cfg.StackdriverProjectID &&
-		other.ZipkinEndpoint == cfg.ZipkinEndpoint &&
-		other.Debug == cfg.Debug &&
-		other.SampleRate == cfg.SampleRate
+	return reflect.DeepEqual(other, cfg)
 }
 
 // NewTracingConfigFromMap returns a Config given a map corresponding to a ConfigMap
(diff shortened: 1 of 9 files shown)


mattmoor:
cc @greghaynes @mdemirhan @Harwayne for any early feedback they might have.

tracing/config/tracing.go
```diff
@@ -21,55 +21,99 @@ import (
...
+	StackdriverProjectID string
+
+	Debug      bool
+	SampleRate float64
 }
 
 // Equals returns true if two Configs are identical
 func (cfg *Config) Equals(other *Config) bool {
-	return other.Enable == cfg.Enable && other.ZipkinEndpoint == cfg.ZipkinEndpoint && other.Debug == cfg.Debug && other.SampleRate == cfg.SampleRate
+	return other.Backend == cfg.Backend &&
```
vagababov: why not just return `cmp.equals` or `reflect.DeepEquals`?

tracing/config/tracing.go
```diff
@@ -21,55 +21,99 @@ import (
...
 		enableBool, err := strconv.ParseBool(enable)
 		if err != nil {
 			return nil, fmt.Errorf("failed parsing tracing config %q: %v", enableKey, err)
 		}
-		tc.Enable = enableBool
+		if enableBool {
+			tc.Backend = Zipkin
+		}
+	}
+	if backend, ok := cfgMap[backendKey]; ok {
```
vagababov: why not do the previous as else of this one? Since if this one is specified, the other does not matter.

tracing/config/tracing.go
```diff
@@ -21,55 +21,99 @@ import (
...
+		bt := BackendType(backend)
```
vagababov: `switch bt:=BT(...); bt {` ?

tracing/config/tracing.go
```diff
@@ -21,55 +21,99 @@ import (
...
+		switch bt {
+		case Stackdriver, Zipkin:
+			tc.Backend = bt
+		default:
+			return nil, fmt.Errorf("unsupported tracing backend value %q", backend)
+		}
 	}
 
 	if endpoint, ok := cfgMap[zipkinEndpointKey]; !ok {
```
vagababov: I think those are joinable with `&&`. If backend is not zipkin *and* it's not set then setting `tc.ZE = ""` is fine.

tracing/opencensus.go
```diff
@@ -99,3 +103,46 @@ func createOCTConfig(cfg *config.Config) *trace.Config {
...
+			// TODO(mattmoor): disable tracing.
+		}
+		if exporter != nil {
+			trace.RegisterExporter(exporter)
+		}
+		if globalOct.exporter != nil {
+			trace.UnregisterExporter(globalOct.exporter)
+		}
+		if globalOct.closer != nil {
+			_ = globalOct.closer.Close()
```
vagababov: remove `_ =`?

tracing/config/tracing.go
mattmoor: I had no idea what this was being used for, so I didn't want to change it too much.  I made it `reflect.DeepEquals`.  #yolo.

tracing/config/tracing.go
```diff
@@ -77,6 +77,8 @@ func NewTracingConfigFromMap(cfgMap map[string]string) (*Config, error) {
 		switch bt := BackendType(backend); bt {
 		case Stackdriver, Zipkin:
```
vagababov: Can't this be enumerated same way? `S, Z, N`?

PR #585, opened 2019-08-21
Revert "Remove reflection from generic store code."

mattmoor:
Reverts knative/pkg#583

Go lacks return type covariance, so `func(*ConfigMap) (T, error)` unfortunately does not implement `func(*ConfigMap) (interface{}, error)`.

This breaks (at least) serving.

cc @dprotaso

### DIFF
--- a/configmap/store.go
+++ b/configmap/store.go
@@ -17,6 +17,7 @@ limitations under the License.
 package configmap
 
 import (
+	"reflect"
 	"sync/atomic"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


markusthoemmes:
At least I now learned why it's  using reflection 😂

jonjohnsonjr:
> Go lacks return type covariance, so func(*ConfigMap) (T, error) unfortunately does not implement func(*ConfigMap) (interface{}, error).

This might be useful as a comment in the code to deter future Markus's :smile:

PR #589, opened 2019-08-21
Add RPS constants

taragu:
Part of https://github.com/knative/serving/issues/5228

Adding RPS constants to metrics keys.
/cc @yanweiguo


yanweiguo:
We shouldn't do this as Stackdriver doesn't support "knative.dev/serving/autoscaler/stable_requests_per_second"
"knative.dev/serving/autoscaler/panic_requests_per_second"
yet.

We can report these metrics to Stackdriver custom metrics without this change.

vagababov:
Does adding them here break anything?

yanweiguo:
> Does adding them here break anything?

Yeah. OpenCensus Stackdriver reporter will complain no such metric and fail all the reporting in a batch.

taragu:
SG. Closing this PR

PR #591, opened 2019-08-21
Preserve the original port if it is specified

JRBANCEL:
The SpoofingClient replaces the entire `URL.Host`, which is actually the URL `authority`, so it replaces the `port` as well. We need to preserve the port. This will enable advanced E2E testing on non-standard ports.


JRBANCEL:
This fix is not needed anymore because this code is removed in https://github.com/knative/pkg/pull/620.

PR #595, opened 2019-08-22
Refactor admission webhook

nimakaviani:
- decouple the webhook server from the controller
- move tests to respective files


nimakaviani:
some moving the files around, before I add another admission controller for the knative ConfigMaps

webhook/resource_admission_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	"strings"
+
+	"github.com/markbates/inflect"
+	"github.com/mattbaird/jsonpatch"
+	"go.uber.org/zap"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	// clientadmissionregistrationv1beta1 "k8s.io/client-go/kubernetes/typed/admissionregistration/v1beta1"
```
vagababov: rm?

webhook/resource_admission_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/apis/duck"
+	"knative.dev/pkg/kmp"
+	"knative.dev/pkg/logging"
+)
+
+// ResourceCallback defines a signature for resource specific (Route, Configuration, etc.)
+// handlers that can validate and mutate an object. If non-nil error is returned, object creation
```
vagababov:
```suggestion
// handlers that can validate and mutate an object. If non-nil error is returned, object creation
```
`creation` or just `mutation`?
Same below?

nimakaviani:
keeping the methods that the linter is complaining about as exported. In a next PR, I will be introducing an admission controller interface with these methods in its contract.

PR #596, opened 2019-08-23
Add basic code to alert for performance regression

Fredy-Z:
Add basic code to alert on `Github` and `Slack` when performance regression is detected in automation tests.

/cc @srinivashegde86


test/mako/alerter/error.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"strings"
+)
+
+// CombineErrors combines slice of errors and return a single error
+func CombineErrors(errs []error) error {
+	if len(errs) == 0 {
+		return nil
+	}
+	var errStrs []string
```
vagababov: you know the size, so preallocate, but I think a better patten here would be to use `strings.Builder`  and to do everything in one loop.

test/mako/alerter/error.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, err := range errs {
+		errStrs = append(errStrs, err.Error())
+	}
+	return fmt.Errorf(strings.Join(errStrs, "\n"))
```
vagababov: also this is `errors.New`.

test/mako/alerter/github/issue.go
```diff
@@ -0,0 +1,212 @@
...
+*/
+
+package github
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/google/go-github/github"
+	"knative.dev/pkg/test/mako/alerter"
```
vagababov: nit: move the knative import one line down.

test/mako/alerter/github/issue.go
```diff
@@ -0,0 +1,212 @@
...
+A new regression for this test has been detected:
+%s`
+
+	// closeIssueComment is the comment of an issue when we close it
+	closeIssueComment = `
+The performance regression goes way for this test, closing this issue.`
+)
+
+// IssueOperations defines operations that can be done to github
+type IssueOperations interface {
```
vagababov: Do we need the interface?

test/mako/alerter/github/issue.go
```diff
@@ -0,0 +1,212 @@
...
+	org    string
+	repo   string
+	dryrun bool
+}
+
+// Setup creates the necessary setup to make calls to work with github issues
+func Setup(githubToken string, config config) (IssueOperations, error) {
+	ghc, err := ghutil.NewGithubClient(githubToken)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot authenticate to github: %v", err)
```
vagababov:
```suggestion
		return nil, fmt.Errorf("cannot authenticate to github: %v", err)
```

test/mako/alerter/github/issue.go
```diff
@@ -0,0 +1,212 @@
...
+		// If one issue with the same title has been closed, reopen it and add new comment
+	} else if *issue.State == string(ghutil.IssueCloseState) {
+		if err := gih.reopenIssue(org, repo, *issue.Number, dryrun); err != nil {
+			return err
+		}
+		comment := fmt.Sprintf(reopenIssueCommentTemplate, desc)
+		if err := gih.addComment(org, repo, *issue.Number, comment, dryrun); err != nil {
+			return err
+		}
+		// If the issue hasn't been updated for a long time, add a new comment
```
vagababov: Let's move the comment in the else, since it describes that.

test/mako/alerter/slack/message.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	var config *repoConfig
+	for _, repoConfig := range repoConfigs.repoConfigs {
+		if repoConfig.repo == repo {
+			config = &repoConfig
+			break
+		}
+	}
+	if config == nil {
+		return nil, fmt.Errorf("No channel configuration found for repo %v", repo)
```
vagababov:
```suggestion
		return nil, fmt.Errorf("no channel configuration found for repo %v", repo)
```

test/mako/alerter/slack/message.go
```diff
@@ -0,0 +1,83 @@
...
+	client slackutil.Operations
+	config repoConfig
+	dryrun bool
+}
+
+// Setup creates the necessary setup to make calls to work with slack
+func Setup(userName, tokenPath, repo string, dryrun bool) (MessageOperations, error) {
+	client, err := slackutil.NewClient(userName, tokenPath)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot authenticate to slack: %v", err)
```
vagababov:
```suggestion
		return nil, fmt.Errorf("cannot authenticate to slack: %v", err)
```

test/mako/alerter/slack/message.go
```diff
@@ -0,0 +1,83 @@
...
+	errs := make([]error, 0)
+	channels := smh.config.channels
+	var wg sync.WaitGroup
+	for i := range channels {
+		channel := channels[i]
+		wg.Add(1)
+		go func() {
+			message := fmt.Sprintf(messageTemplate, time.Now(), text)
+			if err := smh.client.Post(message, channel.identity); err != nil {
+				errs = append(errs, fmt.Errorf("Failed to send message to channel %v", channel))
```
vagababov:
```suggestion
				errs = append(errs, fmt.Errorf("failed to send message to channel %v", channel))
```

vagababov: This is obviously non thread safe.

test/mako/alerter/github/issue.go
Fredy-Z: It's required if we want the struct to be private, people will have to use the `Setup` function to construct `issueHandler`, in which we enforce the validations. It's the same factory pattern as what I used before - https://github.com/knative/pkg/blob/366ab85660d4e1c260795bd23c2aa09a4721c282/test/vegeta/pacers/combined_pacer.go#L48

test/mako/alerter/error.go
Fredy-Z: Changed.

test/mako/alerter/error.go
Fredy-Z: Changed to `strings.Builder`

test/mako/alerter/slack/message.go
Fredy-Z: Added `mutex` to ensure thread safety.

test/mako/alerter/slack/message.go
```diff
@@ -65,16 +65,19 @@ func (smh *messageHandler) Post(text string) error {
...
 	var wg sync.WaitGroup
 	for i := range channels {
 		channel := channels[i]
 		wg.Add(1)
 		go func() {
+			defer wg.Done()
 			message := fmt.Sprintf(messageTemplate, time.Now(), text)
 			if err := smh.client.Post(message, channel.identity); err != nil {
-				errs = append(errs, fmt.Errorf("Failed to send message to channel %v", channel))
+				mux.Lock()
```
vagababov: I guess this is not high perf code, but you can just create a channel and send errors there and consume them on the other side after `wg.Wait`. But this is fine for this.

test/mako/alerter/github/issue.go
vagababov: Yeah but you can just return the struct. Interfaces are more of a client side construct in Go (vs Java).
So unless you are sure to have multiple implementations, just return the struct.
E.g.: https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8

test/mako/alerter/slack/message.go
```diff
@@ -0,0 +1,86 @@
...
+	"knative.dev/pkg/test/mako/alerter"
+	"knative.dev/test-infra/shared/slackutil"
+)
+
+const messageTemplate = `
+As of %s, there is a new performance regression detected from automation test:
+%s`
+
+// MessageOperations defines operations that can be done to slack
+type MessageOperations interface {
```
vagababov: Same here.

Fredy-Z: Removed the interface.

PR #598, opened 2019-08-25
Ignore idempotent updates and other cleanups

vagababov:
This makes sure we don't call the configmap update on the idempotent updates, which we get now and that causes us lots of global resyncs. (┛◉Д◉)┛彡┻━┻.
This should improve our performance under heavier loads.

In addition I went ahead and applied some nitpicking to the code.

/assign mattmoor

### DIFF
--- a/configmap/informed_watcher.go
+++ b/configmap/informed_watcher.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	informers "k8s.io/client-go/informers"
 	corev1informers "k8s.io/client-go/informers/core/v1"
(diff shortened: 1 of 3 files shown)


configmap/informed_watcher.go
```diff
@@ -163,8 +161,13 @@ func (i *InformedWatcher) addConfigMapEvent(obj interface{}) {
 	i.OnChange(configMap)
 }
 
-func (i *InformedWatcher) updateConfigMapEvent(old, new interface{}) {
-	configMap := new.(*corev1.ConfigMap)
+func (i *InformedWatcher) updateConfigMapEvent(o, n interface{}) {
+	// Ignore updates that are idempotent. We are seeing those
+	// periodically.
+	if equality.Semantic.DeepEqual(o, n) {
+		return
```
mattmoor: With the resync period set to 10 hours, I'd kinda like to know why this is happening as often as it is rather than just blindly ignoring it.  If we make this a panic and look at the resulting stacktrace, what's making this happen?

If we want to ignore resyncs, then we can disable them by simply passing 0, but I'd rather understand this better as it may affect our reconcilers as well.

vagababov: I checked the implementation. And now I'll need to forget the details of the shared informers. But anyway the k8s ConfigMap client Watch returns updates which are decoded by the internals of the shared informer and passed on to us.
So we can add time as well to check how long since last update and pass along the ones which are > resync period.

configmap/informed_watcher.go
```diff
@@ -51,8 +51,9 @@ func NewInformedWatcherFromFactory(sif informers.SharedInformerFactory, namespac
 func NewInformedWatcher(kc kubernetes.Interface, namespace string) *InformedWatcher {
 	return NewInformedWatcherFromFactory(informers.NewSharedInformerFactoryWithOptions(
 		kc,
-		// This is the default resync period from controller-runtime.
-		10*time.Hour,
+		// We noticed that we're getting updates all the time anyway, due to the
+		// watches being terminated and re-spawned.
```
mattmoor: I'd mention that this is the resync period for the uninitiated.

vagababov: idk, it's in the API.

vagababov:
Notified the channel (today though)

configmap/informed_watcher.go
dprotaso: I’m not seeing a re-listing & informer updates (Add/Update/Delete events) when watches time out - I just see the watch reconnect. What K8s version do you see this with?

vagababov:
1.13.
My original idea was to just filter out the updates where new.eq.old, but
we went all in.

PR #607, opened 2019-08-27
Replace string pointer with plain string

JRBANCEL:
There is no reason to use a string pointer here.
It makes the calling code weird since string pointers are pretty much never used in Go.


vagababov:
Something that is 99% ready and only needs to update the deps, so we can
merge it within few minutes of each other.

PR #617, opened 2019-08-28
tracing.HTTPSpanMiddlewareIgnoringPaths

Harwayne:
Add tracing.HTTPSpanMiddlewareIgnoringPaths which will turn off tracing for the specified paths.

This is intended to allow us to ignore automated requests, such as liveness and readiness checks.

### DIFF
--- a/tracing/http.go
+++ b/tracing/http.go
@@ -32,16 +32,15 @@ var (
 	// underlyingSampling uses the underlying sampling configuration (normally via the ConfigMap
 	// config-tracing).
 	underlyingSampling = trace.StartOptions{}
+
+	// HTTPSpanMiddleware is an http.Handler middleware to create spans for the HTTP endpoint.
+	HTTPSpanMiddleware = HTTPSpanIgnoringPaths()
 )
 
-// HTTPSpanMiddleware is a http.Handler middleware to create spans for the HTTP endpoint
-func HTTPSpanMiddleware(next http.Handler) http.Handler {
-	return &ochttp.Handler{Handler: next}
-}
 
-// HTTPSpanMiddlewareIgnoringPaths is an http.Handler middleware to create spans for the HTTP
+// HTTPSpanIgnoringPaths is an http.Handler middleware to create spans for the HTTP
 // endpoint, not sampling any request whose path is in pathsToIgnore.
-func HTTPSpanMiddlewareIgnoringPaths(pathsToIgnore ...string) func(http.Handler) http.Handler {
+func HTTPSpanIgnoringPaths(pathsToIgnore ...string) func(http.Handler) http.Handler {
 	pathsToIgnoreSet := sets.NewString(pathsToIgnore...)
 	return func(next http.Handler) http.Handler {
 		return &ochttp.Handler{
--- a/tracing/http.go
+++ b/tracing/http.go
@@ -20,9 +20,38 @@ import (
 	"net/http"
 
 	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/trace"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var (
+	// neverSample forcibly turns off sampling.
+	neverSample = trace.StartOptions{
+		Sampler: trace.NeverSample(),
+	}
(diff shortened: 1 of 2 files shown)


Harwayne:
Ping.

tracing/http.go
```diff
@@ -20,9 +20,38 @@ import (
...
 )
 
 // HTTPSpanMiddleware is a http.Handler middleware to create spans for the HTTP endpoint
 func HTTPSpanMiddleware(next http.Handler) http.Handler {
 	return &ochttp.Handler{Handler: next}
 }
+
+// HTTPSpanMiddlewareIgnoringPaths is an http.Handler middleware to create spans for the HTTP
+// endpoint, not sampling any request whose path is in pathsToIgnore.
+func HTTPSpanMiddlewareIgnoringPaths(pathsToIgnore ...string) func(http.Handler) http.Handler {
```
evankanderson: Why not merge this with the previous function, and make this a private method?

If not, it's particularly strange to have:

`HTTPSpanMiddleware` takes an `http.Handler` as an argument and returns another `http.Handler`.

`HTTPSpanMiddlewareIgnoringPaths` **returns a function** with the same signature as `HTTPSpanMiddleware` uses.

These names are really long -- can we shorten them to something like: `FilteredHttpTrace` (and `HttpTrace`)?

tracing/http_test.go
```diff
@@ -94,3 +95,77 @@ func TestHTTPSpanMiddleware(t *testing.T) {
...
+			path:   paths[0],
+			traced: false,
+		},
+	}
+	for n, tc := range testCases {
+		t.Run(n, func(t *testing.T) {
+			var lastWrite []byte
+			fw := fakeWriter{lastWrite: &lastWrite}
+
+			url := url.URL{
```
evankanderson: Ugh -- not a fan of overloading the `url` name like this.

It looks like it's only used on the following line (141), so maybe just put it inline?

tracing/http_test.go
```diff
@@ -94,3 +95,77 @@ func TestHTTPSpanMiddleware(t *testing.T) {
...
+				Scheme: "http",
+				Host:   "test.example.com",
+				Path:   tc.path,
+			}
+			req, err := http.NewRequest("GET", url.String(), nil)
+			if err != nil {
+				t.Errorf("Failed to make fake request: %v", err)
+			}
+			traceID := "821e0d50d931235a5ba3fa42eddddd8f"
+			req.Header["X-B3-Traceid"] = []string{traceID}
```
evankanderson: Prefer:

```golang
req.Header.Set("X-B3-Traceid", traceId)
req.Header.Set("X-B3-Spanid", "....")
```

Set will create the slice and so forth for you.

Harwayne: Done.

tracing/http.go
Harwayne: I need to match the [Middleware type](https://github.com/cloudevents/sdk-go/blob/3eb440c338c784a56b16c9785d872fe8b0ed379f/pkg/cloudevents/transport/http/options.go#L225) from the CloudEvents SDK:
```golang
type Middleware func(next nethttp.Handler) nethttp.Handler
```

So, I need to return a function. Or I suppose I could make a struct and the function just hangs off the struct?

I could combine the two functions by simply removing the existing `HTTPSpanMiddleware` and renaming the new function to the old name. The downside is that we will have code in [serving](https://github.com/knative/serving/blob/24ae126893d0b26ba621b5d874ed76d8d67cfcd3/cmd/activator/main.go#L235) that looks like:
```golang
ah = tracing.HTTPSpanMiddleware()(ah)
```

What is your preference?

tracing/http_test.go
evankanderson: I'd suggest just using `fmt.Sprintf("http://test.example.com/%s", tc.path)` (given the actual `http.NewRequest` interface), but I don't feel strongly.

tracing/http.go
evankanderson: Discussed this in person; we settled on making `HTTPSpanMiddleware` be an instance of `HTTPSpanIgnoringPaths()`.

Harwayne: Talked in person and made the requested changes.

PR #620, opened 2019-08-29
Spoof at the Resolver layer, not the Request layer

### LINKED ISSUE #621: Support spoofing for HTTPS
Currently, Spoofing is done at the request layer, meaning `http.Request.Host` and `http.Request.URL.Host` are modified. This doesn't work with HTTPS because `http.Request.URL.Host` must match the TLS certificate Common Name or Subject Alternative Names, and therefore cannot be tempered. The solution is to spoof at the DNS resolution.

See: https://github.com/knative/pkg/blob/master/test/spoof/spoof.go#L148 and https://github.com/google/certificate-transparency-go/blob/d5650005f3fc3b87ef669ac9b3a7e9d9f659ef69/x509/verify.go#L109.

## Expected Behavior
`SpoofingClient` will fail to contact a `https://` URL. 

## Actual Behavior
`SpoofingClient` will successfully contact an `https://` URL.

JRBANCEL:
Fixes #621.

Currently, Spoofing is done at the request layer, meaning `http.Request.Host` and `http.Request.URL.Host` are modified. This doesn't work with HTTPS because `http.Request.URL.Host` must match the TLS certificate Common Name or Subject Alternative Names, and therefore cannot be tempered. The solution is to spoof at the DNS resolution.

This also contains a minor change to accept a custom http.Transport that will be used in HTTPS E2E tests.

### DIFF
--- a/test/clients.go
+++ b/test/clients.go
@@ -20,6 +20,7 @@ package test
 
 import (
 	"fmt"
+	"net/http"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -38,8 +39,8 @@ type KubeClient struct {
 }
 
 // NewSpoofingClient returns a spoofing client to make requests
-func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool) (*spoof.SpoofingClient, error) {
-	return spoof.New(client.Kube, logf, domain, resolvable, Flags.IngressEndpoint)
+func NewSpoofingClient(client *KubeClient, transport *http.Transport, logf logging.FormatLogger, domain string, resolvable bool) (*spoof.SpoofingClient, error) {
+	return spoof.New(client.Kube, transport, logf, domain, resolvable, Flags.IngressEndpoint)
 }
 
 // NewKubeClient instantiates and returns several clientsets required for making request to the
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -19,9 +19,12 @@ limitations under the License.
 package spoof
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


test/clients.go
```diff
@@ -38,8 +39,8 @@ type KubeClient struct {
 }
 
 // NewSpoofingClient returns a spoofing client to make requests
-func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool) (*spoof.SpoofingClient, error) {
-	return spoof.New(client.Kube, logf, domain, resolvable, Flags.IngressEndpoint)
+func NewSpoofingClient(client *KubeClient, transport *http.Transport, logf logging.FormatLogger, domain string, resolvable bool) (*spoof.SpoofingClient, error) {
```
vagababov: Why not accept roundtripper?

test/spoof/spoof.go
```diff
@@ -82,72 +85,75 @@ type SpoofingClient struct {
...
+	// Spoofing
+	endpoint, err := ResolveEndpoint(kubeClientset, domain, resolvable, endpointOverride)
+	if err != nil {
+		fmt.Errorf("failed get the cluster endpoint: %v", err)
 	}
-
-	var err error
-	if sc.endpoint, err = ResolveEndpoint(kubeClientset, domain, resolvable, endpointOverride); err != nil {
-		return nil, err
+	oldDialContext := transport.DialContext
```
vagababov: Just setup your own. I don't think we can rely that `http.DefaultTransport` is `http.Transport`.

JRBANCEL:
Related knative.dev/serving change: https://github.com/knative/serving/pull/5329

test/spoof/spoof.go
```diff
@@ -82,72 +85,75 @@ type SpoofingClient struct {
...
+	if oldDialContext == nil {
+		oldDialContext = http.DefaultTransport.(*http.Transport).DialContext
+	}
+	transport.DialContext = func(ctx context.Context, network, addr string) (conn net.Conn, e error) {
+		spoofed := addr
+		if i := strings.LastIndex(addr, ":"); i != -1 {
```
vagababov: `... i != 1 && domain == addr[:i]`
Also a comment describing why are we doing this would not hurt at all.

test/spoof/spoof.go
```diff
@@ -82,72 +85,75 @@ type SpoofingClient struct {
...
+				logf("Spoofing %s -> %s", addr, spoofed)
+			}
+		}
+		return oldDialContext(ctx, network, spoofed)
 	}
 
-	if !resolvable {
-		sc.domain = domain
+	// Zipkin tracing
+	roundTripper := &ochttp.Transport{
```
vagababov: Should this be conditional on zipkin being enabled?
OTOH, it's test so 🤷‍♂

test/spoof/spoof.go
```diff
@@ -82,72 +85,75 @@ type SpoofingClient struct {
...
-		RequestTimeout:  RequestTimeout,
-		logf:            logf,
+func New(
+	kubeClientset *kubernetes.Clientset,
+	transport *http.Transport,
+	logf logging.FormatLogger,
+	domain string,
+	resolvable bool,
+	endpointOverride string) (*SpoofingClient, error) {
+	// Spoofing
```
vagababov: This comment is redundant or incomplete.

test/clients.go
JRBANCEL: Because DialContext needs to be modified.

test/spoof/spoof.go
JRBANCEL: No idea, it was there in the original code. I just kept it.

test/clients.go
vagababov: yeah but you can cast inside `accept interface, return struct` :-)

test/spoof/spoof.go
JRBANCEL: It is mirroring the other comment below `Zipkin tracing`, like in the chain of http.RoundTripper, there is first Tracing and then Spoofing. I don't really care, what should be named? Just delete it?

test/spoof/spoof.go
JRBANCEL: What do you mean?
We have to use the DialContext that is provided to us.
We just spoof and delegate to it.

test/spoof/spoof.go
JRBANCEL: Done.

test/clients.go
JRBANCEL: Not sure this applies here. What do I do if it is not actually a http.Transport?
I really need an http.Transport, not an http.RoundTripper.

test/spoof/spoof.go
vagababov: Just delete it. Word `Spoofing` in a `spoof.go` is less than usefull.

test/spoof/spoof.go
vagababov: SG.

test/spoof/spoof.go
vagababov: What I mean is that you can't rely that this cast will always work. It's an implementation detail of `net/http`. Today it's `http.Transport`, tomorrow it's `http.NewFancyTransport`. Instead just setup up your own dial context (like we do in `networking/transports.go`).

test/clients.go
vagababov: Hmm, it simplifies a few things, but ohwell.
I just hate the cast in the `wait` function below.

JRBANCEL: Me too, http.DefaultTransport should be an *http.Transport, not an http.RoundTripper.
`accept interface, return struct`, so much for returning structs :-|

vagababov: :-) https://en.wikipedia.org/wiki/Brad_Fitzpatrick -- sits in our building and maintains that library.

test/spoof/spoof.go
JRBANCEL: Makes sense. Fixed.

JRBANCEL:
@chaodaiG please take a look and `/approve` when you can :smiley:

chaodaiG:
sorry for the delay, missed the email notification

PR #622, opened 2019-08-30
introduce an interface for AdmissionControllers

nimakaviani:


nimakaviani:
Exactly. I made this PR to serving to get feedback on the shape of it. The ConfigMapAdmissionController would be consumed similarly: https://github.com/knative/serving/pull/5338

mattmoor:
Yeah, that looks pretty straightforward.  I'd like to keep pkg -> other repos unblocked, so if you could stage things for eventing / sample-controller, I'll stamp this and we can land those in quick succession.

nimakaviani:
@mattmoor, here is the list of all staging PRs:

https://github.com/knative/serving/pull/5338
https://github.com/knative/eventing/pull/1798
https://github.com/knative/sample-controller/pull/92

I still need to bump the ref to `knative.dev/pkg` in all the above when this gets merged.

Also would be great if I can get `/ok-to-test` on https://github.com/knative/sample-controller/pull/92

thanks!

mattmoor:
/lgtm
/approve
/hold

@nimakaviani I put a `/hold` here to try and minimize the window where the PKG import is broken.  If you can sync with me on slack tomorrow, we can try to hammer out landing the downstream changes shortly after we let this land.

nimakaviani:
@mattmoor sounds good. will ping you first thing in the morning.

PR #626, opened 2019-08-31
Add the ability to scope informers to a namespace with injection.

mattmoor:
_I've imported it into serving [here](https://github.com/knative/serving/compare/master...mattmoor:namespace-scoping?expand=1) to verify the code generator works._

Sometimes controller authors want to restrict the scope of a controller to a single namespace, which can dramatically reduce the level of privilege the controller needs from cluster to namespace scope.  This adds some logic that enables folks building around the injection-based setup to plumb through a namespace, which will be supplied to the various `informers.NewSharedInformerWithOptions()` calls via `informers.WithNamespace()`.

It would be nice to have a way to attach more generalized `informers.SharedInformerOption`, but unfortunately each of these types seem bound to particular code-generation, so I went with this more focused approach.


n3wscott:
/lgtm

Now with options!!

PR #630, opened 2019-09-03
Spoof at the Resolver layer, not the Request layer

### LINKED ISSUE #621: Support spoofing for HTTPS

JRBANCEL:
Same as https://github.com/knative/pkg/pull/620 that was reverted.
The differences:
- Fixes the race condition by creating a new http.Transport before modifying it
- Public functions signatures are modified in a backward compatible way to minimize changes in knative.dev/serving
- Instead of requiring an http.Transport, use the functional options pattern

---

Fixes #621.

Currently, Spoofing is done at the request layer, meaning http.Request.Host and http.Request.URL.Host are modified. This doesn't work with HTTPS because http.Request.URL.Host must match the TLS certificate Common Name or Subject Alternative Names, and therefore cannot be tempered. The solution is to spoof at the DNS resolution.

### DIFF
--- a/test/request.go
+++ b/test/request.go
@@ -134,7 +134,14 @@ func MatchesAllOf(checkers ...spoof.ResponseChecker) spoof.ResponseChecker {
 // the domain in the request headers, otherwise it will make the request directly to domain.
 // desc will be used to name the metric that is emitted to track how long it took for the
 // domain to get into the state checked by inState.  Commas in `desc` must be escaped.
-func WaitForEndpointState(kubeClient *KubeClient, logf logging.FormatLogger, theURL string, inState spoof.ResponseChecker, desc string, resolvable bool, opts ...RequestOption) (*spoof.Response, error) {
+func WaitForEndpointState(
+	kubeClient *KubeClient,
+	logf logging.FormatLogger,
+	theURL string,
+	inState spoof.ResponseChecker,
+	desc string,
+	resolvable bool,
+	opts ...interface{}) (*spoof.Response, error) {
 	return WaitForEndpointStateWithTimeout(kubeClient, logf, theURL, inState, desc, resolvable, spoof.RequestTimeout, opts...)
 }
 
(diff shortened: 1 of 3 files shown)


test/request.go
```diff
@@ -163,11 +176,19 @@ func WaitForEndpointStateWithTimeout(
 		return nil, err
 	}
 
+	var tOpts []spoof.TransportOption
 	for _, opt := range opts {
-		opt(req)
+		rOpt, ok := opt.(RequestOption)
```
vagababov: ``
if r, ok := opt.(RO); ok {
  r(req)
} else if r, ok := opt.(s.TO); ok {
  to = append(to, r)
}
```
It's one or the other, so if else or continue.

File: test/spoof/spoof.go [hunk 2]
```diff
@@ -82,72 +85,81 @@ type SpoofingClient struct {
...
+		fmt.Errorf("failed get the cluster endpoint: %v", err)
 	}
 
-	var err error
-	if sc.endpoint, err = ResolveEndpoint(kubeClientset, domain, resolvable, endpointOverride); err != nil {
-		return nil, err
+	// Spoof the hostname at the resolver level
+	dialContext := (&net.Dialer{}).DialContext
+	transport := &http.Transport{}
+	transport.DialContext = func(ctx context.Context, network, addr string) (conn net.Conn, e error) {
```
vagababov: These several lines _are_ a TransportOption ;-)

test/spoof/spoof.go
```diff
@@ -82,72 +85,81 @@ type SpoofingClient struct {
...
+	endpoint, err := ResolveEndpoint(kubeClientset, domain, resolvable, endpointOverride)
+	if err != nil {
+		fmt.Errorf("failed get the cluster endpoint: %v", err)
 	}
 
-	var err error
-	if sc.endpoint, err = ResolveEndpoint(kubeClientset, domain, resolvable, endpointOverride); err != nil {
-		return nil, err
+	// Spoof the hostname at the resolver level
+	dialContext := (&net.Dialer{}).DialContext
```
vagababov: probably can be a package level var, since `DialContext` is safe for concurrent usage

test/spoof/spoof.go
JRBANCEL: I somewhat agree, but if I do that, SpoofingClient becomes useless. It becomes just a empty wrapper around http.Client.
Ideally, we wouldn't need SpoofingClient and we could reuse the Prober code with options, but that's a much bigger change.

vagababov: Deleting code is always good :)

JRBANCEL: I'll tackle that later :-)

PR #638, opened 2019-09-06
Sometimes valid value becomes invalid value by combination.

nak3:
example 1. https://github.com/knative/serving/issues/5382
example 2. following combination in `spec.traffic`.

```
  traffic:
  - latestRevision: true
    revisionName: hello-example-dk7nd
    percent: 100
```

But there are no error util for them, so we need to create
custom error like https://github.com/knative/serving/commit/c1583f304597df76094da30295b613ac9f01353d
or `ErrInvalidValue`.

The custom error will make code complicated and `ErrInvalidValue` is
not debug friendly.

To solve it, this patch introduces an util func `ErrInvalidCombination`.


apis/field_error.go
```diff
@@ -312,20 +312,29 @@ func ErrDisallowedUpdateDeprecatedFields(fieldPaths ...string) *FieldError {
...
+// invalid value.
 func ErrInvalidValue(value interface{}, fieldPath string) *FieldError {
 	return &FieldError{
 		Message: fmt.Sprintf("invalid value: %v", value),
 		Paths:   []string{fieldPath},
 	}
 }
 
+// ErrInvalidCombination constructs a FieldError for a field that has received an
+// invalid combination of values.
```
evankanderson: I think this is a cross-field validation -- it seems like this might be better represented (given your example) as:

> Can't set both latestRevision and revisionName in <tuple>

What do you think?

nak3: That is good idea, but it cannot be accurate. I would like to represent
following differences.

example1:
- errror (`failureThreshold` is invalid when `periodSeconds` is `zero`)
```
        readinessProbe:
          periodSeconds: 0
          failureThreshold: 3
```

- OK (`failureThreshold` is valid when `periodSeconds` is `not zero`.)
```
        readinessProbe:
          periodSeconds: 5
          failureThreshold: 3
```

example2:
- error (`revisionName` is invalid when `latestRevision` is `true`)

```
  traffic:
  - latestRevision: true
    revisionName: hello-example-dk7nd
    percent: 100
```

- OK (`revisionName` is valid when `latestRevision` is `false`.)

```
  traffic:
  - latestRevision: false
    revisionName: hello-example-dk7nd
    percent: 100
```

`Can't set both` message conflicts with OK scenario above - especially example1.

apis/field_error.go
```diff
@@ -312,20 +312,29 @@ func ErrDisallowedUpdateDeprecatedFields(fieldPaths ...string) *FieldError {
...
+func ErrInvalidCombination(bad, value interface{}, key, fieldPath string) *FieldError {
+	return &FieldError{
+		Message: fmt.Sprintf("invalid value %v, when %s is %v", bad, key, value),
```
dgerd: This could be confusing as `failureThreshold` is a key and not a value. To move towards Evan's suggestion, perhaps rephrase as: "%s must be unset when %s is %v"

For example,
`failureThreshold` must be unset when `periodSeconds` is `zero` <-- Works for current validation errors
`foo` must be unset when `bar` is `present` <-- Works for strict A or B errors where value does not matter

apis/field_error.go
```diff
@@ -312,20 +312,29 @@ func ErrDisallowedUpdateDeprecatedFields(fieldPaths ...string) *FieldError {
...
 func ErrInvalidValue(value interface{}, fieldPath string) *FieldError {
 	return &FieldError{
 		Message: fmt.Sprintf("invalid value: %v", value),
 		Paths:   []string{fieldPath},
 	}
 }
 
+// ErrInvalidCombination constructs a FieldError for a field that has received an
+// invalid combination of values.
+func ErrInvalidCombination(bad, value interface{}, key, fieldPath string) *FieldError {
```
dgerd: The ordering of fields seems confusing here and easy to get wrong. 

```suggestion
func ErrInvalidCombination(invalidKey, conflictingKey, conflictingValue interface{}, fieldPath string) *FieldError {
```

apis/field_error.go
evankanderson: Should example1 and example2 have the same error message?

I think the top priority is clear error messages; the current message "invalid value 0, when failureThreshold is 3" seems confusing for the following case:

```yaml
   readinessProbe:
    failureThreshold: 3
```

I think at a minimum you need to mention all the field names which have conflicts.

nak3: @dgerd @evankanderson thank you for your comments.

> I think at a minimum you need to mention all the field names which have conflicts.

Yes, all the field names are covered by [`Paths: []string{fieldPath}`](https://github.com/knative/pkg/pull/638/files#diff-d4c2f4719b228e9168c970edd02d87a3R334). For example, errror message for

```
   readinessProbe:
    failureThreshold: 3
```

... is 
```
invalid value 3, when periodSeconds is 0: spec.template.spec.containers[0].readinessProbe.failureThreshold
```
 or
```
invalid value 0, when failureThreshold is 3: spec.template.spec.containers[0].readinessProbe.periodSeconds
```

apis/field_error.go
nak3: No, I wanted to use the arg `bad` as a value, not key. Invalid value's key could be printed by [`Paths: []string{fieldPath}`](https://github.com/knative/pkg/pull/638/files#diff-d4c2f4719b228e9168c970edd02d87a3R334).

apis/field_error.go
nak3: > I am wondering if we have enough cases of this error to make a factory function necessary.

I just quickly checked serving's error and found five cases for this function. Um.. not sure it is enough or not :sweat_smile:

https://github.com/knative/serving/blob/9162139d768af9e2463ab007a43c385ae47ba90d/pkg/apis/serving/v1alpha1/service_validation.go#L173
https://github.com/knative/serving/blob/9162139d768af9e2463ab007a43c385ae47ba90d/pkg/apis/serving/v1beta1/route_validation.go#L149
https://github.com/knative/serving/blob/9162139d768af9e2463ab007a43c385ae47ba90d/pkg/apis/serving/v1/route_validation.go#L149
https://github.com/knative/serving/blob/9162139d768af9e2463ab007a43c385ae47ba90d/pkg/apis/serving/k8s_validation.go#L460
https://github.com/knative/serving/blob/9162139d768af9e2463ab007a43c385ae47ba90d/pkg/apis/serving/k8s_validation.go#L467

evankanderson: Thanks for doing this! The first three cases can definitely be improved:

* v1alpha1/service_validation.go:
   Should probably read: "May not set rolloutPercent for a single revision" (The error is on the length of another field, not the values in the list.)
* v1beta1/route_validation.go:
   Should probably read: "May not set revisionName and latestRevision"
* v1/route_validation.go:
   A duplicate of the above code, because Go inheritance is hard.
* serving/k8s_validation.go:
   Both of these use the `&apis.FieldError{Message: "...", Paths: []string{...}}` format I suggested, with the message "... is disallowed when periodSeconds is zero"

If you can get a consistently-good message for these use cases, go ahead. Right now, I think these error messages look like:
* "invalid value 33, when len(revisions) is 1: spec.release.rolloutPercent"
* "invalid value true, when revisionName is foo: spec.traffic[2].latestRevision"
* "invalid value 3, when periodSeconds is 0: spec.template.spec.containers[0].readinessProbe.failureThreshold"

More generically, I think the problem is that there are two field values that conflict -- this code suggests that users need to change one field (`fieldPath`), rather than suggesting that either `fieldPath` or `key` could be changed to get a working system.

You can also add a `func GenericError(diagnostic string, paths ...string) *FieldError` if that would help in these cases -- that would allow for the different error strings for the different cases, and would look like:

```golang
  err.Also(api.GenericError("May not set rolloutPercent for a single revision", apis.CurrentField))
```

PR #639, opened 2019-09-06
cluster lib: testing setup

chaodaiG:
Separating `Setup` to `Setup` and `Initialize` so that they can be unit tested. Adding unit tests for both methods

/cc @adrcunha

### DIFF
--- a/testutils/clustermanager/gke_test.go
+++ b/testutils/clustermanager/gke_test.go
@@ -18,6 +18,8 @@ package clustermanager
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


testutils/clustermanager/gke_test.go
```diff
@@ -90,6 +92,234 @@ func (fgsc *FakeGKESDKClient) get(project, location, cluster string) (*container
...
+		gotCo.operations = nil
+		if !reflect.DeepEqual(co, data.expClusterOperations) {
+			t.Fatalf("%s\nwant GKECluster:\n'%v'\ngot GKECluster:\n'%v'", errPrefix, data.expClusterOperations, co)
+		}
+	}
+}
+
+func TestInitialize(t *testing.T) {
+	customProj := "customproj"
+	datas := []struct {
```
adrcunha: This test should also exercise the Prow case.

chaodaiG: This test would actually acquire a Boskos project in Prow, I have updated to mock it always return false for IsProw(), so that there is no more accident of acquiring Boskos projects before my next update. Will explore how to mock boskos for this test

PR #648, opened 2019-09-09
Initial StatsReporter for Source 

nachocano:
- Adding a `source_stats_reporter` within metrics package. As with the `source` duck type, eventing sources might not want to depend on eventing to construct a source/importer. Thus, we are adding a source reporter herein. Note the name is source instead of importer, to match the duck type. Label tags are importer though. We may need to change them later on..


nachocano:
/assign @yanweiguo 
/assign @n3wscott 

fyi @sayanh, @Harwayne

metrics/source_stats_reporter.go
```diff
@@ -0,0 +1,146 @@
...
+	Namespace     string
+	EventType     string
+	EventSource   string
+	Name          string
+	ResourceGroup string
+}
+
+// StatsReporter defines the interface for sending source metrics.
+type StatsReporter interface {
+	ReportEventCount(args *ReportArgs, responseCode int) error
```
yanweiguo: Could you add a comment for this func? Looks like it only reports one per call?

metrics/config_test.go
```diff
@@ -1,9 +1,12 @@
 /*
-Copyright 2018 The Knative Authors.
+Copyright 2018 The Knative Authors
```
mattmoor-sockpuppet: Incorrect copyright year:
```suggestion
Copyright 2019 The Knative Authors
```

nachocano: Done

metrics/source_stats_reporter.go
nachocano: Yes, added a comment. Thanks!

logging/config.go
```diff
@@ -196,3 +199,58 @@ func ConfigMapName() string {
 	}
 	return cm
 }
+
+// Base64ToLoggingConfig converts a json+base64 string of a Config.
+// Returns a non-nil Config always.
+func Base64ToLoggingConfig(base64 string) (*Config, error) {
```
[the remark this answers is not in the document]
nachocano: I think we can do that. I'm fine with getting this in, and then do the change...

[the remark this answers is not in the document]
nachocano: @Harwayne suggested the change, let's see what he thinks...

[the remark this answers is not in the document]
nachocano: removed it, it was easier...

n3wscott:
I would break this into at least 2 PRs.

metrics/config_test.go
```diff
@@ -1,9 +1,12 @@
 /*
-Copyright 2018 The Knative Authors.
+Copyright 2019 The Knative Authors
```
n3wscott: this is not a new file so the CR should stay right?

nachocano: > I would break this into at least 2 PRs.

will do

nachocano: > this is not a new file so the CR should stay right?

yeah, sock-puppet suggestion, so I changed it, will roll it back

nachocano:
@n3wscott  splitted in this one and https://github.com/knative/pkg/pull/656
ready for another look

yanweiguo:
/lgtm

I'll approve if @n3wscott has no more comments.

nachocano:
synced offline with @n3wscott and he is willing to accept this. Let's wait to hear from him here, and we are good to go... thanks @yanweiguo

PR #636, opened 2019-09-05
add config validation as admission controller

nimakaviani:
This PR implements the proposal discussed in https://github.com/knative/serving/issues/4610

It adds a validating webhook on ConfigMaps in a knative component namespace.

PR #653, opened 2019-09-10
rename the variable for resource validation webhook

nimakaviani:
creating the PR to address breaking change in the webhook following this conversation: https://github.com/knative/pkg/pull/636#pullrequestreview-285248410

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -52,9 +52,9 @@ var (
 
 // ControllerOptions contains the configuration for the webhook
 type ControllerOptions struct {
-	// WebhookName is the name of the webhook we create to handle
+	// ResourceMutatingWebhookName is the name of the webhook we create to handle
 	// mutations before they get stored in the storage.
-	WebhookName string
+	ResourceMutatingWebhookName string
 
 	// ServiceName is the service name of the webhook.
 	ServiceName string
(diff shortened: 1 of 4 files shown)


nimakaviani:
staged changes in depending repos in the following PRs:

https://github.com/knative/serving/pull/5474
https://github.com/knative/eventing/pull/1853
https://github.com/knative/sample-controller/pull/99

/cc @mattmoor

webhook/webhook.go
```diff
@@ -52,9 +52,9 @@ var (
 
 // ControllerOptions contains the configuration for the webhook
 type ControllerOptions struct {
-	// WebhookName is the name of the webhook we create to handle
+	// ResourceMutationWebhookName is the name of the webhook we create to handle
```
n3wscott: This are normally called "Mutating Admision Webhooks", so `MutatingWebhookName`?

https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/

nimakaviani: I added `Resource` to qualify it against mutating webhooks that we may add at some other point. `ResourceMutatingWebhookName`?

nimakaviani:
@matzew ah thanks for pointing that out. wasn't aware of that repo. will apply the fix.

PR #656, opened 2019-09-10
Serializing/Deserializing Logging and Metrics configs

nachocano:
- Moving to pkg a way to serialize/deserialize logging and metric configMaps. This is a modified version of the code written by @n3wscott [here](https://github.com/google/knative-gcp/blob/master/pkg/reconciler/pullsubscription/resources/config.go), but as will be used in many places, we are moving it down to pkg.


nachocano:
@yanweiguo this one is ready for approval. I think @n3wscott doesn't have approval in these folder

PR #657, opened 2019-09-10
ConfigOptions return an error

Harwayne:
ConfigOptions return an error, so we can react to things failing, rather than only reading it in the logs.

### DIFF
--- a/tracing/opencensus_test.go
+++ b/tracing/opencensus_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package tracing_test
 
 import (
+	"errors"
 	"testing"
 
 	. "knative.dev/pkg/tracing"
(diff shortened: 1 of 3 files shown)


Harwayne:
> Feels like a test is missing?

Good catch, added.

tracing/opencensus_test.go
```diff
@@ -47,3 +48,43 @@ func TestOpenCensusTracerGlobalLifecycle(t *testing.T) {
 	}
 	otherOCT.Finish()
 }
+
+func TestOpenCensusTrace_ApplyConfig_FailingConfigOption(t *testing.T) {
```
vagababov: Remove `_` to be consistent with the rest of the tests in the file.

tracing/opencensus_test.go
```diff
@@ -47,3 +48,43 @@ func TestOpenCensusTracerGlobalLifecycle(t *testing.T) {
...
+	coErr := errors.New("configOption error")
+	oct := NewOpenCensusTracer(func(c *config.Config) error {
+		if c != nil {
+			return coErr
+		}
+		return nil
+	})
+	if err := oct.ApplyConfig(&config.Config{}); err != coErr {
+		t.Errorf("Expected error not seen. Expected %q. Actual %q", coErr, err)
```
vagababov: Here and elsewhere: "got before want", e.g. `error = %v, want: %v`.

tracing/opencensus_test.go
Harwayne: Done.

tracing/opencensus_test.go
Harwayne: Done.

Harwayne:
> I'd go with `%v` for errors, but I don't think it matters much

For production code I agree. For test code, I like `%q` as it shows any pesky trailing spaces.

tracing/zipkin.go
```diff
@@ -44,7 +44,7 @@ func WithZipkinExporter(reporterFact ZipkinReporterFactory, endpoint *zipkinmode
 			reporter, err := reporterFact(cfg)
 			if err != nil {
 				// TODO(greghaynes) log this error
```
mattmoor: This comment seems stale now.

Harwayne: Done.

PR #658, opened 2019-09-10
JSONTrace returns a structured object, not just a string.

Harwayne:
Improve the testing with Zipkin utilities:
- JSONTrace returns a structured object, not just a string.
- JSONTrace no longer sleeps, it retries until a specified timeout.
- Document how CleanupZipkinTracingSetup.

### DIFF
--- a/test/zipkin/util.go
+++ b/test/zipkin/util.go
@@ -14,17 +14,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//util has constants and helper methods useful for zipkin tracing support.
+// util has constants and helper methods useful for zipkin tracing support.
 
 package zipkin
 
 import (
-	"bytes"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"sync"
+	"time"
 
+	"github.com/openzipkin/zipkin-go/model"
 	"go.opencensus.io/trace"
 	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/test/logging"
@@ -32,7 +34,7 @@ import (
 )
 
 const (
-	//ZipkinTraceIDHeader HTTP response header key to be used to store Zipkin Trace ID.
+	// ZipkinTraceIDHeader HTTP response header key to be used to store Zipkin Trace ID.
 	ZipkinTraceIDHeader = "ZIPKIN_TRACE_ID"
 
 	// ZipkinPort is port exposed by the Zipkin Pod
@@ -43,11 +45,11 @@ const (
 	ZipkinTraceEndpoint = "http://localhost:9411/api/v2/trace/"
 
 	// App is the name of this component.
-	// This will be used as a label selector
+	// This will be used as a label selector.
 	app = "zipkin"
 
-	// Namespace we are using for istio components
-	istioNs = "istio-system"
+	// istioNS is the namespace we are using for istio components.
+	istioNS = "istio-system"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


test/zipkin/util.go
```diff
@@ -93,17 +95,33 @@ func SetupZipkinTracing(kubeClientset *kubernetes.Clientset, logf logging.Format
...
 		ZipkinTracingEnabled = true
 	})
+	return ZipkinTracingEnabled
 }
 
 // CleanupZipkinTracingSetup cleans up the Zipkin tracing setup on the machine. This involves killing the process performing port-forward.
+// This should be called exactly once in TestMain. Likely in the form:
+//
+// func TestMain(m *testing.M) {
+//     os.Exit(wrappedMain(m))
```
vagababov:
```
os.Exit(func(m *testing.M) {
...
})
```

test/zipkin/util.go
```diff
@@ -93,17 +95,33 @@ func SetupZipkinTracing(kubeClientset *kubernetes.Clientset, logf logging.Format
...
+// func wrappedMain(m *testing.M) {
+//    // Any setup required for the tests.
+//    defer zipkin.CleanupZipkinTracingSetup(logger)
+//    return m.Run()
+// }
 func CleanupZipkinTracingSetup(logf logging.FormatLogger) {
 	teardownOnce.Do(func() {
+		// Because CleanupZipkinTracingSetup only runs once, make sure that now that it has been
+		// run, SetupZipkinTracing will no longer setup any port forwarding.
+		setupOnce.Do(func() {})
```
vagababov: Not sure if this does what you want?

test/zipkin/util.go
```diff
@@ -117,29 +135,53 @@ func CheckZipkinPortAvailability() error {
...
 
-// JSONTrace returns a trace for the given traceId in JSON format
-func JSONTrace(traceID string) (string, error) {
+// JSONTrace returns a trace for the given traceId. It will continually try to get the trace. If the
+// trace it gets has the expected number of spans, then it will be returned. If not, it will try
+// again. If it reaches timeout, then it returns everything it has so far with an error.
+func JSONTrace(traceID string, expected int, timeout time.Duration) ([]model.SpanModel, error) {
+	var traceSoFar []model.SpanModel
+	t := time.After(timeout)
+	for {
```
vagababov: `for ;len(tracesSoFar) != expected; {`

test/zipkin/util.go
```diff
@@ -117,29 +135,53 @@ func CheckZipkinPortAvailability() error {
 	return monitoring.CheckPortAvailability(ZipkinPort)
 }
 
-// JSONTrace returns a trace for the given traceId in JSON format
-func JSONTrace(traceID string) (string, error) {
+// JSONTrace returns a trace for the given traceId. It will continually try to get the trace. If the
+// trace it gets has the expected number of spans, then it will be returned. If not, it will try
+// again. If it reaches timeout, then it returns everything it has so far with an error.
+func JSONTrace(traceID string, expected int, timeout time.Duration) ([]model.SpanModel, error) {
```
vagababov:
```suggestion
func JSONTrace(traceID string, expected int, timeout time.Duration) (tracesSoFar []model.SpanModel, err error) {
```

Then you can remove all the definitions below.

test/zipkin/util.go
```diff
@@ -117,29 +135,53 @@ func CheckZipkinPortAvailability() error {
 	return monitoring.CheckPortAvailability(ZipkinPort)
 }
 
-// JSONTrace returns a trace for the given traceId in JSON format
-func JSONTrace(traceID string) (string, error) {
+// JSONTrace returns a trace for the given traceId. It will continually try to get the trace. If the
```
vagababov:
```suggestion
// JSONTrace returns a trace for the given traceID. It will continually try to get the trace. If the
```

test/zipkin/util.go
Harwayne: My reading of sync.Once is that it will allow Do() to be run at most once. So there are three cases:
1. Something has already called Do(). Then this is a no-op. It will return quickly.
2. Something is currently calling Do(). Then this will block until that call has finished. Thus ensuring we clean things up appropriately.
3. Nothing has ever called Do(). Then this will ensure nothing can do so in the future.

I'm pretty sure this does what I want, doesn't it?

test/spoof/spoof.go
```diff
@@ -254,11 +254,8 @@ func (sc *SpoofingClient) logZipkinTrace(spoofResp *Response) {
 	traceID := spoofResp.Header.Get(zipkin.ZipkinTraceIDHeader)
 	sc.logf("Logging Zipkin Trace for: %s", traceID)
 
-	// Sleep to ensure all traces are correctly pushed on the backend.
-	time.Sleep(5 * time.Second)
-
-	json, err := zipkin.JSONTrace(traceID)
-	if err != nil {
+	json, err := zipkin.JSONTrace(traceID, -1, 5 * time.Second)
```
vagababov:
```suggestion
	json, err := zipkin.JSONTrace(traceID, -1/*what is this -1?!*/, 5 * time.Second)
```

test/zipkin/util.go
vagababov: But your `Do` is a no-op...

Harwayne: There are two sync.Onces, `teardownOnce` and `setupOnce`. If `teardownOnce` is being called, we need to make sure that `setupOnce` is not called afterwards. That is what this line is doing, by doing this no-op, anyone calling `zipkin.SetupZipkinTracing` in the future will _not_ create a proxy-forwarding process.

test/zipkin/util.go
Harwayne: Done.

test/zipkin/util.go
Harwayne: Done.

srinivashegde86:
/approve

Thanks for fixing this.

Issue #643, opened 2019-09-08
Stop caching test loggers


vagababov:
Currently `TestLogger(t)` caches the loggers and then we need to call `ClearAll` to get rid of them.
There is not reason why we should cache the TestLogger in the library and then clean it up, rather than instantiate the logger in the test and give it to anything that needs that logger.

While this might bring some additional  work in the integration tests, unit tests should  mostly remain the same.
And some tests forget to clear test loggers and this causes panics when `count=2`.
 
## In what area(s)?

/area test-and-release
/kind good-first-issue

mattmoor:
This is mostly relevant to unit testing which sets up controllers expecting a zap logger.  The integration tests have all moved to `t.Logf`

taragu:
@vagababov the 3 PRs to remove test logger caching is ready for another review. Would you please take a look?

mattmoor:
Is this done?

taragu:
@mattmoor yep, this is done

coryrc:
Based on the previous comment:
/close

PR #661, opened 2019-09-11
Remove caching in test logger setup

taragu:
Part of #643

### DIFF
--- a/logging/testing/util.go
+++ b/logging/testing/util.go
@@ -18,7 +18,6 @@ package testing
 
 import (
 	"context"
-	"sync"
 	"testing"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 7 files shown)


logging/testing/util.go
```diff
@@ -27,39 +26,14 @@ import (
...
-	return logger
-}
-
-// ClearAll removes all the testing loggers.
-// `go test -count=X` executes runs in the same process, thus the map
-// persists between the runs, but the `t` will no longer be valid and will
-// cause a panic deep inside testing code.
-func ClearAll() {
-	loggers = make(map[string]*zap.SugaredLogger)
+	return zaptest.NewLogger(t, opts).Sugar().Named(t.Name())
```
vagababov: Do we need named anymore?

taragu: Yeah we don't need it anymore, since we are not caching the loggers by name.

vagababov:
/lgtm
/hold

/assign mattmoor 

Once Matt approves and it merges you need to execute ` dep ensure -update knative.dev/pkg && ./hack/update-codege.sh` in fast succession on all the dependent repos.
So unhold when you can do it and someone is around to lgtm :-) (you'll need @vaikas-google and @mattmoor  since the changes are across the whole repo).

logging/testing/util.go
```diff
@@ -27,39 +26,14 @@ import (
...
-	loggers[t.Name()] = logger
-
-	return logger
-}
-
-// ClearAll removes all the testing loggers.
-// `go test -count=X` executes runs in the same process, thus the map
-// persists between the runs, but the `t` will no longer be valid and will
-// cause a panic deep inside testing code.
-func ClearAll() {
```
mattmoor: Perhaps we should leave this for now, since it's breaking.  Once the no-caching change lands we can remove all of the ClearAll() calls and safely remove this.

taragu: @mattmoor Sounds good, I can leave the `ClearAll()` for now. What about `SetupFakeContextWithCancel`? Currently it creates a new logger every time it's called, so I think we need to update it to not create a new logger

mattmoor: Sorry, what I meant was: if we leave an *empty* `ClearAll()` then this stages more cleanly.  Once the downstream repos import this (automatically, unless you are quick) we can just remove all the ClearAll calls, and then come back and clean up this empty stub.  Sorry for not being more clear before.

taragu: I see. I've updated this PR with the change. That also means for a short while (before changes to serving and eventing lands to remove multiple calls to create loggers in the same test), we could have multiple loggers for the same test?

vaikas-google:
Just checking on this @mattmoor @vagababov should we go ahead and update eventing / serving with this? I can take point on eventing.
/approve

mattmoor:
@vaikas-google Tara may have started on PRs already, but I'd coordinate with her.

taragu:
@vaikas-google @mattmoor yep I've started the PRs for eventing and serving. Thanks for the review!

PR #664, opened 2019-09-11
Moving Source StatsReporter to its own package

nachocano:
- Similar to the the webhook statsReporter, we are moving the source statsReporter to its own package. 
- Minor refactor for better readability.


source/source_stats_reporter_test.go
```diff
@@ -33,7 +33,7 @@ func unregister() {
 }
 
 func TestStatsReporter(t *testing.T) {
-	t.Skip("Fails in PROW but not locally, needs further investigation")
+	t.Skip("Fails in PROW but not locally, possibly due to the webhook stats_reporter test. Needs further investigation.")
```
vaikas-google: Is there an issue tracking this work? If not, could you create one?

nachocano: Done, https://github.com/knative/pkg/issues/665
Thanks!

PR #666, opened 2019-09-11
Cluster operations wait for *Operation object directly

chaodaiG:
Previous implementation waits on cluster creation operations indirectly by querying cluster status, which might not work all the time. Switch to wait on `*container.Operation` returned by cluster creation method instead, this should be much more reliable. Also, the same wait mechanism can be reused by `delete` and any other future operations.

By the way introduce protection for protected projects and cluster as currently supported

/cc @adrcunha

### DIFF
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -17,6 +17,7 @@ limitations under the License.
 package clustermanager
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -174,12 +175,12 @@ func (gc *GKECluster) Initialize() error {
 		}
 	}
 	if nil == gc.Project || "" == *gc.Project {
-		return fmt.Errorf("gcp project must be set")
+		return errors.New("gcp project must be set")
 	}
 	if !common.IsProw() && nil == gc.Cluster {
 		gc.NeedCleanup = true
 	}
-	log.Printf("use project '%s' for running test", *gc.Project)
+	log.Printf("Using project %q for running test", *gc.Project)
 	return nil
 }
 
@@ -233,7 +234,7 @@ func (gc *GKECluster) Acquire() error {
 		clusterLoc := getClusterLocation(region, gc.Request.Zone)
 		// TODO(chaodaiG): add deleting logic once cluster deletion logic is done
 
-		log.Printf("Creating cluster '%s' in '%s'", clusterName, clusterLoc)
+		log.Printf("Creating cluster %q' in %q", clusterName, clusterLoc)
 		var createOp *container.Operation
 		createOp, err = gc.operations.create(*gc.Project, clusterLoc, rb)
 		if nil == err {
@@ -252,7 +253,7 @@ func (gc *GKECluster) Acquire() error {
 			}
 			log.Printf(errMsg)
 		} else {
-			log.Printf("cluster creation completed")
+			log.Print("Cluster creation completed")
 			gc.Cluster = cluster
 			break
 		}
@@ -264,8 +265,11 @@ func (gc *GKECluster) Acquire() error {
 // wait depends on unique opName(operation ID created by cloud), and waits until
 // it's done
 func (gc *GKECluster) wait(location, opName string, wait time.Duration) error {
-	validStatuses := []string{"PENDING", "RUNNING", "DONE"}
-	doneStatus := "DONE"
+	const (
+		pendingStatus = "PENDING"
+		runningStatus = "RUNNING"
+		doneStatus    = "DONE"
+	)
 	var op *container.Operation
 	var err error
 
@@ -275,26 +279,21 @@ func (gc *GKECluster) wait(location, opName string, wait time.Duration) error {
 		select {
 		// Got a timeout! fail with a timeout error
 		case <-timeout:
-			return fmt.Errorf("timed out waiting")
+			return errors.New("timed out waiting")
 		case <-tick:
 			// Retry 3 times in case of weird network error, or rate limiting
-			r := 0
-			for w := 50 * time.Microsecond; r < 3; w *= 2 {
-				r++
+			for r, w := 0, 50*time.Microsecond; r < 3; r, w = r+1, w*2 {
 				op, err = gc.operations.getOperation(*gc.Project, location, opName)
 				if nil == err {
-					valid := false
-					for _, validStatus := range validStatuses {
-						if op.Status == validStatus {
-							valid = true
-						}
-					}
-					if !valid {
-						err = fmt.Errorf("unexpected operation status: '%s'", op.Status)
-					} else if op.Status == doneStatus {
+					if op.Status == doneStatus {
 						return nil
-					} else { // valid operation, no need to retry
+					} else if op.Status == pendingStatus || op.Status == runningStatus {
+						// Valid operation, no need to retry
 						break
+					} else {
+						// Have seen intermittent error state and fixed itself,
+						// let it retry to avoid too much flakiness
+						err = fmt.Errorf("unexpected operation status: %q", op.Status)
 					}
 				}
 				time.Sleep(w)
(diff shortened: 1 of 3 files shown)


testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
 
-// Delete deletes a GKE cluster
-func (gc *GKECluster) Delete() error {
-	if !gc.NeedCleanup {
-		return nil
+// wait depends on unique opName(operation ID created by cloud), and waits until
+// it's done
+func (gc *GKECluster) wait(location, opName string, wait time.Duration) error {
+	validStatuses := []string{"PENDING", "RUNNING", "DONE"}
+	doneStatus := "DONE"
```
vagababov: const.

testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
+	var op *container.Operation
+	var err error
+
+	timeout := time.After(wait)
+	tick := time.Tick(500 * time.Millisecond)
+	for {
+		select {
+		// Got a timeout! fail with a timeout error
+		case <-timeout:
+			return fmt.Errorf("timed out waiting")
```
vagababov: fmt.Errorf -> errors.New

testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
+		case <-tick:
+			// Retry 3 times in case of weird network error, or rate limiting
```
vagababov: Wouldn't retries make rate limiting even more serious?

testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
+			r := 0
+			for w := 50 * time.Microsecond; r < 3; w *= 2 {
```
vagababov: `for r, w := 0, 50; r < 3; r, w := r+1,w*2 {}`

testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
+				r++
+				op, err = gc.operations.getOperation(*gc.Project, location, opName)
+				if nil == err {
```
vagababov: Feels like C/C++ :)

testutils/clustermanager/gke.go
```diff
@@ -260,13 +261,70 @@ func (gc *GKECluster) Acquire() error {
...
+					valid := false
+					for _, validStatus := range validStatuses {
+						if op.Status == validStatus {
+							valid = true
+						}
+					}
+					if !valid {
+						err = fmt.Errorf("unexpected operation status: '%s'", op.Status)
```
vagababov: Is there any reason to continue looping?
Also if first iteration succeeds but the second one fails, `err` will be non-nil.

testutils/clustermanager/gke.go

testutils/clustermanager/gke.go

testutils/clustermanager/gke.go
chaodaiG: That's why the interval doubles every cycle. I haven't ran into the rate limit yet, will make changes to this if it does

testutils/clustermanager/gke.go
chaodaiG: seems like python, good to learn. Done

testutils/clustermanager/gke.go
chaodaiG: :)

testutils/clustermanager/gke.go
chaodaiG: I have seen it once there was an intermittent wrong state during waiting, but it fixed itself, comment added. If the first iteration succeeded it shouldn't go to second one, PTAL.

testutils/clustermanager/gke.go
```diff
@@ -166,7 +175,10 @@ func (gc *GKECluster) Initialize() error {
 		}
 	}
 	if nil == gc.Project || "" == *gc.Project {
-		return fmt.Errorf("gcp project must be set")
+		return errors.New("gcp project must be set")
+	}
+	if !common.IsProw() && nil == gc.Cluster {
+		gc.NeedCleanup = true
 	}
 	log.Printf("use project '%s' for running test", *gc.Project)
```
vagababov:
```suggestion
	log.Printf("Using project %q for running test", *gc.Project)
```

testutils/clustermanager/gke.go
```diff
@@ -217,41 +230,30 @@ func (gc *GKECluster) Acquire() error {
 			},
 			ProjectId: *gc.Project,
 		}
-		log.Printf("Creating cluster in %s", getClusterLocation(region, gc.Request.Zone))
-		_, err = gc.operations.create(*gc.Project, getClusterLocation(region, gc.Request.Zone), rb)
+
+		clusterLoc := getClusterLocation(region, gc.Request.Zone)
+		// TODO(chaodaiG): add deleting logic once cluster deletion logic is done
+
+		log.Printf("Creating cluster '%s' in '%s'", clusterName, clusterLoc)
```
vagababov:
```suggestion
		log.Printf("Creating cluster %q' in %q", clusterName, clusterLoc)
```

testutils/clustermanager/gke.go
```diff
@@ -217,41 +230,30 @@ func (gc *GKECluster) Acquire() error {
...
+				// TODO(chaodaiG): add this part when deletion logic is done
+			}
 			// TODO(chaodaiG): catch specific errors as we know what the error look like for stockout etc.
 			if len(regions) != i+1 {
 				errMsg = fmt.Sprintf("%s. Retry another region '%s' for cluster creation", errMsg, regions[i+1])
 			}
 			log.Printf(errMsg)
 		} else {
-			log.Printf("cluster creation succeeded")
+			log.Printf("cluster creation completed")
```
vagababov:
```suggestion
			log.Print("Cluster creation completed")
```

testutils/clustermanager/gke.go
```diff
@@ -260,13 +262,68 @@ func (gc *GKECluster) Acquire() error {
...
+				if nil == err {
+					if op.Status == doneStatus {
+						return nil
+					} else if op.Status == pendingStatus || op.Status == runningStatus {
+						// Valid operation, no need to retry
+						break
+					} else {
+						// Have seen intermittent error state and fixed itself,
+						// let it retry to avoid too much flakiness
+						err = fmt.Errorf("unexpected operation status: '%s'", op.Status)
```
vagababov: should we break here?, aso same '%s'->%q.

testutils/clustermanager/gke.go
```diff
@@ -260,13 +262,68 @@ func (gc *GKECluster) Acquire() error {
...
+
+	return err
+}
+
+// ensureProtected ensures not operating on protected project/cluster
+func (gc *GKECluster) ensureProtected() {
+	if nil != gc.Project {
+		for _, pp := range protectedProjects {
+			if *gc.Project == pp {
+				log.Fatalf("project '%s' is protected", *gc.Project)
```
vagababov:
```suggestion
				log.Fatalf("project %q is protected", *gc.Project)
```

testutils/clustermanager/gke.go
```diff
@@ -260,13 +262,68 @@ func (gc *GKECluster) Acquire() error {
...
+			}
+		}
+	}
+	if nil != gc.Cluster {
+		for _, pc := range protectedClusters {
+			if gc.Cluster.Name == pc {
+				log.Fatalf("cluster '%s' is protected", gc.Cluster.Name)
```
vagababov:
```suggestion
				log.Fatalf("cluster %q is protected", gc.Cluster.Name)
```

testutils/clustermanager/gke.go
chaodaiG: Purposely not breaking here, as if the getOperation API intermittently spit out random status here and we don't do retry, then the entire function will stop waiting and report operation failed. 
format done

PR #670, opened 2019-09-12
Use *url.URL instead of string in request.go

JRBANCEL:
Fixes https://github.com/knative/serving/issues/5419 - Part I

See corresponding `knative.dev/serving` change: https://github.com/knative/serving/pull/5503

/hold

### DIFF
--- a/apis/url.go
+++ b/apis/url.go
@@ -76,3 +76,9 @@ func (u *URL) String() string {
 	uu := url.URL(*u)
 	return uu.String()
 }
+
+// URL returns the URL as a url.URL.
+func (u *URL) URL() *url.URL {
+	url := url.URL(*u)
+	return &url
+}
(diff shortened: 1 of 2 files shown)


mattmoor:
/approve
/hold

This change is breaking, and so we should wait until after the downstream repos are picking up changes again before landing it, and stage changes to import this into repos that it breaks.

JRBANCEL:
@mattmoor, the corresponding `knative.dev/serving` is https://github.com/knative/serving/pull/5503.
These methods are not used in `knative.dev/eventing`.

Do you I something do something else?

mattmoor:
@JRBANCEL no, that's the usual routine, but serving is parked on the knative/pkg `release-0.9` branch until tomorrow morning.  I left a couple comments on the serving PR, but once that's ready then this is good to go, I just want to minimize the amount of downtime due to breaking changes here.

JRBANCEL:
https://github.com/knative/serving/pull/5503 tests are passing.
/hold cancel

PR #676, opened 2019-09-13
StackDriver CustomMetrics with Custom Domain Prefix

nachocano:
- Adding an additional configuration variable in configMaps to be able to specify whether to enable SD custom metrics as well as whether they should be prefixed with the configured metrics domain (instead of the hardcoded `knative.dev` one). 
- This is needed for eventing, as we will have multiple resources reporting metrics from different domains (e.g., cloud.run/events). 
- This is an additive change, it should not break existing functionality in serving, where the config-observability map doesn't even have an example of how to set this up.


nachocano:
/hold until this one gets in https://github.com/knative/pkg/pull/675
and we rebase to minimize changes

metrics/config.go
```diff
@@ -103,19 +104,27 @@ type metricsConfig struct {
...
 	// flag to "true" could cause extra Stackdriver charge.
 	// If backendDestination is not Stackdriver, this is ignored.
 	allowStackdriverCustomMetrics bool
+	// allowStackdriverCustomMetricsDomain indicates whether it is allowed to send metrics to
+	// Stackdriver using "global" resource type and custom metric type if the
+	// metrics are not supported by the registered monitored resource types. It allows to
+	// also include the configured metrics domain as part of the metric name. Setting this
+	// flag to "true" could cause extra Stackdriver charge.
+	// If backendDestination is not Stackdriver, this is ignored.
+	allowStackdriverCustomMetricsDomain bool
```
yanweiguo: Instead of introduces a bool flag to allow custom metrics domain, could we introduce a string called `subDomain` or something similar, then the full custom metric prefix is "custom.googleapis.com/<sub-domain>/<component>".

The subDomain defaults to `knative.dev/` at first. Then we use it in serving and make it as a requirement input.

nachocano: Makes sense!
So a new config variable named `metrics.stackdriver-subdomain` works? 
And if not specified, then we default to `knative.dev/`?

nachocano: `metrics.stackdriver-custom-metrics-domain` ?

yanweiguo: the name sounds good to me

nachocano:
@yanweiguo ready for another look

metrics/config.go
```diff
@@ -103,19 +104,23 @@ type metricsConfig struct {
...
 	// True if backendDestination equals to "stackdriver". Store this in a variable
 	// to reduce string comparison operations.
 	isStackdriverBackend bool
 	// stackdriverMetricTypePrefix is the metric domain joins component, e.g.
 	// "knative.dev/serving/activator". Store this in a variable to reduce string
 	// join operations.
 	stackdriverMetricTypePrefix string
-	// stackdriverCustomMetricTypePrefix is "custom.googleapis.com/knative.dev" joins
-	// component, e.g. "custom.googleapis.com/knative.dev/serving/activator".
+	// stackdriverCustomMetricTypePrefix is "custom.googleapis.com joined with the subdomain and component.
```
yanweiguo: Miss right double quote.

nachocano: fixed! thanks!

Issue #522, opened 2019-07-11
Call workqueue.SetProvider to work with opencensus metrics


mattmoor:
The Kubernetes workqueue package defines a way to register a callback for synthesizing metrics for each of the controller workqueues in a binary process [here](https://github.com/kubernetes/client-go/blob/1a4f03865c3476aab9f565c69e63be8f75edb97f/util/workqueue/metrics.go#L330-L334)

To integrate with this, we must implement the [MetricsProvider](https://github.com/kubernetes/client-go/blob/1a4f03865c3476aab9f565c69e63be8f75edb97f/util/workqueue/metrics.go#L194-L210) interface and call it from our sharedmain before we call our [controller constructors](https://github.com/knative/pkg/blob/e2418a08c195ff0acda037c11c21882ca5453d96/injection/sharedmain/main.go#L85-L89).

mattmoor:
cc @mdemirhan

PR #678, opened 2019-09-14
Implement workqueue.MetricsProvider

mattmoor:
The kubernetes workqueue provides a facility for collecting metrics by registering a `workqueue.MetricsProvider` via `workqueue.SetProvider`.

This change implements that interface to expose the workqueue metrics into opencensus.

This is loosely based on some work started by @grantr [here](https://github.com/kubernetes-sigs/controller-runtime/compare/master...grantr:opencensus-replace#diff-bb94124aff8d568cb4e82854c7d44fd1)

Fixes: https://github.com/knative/pkg/issues/522

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	"sync/atomic"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type counterMetric struct {
+	mutators []tag.Mutator
+	measure  *stats.Int64Measure
+}
+
+var (
+	_ workqueue.CounterMetric = (*counterMetric)(nil)
+)
+
+// Inc implements CounterMetric
+func (m counterMetric) Inc() {
+	Record(context.Background(), m.measure.M(1), stats.WithTags(m.mutators...))
+}
+
+type gaugeMetric struct {
+	mutators []tag.Mutator
+	measure  *stats.Int64Measure
+	total    int64
+}
+
+var (
+	_ workqueue.GaugeMetric = (*gaugeMetric)(nil)
+)
+
+// Inc implements CounterMetric
+func (m *gaugeMetric) Inc() {
+	total := atomic.AddInt64(&m.total, 1)
+	Record(context.Background(), m.measure.M(total), stats.WithTags(m.mutators...))
+}
+
+// Dec implements GaugeMetric
+func (m *gaugeMetric) Dec() {
+	total := atomic.AddInt64(&m.total, -1)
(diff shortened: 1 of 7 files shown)


metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/tag"
+	"k8s.io/client-go/util/workqueue"
+)
+
+var tagName = tag.MustNewKey("name")
+
+type intMetric struct {
```
vagababov: It's more of a counter metric than generic int metric

metrics/metricstest/metricstest.go
```diff
@@ -58,7 +58,7 @@ func CheckCountData(t *testing.T, name string, wantTags map[string]string, wantV
 		checkRowTags(t, row, name, wantTags)
 
 		if s, ok := row.Data.(*view.CountData); !ok {
-			t.Errorf("For metric %s: Reporter expected a CountData type", name)
+			t.Errorf("For metric %s: Reporter expected a CountData type, but got %T", name, row.Data)
```
vagababov: Since you're changing :)
"%s: got = %T, want:...".

metrics/workqueue.go
```diff
@@ -0,0 +1,134 @@
...
+		mutators: []tag.Mutator{tag.Insert(tagName, name)},
+		measure:  wp.Latency,
+	}
+}
+
+// LatencyView returns a view of the Latency metric.
+func (wp *WorkqueueProvider) LatencyView() *view.View {
+	return measureView(wp.Latency,
+		// 10x buckets
+		view.Distribution(1e-08, 1e-07, 1e-06, 1e-05, 1e-04, 0.001, 0.01, 0.1, 1, 10))
```
vagababov: We have a good buckets function I think.

vagababov:
Looks good to me, but I'll let @yanweiguo , since I am not proficient enough in metrics :)

metrics/metrics.go
mattmoor: There's nothing stopping us from expanding the interface as needed to implement K8s metric provider interfaces ;-)

metrics/metricstest/metricstest.go
mattmoor: Done

metrics/workqueue.go
mattmoor: We have Buckets125, but I created a new one for this.

metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var (
+	_ workqueue.GaugeMetric   = (*intMetric)(nil)
+	_ workqueue.CounterMetric = (*intMetric)(nil)
+)
+
+// Inc implements CounterMetric
+func (m intMetric) Inc() {
+	stats.RecordWithTags(context.Background(), m.mutators, m.measure.M(1))
```
yanweiguo: Could you implement something similar to `Record` func in `metrics/record.go`? If we record the data unconditionally, they will be sent to custom metrics domain when Stackdriver is used, which will cause Stackdriver charge.

mattmoor: I adapted this method to support options and the code to use it.

yanweiguo:
/lgtm
/approve
/hold

Could you verify this doesn't break Knative serving reporting metrics to Stackdriver?

mattmoor:
The original metrics are there, but I think changing the `Record` seems to have messed things up.  I'm investigating...

mattmoor:
@yanweiguo do we have any indication that custom metrics work properly with stackdriver today w/ HEAD?

yanweiguo:
> @yanweiguo do we have any indication that custom metrics work properly with stackdriver today w/ HEAD?

In OSS, no. You have to manually test it. We really need e2e tests for these metrics reported to both Prometheus and Stackdriver.

mattmoor:
/hold cancel

I reworked this a bit so that we could keep the depth metric using `view.LastValue()` (I couldn't get it to agree with stackdriver any other way), and things look right:
image

Issue #680, opened 2019-09-14
Call metrics.Register to expose client metrics to opencensus.


mattmoor:
_This is a similar issue to [this one](https://github.com/knative/pkg/issues/522) but for client metrics, which is implemented in [this PR](https://github.com/knative/pkg/pull/678)_

This issue covers the work to integrate with [this method](https://godoc.org/k8s.io/client-go/tools/metrics#Register).

Issue #679, opened 2019-09-14
Call cache.SetReflectorMetricsProvider to work with opencensus metrics


mattmoor:
_This is a similar issue to [this one](https://github.com/knative/pkg/issues/522) but for reflector metrics, which is implemented in [this PR](https://github.com/knative/pkg/pull/678)_

This issue covers the work to integrate with [this method](https://godoc.org/k8s.io/client-go/tools/cache#SetReflectorMetricsProvider).

mattmoor:
It would appear that this isn't hooked up to anything 🤔

PR #682, opened 2019-09-16
Hook into two other Kubernetes metric subsystems.

mattmoor:
This adds logic to hook into two other metric systems:
1. `cache.SetReflectorMetricsProvider`, which doesn't seem hooked up in Kubernetes yet, but would theoretically give us metrics about the mechanisms underpinning informers.
2. `metrics.Register`, which hooks us into the rest client infrastructure to give us metrics about low-level API server calls.

Fixes: https://github.com/knative/pkg/issues/679
Fixes: https://github.com/knative/pkg/issues/680

This is WIP until https://github.com/knative/pkg/pull/678 lands.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -18,11 +18,35 @@ package metrics
 
 import (
 	"context"
+	"net/url"
 	"sync/atomic"
+	"time"
 
 	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/metrics"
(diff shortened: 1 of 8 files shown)


metrics/metrics.go
```diff
@@ -68,9 +93,60 @@ type floatMetric struct {
...
+
+type resultMetric struct {
+	measure *stats.Int64Measure
+}
+
+var (
+	_ metrics.ResultMetric = (*resultMetric)(nil)
+)
+
+// Observe implements ResultMetric
```
yanweiguo: Function name doesn't match.

mattmoor: fixed

PR #689, opened 2019-09-17
Add SIGPIPE signal handler to the signals package

skaslev:
By default raised SIGPIPE signal will cause the program to exit immediately[1]
and bypass any shutdown logic.

Add syscall.SIGPIPE to the signals we get notified to disable this behavior.

[1] https://golang.org/pkg/os/signal/#hdr-SIGPIPE

### DIFF
--- a/signals/signal_posix.go
+++ b/signals/signal_posix.go
@@ -23,4 +23,4 @@ import (
 	"syscall"
 )
 
-var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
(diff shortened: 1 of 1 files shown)


mattmoor:
/ok-to-test

What's the motivation for this (source of the SIGPIPE)?  This is generally used to handle the K8s lifecycle events, which are SIGTERM -> SIGKILL

skaslev:
I ran into this while using the `signals` package in a CLI tool. Launching it in a pipeline and hitting Ctrl-C resulted in a SIGPIPE and the tool exited immediately.

I don't think any Knative containers are started in a pipeline command atm but if they do at some point this change might save a few head scratches for why cleanup logic doesn't get triggered.

PR #691, opened 2019-09-17
Watch the configMaps when they are available

houshengbo:


injection/sharedmain/main.go
```diff
@@ -140,12 +140,16 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
...
-	cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	if _, err = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{}); err != nil {
+		cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	}
 
 	// Watch the observability config map
-	cmw.Watch(metrics.ConfigMapName(),
-		metrics.UpdateExporterFromConfigMap(component, logger),
-		profilingHandler.UpdateFromConfigMap)
+	if _, err = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(metrics.ConfigMapName(), metav1.GetOptions{}); err != nil {
```
mattmoor: I think @Harwayne had a change to let this fallback on defaults when the initial configmap wasn't found, which I think I generally prefer to the approach we're taking here.

mattmoor: Ping @Harwayne

Harwayne: Sorry, had mentioned it slack rather than here. 

https://github.com/knative/eventing/blob/09346d5b53f34d4d7910b8d0384be274431a458a/pkg/configmap/store.go#L30 is the code eventing uses to default. 

Most of the defaulting behavior is here in pkg, I think this is the [key](https://github.com/knative/pkg/blob/master/configmap/informed_watcher.go#L176-L178).

matzew:
What;s the status here @houshengbo ?

vagababov:
Since no reply for more than two weeks, I think we can close this one.

PR #693, opened 2019-09-18
Remove CheckZipkinPortAvailability.

Harwayne:
Remove CheckZipkinPortAvailability.

The check did not work well on Macs, where calling JSONTrace would fail immediately because the check failed.

---

The problem was found and fixed by @Fredy-Z, I am merely the one merging it.


Fredy-Z:
/cc @srinivashegde86 
For approval.

PR #696, opened 2019-09-18
Ban importing aliased domains.

n3wscott:
Eventing had some bad merges where we almost shipped `github.com/knative/eventing/` imports in `0.9.0`.

Adding a test for Knative to never have this be possible again.

### DIFF
--- a/test/imports/banned_deps_test.go
+++ b/test/imports/banned_deps_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package test
+package imports
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


test/imports/banned_deps_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if !strings.Contains(dir, "vendor") {
+		t.Skipf("knative.dev/pkg not vendored, skipping...")
+	}
+
+	parts := strings.Split(dir, "vendor")
+
+	dir = fmt.Sprintf("%s%s%s", parts[0], string(filepath.Separator), "vendor")
+
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
evankanderson: Why not use [Build.ImportDir](https://golang.org/pkg/go/build/#Context.ImportDir) ?

evankanderson: (That way, you could have a tool that checks that a go package spec does not have certain includes.)

n3wscott: I don't want to check anything except the vendor dir. They are have bad thing on their import path and that is fine as long as it is not in our vendor dir.

adrcunha:
I think this should be converted into a tool and added as a common build test step to all repos.

/lgtm
/approve

PR #697, opened 2019-09-18
Close the stop channel only once

### LINKED ISSUE #694: Mako stub crash when too much data are pushed in
## Expected Behaviour
It outputs all data

## Actual Behaviour
It doesn't output anything, the container exits with status code 0 and `q.Store()` hangs returning any error

1. Try to run the broker tests in eventing repo with increased parallelism and longer pace configuration, like `1000:30,2000:30,1000:30,2000:30`
2. Follow both the log of aggregator and mako stub container
3. When the aggregator will start to publish data to store, the mako-stub will quit with status code 0

## Additional Info
You can try with PR https://github.com/knative/eventing/pull/1917 too, since it doubles the generated data

grantr:
Without this, calling the ShutdownMicroservice API twice will cause a panic due to closing a closed channel.

Related to #694

/cc @slinkydeveloper @antoineco

### DIFF
--- a/test/mako/stub-sidecar/main.go
+++ b/test/mako/stub-sidecar/main.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/golang/protobuf/jsonpb"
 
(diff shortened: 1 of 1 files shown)


antoineco:
The fix should prevent the double close, which is a step forward 👍  I can't figure out how [`qclose`](https://github.com/knative/eventing/blob/74ad1d9/test/test_images/latencymako/aggregator.go#L105-L113) can be closed twice though. It's either deferred at the end of the run, or performed inside `fatalf()`, but it shouldn't be both due to `os.Exit`.

I would remove the "fixes # 694" from the issue description though, because we haven't determined why the call to `quickstore.Store()` causes the sidecar to terminate when a lot of data points are submitted.

PR #711, opened 2019-09-21
Expose Go memstats as metrics.

mattmoor:
This exposes metrics for the properties available through [here](https://godoc.org/runtime#MemStats) so that we can get some better visibility into things like GC pressure.

Here's an example of Prometheus showing `HeapAlloc` for our controller process:
image

### DIFF
--- a/metrics/memstats.go
+++ b/metrics/memstats.go
@@ -1,0 +1,539 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mattmoor:
cc @Harwayne @mdemirhan

metrics/memstats.go
```diff
@@ -0,0 +1,539 @@
...
+	// This is the same as the fraction of CPU reported by
+	// GODEBUG=gctrace=1.
+	GCCPUFraction *stats.Float64Measure
+}
+
+// Start initiates a Go routine that starts pushing metrics into
+// the provided measures.
+func (msp *MemStatsProvider) Start(ctx context.Context, period time.Duration) {
+	go func() {
+		ticker := time.NewTicker(period)
```
vagababov:
```
for time.NewTicker(period) {
}
```

mattmoor: That doesn't work with cancellation?

vagababov: It does, but I guess it does not simply things

PR #717, opened 2019-09-24
enable list options for inject informers

lvjing2:


vagababov:
Needs merge
/ok-to-test
/assign @n3wscott @mattmoor

codegen/cmd/injection-gen/generators/factory.go
```diff
@@ -91,13 +92,36 @@ func init() {
 
 // Key is used as the key for associating information with a context.Context.
 type Key struct{}
+type ListOptionsKey struct{}
```
mattmoor: ListOptions aren't a code-gen output, so these shouldn't be here.  There should be a single definition (in `package injection`) of this that's invoked by the generated code.

vagababov:
Feels very stale with merge conflicts and no active progress.
/close

PR #719, opened 2019-09-24
test/webhook-apicoverage: Junit cleanup utility

dushyanthsc:
Change to add a utility to cleanup Junit test summary files.

### DIFF
--- a/test/webhook-apicoverage/tools/tools.go
+++ b/test/webhook-apicoverage/tools/tools.go
@@ -22,8 +22,11 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"os/user"
 	"path"
+	"path/filepath"
+	"regexp"
 
 	"github.com/pkg/errors"
 	"knative.dev/pkg/test/webhook-apicoverage/coveragecalculator"
@@ -52,6 +55,10 @@ const (
 	// WebhookResourcePercentageCoverageEndPoint constant for
 	// ResourcePercentageCoverage API endpoint.
 	WebhookResourcePercentageCoverageEndPoint = "https://%s:443" + webhook.ResourcePercentageCoverageEndPoint
+)
+
+var (
+	jUnitFileRegexExpr = regexp.MustCompile(`junit_.*\.xml`)
 )
 
 // GetDefaultKubePath helper method to fetch kubeconfig path.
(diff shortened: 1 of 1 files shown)


test/webhook-apicoverage/tools/tools.go
```diff
@@ -54,6 +57,10 @@ const (
 	WebhookResourcePercentageCoverageEndPoint = "https://%s:443" + webhook.ResourcePercentageCoverageEndPoint
 )
 
+var (
+	JunitFileRegexExpr = regexp.MustCompile(`junit_.*\.xml`)
```
vagababov:
```suggestion
	jUnitFileRegexExpr = regexp.MustCompile(`junit_.*\.xml`)
```

test/webhook-apicoverage/tools/tools.go
```diff
@@ -229,3 +236,15 @@ func WriteResourcePercentages(outputFile string,
 
 	return ioutil.WriteFile(outputFile, []byte(htmlData), 0400)
 }
+
+// Helper function to cleanup any existing Junit XML files.
+// This is done to ensure that we only have one Junit XML file providing the
+// API Coverage summary.
+func CleanupJunitFiles(artifactsDir string) {
+	filepath.Walk(artifactsDir, func(path string, info os.FileInfo, err error) error {
+		if !info.IsDir() && JunitFileRegexExpr.MatchString(info.Name()) {
```
vagababov:
```suggestion
		if !info.IsDir() && jUnitFileRegexExpr.MatchString(info.Name()) {
```

test/webhook-apicoverage/tools/tools.go
```diff
@@ -229,3 +236,15 @@ func WriteResourcePercentages(outputFile string,
 
 	return ioutil.WriteFile(outputFile, []byte(htmlData), 0400)
 }
+
+// Helper function to cleanup any existing Junit XML files.
+// This is done to ensure that we only have one Junit XML file providing the
+// API Coverage summary.
+func CleanupJunitFiles(artifactsDir string) {
```
Fredy-Z: Can we make this function internal and call it before/after we create the Junit XML file for API Coverage? Maybe here https://github.com/knative/pkg/blob/a1b3961a1d0badabee1e5e0fd3a18c3fbace4162/test/webhook-apicoverage/tools/tools.go#L223 ?

dushyanthsc: That be forcing all repos to follow a pattern. Its better to have it as composable rather than forcing repos to follow one, so I would keep this way.

Fredy-Z:
/lgtm
/cc @adrcunha 
In case you have some comments.

PR #722, opened 2019-09-25
Move more benchmark configs to pkg/test

chizhg:
Move more benchmark configs to pkg/test, based on the feedbacks from https://github.com/knative/serving/pull/5520:
1. Get `Repository` and `SlackConfig` from the configmap
2. `Setup` returns a `Client` that wraps all the thing we need.
3. Set up alerter as part of Mako setup process

NOTE:
This is a breaking change for both serving and eventing, the two repos will need to:
1. Add `repository` and `slackConfig` to the configmap
2. Update the code that calls `mako.Setup`

/cc @mattmoor 
/cc @vagababov 
/cc @srinivashegde86


test/mako/alerter/alerter.go
```diff
@@ -49,20 +51,28 @@ func (alerter *Alerter) SetupSlack(userName, readTokenPath, writeTokenPath strin
 }
 
 // HandleBenchmarkResult will handle the benchmark result which returns from `q.Store()`
-func (alerter *Alerter) HandleBenchmarkResult(testName string, output qpb.QuickstoreOutput, err error) {
+func (alerter *Alerter) HandleBenchmarkResult(testName string, output qpb.QuickstoreOutput, err error) error {
 	if err != nil {
 		if output.GetStatus() == qpb.QuickstoreOutput_ANALYSIS_FAIL {
+			errs := make([]error, 0)
```
vagababov:
```suggestion
			var errs []error
```

PR #723, opened 2019-09-25
Cluster management lib: allow custom cluster name

chaodaiG:
Allow custom cluster name to be used for cluster creation. 
Cluster management lib switch to accept `GKERequst` struct as input rather than more than 5 positional string + int64 + []string params.

Fixes https://github.com/knative/test-infra/issues/1393

/cc @adrcunha 
/cc @Fredy-Z

### DIFF
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -57,12 +57,12 @@ type GKEClient struct {
 
 // GKERequest contains all requests collected for cluster creation
 type GKERequest struct {
-	// Project: GKE project, no default. Fall back to get project from kube
-	// config then gcloud config
+	// Project: GKE project, no default. Fall back to get project from kubeconfig
+	// then gcloud config
 	Project string
 
 	// ClusterName: custom cluster name to use. Fall back to cluster set by
-	// kube config, else composed as k[REPO]-cls-e2e-[BUILD_ID]
+	// kubeconfig, else composed as k[REPO]-cls-e2e-[BUILD_ID]
 	ClusterName string
 
 	// MinNodes: default to 1 if not provided
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -66,10 +66,10 @@ type GKERequest struct {
 	ClusterName string
 
 	// MinNodes: default to 1 if not provided
-	MinNodes uint64
+	MinNodes int64
 
 	// MaxNodes: default to max(3, MinNodes) if not provided
-	MaxNodes uint64
+	MaxNodes int64
 
 	// NodeType: default to n1-standard-4 if not provided
 	NodeType string
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -57,13 +57,35 @@ type GKEClient struct {
 
 // GKERequest contains all requests collected for cluster creation
 type GKERequest struct {
-	MinNodes      int64
-	MaxNodes      int64
-	NodeType      string
-	Region        string
-	Zone          string
+	// Project: GKE project, no default. Fall back to get project from kube
+	// config then gcloud config
+	Project string
+
+	// ClusterName: custom cluster name to use. Fall back to cluster set by
+	// kube config, else composed as k[REPO]-cls-e2e-[BUILD_ID]
+	ClusterName string
+
+	// MinNodes: default to 1 if not provided
+	MinNodes int64
+
+	// MaxNodes: default to 3 if not provided
+	MaxNodes int64
+
+	// NodeType: default to n1-standard-4 if not provided
+	NodeType string
+
+	// Region: default to regional cluster if not provided, and use default backup regions
+	Region string
+
+	// Zone: default is none, must be provided together with region
+	Zone string
+
+	// BackupRegions: fall back regions to try out in case of cluster creation
+	// failure due to regional issue(s)
 	BackupRegions []string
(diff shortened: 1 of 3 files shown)


testutils/clustermanager/gke.go
```diff
@@ -123,55 +145,46 @@ func (gsc *GKESDKClient) setAutoscaling(project, clusterName, location, nodepool
...
 	}
-	if region != nil {
-		gc.Request.Region = *region
+	if r.NodeType == "" {
+		r.NodeType = DefaultGKENodeType
 	}
-	if common.GetOSEnv(regionEnv) != "" {
-		gc.Request.Region = common.GetOSEnv(regionEnv)
+	// Only use default backup regions if region is not provided
+	if len(r.BackupRegions) == 0 && r.Region == "" {
```
Fredy-Z: So we won't have `BackupRegions` if we provide `Region`, is it the same logic as the current shell library?

testutils/clustermanager/gke.go
```diff
@@ -123,55 +145,46 @@ func (gsc *GKESDKClient) setAutoscaling(project, clusterName, location, nodepool
...
+		gc.Project = &r.Project
 		gc.NeedCleanup = true
 	}
 
-	if minNodes != nil {
-		gc.Request.MinNodes = *minNodes
-	}
-	if maxNodes != nil {
-		gc.Request.MaxNodes = *maxNodes
+	if r.MinNodes == 0 {
```
Fredy-Z: There can be a chance that user sets `MinNodes` as a value that is >3, but does not set `MaxNodes`, then the cluster creation will result in an error. How about setting `MaxNodes` to be equal to `MinNodes` for such case, or adding some kind of verification?

testutils/clustermanager/gke.go
```diff
@@ -123,55 +145,46 @@ func (gsc *GKESDKClient) setAutoscaling(project, clusterName, location, nodepool
...
-			MinNodes:      DefaultGKEMinNodes,
-			MaxNodes:      DefaultGKEMaxNodes,
-			NodeType:      DefaultGKENodeType,
-			Region:        DefaultGKERegion,
-			Zone:          DefaultGKEZone,
-			BackupRegions: DefaultGKEBackupRegions,
-			Addons:        addons,
-		},
-	}
+// Setup sets up a GKECluster client, takes GEKRequst as parameter and applies
```
Fredy-Z:
```suggestion
// Setup sets up a GKECluster client, takes GKERequst as parameter and applies
```

testutils/clustermanager/gke.go
chizhg: Also the document says `--min-nodes` and `--max-nodes` will be ignored if autoscaling is disabled, see https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#. Though it's for `gcloud` command, I think it's the same for the SDK.
Do we have autoscaling enabled by default?

testutils/clustermanager/gke.go
chaodaiG: your suggestion is also missing the letter :)
But I'll update it

testutils/clustermanager/gke.go
chaodaiG: yes we have autoscaling enabled by default. It's a good idea to avoid only set minNodes

testutils/clustermanager/gke.go
chaodaiG: It's actually not. Adding this logic as I think it makes more sense, for example in performance test, you do want to keep in the same region so results are comparable, and in this case we shouldn't have hidden `BackupRegions`

testutils/clustermanager/gke.go

testutils/clustermanager/gke.go
```diff
@@ -57,13 +57,35 @@ type GKEClient struct {
...
+	// Project: GKE project, no default. Fall back to get project from kube
+	// config then gcloud config
+	Project string
+
+	// ClusterName: custom cluster name to use. Fall back to cluster set by
+	// kube config, else composed as k[REPO]-cls-e2e-[BUILD_ID]
+	ClusterName string
+
+	// MinNodes: default to 1 if not provided
+	MinNodes int64
```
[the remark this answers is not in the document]
chaodaiG: Had to revert this change. As GKE SDK expects int64

testutils/clustermanager/gke.go
```diff
@@ -57,13 +57,35 @@ type GKEClient struct {
 
 // GKERequest contains all requests collected for cluster creation
 type GKERequest struct {
-	MinNodes      int64
-	MaxNodes      int64
-	NodeType      string
-	Region        string
-	Zone          string
+	// Project: GKE project, no default. Fall back to get project from kube
```
```suggestion
	// Project: GKE project, no default. Fall back to get project from kubeconfig
```

testutils/clustermanager/gke.go
```diff
@@ -57,13 +57,35 @@ type GKEClient struct {
...
+	// config then gcloud config
```
```suggestion
	// then gcloud config
```

testutils/clustermanager/gke.go
```diff
@@ -57,13 +57,35 @@ type GKEClient struct {
...
+	Project string
+
+	// ClusterName: custom cluster name to use. Fall back to cluster set by
+	// kube config, else composed as k[REPO]-cls-e2e-[BUILD_ID]
```
```suggestion
	// kubeconfig, else composed as k[REPO]-cls-e2e-[BUILD_ID]
```

PR #726, opened 2019-09-27
Fix spoof to not always append port

greghaynes:
In #630 we changed transport spoofer behavior to always
append a port to our resolved host. A valid use case is to
supply a full l4 dest as the spoof targed (myhost:port). In
this case we now append an extra port resulting in an invalid
dest. Lets keep our original behavior where assume our spoofed
endpoint is a l4 dest, not a hostname.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -116,10 +116,9 @@ func New(
 	transport := &http.Transport{
 		DialContext: func(ctx context.Context, network, addr string) (conn net.Conn, e error) {
 			spoofed := addr
-			if i := strings.LastIndex(addr, ":"); i != -1 && domain == addr[:i] {
-				// The original hostname:port is spoofed by replacing the hostname by the value
-				// returned by ResolveEndpoint.
-				spoofed = endpoint + ":" + addr[i+1:]
+			i := strings.LastIndex(addr, ":")
+			if (i == -1 && domain == addr) || (i != -1 && domain == addr[:i]) {
+				spoofed = endpoint
 			}
 			return dialContext(ctx, network, spoofed)
 		},


test/spoof/spoof.go
```diff
@@ -116,10 +116,9 @@ func New(
 	transport := &http.Transport{
 		DialContext: func(ctx context.Context, network, addr string) (conn net.Conn, e error) {
 			spoofed := addr
-			if i := strings.LastIndex(addr, ":"); i != -1 && domain == addr[:i] {
-				// The original hostname:port is spoofed by replacing the hostname by the value
-				// returned by ResolveEndpoint.
-				spoofed = endpoint + ":" + addr[i+1:]
+			i := strings.LastIndex(addr, ":")
+			if (i == -1 && domain == addr) || (i != -1 && domain == addr[:i]) {
+				spoofed = endpoint
```
JRBANCEL: This won't work. `endpoint` doesn't contain a port, so the port is stripped.

JRBANCEL: You should check if endpoint contains a port in which case, this is correct, but if it doesn't let's keep the original port.

greghaynes: hrm, so previously (before #630) I was able to specify -ingressendpoint with a port and everything worked. Not sure what exactly we are going for but clearly we changed some assumptions that endpointOverride could be an L4 dest and now can only be a hostname...

JRBANCEL: We can keep that feature, all I am saying is you need to do something like:

```
            if j := strings.LastIndex(endpoint, ":"); j != -1 {
                // Override contains port, keep it
                spoofed = endpoint
            } else {
                // Keep original port if any
                if i := strings.LastIndex(addr, ":"); i != -1 {
                    spoofed = endpoint + ":" + addr[i+1:]
                } else {
                    spoofed = endpoint
                }
            }
```

greghaynes: I dont really understand the use case for keeping the port of the request here. I'd expect two scenarios - either we are forwarding to "typical" ingress where port 80/443 are configured, or we are forwarding to a custom nodeport for someones development/test cluster. In both cases I'd expect us to be able to treat the resolved dest as the full L4 dest?

JRBANCEL: Canonical example, `spoof foo-bar.example.com` to `CLUSTER_IP`. If you query `http://foo-bar.example.com`, you want to actually point to `CLUSTER_IP:80`. When querying `https://foo-bar.example.com` you want to actually point to `CLUSTER_IP:443`. the port matters, it has to be preserved.

That's why I think spoofing should not mess with ports. It is basically just mocking DNS resolution.

itsmurugappan: when I use spoofing to do e2e, i run something like below 
```
go test -v -tags=e2e -count=1 ./e2e -run ^TestProbeWhitelist$ --dockerrepo docker.io/murugappans --ingressendpoint 127.0.0.1:8081
```
I have my istio-ingressgateway port forwarded to 127.0.0.1:8081. I get the below error.

```
127.0.0.1:8081:80: too many colons in address
Spoofing probe-whitelist-mbaqsmga.serving-tests-sidecar-enabled.example.com -> 127.0.0.1:8081
```

With this PR fix am able to successfully spoof, without this how can I get it working ?

JRBANCEL: As I said in my various comments above, I am not necessarily against this PR, but the problem it assumes that spoofing is done at the IP:Port level and I am arguing it should be done at the IP level only because a Knative service can serve traffic on multiple ports at the same time.
Doesn't your cluster have a public IP?

itsmurugappan: I agree with you.  [This](https://github.com/knative/pkg/pull/726#discussion_r329258826) works for me I am actually testing from my machine using [kind](https://github.com/kubernetes-sigs/kind), wherein the only way to hit the service is through port forwarding.

vagababov:
Greg left project
/close

PR #734, opened 2019-10-01
Clustermanager lib: don't expose initialize, calling it internally only, also allow skip cluster creation and force cleanup options

chaodaiG:
`initialize` is a prerequisite for calling `Acquired` and `Delete`, which is not supposed to serve any other purpose any way, so make it private and call it internally instead of exposing it to save one step in using this lib. By the way also allows:
- Skip cluster creation: which can be used for acquiring cluster from environment
- Force cluster cleanup: there is a `NeedsCleanup` switch, which was implemented to be true only if the cluster is created by the same client. However this is never true in CI since cluster creation and deletion will be invoked by separate processes. Adding this additional flag so that we can enforce cleanup

Part of: https://github.com/knative/test-infra/issues/1186
/cc @adrcunha

### DIFF
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -222,7 +222,7 @@ func (gc *GKECluster) initialize() error {
 		if err := gc.checkEnvironment(); err != nil {
 			return fmt.Errorf("failed checking existing cluster: '%v'", err)
 		} else if gc.Cluster != nil { // Return if Cluster was already set by kubeconfig
-			// If clustername provided is different from kubeconfig, don't use it
+			// If clustername provided and kubeconfig set, ignore kubeconfig
 			if gc.Request != nil && gc.Request.ClusterName != "" && gc.Cluster.Name != gc.Request.ClusterName {
 				gc.Cluster = nil
 			}
--- a/testutils/clustermanager/gke.go
+++ b/testutils/clustermanager/gke.go
@@ -87,6 +87,13 @@ type GKERequest struct {
 
 	// Addons: cluster addons to be added to cluster, such as istio
 	Addons []string
+
+	// SkipCreation: skips cluster creation
+	SkipCreation bool
+
+	// NeedsCleanup: enforce clean up if given this option, used when running
+	// locally
+	NeedsCleanup bool
 }
 
 // GKECluster implements ClusterOperations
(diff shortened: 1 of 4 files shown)


testutils/clustermanager/gke_test.go
```diff
@@ -739,7 +741,10 @@ func TestAcquire(t *testing.T) {
 			}, nil, false,
 		}, {
 			// cluster creation succeeded
-			nil, "", false, []string{}, []string{}, &container.Cluster{
+			nil, "", false, []string{}, []string{}, true, nil, nil, false,
```
adrcunha: Shouldn't this be two separate test cases, one with skip and one without?

testutils/clustermanager/gke.go
```diff
@@ -207,15 +214,21 @@ func (gs *GKEClient) Setup(r GKERequest) ClusterOperations {
...
+func (gc *GKECluster) initialize() error {
 	// Try obtain project name via `kubectl`, `gcloud`
 	if gc.Project == nil {
 		if err := gc.checkEnvironment(); err != nil {
 			return fmt.Errorf("failed checking existing cluster: '%v'", err)
-		} else if gc.Cluster != nil { // return if Cluster was already set by kubeconfig
-			return nil
+		} else if gc.Cluster != nil { // Return if Cluster was already set by kubeconfig
+			// If clustername provided is different from kubeconfig, don't use it
+			if gc.Request != nil && gc.Request.ClusterName != "" && gc.Cluster.Name != gc.Request.ClusterName {
```
adrcunha: I'm not sure I understand the rationale here... :/

chaodaiG: The use case in my mind is that for local runs, I want the cluster to be created with name `foo`,  and my local kubeconfig is pointing at `bar` cluster, by passing `ClusterName = foo` as argument I would expect not to use the cluster pointed by my kubeconfig, but create a new cluster for me. Open to discussion though

testutils/clustermanager/gke_test.go
chaodaiG: This one is false, and the testcase below is  true. Maybe I didn't get your question correctly, so leave it up for you to resolve

testutils/clustermanager/gke.go
```diff
@@ -207,15 +214,21 @@ func (gs *GKEClient) Setup(r GKERequest) ClusterOperations {
...
-func (gc *GKECluster) Initialize() error {
+func (gc *GKECluster) initialize() error {
 	// Try obtain project name via `kubectl`, `gcloud`
 	if gc.Project == nil {
 		if err := gc.checkEnvironment(); err != nil {
 			return fmt.Errorf("failed checking existing cluster: '%v'", err)
-		} else if gc.Cluster != nil { // return if Cluster was already set by kubeconfig
-			return nil
+		} else if gc.Cluster != nil { // Return if Cluster was already set by kubeconfig
+			// If clustername provided is different from kubeconfig, don't use it
```
```suggestion
			// If clustername provided and kubeconfig set, ignore kubeconfig
```

testutils/clustermanager/gke.go
adrcunha: Makes sense. The comment isn't clear to me to illustrate this scenario, left a suggestion, hopefully it makes sense to you.

chaodaiG: Your suggestion makes sense to me, submitted it

PR #738, opened 2019-10-02
allow for custom assertions to be made in TableTests

dprotaso:


dprotaso:
Asserting tracking would be one available option now 

https://github.com/knative/pkg/pull/739

PR #739, opened 2019-10-02
make the NullTracker a fake to capture Track calls

dprotaso:


reconciler/testing/tracker.go
```diff
@@ -17,18 +17,42 @@ limitations under the License.
...
 	corev1 "k8s.io/api/core/v1"
 
 	"knative.dev/pkg/tracker"
 )
 
-// NullTracker implements Tracker.
-type NullTracker struct{}
+// NullTracker implements Tracker
+//
+// Alias is preserved for backwards compatability
```
mattmoor-sockpuppet: Found potential misspelling(s):
```suggestion
// Alias is preserved for backwards compatibility
```

PR #743, opened 2019-10-02
Use spec.env to get benchmark_key for eventing

### LINKED ISSUE #721: Allow passing benchmark ids to mako setup
`mako.Setup` tries to get the Mako benchmark id by reading a Mako config file from kodata. This is a totally reasonable approach, but it doesn't work for eventing because there's a single image that's used for multiple benchmarks. The kodata for the single image can't have all the benchmark ids, especially since we'll have tests in both eventing and eventing-contrib.

Eventing may eventually refactor tests to use the layout `mako.Setup` expects, but for now it's blocking Mako support.

I'd like an alternate mako setup method (with a different name) that can choose a benchmark id at runtime instead of compile time. Ideas for this:
- take a map of environment strings to benchmark ids and pass the map in as an env var or flag
- create a configmap containing all the benchmark config files, mount it in the container, and use it similarly to kodata

/area test-and-release
/kind feature

grac3gao:
Fixes #721 

mako.Setup tries to get the Mako benchmark id by reading a Mako config file from kodata in getBenchmark() method. This is a totally reasonable approach, but it doesn't work for eventing because there's a single image that's used for multiple benchmarks. 

Thus, for eventing, it will get benchmark_key and benchmark_name from spec.env (which will be set in the pod's yaml file)

### DIFF
--- a/test/mako/config/benchmark.go
+++ b/test/mako/config/benchmark.go
@@ -18,10 +18,12 @@ package config
 
 import (
 	"fmt"
+	"github.com/kelseyhightower/envconfig"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+	"reflect"
 
 	"github.com/golang/protobuf/proto"
 	mpb "github.com/google/mako/spec/proto/mako_go_proto"
@@ -38,12 +40,37 @@ func MustGetBenchmark() (*string, *string) {
 	return benchmarkKey, benchmarkName
 }
 
+type BenchmarkEnv struct {
+	DevKey   string `envconfig:"BENCHMARK_KEY_DEV"`
+	DevName  string `envconfig:"BENCHMARK_NAME_DEV"`
+	ProdKey  string `envconfig:"BENCHMARK_KEY_PROD"`
+	ProdName string `envconfig:"BENCHMARK_NAME_PROD"`
+}
+
 // getBenchmark fetches the appropriate benchmark_key for this configured environment.
 func getBenchmark() (*string, *string, error) {
 	// Figure out what environment we're running in from the Mako configmap.
 	env, err := getEnvironment()
 	if err != nil {
 		return nil, nil, err
+	}
+	// If pod has benchmark related env
+	benchmarkEnv := BenchmarkEnv{}
+	if err := envconfig.Process("", &benchmarkEnv); err != nil {
+		return nil, nil, err
+	}
+	if !reflect.DeepEqual(benchmarkEnv, BenchmarkEnv{}) {
+		if env == "dev" {
+			if len(benchmarkEnv.DevKey) != 0 && len(benchmarkEnv.DevName) != 0 {
+				return &benchmarkEnv.DevKey, &benchmarkEnv.DevName, nil
+			}
+			return nil, nil, fmt.Errorf(("not enough env in aggregator container to define a benchmark"))
+		} else if env == "prod" {
+			if len(benchmarkEnv.ProdKey) != 0 && len(benchmarkEnv.ProdName) != 0 {
+				return &benchmarkEnv.ProdKey, &benchmarkEnv.ProdName, nil
+			}
+			return nil, nil, fmt.Errorf(("not enough env in aggregator container to define a benchmark"))
+		}
 	}
 	// Read the Mako config file for this environment.
 	data, err := readFileFromKoData(env + ".config")
@@ -55,7 +82,6 @@ func getBenchmark() (*string, *string, error) {
 	if err := proto.UnmarshalText(string(data), bi); err != nil {
 		return nil, nil, err
 	}
-
 	// Return the benchmark_key from this environment's config file.
 	return bi.BenchmarkKey, bi.BenchmarkName, nil
 }
--- a/test/mako/config/benchmark.go
+++ b/test/mako/config/benchmark.go
@@ -61,15 +61,15 @@ func getBenchmark() (*string, *string, error) {
 	}
 	if !reflect.DeepEqual(benchmarkEnv, BenchmarkEnv{}) {
 		if env == "dev" {
-			if len(benchmarkEnv.DevKey) != 0 && len(benchmarkEnv.DevName) != 0 {
+			if benchmarkEnv.DevKey != "" && benchmarkEnv.DevName != "" {
 				return &benchmarkEnv.DevKey, &benchmarkEnv.DevName, nil
 			}
-			return nil, nil, fmt.Errorf(("not enough env in aggregator"))
+			return nil, nil, fmt.Errorf("failed to get benchmark config from environment variables")
 		} else if env == "prod" {
-			if len(benchmarkEnv.ProdKey) != 0 && len(benchmarkEnv.ProdName) != 0 {
+			if benchmarkEnv.ProdKey != "" && benchmarkEnv.ProdName != "" {
 				return &benchmarkEnv.ProdKey, &benchmarkEnv.ProdName, nil
 			}
-			return nil, nil, fmt.Errorf(("not enough env in aggregator"))
+			return nil, nil, fmt.Errorf("failed to get benchmark config from environment variables")
 		}
 	}
 	// Read the Mako config file for this environment.
--- a/test/mako/config/benchmark.go
+++ b/test/mako/config/benchmark.go
@@ -41,10 +41,10 @@ func MustGetBenchmark() (*string, *string) {
 }
 
 type BenchmarkEnv struct {
-	DevKey   string `envconfig:"BENCHMARK_KEY_DEV"`
-	DevName  string `envconfig:"BENCHMARK_NAME_DEV"`
-	ProdKey  string `envconfig:"BENCHMARK_KEY_PROD"`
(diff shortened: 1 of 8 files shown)


grac3gao:
/ assign @grantr

grac3gao:
ready for review

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+
 // getBenchmark fetches the appropriate benchmark_key for this configured environment.
 func getBenchmark() (*string, *string, error) {
 	// Figure out what environment we're running in from the Mako configmap.
 	env, err := getEnvironment()
 	if err != nil {
 		return nil, nil, err
 	}
+	// If pod has benchmark related env
+	benchmarkEnv := BenchmarkEnv{}
```
chizhg: It seems this piece of code should only be executed if the benchmark config files are not added in `kodata`?

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+	if err := envconfig.Process("", &benchmarkEnv); err != nil {
+		return nil, nil, err
+	}
+	if !reflect.DeepEqual(benchmarkEnv, BenchmarkEnv{}) {
+		if env == "dev" {
+			if len(benchmarkEnv.DevKey) != 0 && len(benchmarkEnv.DevName) != 0 {
```
chizhg: `benchmarkEnv.DevKey != "" && benchmarkEnv.DevName != ""` is more intuitive for checking empty strings.

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+				return &benchmarkEnv.DevKey, &benchmarkEnv.DevName, nil
+			}
+			return nil, nil, fmt.Errorf(("not enough env in aggregator"))
```
chizhg: Please change the error message to something like `failed to get benchmark config from environment variables`, since this function is used by multiple projects and only `eventing` has `aggregator`.
Also there should only be one pair of brackets :)

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+		} else if env == "prod" {
+			if len(benchmarkEnv.ProdKey) != 0 && len(benchmarkEnv.ProdName) != 0 {
+				return &benchmarkEnv.ProdKey, &benchmarkEnv.ProdName, nil
+			}
+			return nil, nil, fmt.Errorf(("not enough env in aggregator"))
```
chizhg: Ditto.

test/mako/config/benchmark.go
grac3gao: This piece of code will be executed only if environment var defined in struct BenchmarkEnv{} are added under the  aggregator container's spec.env in eventing. If these environment var are not presented, envconfig.Process() will return an empty BenchmarkEnv{} struct, and won't go through this piece of code.

grac3gao:
ready for new round of review

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+	benchmarkEnv := BenchmarkEnv{}
+	if err := envconfig.Process("", &benchmarkEnv); err != nil {
+		return nil, nil, err
+	}
+	if !reflect.DeepEqual(benchmarkEnv, BenchmarkEnv{}) {
+		if env == "dev" {
+			if benchmarkEnv.DevKey != "" && benchmarkEnv.DevName != "" {
+				return &benchmarkEnv.DevKey, &benchmarkEnv.DevName, nil
+			}
+			return nil, nil, fmt.Errorf("failed to get benchmark config from environment variables")
```
chaodaiG:
```suggestion
			return nil, nil, errors.New("failed to get benchmark config from environment variables")
```

nit.  Call `errors.New` directly since `fmt.Errorf` is wrapping `errors.New` when there is no formatting needs

test/mako/config/benchmark.go
```diff
@@ -38,13 +40,38 @@ func MustGetBenchmark() (*string, *string) {
...
+		} else if env == "prod" {
+			if benchmarkEnv.ProdKey != "" && benchmarkEnv.ProdName != "" {
+				return &benchmarkEnv.ProdKey, &benchmarkEnv.ProdName, nil
+			}
+			return nil, nil, fmt.Errorf("failed to get benchmark config from environment variables")
```
chaodaiG:
```suggestion
			return nil, nil, errors.New("failed to get benchmark config from environment variables")
```

ditto

grac3gao:
New change added after code review. Ready to review.

test/mako/config/benchmark.go
```diff
@@ -38,13 +41,38 @@ func MustGetBenchmark() (*string, *string) {
 	return benchmarkKey, benchmarkName
 }
 
+type BenchmarkEnv struct {
```
evankanderson: Why do you need separate dev+prod benchmark data?

It seems like you'd want to set the key/name based on dev/prod outside the container, so the container doesn't need to know about "dev" vs "prod" (so that adding "release" later, for example, requires no code changes).

test/mako/config/benchmark.go
```diff
@@ -38,13 +41,38 @@ func MustGetBenchmark() (*string, *string) {
...
+	benchmarkEnv := BenchmarkEnv{}
+	if err := envconfig.Process("", &benchmarkEnv); err != nil {
+		return nil, nil, err
+	}
+	if !reflect.DeepEqual(benchmarkEnv, BenchmarkEnv{}) {
+		if env == "dev" {
+			if benchmarkEnv.DevKey != "" && benchmarkEnv.DevName != "" {
+				return &benchmarkEnv.DevKey, &benchmarkEnv.DevName, nil
+			}
+			return nil, nil, errors.New("failed to get benchmark config from environment variables")
```
evankanderson: I'm not a fan of this. It seems like you could end up with the following:

(empty config file)

> Crashes if only PROD_BENCHMARK_NAME and PROD_BENCHMARK_KEY are set.

> Continues if only DEV_BENCHMARK_NAME and DEV_BENCHMARK_KEY are set.

> Continues if both DEV_BENCHMARK_NAME/KEY and PROD_BENCHMARK_NAME/KEY are set, but always uses the "dev" value.


My $0.02 would be to move this from files in the container via `kodata` to environment variables in most/all cases.

evankanderson: Additionally, if you use only a single set of env vars, you could name them "Name" and "Key", and use "benchmark" as the first argument to `envconfig.Process()`.

evankanderson: I think that would make the error handling here nicer, too.

test/mako/config/benchmark.go
grac3gao: Currently, we have two benchmark for one perf test (dev and prod). When doing a perf test, it will install a configMap in the corresponding namespace to store the environment value (`dev` or `prod`) in configMap’s YAML file. 

Before my code change, this value is used in this function by `env, err := getEnvironment())` 
to check whether the container should publish data to dev benchmark or prod benchmark (by looking at benchmark_id from `env + “.config”`)

In the piece of code I added, I utilize this `env` variable to check whether I need to use `DEV_BENCHMARK_*` or `PROD_BENCHMARK_*` environment variable in the container.

test/mako/config/benchmark.go
grac3gao: Would it be possible to give me more explanation for the single set of environment vars in container? 

I use the `env` variable got from configMap to check whether I need to use `DEV_BENCHMARK_*` values or `PROD_BENCHMARK_*` values in this part of code. It will go for the following scenario:
1. `env` is not set in configMap (error is handled in `env, err := getEnvironment())` at the beginning of this function)
2. `env` is `dev` and there is at least one `*_BENCHMARK_*` value set in the container
    	(1).`DEV_BENCHMARK_NAME` and `DEV_BENCHMARK_KEY` are set, return `DEV_BENCHMARK_NAME/KEY` values
        (2). Only one value from `DEV_BENCHMARK_NAME` and `DEV_BENCHMARK_KEY` is set, crash
        (3).`DEV_BENCHMARK_NAME/KEY` are not set, crash
3. `env` is `prod` and there is at least one `*_BENCHMARK_*` value set in the container
      ...logic is the same as `env` is `dev`
4. `env` is other value (continue, error is handled in the later part of this function)

evankanderson: IMO, we should either use environment variables or file contents; mixing the two is a bit strange.

If you want to use the ConfigMap with env vars, you can reflect ConfigMap data into environment vars:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-container-environment-variables-using-configmap-data

Alternatively, you could extract the name and key from the config file if present, rather than extracting it from env vars.

grac3gao:
close this PR now, need to change the method.

PR #746, opened 2019-10-04
Handle and retry 'connection reset by peer' errors

JRBANCEL:
Some E2E are flaky because we do not handle the transient `connection reset by peer` errors.
Examples: [[1]](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.3-mesh/1179951947675865088), [[2]](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.2-mesh/1179919241718009856), [[3]](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-istio-1.2-mesh/1179979633756475392).

These errors are expected and should simply be retried.

### DIFF
--- a/test/spoof/error_checks_test.go
+++ b/test/spoof/error_checks_test.go
@@ -19,6 +19,7 @@ limitations under the License.
 package spoof
 
 import (
+	"errors"
 	"net/http"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


test/spoof/error_checks_test.go
```diff
@@ -57,31 +58,53 @@ func TestDNSError(t *testing.T) {
...
+		})
+	}
+}
+
+func TestConnectionReset(t *testing.T) {
+	for _, tt := range []struct {
+		name      string
+		err       error
+		connReset bool
+	}{{
```
chaodaiG: missing a testcase of nil error

test/spoof/error_checks_test.go
```diff
@@ -57,31 +58,53 @@ func TestDNSError(t *testing.T) {
...
-			if tcpRefuse := isTCPConnectRefuse(err); tt.tcpRefuse != tcpRefuse {
-				t.Errorf("Expected tcpRefuse=%v, got %v", tt.tcpRefuse, tcpRefuse)
+			if connRefused := isConnectionRefused(err); tt.connRefused != connRefused {
+				t.Errorf("Expected connRefused=%v, got %v", tt.connRefused, connRefused)
+			}
+		})
+	}
+}
+
+func TestConnectionReset(t *testing.T) {
```
chaodaiG: Like this unit test better than the one above, IMHO the one above should also be something like this test

chaodaiG:
Great! Left a couple of nits, feel free ignore them
/lgtm
/approve

/hold
in case you want to deal with my nit

test/spoof/error_checks_test.go
JRBANCEL: Well, on one hand, the one above is doing more than just checking errors (using HttpClient) but at the same time it uses the real errors from Go library.

JRBANCEL:
I addressed the comments.
/hold cancel
Feel free to /lgtm if it looks good

PR #750, opened 2019-10-07
Move prow and gcs package from test-infra to pkg

chizhg:
As part of the work to get rid of the dependency cycle, move `prow` and `gcs` package from test-infra to pkg, since they are vendored in serving.

/cc @chaodaiG


chaodaiG:
/lgtm
Please be careful when removing these from test-infra, make sure we don't break release branches in other repos

/cc @evankanderson

PR #751, opened 2019-10-07
Spoof.New should fail when ResolveEndpoint fails

JRBANCEL:


chaodaiG:
Interesting, how did it get merged in the first place

/approve

PR #752, opened 2019-10-08
Support listing all clusters under a project'

chizhg:
For performance test cluster lifecycle management, we need to list all clusters under a project. Add the support in our library.

/cc @chaodaiG

### DIFF
--- a/testutils/gke/client.go
+++ b/testutils/gke/client.go
@@ -96,8 +96,8 @@ func (gsc *sdkClient) DeleteClusterAsync(project, region, zone, clusterName stri
 	if zone != "" {
 		return gsc.Projects.Zones.Clusters.Delete(project, location, clusterName).Context(context.Background()).Do()
 	}
-	opsFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
-	return gsc.Projects.Locations.Clusters.Delete(opsFullPath).Context(context.Background()).Do()
+	clusterFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
+	return gsc.Projects.Locations.Clusters.Delete(clusterFullPath).Context(context.Background()).Do()
 }
 
 // GetCluster gets the GKE cluster with the given cluster name.
@@ -106,15 +106,15 @@ func (gsc *sdkClient) GetCluster(project, region, zone, clusterName string) (*co
 	if zone != "" {
 		return gsc.Projects.Zones.Clusters.Get(project, location, clusterName).Context(context.Background()).Do()
 	}
-	opsFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
-	return gsc.Projects.Locations.Clusters.Get(opsFullPath).Context(context.Background()).Do()
+	clusterFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
+	return gsc.Projects.Locations.Clusters.Get(clusterFullPath).Context(context.Background()).Do()
 }
 
 // ListClustersInProject lists all the GKE clusters created in the given project.
 func (gsc *sdkClient) ListClustersInProject(project string) ([]*container.Cluster, error) {
 	var clusters []*container.Cluster
-	projectPath := fmt.Sprintf("projects/%s/locations/-", project)
-	resp, err := gsc.Projects.Locations.Clusters.List(projectPath).Do()
+	projectFullPath := fmt.Sprintf("projects/%s/locations/-", project)
+	resp, err := gsc.Projects.Locations.Clusters.List(projectFullPath).Do()
 	if err != nil {
 		return clusters, fmt.Errorf("failed to list clusters under project %s: %v", project, err)
 	}
--- a/testutils/gke/client.go
+++ b/testutils/gke/client.go
@@ -33,6 +33,7 @@ type SDKOperations interface {
 	DeleteClusterAsync(string, string, string, string) (*container.Operation, error)
 	GetCluster(string, string, string, string) (*container.Cluster, error)
 	GetOperation(string, string, string, string) (*container.Operation, error)
+	ListClustersInProject(string) ([]*container.Cluster, error)
 }
 
 // sdkClient Implement SDKOperations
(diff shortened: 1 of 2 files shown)


chizhg:
/hold
Let me test it first.

testutils/gke/client.go
```diff
@@ -109,12 +110,23 @@ func (gsc *sdkClient) GetCluster(project, region, zone, clusterName string) (*co
 	return gsc.Projects.Locations.Clusters.Get(clusterFullPath).Context(context.Background()).Do()
 }
 
+// ListClustersInProject lists all the GKE clusters created in the given project.
+func (gsc *sdkClient) ListClustersInProject(project string) ([]*container.Cluster, error) {
+	var clusters []*container.Cluster
```
vagababov: should this follow the pattern of
```
if zonal:
  list eveything in zone
else:
  list everything in project
```
?

testutils/gke/fake/client.go
```diff
@@ -172,6 +173,19 @@ func (fgsc *GKESDKClient) GetCluster(project, region, zone, cluster string) (*co
 	return nil, fmt.Errorf("cluster not found")
 }
 
+// ListClustersInProject lists all the GKE clusters created in the given project.
+func (fgsc *GKESDKClient) ListClustersInProject(project string) ([]*container.Cluster, error) {
+	allClusters := make([]*container.Cluster, 0)
```
vagababov: you can do `make(x, 0, len(fgsc.clusters))`

testutils/gke/client.go
chaodaiG: or maybe we can list both

testutils/gke/client.go
```diff
@@ -109,12 +110,23 @@ func (gsc *sdkClient) GetCluster(project, region, zone, clusterName string) (*co
...
+
 // GetOperation gets the operation ref with the given operation name.
 func (gsc *sdkClient) GetOperation(project, region, zone, opName string) (*container.Operation, error) {
 	location := GetClusterLocation(region, zone)
 	if zone != "" {
 		return gsc.Service.Projects.Zones.Operations.Get(project, location, opName).Do()
 	}
-	name := fmt.Sprintf("projects/%s/locations/%s/operations/%s", project, location, opName)
-	return gsc.Service.Projects.Locations.Operations.Get(name).Do()
+	clusterFullPath := fmt.Sprintf("projects/%s/locations/%s/operations/%s", project, location, opName)
```
chaodaiG: this should be named `opsFullPath` to be accurate

testutils/gke/client.go
chizhg: One difference between this function and others is `region` is not required. 

We can define the function as
```
func ListClusters(project, region, zone string) ([]*container.Cluster, error) {
  if zone != "" {
    return ...
  }
  
  if region != "" {
    return ...
  }

  return ...
}
```
But then if we want to get all clusters for this project, we'll have to pass two empty strings.

If you think this is better I can make the change.

testutils/gke/fake/client.go
chizhg: Hmm, we can not predetermine its length, because in the following for-loop, we need to append all elements in the slices to it.

testutils/gke/client.go
chizhg: Done.

testutils/gke/client.go
```diff
@@ -95,8 +96,8 @@ func (gsc *sdkClient) DeleteClusterAsync(project, region, zone, clusterName stri
 	if zone != "" {
 		return gsc.Projects.Zones.Clusters.Delete(project, location, clusterName).Context(context.Background()).Do()
 	}
-	parent := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
-	return gsc.Projects.Locations.Clusters.Delete(parent).Context(context.Background()).Do()
+	opsFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
```
chaodaiG: this one should be `clusterFullPath` still, based on the last word in it's path

testutils/gke/client.go
```diff
@@ -105,8 +106,19 @@ func (gsc *sdkClient) GetCluster(project, region, zone, clusterName string) (*co
 	if zone != "" {
 		return gsc.Projects.Zones.Clusters.Get(project, location, clusterName).Context(context.Background()).Do()
 	}
-	clusterFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
-	return gsc.Projects.Locations.Clusters.Get(clusterFullPath).Context(context.Background()).Do()
+	opsFullPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", project, location, clusterName)
```
chaodaiG: ditto

testutils/gke/client.go
```diff
@@ -105,8 +106,19 @@ func (gsc *sdkClient) GetCluster(project, region, zone, clusterName string) (*co
...
+	return gsc.Projects.Locations.Clusters.Get(opsFullPath).Context(context.Background()).Do()
+}
+
+// ListClustersInProject lists all the GKE clusters created in the given project.
+func (gsc *sdkClient) ListClustersInProject(project string) ([]*container.Cluster, error) {
+	var clusters []*container.Cluster
+	projectPath := fmt.Sprintf("projects/%s/locations/-", project)
+	resp, err := gsc.Projects.Locations.Clusters.List(projectPath).Do()
```
chaodaiG: so you are only listing regional clusters? Or this API returns both regional and zonal clusters?

testutils/gke/fake/client.go
```diff
@@ -172,6 +173,19 @@ func (fgsc *GKESDKClient) GetCluster(project, region, zone, cluster string) (*co
...
+	projectPath := fmt.Sprintf("projects/%s", project)
```
chaodaiG: to be consistent, `projectFullPath`

testutils/gke/client.go
chizhg: Done.

testutils/gke/client.go
chizhg: Done.

testutils/gke/fake/client.go
chizhg: Done.

testutils/gke/client.go
chizhg: I've tested it, it can also return zonal clusters.

chaodaiG: seems like the GKE API is a little messed up

testutils/gke/fake/client.go
chaodaiG: this one is not done yet

testutils/gke/client.go
chizhg: Ah yeah that's true. But only this path uses `-` to represent all `region`s and `zone`s, I guess  it's fine :)

chaodaiG:
/lgtm
/approve
The UI is somehow misleading

PR #753, opened 2019-10-08
Clean up websocket code a bit.

vagababov:
1. %q are very annoying when actually trying to read logs and the address is obvious (and zap quotes them)
2. there's no reason to have external error (and we were swallowing the timeout error, should it happen)

/assign @evankanderson @markusthoemmes

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -125,20 +125,20 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 		for {
 			select {
 			default:
-				logger.Infof("Connecting to %q", target)
+				logger.Infof("Connecting to %s", target)
 				if err := c.connect(); err != nil {
-					logger.Errorw(fmt.Sprintf("Connecting to %q failed", target), zap.Error(err))
+					logger.Errorw(fmt.Sprintf("Connecting to %s failed", target), zap.Error(err))
 					continue
 				}
-				logger.Infof("Connected to %q", target)
+				logger.Infof("Connected to %s", target)
 				if err := c.keepalive(); err != nil {
-					logger.Errorw(fmt.Sprintf("Connection to %q broke down, reconnecting...", target), zap.Error(err))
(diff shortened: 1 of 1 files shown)


websocket/connection.go
```diff
@@ -125,20 +125,20 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 		for {
 			select {
 			default:
-				logger.Infof("Connecting to %q", target)
+				logger.Infof("Connecting to %s", target)
 				if err := c.connect(); err != nil {
-					logger.Errorw(fmt.Sprintf("Connecting to %q failed", target), zap.Error(err))
+					logger.Errorw(fmt.Sprintf("Connecting to %s failed", target), zap.Error(err))
```
evankanderson: Slight preference for:

```golang
logger.With(zap.Error(err)).Errorf("Connecting to %s failed", target)
```

websocket/connection.go
```diff
@@ -125,20 +125,20 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
...
 					continue
 				}
-				logger.Infof("Connected to %q", target)
+				logger.Infof("Connected to %s", target)
 				if err := c.keepalive(); err != nil {
-					logger.Errorw(fmt.Sprintf("Connection to %q broke down, reconnecting...", target), zap.Error(err))
+					logger.Errorw(fmt.Sprintf("Connection to %s broke down, reconnecting...", target), zap.Error(err))
```
evankanderson: And here.

websocket/connection.go
```diff
@@ -125,20 +125,20 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 		for {
 			select {
 			default:
-				logger.Infof("Connecting to %q", target)
+				logger.Infof("Connecting to %s", target)
```
evankanderson: This seems spammy to have 3 of these logs in success cases. Convert one or two into `.Debugf`?

vagababov: Downgraded one that seemed redundant.

vagababov:
Done.

PR #754, opened 2019-10-08
Add OWNERS to the websocket which was missing before.

vagababov:
Alias it to the controller approvers which seem more or less reasonable

/assign @mattmoor @vaikas-google

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -46,6 +46,11 @@ aliases:
   - mdemirhan
   - yanweiguo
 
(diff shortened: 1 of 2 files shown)


evankanderson:
I'm wondering if this should be its own group, mostly because the websocket (and gRPC) APIs are a bit different than controllers in general.

/lgtm

vagababov:
Sure, can go that way as well.

vagababov:
Sounds good.
Added a more generic.

PR #757, opened 2019-10-11
Rename the ObjectReference field to Ref in Destination object

### LINKED ISSUE #756: Rename the objectReference field in Destination
## Expected Behavior
Currently we put all the fields at the top level of the Destination struct makes validation complex for the user. Similar concerns were expressed in https://github.com/knative/pkg/pull/561. The validation message [here](https://github.com/knative/pkg/blob/0a0c4972cd7134ce54b1ce6a6de8f189257a64fb/apis/v1alpha1/destination.go#L121-L123) considers the three fields `apiVersion`, `kind`, and `name` as a single object distinct from `uri`, but that's not reflected in the api structure. 

Something like this would be backward compatible with trigger and subscription:

```go
type Destination struct {
	Ref *corev1.ObjectReference `json:"ref,inline"`
	URI *apis.URL `json:"uri,omitempty"`
	Path *string `json:"path,omitempty"`
}
```

and is more obvious that Ref member fields are logically distinct from URI and Path fields.

capri-xiyue:
Fixes #756 
## Proposed Changes

- Rename the ObjectReference field to Ref in Destination object
-
-

**Release Note**

```release-note

```


apis/v1alpha1/destination.go
```diff
@@ -31,7 +31,7 @@ import (
 // Destination represents a target of an invocation over HTTP.
 type Destination struct {
 	// ObjectReference points to an Addressable.
-	*corev1.ObjectReference `json:",inline"`
+	Ref *corev1.ObjectReference `json:",inline"`
```
n3wscott: Why? Leave it anonymous.

grantr: You'll need to change the struct tag as well to get ref into the json:

```go
Ref *corev1.ObjectReference `json:"ref,omitempty"`
```

grantr:
This PR shouldn't go in until the discussion in https://github.com/knative/eventing/issues/1918 is resolved. Please add any thoughts about the API decision to that issue.

capri-xiyue:
In knative/eventing#1918, people have a discussion regarding the Destination API.  The conclusion is a new Destination API. Created another issue https://github.com/knative/pkg/issues/772 to change the Destination API. Will close this PR and relative issue.

PR #758, opened 2019-10-11
Add RunInformers function

skaslev:
Add RunInfomers which is similar to the StartInformers function but allows for
users to get notified when the informers have finished running.

This function will be mainly used in tests to fix the race described in
knative/serving#5351

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -445,6 +445,27 @@ func StartInformers(stopCh <-chan struct{}, informers ...Informer) error {
 	return nil
 }
 
+// RunInformers kicks off all of the passed informers and then waits for all of
+// them to synchronize. Returned function will wait for all informers to finish.
+func RunInformers(stopCh <-chan struct{}, informers ...Informer) (func(), error) {
+	var wg sync.WaitGroup
+	wg.Add(len(informers))
+	for _, informer := range informers {
+		informer := informer
+		go func() {
+			defer wg.Done()
+			informer.Run(stopCh)
+		}()
+	}
+
+	for i, informer := range informers {
+		if ok := cache.WaitForCacheSync(stopCh, informer.HasSynced); !ok {
+			return wg.Wait, fmt.Errorf("failed to wait for cache at index %d to sync", i)
+		}
+	}
+	return wg.Wait, nil
+}
+
 // StartAll kicks off all of the passed controllers with DefaultThreadsPerController.
 func StartAll(stopCh <-chan struct{}, controllers ...*Impl) {
 	wg := sync.WaitGroup{}
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -852,11 +852,19 @@ func checkStats(t *testing.T, r *FakeStatsReporter, reportCount, lastQueueDepth,
 type fixedInformer struct {
 	m    sync.Mutex
 	sunk bool
+	done bool
 }
 
 var _ Informer = (*fixedInformer)(nil)
 
-func (fi *fixedInformer) Run(<-chan struct{}) {}
+func (fi *fixedInformer) Run(stopCh <-chan struct{}) {
+	<-stopCh
+
+	fi.m.Lock()
+	defer fi.m.Unlock()
+	fi.done = true
+}
+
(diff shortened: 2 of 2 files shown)


skaslev:
See knative/serving#5795 for how this is intended to be used.

controller/controller.go
```diff
@@ -445,6 +445,34 @@ func StartInformers(stopCh <-chan struct{}, informers ...Informer) error {
 	return nil
 }
 
+// RunInformers kicks off all of the passed informers and then waits for all of
+// them to synchronize. Returned channel will get closed after all informers
+// finish running.
+func RunInformers(stopCh <-chan struct{}, informers ...Informer) (<-chan struct{}, error) {
```
vagababov: `RunWithWait/RunAndWait`?

controller/controller.go
```diff
@@ -445,6 +445,34 @@ func StartInformers(stopCh <-chan struct{}, informers ...Informer) error {
...
+		}()
+	}
+	go func() {
+		wg.Wait()
+		close(doneCh)
+	}()
+
+	for i, informer := range informers {
+		if ok := cache.WaitForCacheSync(stopCh, informer.HasSynced); !ok {
+			return doneCh, fmt.Errorf("failed to wait for cache at index %d to sync", i)
```
vagababov: do we want to close `doneCh` here?

controller/controller.go
skaslev: I pushed a new implementation using `errgroup.Group` before having a chance to see those comments. Sorry about that.

I don't have strong opinion about the name as long as this does unflake the data race.

controller/controller.go
```diff
@@ -445,6 +446,25 @@ func StartInformers(stopCh <-chan struct{}, informers ...Informer) error {
 	return nil
 }
 
+// RunInformers kicks off all of the passed informers in the given error group
+// and then waits for all of them to synchronize.
+func RunInformers(stopCh <-chan struct{}, grp *errgroup.Group, informers ...Informer) error {
```
vagababov: I liked previous approach better.
But the best would probably be to return a call back. Something like

```
func RI(stop, inf...) (cb, error) {
  var g errgroup.Group
  for i := range {
    i:=i
    grp.Go()
  }

  for ... {sync looooooop}
  return func() { wg.Wait()}, nil
```

Now the caller can do
```
ctx, cancel, inf := ....
cb, er := RunInformers(ctx.Done(), inf)
if err { cancel(); handleError }
defer func() {
  cancel()
  cb()
}
cancel()
```

skaslev:
> Looks good, but we need a test that checks that the returned callback is invoked.

I don't quite understand. The callback should be called by users of `RunInformers` and the test is a user that either calls or doesn't the callback depending on does it want wait. 

Do you mean making sure that `waitInformers` is called when used in such a way?

```
			ctx, cancel, _ := rtesting.SetupFakeContextWithCancel(t)
			waitInformers := func() {}
			defer func() {
				cancel()
				waitInformers()
			}()

			waitInformers, err := controller.RunInformers(ctx.Done(), endpointsInformer.Informer())
			if err != nil {
				t.Fatalf("Failed to start informers: %v", err)
			}
```

vagababov:
Yes, we need  to make sure it works (at least make sure we can call it and it won't hang).

skaslev:
@vagababov, I added a new test. Please take another look.

controller/controller.go
```diff
@@ -445,6 +445,27 @@ func StartInformers(stopCh <-chan struct{}, informers ...Informer) error {
...
+// RunInformers kicks off all of the passed informers and then waits for all of
+// them to synchronize. Returned function will wait for all informers to finish.
+func RunInformers(stopCh <-chan struct{}, informers ...Informer) (func(), error) {
+	var wg sync.WaitGroup
+	wg.Add(len(informers))
+	for _, informer := range informers {
+		informer := informer
+		go func() {
+			informer.Run(stopCh)
+			wg.Done()
```
vagababov:
```
defer wg.Done()
informer.Run(stopCh)
```

skaslev: Done.

PR #760, opened 2019-10-14
jsonTrace errors are retried until the timeout.

Harwayne:
Helps with knative/eventing#2040.

Changes:
* jsonTrace errors are retried until the timeout.
* jsonTrace json.Unmarshal errors include the string being unmarshaled.

### DIFF
--- a/test/zipkin/util.go
+++ b/test/zipkin/util.go
@@ -20,6 +20,7 @@ package zipkin
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"sync"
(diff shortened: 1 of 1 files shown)


test/zipkin/util.go
```diff
@@ -171,7 +175,7 @@ func jsonTrace(traceID string) ([]model.SpanModel, error) {
 	var models []model.SpanModel
 	err = json.Unmarshal(body, &models)
 	if err != nil {
-		return empty, err
+		return empty, fmt.Errorf("unmarshalling JSON, %q: %v", body, err)
```
chizhg:
```suggestion
		return empty, fmt.Errorf("got an error in unmarshalling JSON %q: %v", body, err)
```

Harwayne: Done, PTAL.

Although I go back and forth on this. In the future, I think I will try to follow https://github.com/uber-go/guide/blob/master/style.md#error-wrapping, where essentially I don't add the 'failed to' at every step and just let the final log line add the 'failed to' prefix.

chizhg: Thanks. That doc looks pretty useful BTW.

PR #762, opened 2019-10-14
Add a command package to run command line

chizhg:
Add a `cmd` package under `pkg/test` to run command lines.
It includes 3 functions - `Run`, `RunBatchSequentially` and `RunBatchParallel`
1. `Run` can be used to run a command in the similar way as we run from command line, e.g. `cmd.Run("kubectl get pods -n knative-serving")`, which is more understandable than `exec.Command("kubectl", "get", "pods", "-n", "knative-serving").Output()`
2. `RunBatchSequentially` will run a list of commands sequentially, it can be used in the scenario when we want to run a series of commands, and the latter commands rely on the previous commands to successfully finish.
3. `RunBatchParallel` will run a list of commands in parallel, it can be used in the scenario when we want to run a series of commands that are independent.

This is a part of the implementation of https://docs.google.com/document/d/1fA5v5qCv00K4Mheti45VpkacMUYaD1YOVVFUw0Nx3AE/edit#heading=h.8817xwi5yv44

/cc @chaodaiG 
/cc @adrcunha


chaodaiG:
@chizhg , can you elaborate the reasoning/usecases for implementing these methods?

chizhg:
> @chizhg , can you elaborate the reasoning/usecases for implementing these methods?

Done.

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+	"sync"
+
+	"knative.dev/pkg/test/helpers"
+)
+
+// Run will run the command and return the standard output, plus error if there is one.
+func Run(cmdLine string) (string, error) {
+	cmdSplit := strings.Fields(cmdLine)
+	if len(cmdSplit) == 0 {
+		return "", fmt.Errorf("the command line %q cannot be empty", cmdLine)
```
adrcunha: %q is unnecessary here, cmdSplit is empty.

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	cmd := cmdSplit[0]
+	args := cmdSplit[1:]
+	cmdOut, err := exec.Command(cmd, args...).Output()
```
adrcunha: You're swallowing stderr (actually, letting it flow to the terminal), you shouldn't.

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+
+// RunBatchSequentially will run the commands sequentially.
+// If there is an error when running a command, it will return directly with all standard output so far and the error.
+func RunBatchSequentially(cmdLines ...string) (string, error) {
+	var outputs []string
+	for _, cmdLine := range cmdLines {
+		output, err := Run(cmdLine)
+		outputs = append(outputs, output)
+		if err != nil {
+			return combineOutputs(outputs), fmt.Errorf("error happened when running %q: %v", cmdLine, err)
```
adrcunha:
```suggestion
			return combineOutputs(outputs), fmt.Errorf("error running %q: %v", cmdLine, err)
```

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+		outputs = append(outputs, output)
+	}
+	for err := range errCh {
+		errs = append(errs, err)
+	}
+
+	return combineOutputs(outputs), helpers.CombineErrors(errs)
+}
+
+// combineOutputs will combine the slice of output strings to a single string
```
adrcunha: You for
```suggestion
// combineOutputs will combine the slice of output strings into a single string, LF-separated
```

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+func combineOutputs(outputs []string) string {
+	var sb strings.Builder
+	for _, output := range outputs {
+		sb.WriteString(output)
+		sb.WriteRune('\n')
+	}
+	return strings.TrimRight(sb.String(), "\n")
+}
+
+// getErrorCode returns the exit code of the command run
```
adrcunha:
```suggestion
// getErrorCode extracts the exit code of an *ExitError type
```

test/cmd/error.go
```diff
@@ -0,0 +1,30 @@
...
+
+// CommandLineError is a custom error we use for errors got from running commands
+type CommandLineError struct {
+	Command     string
+	ErrorCode   int
+	ErrorOutput string
+}
+
+func (c CommandLineError) Error() string {
+	return fmt.Sprintf("[%d]running %q got error: %q", c.ErrorCode, c.Command, c.ErrorOutput)
```
adrcunha:
```suggestion
	return fmt.Sprintf("error %d when running %q: %q", c.ErrorCode, c.Command, c.ErrorOutput)
```

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
+/*
```
adrcunha: Tests, please. Here and everywhere.

test/cmd/command.go
vagababov: this is called `strings.Join()` call.

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+			defer wg.Done()
+			output, err := Run(cmdLine)
+			outputCh <- output
+			if err != nil {
+				errCh <- err
+			}
+		}()
+	}
+
+	go func() {
```
vagababov: It might be simpler 
```
errCh := make(chan error, len(cmdLines))
outputCh := make(chan string, len(cmdLines))
...
wg.Wait()
close(outputCh)
close(errCh)
for  o := range outputCh {}
for e := range errCh {}
```

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+		wg.Wait()
+		close(outputCh)
+		close(errCh)
+	}()
+
+	outputs := make([]string, 0)
```
vagababov:
```suggestion
	outputs := make([]string, 0, len(cmdLines))
```

test/cmd/command.go
```diff
@@ -0,0 +1,115 @@
...
+	errs := make([]error, 0)
```
vagababov:
```suggestion
	errs := make([]error, 0, len(cmdLines))
```

test/cmd/error.go
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import "fmt"
+
+// CommandLineError is a custom error we use for errors got from running commands
+type CommandLineError struct {
```
chaodaiG: I assume this is used for returning `error`, how easy is it to use the fields like `ErrorCode` etc.? Do we need to cast the error to `CommandlineError` every time using this?

test/cmd/command_test.go
```diff
@@ -0,0 +1,129 @@
...
+			t.Fatalf("Expect output in %v but actual is %q", c.possibleOutput, out)
+		}
+		if err == nil && c.shouldGetError {
+			t.Fatal("Expect to get an error but got nil")
+		}
+		if err != nil && !c.shouldGetError {
+			t.Fatalf("Got an error %v but should get nil", err)
+		}
+	}
+}
```
mattmoor: Add trailing newline:
```suggestion
}

```

test/cmd/command.go
chizhg: Fixed.

test/cmd/command.go
chizhg: Nice catch. Added the stderr.

test/cmd/command.go
chizhg: Added.

test/cmd/command.go
chizhg: Done.

test/cmd/command.go
chizhg: Done.

test/cmd/error.go
chizhg: Yes we'll need to cast the error to `CommandlineError` in order to get its fields. But I believe `ErrorCode` will not be used quite often, so it should be fine, and this is also how `exec.ExitError` works. I'm not directly using `exec.ExitError` since I want to add some other fields in `CommandlineError`.

chizhg:
Ping.

Sorry for the delay, can I get a second round of review for this PR?

test/helpers/error.go
```diff
@@ -28,10 +28,17 @@ func CombineErrors(errs []error) error {
...
 	for _, err := range errs {
-		sb.WriteString(err.Error())
-		sb.WriteString("\n")
+		if err != nil {
+			isErr = true
+			sb.WriteString(err.Error())
+			sb.WriteString("\n")
+		}
+	}
+	if !isErr {
```
adrcunha: Can't you just check `sb.Len()` instead of having a boolean flag?

test/cmd/command.go
```diff
@@ -0,0 +1,107 @@
...
+	for e := range errCh {
+		es = append(es, e)
+	}
+
+	return strings.Join(os, "\n"), helpers.CombineErrors(es)
+}
+
+// getErrorCode extracts the exit code of an *ExitError type
+func getErrorCode(err error) int {
+	errorCode := -1
```
adrcunha: I suggest defaulting to 1, the general error indication. You *will* return an error code, along with stdErr, so I think it's better signaling "an error occurred" than a specific unknown code (and anyways, -1 is not a valid value for exit codes anyway).

test/cmd/command_test.go
```diff
@@ -0,0 +1,130 @@
...
+		{
+			[]string{"echo 123", "", "echo 234"},
+			"123\n\n",
+			true,
+		},
+		{
+			[]string{"unknowncommand"},
+			"",
+			true,
+		},
```
adrcunha: nit 1: add the case `[]string{"unknowncommand", "echo 123"}` which should not return `123`.
nit 2: missing the return code test.

test/cmd/command_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func TestRunCommand(t *testing.T) {
+	testCases := []struct {
+		command string
+		expectedOutput string
+		shouldGetError bool
```
adrcunha: This should be the expected error code.

test/cmd/command_test.go
```diff
@@ -0,0 +1,130 @@
...
+	}{
```
adrcunha: Can we have a test for error output? That's an important piece of these helpers.

test/cmd/command_test.go
```diff
@@ -0,0 +1,130 @@
...
+		if err == nil && c.shouldGetError {
+			t.Fatal("Expect to get an error but got nil")
+		}
+		if err != nil && !c.shouldGetError {
+			t.Fatalf("Got an error %v but should get nil", err)
+		}
+	}
+}
+
+func TestRunCommandsInParallel(t *testing.T) {
```
adrcunha: Same nits from RunCommands.

test/cmd/command.go
```diff
@@ -0,0 +1,107 @@
...
+	args := cmdSplit[1:]
+	cmd := exec.Command(cmdName, args...)
+	var eb bytes.Buffer
+	cmd.Stderr = &eb
+
+	out, err := cmd.Output()
+	if err != nil {
+		errorCode := getErrorCode(err)
+		commandLineErr := CommandLineError{Command: cmdLine, ErrorOutput: eb.Bytes(), ErrorCode: errorCode}
+		return string(out), commandLineErr
```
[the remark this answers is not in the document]
chizhg: Done.

test/cmd/command.go
```diff
@@ -0,0 +1,107 @@
...
+	errCh := make(chan error, len(cmdLines))
+	outputCh := make(chan string, len(cmdLines))
+	wg := sync.WaitGroup{}
+	for i := range cmdLines {
+		cmdLine := cmdLines[i]
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			output, err := RunCommand(cmdLine)
+			outputCh <- output
```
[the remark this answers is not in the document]
chizhg: Good point. Added the lock.

test/helpers/error.go
chizhg: Changed to using an array to save the error messages, and check `if len(msgs) == 0`

test/cmd/command.go
chizhg: Changed to defaulting to 1

test/cmd/command_test.go
chizhg: 1. Added
2. Added the return code test

test/cmd/command_test.go
chizhg: Changed to `expectedErrorCode`.

test/cmd/command_test.go
chizhg: I feel not like the tests for error output, since it's tied to the OS. And it will be confusing if the tests pass on Prow and failed when running on local (a different OS)

test/cmd/command_test.go
chizhg: `RunCommandsInParallel` will return the combination of multiple errors if there are, so there will not be error code in the error it returns. Do you want me to change the behavior?

test/cmd/command.go
```diff
@@ -0,0 +1,107 @@
...
+
+// RunCommands will run the commands sequentially.
+// If there is an error when running a command, it will return directly with all standard output so far and the error.
+func RunCommands(cmdLines ...string) (string, error) {
+	var outputs []string
+	for _, cmdLine := range cmdLines {
+		output, err := RunCommand(cmdLine)
+		outputs = append(outputs, output)
+		if err != nil {
+			return strings.Join(outputs, "\n"), fmt.Errorf("error running %q: %v", cmdLine, err)
```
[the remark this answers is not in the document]
chizhg: I feel it redundant to change it that way, since the `outputs` array will be extended in the for loop. Or are you suggesting using a `StringBuilder` here?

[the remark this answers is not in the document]
adrcunha: Indeed, it's cleaner this way, nevermind my comment.

test/cmd/command_test.go
adrcunha: Error output is not tied to the OS, but to the command you run. Use standard error codes to take care of that; for example, running the nonexisting command "foobar" returns the error code 127, or you can exit a specific error code. For example, if you run "bash -c 'echo foo > /dev/stderr; exit 4'" you will get the error code 4 and just "foo" in stderr no matter the OS, Prow or local.

test/cmd/command_test.go
```diff
@@ -0,0 +1,150 @@
...
+func TestRunCommand(t *testing.T) {
+	testCases := []struct {
+		command string
+		expectedOutput string
+		expectedErrorCode int
+	}{
+		{"", "", 1},
+		{"   ", "", 1},
+		{"echo hello, world", "hello, world\n", 0},
+		{"unknowncommand", "", 1},
```
adrcunha: This looks fishy. I would expect 127 for an unknown binary.

test/cmd/command_test.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+		{
+			[]string{"echo 123", "", "echo 234"},
+			"123\n\n",
+			1,
+		},
+		{
+			[]string{"unknowncommand"},
+			"",
+			1,
```
adrcunha: Ditto regarding the return code value.

test/cmd/command_test.go
adrcunha: No, it makes sense returning them separate, but it would be good to test the output more strictly.

test/cmd/command_test.go
chizhg: Got it. Added the tests.
Thanks.

test/cmd/command_test.go
chizhg: Added tests for the error output.

test/cmd/command_test.go
chizhg: The `os/exec` library does not return `exec.ExitError` for unknown commands, so we return 1 as the default error code.

This test case is not that useful as it does not cover the `exec.ExitError`, changed to the command like `bash -c 'echo foo > /dev/stderr; exit 4'`

test/cmd/command_test.go
chizhg: Added the test case.

chizhg:
Thanks for the review! @adrcunha 

Since I introduced a new dependency
/assign @mattmoor

PR #763, opened 2019-10-15
Some refactors of the pkg/test code

chizhg:
Some refactors to the pkg/test and pkg/testutils code:
1. Move `gke` package from `pkg/testutils` to `pkg/test`
2. Consolidate the functions under `pkg/test/helpers` (we may need a better package name other than `helpers`, according to https://blog.golang.org/package-names, but it's pending discussion and it will be a breaking change)
3. Rename `pkg/testutils/clustermanager` to `pkg/testutils/prow/clustermanager` since there will be a `pkg/testutils/benchmark/clustermanager` package (open to discussion)

NOTE:
I feel it makes sense to leave `clustermanager` under a separate package other than `pkg/test`, since it is only used by test-infra 2.0, but we may need a more intuitive name than `testutils`.

/cc @chaodaiG 
/cc @adrcunha


test/helpers/name.go
```diff
@@ -17,14 +17,24 @@ limitations under the License.
...
 
 const (
+	letterBytes    = "abcdefghijklmnopqrstuvwxyz"
+	randSuffixLen  = 8
+	sep            = '-'
 	testNamePrefix = "Test"
 )
 
+func init() {
+	rand.Seed(time.Now().UTC().UnixNano())
```
adrcunha: Please document why you explicitly need to seed the random number generator.

test/helpers/name.go
```diff
@@ -34,3 +44,46 @@ func ObjectPrefixForTest(t *testing.T) string {
 func ObjectNameForTest(t *testing.T) string {
 	return AppendRandomString(ObjectPrefixForTest(t))
 }
+
+// AppendRandomString will generate a random string that begins with prefix.
+// This is useful if you want to make sure that your tests can run at the same
+// time against the same environment without conflicting.
+// This method will use "-" as the separator between the prefix and
+// the random suffix.
+// This method will seed rand with the current time when the package is initialized.
```
adrcunha: This is not true.

test/helpers/name.go
```diff
@@ -34,3 +44,46 @@ func ObjectPrefixForTest(t *testing.T) string {
...
+	for i := range suffix {
+		suffix[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+
+	return strings.Join([]string{prefix, string(suffix)}, string(sep))
+}
+
+// MakeK8sNamePrefix converts each chunk of non-alphanumeric character into a single dash
+// and also convert camelcase tokens into dash-delimited lowercase tokens.
+func MakeK8sNamePrefix(s string) string {
```
adrcunha: I've seen this methods before, and wanted to make sure they're not another set of duplicates.

test/helpers/name.go
```diff
@@ -17,14 +17,24 @@ limitations under the License.
 package helpers
 
 import (
+	"math/rand"
```
chaodaiG: I think it's recommended to use `https://golang.org/pkg/crypto/rand/`

test/helpers/name.go
chizhg: Done.

test/helpers/name.go
chizhg: I moved this function from the file I deleted, and did not make any change.
If what you meant was seed rand is NOT done by this method, I've removed this line, and added the comment as per your last comment.

test/helpers/name.go
chizhg: I'm pretty sure it's not, since initially it was I that moved this method to this package from `serving` :)

test/helpers/name.go
chizhg: I agree with it. But as per https://yourbasic.org/golang/crypto-rand-int/, since `crypto/rand` does not have `Seed` method, it's a bit verbose if we want to use it. So I still prefer `math/rand` here.

vagababov: They can be combined, but also for reasons we're doing here (generating random service names for tests) regular rand should suffice.

test/helpers/name.go
```diff
@@ -34,3 +45,45 @@ func ObjectPrefixForTest(t *testing.T) string {
...
+		sb.WriteRune(unicode.ToLower(c))
+		newToken = false
+	}
+	return sb.String()
+}
+
+// GetBaseFuncName returns the baseFuncName parsed from the fullFuncName.
+// eg. test/e2e.TestMain will return TestMain.
+func GetBaseFuncName(fullFuncName string) string {
+	baseFuncName := fullFuncName[strings.LastIndex(fullFuncName, "/")+1:]
```
vagababov: are we guaranteed a slash here?

test/helpers/name.go
chaodaiG: sgtm

test/helpers/name.go
adrcunha: Please document "why", not "how". "How" is clear from the code. What's the big deal with a pseudo random generator for the "appendsuffix" method? Actually I would even prefer that so reproducing scenarios would be easier.

chizhg:
> I'm not sure whether moving `clustermanager` under `prow` is the best idea,since there is already another package called `prow`

As per discussion, moved to `testutils/clustermanager/e2e-tests`, and the clustermanager for perf tests will be `testutils/clustermanager/perf-tests`

chaodaiG:
/lgtm
from my side

test/helpers/name.go
chizhg: It's not guaranteed, so I modified it a bit to make the logic clearer, does it look better to you?

adrcunha:
/hold

There are open comments.

test/helpers/name.go
chizhg: The comment on the method `AppendRandomString` explained the reason:
```
// AppendRandomString will generate a random string that begins with prefix.
// This is useful if you want to make sure that your tests can run at the same
// time against the same environment without conflicting.
```
Or do you prefer moving the comment here?

adrcunha: Initializing the random number generator is different from generating a random suffix. Please re-read my comment and reasoning.

test/helpers/name.go
vagababov: It's fine.

test/helpers/name.go
chizhg: Sorry for making this back and forth. 
Suppose there is a scenario as below:
One user was running the tests with `go test ./test/e2e`, and one of the tests created a service called `ksvc-xyzabc` with the `AppendRandomString` function. 
But the user made some new changes, and wanted to test again before the previous test finished. So he/she run `go test ./test/e2e` again. 
Without this `Seed`, the `AppendRandomString` function will create the same name as the previous one, which is `ksvc-xyzabc`, and conflicts with the last test, since the generator will create the same random sequence.
So this `Seed` is added for avoid this scenario.

I also added more comment. Does it make sense to you?

adrcunha: Thanks, now I understand the problem: if you don't seed the random number generator, rerunning a test will likely generate the same random names again, causing conflicts if the resources were not deleted.

I suggest clarifying your comment, though. Something like:
```
// Properly seed the random number generator so AppendRandomString() is actually random.
// Otherwise, rerunning tests will generate the same names afain, causing conflicts with
// already existing conflicts.
```

Also, I strongly suggest you dump the seed, so you can reproduce the name generation during debugging.

chizhg: Done.

chizhg:
Thanks!
/hold cancel

PR #765, opened 2019-10-16
Bump YAML library to latest release

matzew:
As per title, just to make sure we're not vulnerable to CVE-2019-11253. See kubernetes/kubernetes#83253.

It's unclear if we are even vulnerable against this at all. Doesn't hurt to bump/override the dep though.

**Release Note**

```release-note
NONE
```

/assign @mattmoor

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -832,12 +832,12 @@
   version = "v0.9.1"
 
 [[projects]]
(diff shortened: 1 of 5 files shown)


matzew:
related...

fixed in serving:
* https://github.com/knative/serving/pull/5810

updated in eventing (core):
* https://github.com/knative/eventing/pull/2032/files#diff-bd247e83efc3c45ae9e8c47233249f18R832

PR #769, opened 2019-10-17
Fix race condition when updating Stackdriver exporter

### LINKED ISSUE #768: Race condition when updating metrics Stackdriver exporter
It is possible for changes to Stackdriver-related fields in ConfigMap config-observability to be applied out of order to "knative.dev/pkg/metric"s Stackdriver exporter.

There is a small window of opportunity for a race condition between reading the current metrics config, checking if a new  Stackdriver exporter is required, and setting the value of the new Stackdriver exporter. 

## Expected Behavior
When multiple updates to ConfigMap config-observability arrive in a short time window, the changes in configuration are applied to the Stackdriver exporter in the order they are received.

## Actual Behavior
When multiple updates to ConfigMap config-observability arrive in a short time window, the changes are applied in a non-deterministic order due to race condition in metrics package.

1. Update config-observability twice very, very quickly
2. Check logs of a knative component that export metrics to see which change applied

## Additional Info
In reality, this is a theoretical bug because modifying config-observability is unlikely to happen often in succession and it would be unlikely to modify it twice fast enough to reproduce the issue.

anniefu:
Fixes #768 

Lock during entire process of reading ConfigMap "config-observability" and updating Stackdriver exporter.

Also, move functions in config.go that touch variables declared in exporter.go to exporter.go.

### DIFF
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -90,10 +90,10 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
 		return err
 	}
 
-	if isNewExporterRequiredHelper(newConfig, curMetricsConfig) {
+	if isNewExporterRequired(newConfig) {
 		logger.Info("Flushing the existing exporter before setting up the new exporter.")
 		flushExporterHelper(curMetricsExporter)
-		e, err := newMetricsExporterHelper(newConfig, curMetricsExporter, logger)
+		e, err := newMetricsExporter(newConfig, logger)
 		if err != nil {
 			logger.Errorf("Failed to update a new metrics exporter based on metric config %v. error: %v", newConfig, err)
 			return err
@@ -109,12 +109,9 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
 
 // isNewExporterRequired compares the non-nil newConfig against curMetricsConfig. When backend changes,
 // or stackdriver project ID changes for stackdriver backend, we need to update the metrics exporter.
+// This function is not implicitly thread-safe.
 func isNewExporterRequired(newConfig *metricsConfig) bool {
-	cc := getCurMetricsConfig()
-	return isNewExporterRequiredHelper(newConfig, cc)
-}
-
-func isNewExporterRequiredHelper(newConfig *metricsConfig, cc *metricsConfig) bool {
+	cc := curMetricsConfig
 	if cc == nil || newConfig.backendDestination != cc.backendDestination {
 		return true
 	}
@@ -123,12 +120,9 @@ func isNewExporterRequiredHelper(newConfig *metricsConfig, cc *metricsConfig) bo
 }
 
 // newMetricsExporter gets a metrics exporter based on the config.
+// This function is not implicitly thread-safe.
 func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
-	ce := getCurMetricsExporter()
-	return newMetricsExporterHelper(config, ce, logger)
-}
-
-func newMetricsExporterHelper(config *metricsConfig, ce view.Exporter, logger *zap.SugaredLogger) (view.Exporter, error) {
+	ce := curMetricsExporter
 	// If there is a Prometheus Exporter server running, stop it.
 	resetCurPromSrv()
 
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -95,7 +95,7 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
 
 	if isNewExporterRequired(newConfig) {
 		logger.Info("Flushing the existing exporter before setting up the new exporter.")
-		flushExporterHelper(curMetricsExporter)
+		flushExporterUnlocked(curMetricsExporter)
 		e, err := newMetricsExporter(newConfig, logger)
 		if err != nil {
 			logger.Errorf("Failed to update a new metrics exporter based on metric config %v. error: %v", newConfig, err)
@@ -194,10 +194,10 @@ func setCurMetricsConfigUnlocked(c *metricsConfig) {
 // Return value indicates whether the exporter is flushable or not.
 func FlushExporter() bool {
 	e := getCurMetricsExporter()
-	return flushExporterHelper(e)
+	return flushExporterUnlocked(e)
 }
 
-func flushExporterHelper(e view.Exporter) bool {
+func flushExporterUnlocked(e view.Exporter) bool {
 	if e == nil {
 		return false
 	}
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -25,7 +25,7 @@ import (
 var (
 	curMetricsExporter view.Exporter
 	curMetricsConfig   *metricsConfig
-	metricsMux         sync.Mutex
+	metricsMux         sync.RWMutex
 )
 
 type flushable interface {
@@ -75,6 +75,9 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
 }
 
 // UpdateExporter updates the exporter based on the given ExporterOptions.
+// This is a thread-safe function. The entire series of operations is locked
+// to prevent a race condition between reading the current configuration
+// and updating the current exporter.
 func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
 	metricsMux.Lock()
 	defer metricsMux.Unlock()
@@ -148,8 +151,8 @@ func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.
 }
 
 func getCurMetricsExporter() view.Exporter {
-	metricsMux.Lock()
-	defer metricsMux.Unlock()
+	metricsMux.RLock()
+	defer metricsMux.RUnlock()
 	return curMetricsExporter
 }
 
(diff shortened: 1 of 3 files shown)


metrics/exporter.go
```diff
@@ -32,11 +33,107 @@ type flushable interface {
...
+// or stackdriver project ID changes for stackdriver backend, we need to update the metrics exporter.
+func isNewExporterRequired(newConfig *metricsConfig) bool {
+	cc := getCurMetricsConfig()
+	return isNewExporterRequiredHelper(newConfig, cc)
+}
+
+func isNewExporterRequiredHelper(newConfig *metricsConfig, cc *metricsConfig) bool {
+	if cc == nil || newConfig.backendDestination != cc.backendDestination {
+		return true
+	} else if newConfig.backendDestination == Stackdriver && newConfig.stackdriverProjectID != cc.stackdriverProjectID {
```
vagababov: replace whole else if block with
```
if ... { return true }
return a== b && c != d
```

Or just completely flatten it out into
`return a==b  || c != d || e ==f || g != h`.

metrics/exporter.go
```diff
@@ -32,11 +33,107 @@ type flushable interface {
...
+		return true
+	}
+
+	return false
+}
+
 // newMetricsExporter gets a metrics exporter based on the config.
 func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
+	ce := getCurMetricsExporter()
+	return newMetricsExporterHelper(config, ce, logger)
```
vagababov: ditto.

metrics/exporter.go
```diff
@@ -32,11 +33,107 @@ type flushable interface {
...
+	// ConfigMap is the data from config map config-observability. Must be present.
+	// See https://github.com/knative/serving/blob/master/config/config-observability.yaml
+	// for details.
+	ConfigMap map[string]string
+}
+
+// UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
+// when a config map is updated.
+func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	domain := Domain()
```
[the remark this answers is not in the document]
yanweiguo: Required by Stackdriver. Full SD metric name is <domain>/<component>/<metric name>

metrics/exporter.go
```diff
@@ -32,11 +33,107 @@ type flushable interface {
...
+	}
+
+	setCurMetricsConfigUnlocked(newConfig)
+	return nil
+}
+
+// isNewExporterRequired compares the non-nil newConfig against curMetricsConfig. When backend changes,
+// or stackdriver project ID changes for stackdriver backend, we need to update the metrics exporter.
+func isNewExporterRequired(newConfig *metricsConfig) bool {
+	cc := getCurMetricsConfig()
```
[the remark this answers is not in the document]
yanweiguo: Sorry I don't under standard this comment. What do you mean?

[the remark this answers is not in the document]
anniefu: isNewExporterRequired existed before this PR and is used by many of the metrics test. getCurMetricsConfig() here uses a lock, so to make sure all the tests passed w/o modification, I left the behavior of isNewExporterRequired untouched.

metrics/exporter.go
anniefu: Created for the same reason as isNewExporterRequired (to ensure tests pass).

metrics/exporter.go
anniefu: Pending

metrics/exporter.go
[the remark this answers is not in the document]
vagababov: So why just putting directly in the closure won't work?

metrics/exporter.go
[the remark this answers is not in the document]
vagababov: I can see this as a temporal, but this has to be cleaned up.

metrics/exporter.go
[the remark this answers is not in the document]
anniefu: Sure, I can do that.

metrics/exporter.go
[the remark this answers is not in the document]
anniefu: What would be your ideal form of cleanup? I could see moving the functions only used by tests to a separate file.

I could also remove the functions only used by tests completely and remove tests designed to test those functions.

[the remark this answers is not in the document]
vagababov: Basically having a function whose only purpose is to just call another function with a param seems like overhead. I'd like it to go away :)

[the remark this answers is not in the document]
anniefu: Ok, I've consolidated the two half-duplicated functions to just one unlocked function each. This is a change in existing behavior because they are no longer thread-safe. I've added a line in the function comment to point that out.

Tests are still fine because they are single-threaded.

metrics/exporter.go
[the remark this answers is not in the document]
anniefu: It looks like Domain() is dynamically reading an environment variable, so the value at function closure variable binding time could potentially be different from function closure execution time.

I'm going to leave it as-is to preserve existing behavior and avoid any unintended side effects.

metrics/exporter.go
```diff
@@ -80,6 +173,10 @@ func getCurMetricsConfig() *metricsConfig {
 func setCurMetricsConfig(c *metricsConfig) {
 	metricsMux.Lock()
 	defer metricsMux.Unlock()
+	setCurMetricsConfigUnlocked(c)
+}
+
+func setCurMetricsConfigUnlocked(c *metricsConfig) {
```
yanweiguo: How about changing this one to `setCurMetricsConfig` and the above one to `setCurMetricsConfigForTesting` and move it to a test file since it is only used in test now?

metrics/exporter.go
```diff
@@ -67,6 +156,10 @@ func getCurMetricsExporter() view.Exporter {
 func setCurMetricsExporter(e view.Exporter) {
 	metricsMux.Lock()
 	defer metricsMux.Unlock()
+	setCurMetricsExporterUnlocked(e)
+}
+
+func setCurMetricsExporterUnlocked(e view.Exporter) {
```
yanweiguo: Same for the name here.

metrics/exporter.go
```diff
@@ -32,11 +33,99 @@ type flushable interface {
...
+			Domain:    domain,
+			Component: component,
+			ConfigMap: configMap.Data,
+		}, logger)
+	}
+}
+
+// UpdateExporter updates the exporter based on the given ExporterOptions.
+func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
+	metricsMux.Lock()
```
yanweiguo: Could you add the comment here to explain why we lock the whole function?

metrics/exporter.go
```diff
@@ -94,6 +191,10 @@ func setCurMetricsConfig(c *metricsConfig) {
 // Return value indicates whether the exporter is flushable or not.
 func FlushExporter() bool {
 	e := getCurMetricsExporter()
+	return flushExporterHelper(e)
```
yanweiguo: Could we change the lock in `getCurMetricsExporter` to read lock and not split this func? Having two is confusing. Since `UpdateExporterFromConfigMap` is not frequently used, adding a read lock won't cost too much.

anniefu: I'm not sure we can do that because a nested read lock request in UpdateExporter will still cause deadlock since writers still block out all readers.

metrics/exporter.go
anniefu: Pending

metrics/exporter.go
anniefu: It might be best if we just leave the existing functions as locked functions, and introduce these new unlocked versions for the exporter's use.

We have some other non-testing components in the metrics package that are using the existing getCurMetricsConfig(). Those calls do require the lock.

Only tests use the setter functions, but it would be very strange to have getCurMetricsConfig() use a lock, but then have setCurMetricsConfig() not.

metrics/exporter.go
anniefu: I agree a reader-writer lock is more appropriate here though, so I will change the mutex to be a RWMutex though and have the getters just use RLock() and setters use Lock().

It's just for FlushExporter we still need a completely unlocked helper.

metrics/exporter.go
```diff
@@ -94,6 +194,10 @@ func setCurMetricsConfig(c *metricsConfig) {
...
+}
+
+func flushExporterHelper(e view.Exporter) bool {
```
yanweiguo: Change to `flushExporterUnlocked` to keep consistent?

metrics/exporter.go
yanweiguo: To me it's more strange to add testing only functions in non-test files. @vagababov  WDYT?

metrics/exporter.go
anniefu: I can do that. I was using "Helper" to indicate the function stub change, but Unlocked works too.

metrics/exporter.go
anniefu: Since there are existing non-test components that expect thread-safe getCurMetricsConfig, I don't think we can make setCurMetricsConfig unlocked. It would be very confusing for anyone trying to use that function or read the code in the future to reason about thread safety.

I think it's common to implement Getters/Setters in pairs for API completeness, even if they're not both used immediately.

But I can move setCurMetricsConfig to a separate test_utils file for now to make it clear that it's only used in tests?

anniefu: I'm going to leave all the default get/set functions as locked to prevent future confusion.

PR #770, opened 2019-10-17
Name the parameters of functions in GKE SDK interface

chizhg:
The parameters defined in functions of an interface should be named, otherwise users will not know which values to provide if calling the function with an interface.

/cc @chaodaiG 
/cc @adrcunha

### DIFF
--- a/test/gke/client.go
+++ b/test/gke/client.go
@@ -27,13 +27,13 @@ import (
 
 // SDKOperations wraps GKE SDK related functions
 type SDKOperations interface {
-	CreateCluster(string, string, string, *container.CreateClusterRequest) error
-	CreateClusterAsync(string, string, string, *container.CreateClusterRequest) (*container.Operation, error)
-	DeleteCluster(string, string, string, string) error
-	DeleteClusterAsync(string, string, string, string) (*container.Operation, error)
-	GetCluster(string, string, string, string) (*container.Cluster, error)
-	GetOperation(string, string, string, string) (*container.Operation, error)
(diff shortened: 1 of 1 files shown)


vagababov:
Interface is a user part of the API in the Go, usually, which means _users_ know the parameters.
It might be that design of the library API is improper...

chizhg:
> Interface is a user part of the API in the Go, usually, which means _users_ know the parameters.
> It might be that design of the library API is improper...

The real problem for me is, when I write the code and use these functions from my IDE, the autocompletion cannot tell me what values I need to provide for the parameters, e.g. `DeleteCluster(string, string, string, string) error`. Without checking the actual implementation, it's very easy to get confused with the order 😅

chaodaiG:
This looks good to me, but I would also like to learn more about what's
recommended practice for interface in go from Victor

chizhg:
Ping.
Is there any other concern for this PR?

vagababov:
Not from me :)

chizhg:
> In general Go best practices suggest that `return types accept interfaces`, which in effect means that interface is a client side construct, wherewith you specify what you want _to accept_. Which also means that, if e.g. you need just 1 method from a type, you can specify the interface that way in your client side app and unit testing becomes much easier, since your mocks need to implement one method (and what's better they don't even have to know that the actual type has 10 more methods).
> 
> Which brings to my comment that you usually have interface in the client side, which means it's localized and you kind of know the params :-)

That is great to learn. Thanks :-)

PR #774, opened 2019-10-17
change Destination api and addressable resolver, fixed unit tests and…

### LINKED ISSUE #772: Change Destination API and modify addressable resolver based on the new Destination API
## Expected Behavior
In https://github.com/knative/eventing/issues/1918, people have a discussion regarding the Destination API. The conclusion for that discussion is as follows:
```go
type Destination struct{
  // +optional
  Ref *corev1.ObjectReference
  // +optional
  URI *apis.URL
}
```
Validation rules for Destination:
- If Ref is absent and URI is absent, the Destination is invalid.
- If Ref is present and URI is absolute (`URL.IsAbs()` returns true), the Destination is invalid.
- If Ref is absent and URI is relative (`URL.IsAbs()` returns false), the Destination is invalid.
- All other Destinations are valid.

Also need to modify the addressable resolver as follows:
Target URI resolution rules:

If URI field is absent, the target URI is the URI retrieved from Ref.
If URI field is present and absolute, the target URI is the value of the URI field.
If URI field is present and relative, the target URI is the result of calling URL.ResolveReference with the URI retrieved from Ref as the base URI and the value of the URI field as the URI reference.

capri-xiyue:
fix #772 

## Proposed Changes

- Change Destination api and addressable resolver, fixed unit tests and…
- Fixed broker unit tests
- Added unit tests for new business logic

**Release Note**

```release-note
None
```


capri-xiyue:
assign @grantr

capri-xiyue:
assign @grantr

resolver/addressable_resolver.go
```diff
@@ -65,20 +64,28 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
+			}
+			return url.URL().ResolveReference(dest.URI.URL()).String(), nil
+		}
+		return url.URL().String(), nil
 	}
 
 	if dest.URI != nil {
-		return extendPath(dest.URI.DeepCopy(), dest.Path).String(), nil
+		if !dest.URI.URL().IsAbs() || dest.URI.Host == "" {
+			return "", fmt.Errorf("invalid URI: %v", dest.URI.String())
```
grantr: The error string should be `URI is not absolute: %v` to tell the user exactly what's invalid about it.

resolver/addressable_resolver_test.go
```diff
@@ -64,97 +63,99 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
-			ObjectReference: getAddressableRef(),
-			Path:            ptr.String("/foo"),
+			Ref: getAddressableRef(),
+			URI: &apis.URL{
+			Path:   "/foo",
+		},
 		},
 		wantURI: addressableDNS + "/foo",
-	}, "object ref with path without leading slash": {
+	}, "ref with relative URI without leading slash": {
```
grantr: I'd love to see two additional tests that verify the expected behavior for Addressables with paths, both with a trailing slash like `example.com/foo/` and without like `example.com/foo`.

apis/v1alpha1/destination.go
```diff
@@ -18,132 +18,60 @@ package v1alpha1
...
-}
 
 func (current *Destination) Validate(ctx context.Context) *apis.FieldError {
 	if current != nil {
-		errs := validateDestination(*current).ViaField(apis.CurrentField)
-		if current.Path != nil {
-			errs = errs.Also(validateDestinationPath(*current.Path).ViaField("path"))
-		}
-		return errs
+		return validateDestination(*current).ViaField(apis.CurrentField)
```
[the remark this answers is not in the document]
capri-xiyue: `Validate(ctx context.Context)` implements  `Validatable` interface and receives `context.Context` as parameter. `ValidateDestination(dest Destination)` receives `Destination` as parameter. I found that in eventing, we will use this method `ValidateDestination(dest Destination)` to validate the subfields of `Trigger` in Trigger validation. We can't use `Validate(ctx context.Context)` to to validate the subfields of `Trigger` since the parameter doesn't match

> Why do we need `Validate` and `validateDestination`? Seems like these methods could be collapsed into just `Validate`. Any context @n3wscott?

apis/v1alpha1/destination.go
```diff
@@ -18,132 +18,60 @@ package v1alpha1
...
-func NewDestination(obj *corev1.ObjectReference, paths ...string) (*Destination, error) {
-	dest := &Destination{
-		ObjectReference: obj,
-	}
-	err := dest.AppendPath(paths...)
-	if err != nil {
-		return nil, err
-	}
-	return dest, nil
+func NewDestination(obj *corev1.ObjectReference) (*Destination, error) {
```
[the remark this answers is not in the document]
capri-xiyue: I doubled checked it. At first, the function is the helper function for `TestDestinationWithPath`. Now since we delete `Path` and delete `TestDestinationWithPath`. I think we can delete it. I also checked repo of knative-gcp, I don't find any usage of function `NewDestination`. I think it's safe to remove it.

apis/v1alpha1/destination.go
```diff
@@ -18,132 +18,60 @@ package v1alpha1
...
+	return &Destination{
+		Ref: obj,
+	}, nil
 }
 
 // NewDestinationURI constructs a Destination from a URI.
-func NewDestinationURI(uri *apis.URL, paths ...string) (*Destination, error) {
-	dest := &Destination{
+func NewDestinationURI(uri *apis.URL) (*Destination, error) {
```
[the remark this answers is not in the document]
capri-xiyue: Same as above

apis/v1alpha1/destination.go
```diff
@@ -18,132 +18,60 @@ package v1alpha1
...
 }
 
 func validateDestination(dest Destination) *apis.FieldError {
-	if dest.URI != nil {
-		if dest.ObjectReference != nil {
-			return apis.ErrMultipleOneOf("uri", "[apiVersion, kind, name]")
-		}
-		if dest.URI.Host == "" || dest.URI.Scheme == "" {
+	if dest.Ref == nil && dest.URI == nil {
+		return apis.ErrGeneric("expected at least one, got neither", "uri", "[apiVersion, kind, name]")
```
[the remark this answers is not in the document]
capri-xiyue: The error msg of `apis.ErrMissingOneOf("ref", "uri")` is `expected exactly one, got neither`.
But in this case, we need at least one instead of exactly one. `Ref` and `URI' can both exists if `URI` is a relative URI.

apis/v1alpha1/destination.go
```diff
@@ -18,132 +18,60 @@ package v1alpha1
...
+	}
+	if dest.Ref != nil && dest.URI != nil && dest.URI.URL().IsAbs() {
+		return apis.ErrGeneric("URI with absolute URL is not allowed when Ref exists", "uri", "[apiVersion, kind, name]")
+	}
+	if dest.Ref != nil && dest.URI == nil{
+		return validateDestinationRef(*dest.Ref)
+	}
+	if dest.Ref == nil && dest.URI != nil && (!dest.URI.URL().IsAbs() || dest.URI.Host == "") {
```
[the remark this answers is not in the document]
capri-xiyue: The implementation of isAbs is as follows:
`func (u *URL) IsAbs() bool {
	return u.Scheme != ""
}`
As long as the scheme is not empty, the URI is considered to be absolute. That's why I added another check for the host to be non-empty. I will add a comment to explain it.

apis/v1alpha1/destination.go
```diff
@@ -66,8 +51,9 @@ func ValidateDestination(dest Destination) *apis.FieldError {
 	if dest.Ref != nil && dest.URI == nil{
 		return validateDestinationRef(*dest.Ref)
 	}
+	// IsAbs check whether the URL has a non-empty scheme. Besides the non non-empty scheme, we also require dest.URI has a non-empty host
```
grantr: typo nit:`Besides the non-empty scheme`

apis/v1alpha1/destination.go
```diff
@@ -43,18 +43,18 @@ func (current *Destination) Validate(ctx context.Context) *apis.FieldError {
...
-	if dest.Ref != nil && dest.URI == nil{
-		return validateDestinationRef(*dest.Ref)
+		return apis.ErrGeneric("Absolute URI is not allowed when Ref is present", "ref", "uri")
 	}
 	// IsAbs check whether the URL has a non-empty scheme. Besides the non non-empty scheme, we also require dest.URI has a non-empty host
 	if dest.Ref == nil && dest.URI != nil && (!dest.URI.URL().IsAbs() || dest.URI.Host == "") {
-			return apis.ErrInvalidValue("URI with Relative URL is not allowed when Ref is absent", "uri")
+			return apis.ErrInvalidValue("Relative URI is not allowed when Ref is absent",  "uri")
 		}
+	if dest.Ref != nil && dest.URI == nil{
```
grantr: I think this if statement isn't necessary. There are 4 options for the state of these two fields:
1. ref nil, uri nil
1. ref not nil, uri not nil
1. ref nil, uri not nil
1. ref not nil, uri nil

We've already checked the first 3, so the only possibility at this point is 4, and there's no need for an if (or the `return nil` at the end)

capri-xiyue: If we remove the 4th and remove return nil. The logic doesn't work.
For example, with destination with only valid URI, Ref is absent.
It pass the first three check. When there is no 4th check and return nil, it will try to call 
`validateDestinationRef(*dest.Ref)`. However, the Ref of destination is nil,  we will have error 'invalid pointer found on stack'. The prerequisites to call `validateDestinationRef(*dest.Ref)` is the Ref of destination should be present

grantr: You're right, I missed that there were additional restrictions on cases 2 and 3. Ignore this comment!

vaikas-google:
/lgtm
/approve

Thanks for doing this!

PR #780, opened 2019-10-19
Do not override log level if component is undefined in logging config

antoineco:
### Problem description

Given a ConfigMap named "config-logging" with the following data:
```yaml
data:
  zap-logger-config: { "level": "debug" }
```

The log level of the main component is set to `info` regardless of the value configured in the Zap logger config.

### Reason

`config.LoggingLevel[name]` returns a default value of `int8(0)` if `name` is not in the `config.LoggingLevel` map. This value corresponds to `zapcore.InfoLevel`.

We must ensure that `name` is actually an element of the map before passing the override value to `config.NewLogger()`. Same thing inside `UpdateLevelFromConfigMap()`. 

### Proposed changes

- Set the override value to `""` by default, and use an actual log level _only_ if the user explicitly defined a log level for the given component.
- Ensure the atomic level is reset to the global log level when the component log level is deleted from the ConfigMap.
- (stretch) Polish logging tests

/kind bug

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -33,7 +33,12 @@ import (
 
 const ConfigMapNameEnv = "CONFIG_LOGGING_NAME"
 
-var zapLoggerConfig = "zap-logger-config"
+const (
+	loggerConfigKey    = "zap-logger-config"
+	fallbackLoggerName = "fallback-logger"
+)
+
+var emptyLoggerConfigError = errors.New("empty logger configuration")
 
 // NewLogger creates a logger with the supplied configuration.
 // In addition to the logger, it returns AtomicLevel that can
@@ -48,7 +53,7 @@ func NewLogger(configJSON string, levelOverride string, opts ...zap.Option) (*za
 	}
 
 	loggingCfg := zap.NewProductionConfig()
-	if len(levelOverride) > 0 {
+	if levelOverride != "" {
 		if level, err := levelFromString(levelOverride); err == nil {
 			loggingCfg.Level = zap.NewAtomicLevelAt(*level)
 		}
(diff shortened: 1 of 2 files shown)


logging/config.go
```diff
@@ -175,14 +193,31 @@ func levelFromString(level string) (*zapcore.Level, error) {
...
 		if err != nil {
 			logger.Errorw("Failed to parse the logging configmap. Previous config map will be used.", zap.Error(err))
 			return
 		}
 
-		level := loggingConfig.LoggingLevel[levelKey]
+		level, defined := config.LoggingLevel[levelKey]
+		if !defined {
+			// reset to global level
+
```
vagababov: remove empty line.

logging/config.go
```diff
@@ -175,14 +193,31 @@ func levelFromString(level string) (*zapcore.Level, error) {
...
+			loggingCfg, err := zapConfigFromJSON(config.LoggingConfig)
+			switch {
+			case err == emptyLoggerConfigError:
+				level = zap.NewAtomicLevel().Level()
+			case err != nil:
+				logger.With(zap.Error(err)).Errorf("Failed to parse logger configuration. "+
```
vagababov:
```suggestion
				logger.With(zap.Error(err)).Errorf("failed to parse logger configuration; "+
```

logging/config.go
```diff
@@ -175,14 +193,31 @@ func levelFromString(level string) (*zapcore.Level, error) {
...
+					"Previous log level retained for %v", levelKey)
```
vagababov:
```suggestion
					"previous log level retained for %v", levelKey)
```

logging/config.go
antoineco: Care to explain why? All other logged messages in this package start with a capital letter.

vagababov: Well, I guess you can fix the whole package.
From official golang page: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

antoineco: It’s a log message, not an error string. `log.Errorf` is not `fmt.Errorf`.

vagababov: 🤦‍♂ 
sorry about that :)

antoineco: No worries, thanks for the thorough review!

Issue #779, opened 2019-10-19
Metrics name prefix comes as controller name


hrishin:
## Expected Behavior
`sharedmain.Main` accept only one key `componentname` it uses the same name as a prefix for emitting the metrics on `/metrics` endpoint. In `Tekton` case we want to use a different name for the component and different prefix of metrics name. `sharedmain.Main` could accept the additional argument for to pass metrics prefix name 

## Actual Behavior
`sharedmain,Main` accepts only one [argument](https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#L93) and that is used as logging ket and metrics prefix key

/area API
/kind feature
/kind good-first-issue

PR #781, opened 2019-10-20
Allows passing metrics key name in sharedmain.Main()

hrishin:
The sharedmain.Main() accept only one key "component name" as logging key and it uses the
same key name as a prefix for emitting the metrics on /metrics endpoint. In the case of Tekton
we want to use a different name for the component and different prefix of metrics name.
sharedmain.Main could accept the additional argument to pass metrics prefix name.

This patch enhances sharedmain.Main() to accept additional parameter
as a metrics name.

Fixes https://github.com/knative/pkg/issues/779

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -91,11 +91,22 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
 }
 
 func Main(component string, ctors ...injection.ControllerConstructor) {
-	// Set up signals so we handle the first shutdown signal gracefully.
-	MainWithContext(signals.NewContext(), component, ctors...)
+	MainWithMetricName(component, "", ctors...)
 }
 
-func MainWithContext(ctx context.Context, component string, ctors ...injection.ControllerConstructor) {
+func MainWithMetricName(component, metricsName string, ctors ...injection.ControllerConstructor) {
+	if metricsName == "" {
+		metricsName = component
+	}
+	// Set up signals so we handle the first shutdown signal gracefully.
+	MainWithContext(signals.NewContext(), component, metricsName, ctors...)
+}
+
+func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, ctors ...injection.ControllerConstructor) {
+	main(ctx, component, component, cfg, ctors...)
+}
+
+func MainWithContext(ctx context.Context, component, metricsName string, ctors ...injection.ControllerConstructor) {
 	var (
 		masterURL  = flag.String("master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -90,12 +90,12 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
 	return logging.NewConfigFromConfigMap(loggingConfigMap)
 }
 
-func Main(component string, ctors ...injection.ControllerConstructor) {
+func Main(component, metricsName string, ctors ...injection.ControllerConstructor) {
```
vagababov: This is a backwards incompatible change.
What you probably should do instead is create a new `MainWithMetricName` and thread that through there.

hrishin: @vagababov yes you are right.

hrishin:
@vagababov thats valid question :thinking:  I thought `MainWithConfig` by by notion is a kind of auxiliary function (I guess its not).
However, it's better to keep it backward compatible as well (though increasing nesting may not look good in the longterm?).

vagababov:
Well we can just check eventing to see if they use it. I don't think serving does (but we can check as well).
If they don't, then we can just make it private, and stop adding more methods :)

hrishin:
@vagababov it's used in [plenty of places](https://github.com/search?q=org%3Aknative+MainWithConfig&type=Code) such as operator and controllers :sweat_smile:

vagababov:
Well I guess the alternative is to stage the change in all the repos and then do a fast one, since imploding number of constructors is not very useful :

hrishin:
> Well I guess the alternative is to stage the change in all the repos and then do a fast one

Do you mean to raise the PR ready? Could you elaborate on what exactly this means? :thinking: 

How about getting this patch in while marking unwanted constructor as deprecating? Then eventually migrate all other dependencies and deprecate those constructors?

vagababov:
I mean we can create pending PRs in all repos and then fast merge them all
in sequence.
Your suggestion might also might work.

hrishin:
> I mean we can create pending PRs in all repos and then fast merge them all
in sequence.

Alright @vagababov will get that work

mattmoor:
I worry about the ever-expanding arity of constructor overloads, especially when they take N strings.  Perhaps we could provide a way to attach a metric name to context and default it to the component name if omitted.  That way we can simply fallback on the variant that allows a context to be provided?

n3wscott:
+1 to main with context
-1 to adding to the signature.

hrishin:
@mattmoor @n3wscott makes sense. (I was away, procastinating)

Will try to get this in. :+1: Thank you for being patient here. :bowing_man:

vagababov:
Stale with merge conflicts and no activity for 5 months.
/close

PR #785, opened 2019-10-21
Add utility methods for api.URL

evankanderson:
In particular, this adds:

* Two constructors for common cases, one for HTTP and one for HTTPS URLs
* An `IsEmpty` method to automate the `nil`-or-empty checks

### DIFF
--- a/apis/url.go
+++ b/apis/url.go
@@ -103,6 +103,9 @@ func (u *URL) String() string {
 
 // URL returns the URL as a url.URL.
 func (u *URL) URL() *url.URL {
+	if u == nil {
+		return &url.URL{}
+	}
 	url := url.URL(*u)
 	return &url
 }
--- a/apis/url.go
+++ b/apis/url.go
@@ -41,6 +41,30 @@ func ParseURL(u string) (*URL, error) {
 	return (*URL)(pu), nil
 }
 
+// HTTP creates an http:// URL pointing to a known domain.
+func HTTP(domain string) *URL {
+	return &URL{
+		Scheme: "http",
+		Host:   domain,
+	}
+}
+
+// HTTPS creates an https:// URL pointing to a known domain.
+func HTTPS(domain string) *URL {
+	return &URL{
+		Scheme: "https",
+		Host:   domain,
+	}
+}
+
+// IsEmpty returns true if the URL is `nil` or represents an empty URL.
+func (u *URL) IsEmpty() bool {
+	if u == nil {
+		return true
+	}
+	return *u == URL{}
+}
+
 // MarshalJSON implements a custom json marshal method used when this type is
 // marshaled using json.Marshal.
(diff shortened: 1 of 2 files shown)


apis/url_test.go
```diff
@@ -71,6 +81,16 @@ func TestParseURL(t *testing.T) {
 				t.Fatalf("ParseURL() = %v, wanted error", got)
 			}
 
+			if tc.nonEmpty {
```
vagababov: can we just do `if tc.nonEmpty == got.IsEmpty()`, or which is better `nonEmpty -> wantEmpty` and then compare inequality?

evankanderson: My thought process that led to this code:

1. When I started, there were more "empty" than non-empty cases, and I wanted the default to be simpler. At this point, there are more non-empty cases, so I can change this to `wantEmpty`.

   It also looks like the "empty" and "empty string" tests are the same, so I may remove one.

2. I started by writing this as `if tc.nonEmpty == got.IsEmpty()`, but then realized that I wanted different error messages for the two cases. I'm happy to collapse this to a slightly less human error message and make the test shorter if you'd prefer.

apis/url.go
```diff
@@ -41,6 +41,30 @@ func ParseURL(u string) (*URL, error) {
...
+// HTTPS creates an https:// URL pointing to a known domain.
+func HTTPS(domain string) *URL {
+	return &URL{
+		Scheme: "https",
+		Host:   domain,
+	}
+}
+
+// IsEmpty returns true if the URL is `nil` or represents an empty URL.
+func (u *URL) IsEmpty() bool {
```
vaikas-google: how's is it possible for u to be nil below because you have to call through it? Am I confused again?

apis/url_test.go
vagababov: Yeah, it's usually pretty obvious when you say `isempty=%v, wanted: %v`.

apis/url.go
vagababov: You can call on a nil receiver (as long as you don't access any field members).

vaikas-google: Ack, thanks :)

apis/url_test.go
evankanderson: Updated. I also noticed a bug in `URL.URL()` in the `nil` case, so I updated the tests to catch it, and then fixed the bug.

evankanderson:
> The following is the coverage report on the affected files.
> Say `/test pull-knative-pkg-go-coverage` to re-run this coverage report
> 
> File	Old Coverage	New Coverage	Delta
> [apis/url.go](https://storage.cloud.google.com/knative-prow/pr-logs/pull/knative_pkg/785/pull-knative-pkg-go-coverage/1186422682212634624/artifacts/line-cov.html#file0)	91.3%	100.0%	8.7

Oh yeah!

PR #787, opened 2019-10-22
Add support for Stackdriver on any compute platform to metrics

anniefu:
Add ability to configure metrics package Stackdriver exporter from a Stackdriver-specific config map.

Currently, Stackdriver metrics rely on the Google Compute Engine metadata server, which is only accessible on GCE, to configure the Stackdriver client. This allows sending metrics to Stackdriver from anywhere.

Changes
- Add "knative.dev/stackdriver/config"  package to encapsulate necessary Stackdriver metadata to configure a Stackdriver client
- Modify metrics config package to be able to update from a Stackdriver ConfigMap
- Modify Stackdriver exporter to read Stackdriver metadata from metrics config first before automatically falling back to Google Application Credentials

The Stackdriver config is optional and best effort, nothing fails if the config is absent or incomplete. If the values are invalid, the metrics may fail at export time, but the user will get an error from Stackdriver indicating the invalid value.

Part of feature work for knative/serving#5777


anniefu:
Additional tests for stackdriver_exporter.go & stackdriver/config.go on the way. Just wanted to get the bulk of the change out there for review.

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +75,31 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
+		} else {
+			logger.Errorf("Failed to retrieve Secret for authenticating with Stackdriver: ", zap.Error(err))
+			// Automatically fall back on Google application default credentials
+		}
+	}
+
 	e, err := newStackdriverExporterFunc(stackdriver.Options{
-		ProjectID:               config.stackdriverProjectID,
+		ProjectID:               proj,
+		Location:                config.stackdriverConfig.GcpLocation,
```
yanweiguo: Could you please add a comment that if `config.stackdriverConfig.GcpLocation` is empty, automatically fall back on the location provided by the metadata server?

stackdriver/config/config.go
```diff
@@ -0,0 +1,128 @@
...
+	return kubeclient.CoreV1().Secrets(secretNamespace).Get(config.GcpSecretName, metav1.GetOptions{})
+}
+
+// ConvertSecretToExporterOption converts a Kubernetes Secret to an OpenCensus Stackdriver Exporter Option.
+func ConvertSecretToExporterOption(secret *v1.Secret) option.ClientOption {
+	return option.WithCredentialsJSON(secret.Data[secretDataFieldKey])
+}
+
+func ensureKubeclient() {
+	if kubeclient != nil {
```
yanweiguo: Do we need to add lock to make sure reading this var is multiple-threads safe?

stackdriver/config/config.go
```diff
@@ -0,0 +1,128 @@
...
+	GcpSecretName string
+}
+
+// NewStackdriverConfigFromConfigMap returns a Config for the given configmap
+func NewStackdriverConfigFromConfigMap(config *corev1.ConfigMap) (*Config, error) {
+	return NewStackdriverConfigFromMap(config.Data)
+}
+
+// NewStackdriverConfigFromMap returns a Config for the given map
+func NewStackdriverConfigFromMap(config map[string]string) (*Config, error) {
```
yanweiguo: I would expect unit tests for funcs in this file.

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +75,31 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
+	var co []option.ClientOption
+	if config.stackdriverConfig.GcpSecretName != "" {
+		secret, err := getStackdriverSecretFunc(&config.stackdriverConfig)
+
+		if err == nil {
+			co = []option.ClientOption{sdconfig.ConvertSecretToExporterOption(secret)}
+		} else {
+			logger.Errorf("Failed to retrieve Secret for authenticating with Stackdriver: ", zap.Error(err))
+			// Automatically fall back on Google application default credentials
+		}
```
yanweiguo: Tests for these changes?

metrics/exporter.go
```diff
@@ -82,17 +105,48 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
...
 	defer metricsMux.Unlock()
 
-	newConfig, err := createMetricsConfig(ops, logger)
-	if err != nil {
-		if curMetricsExporter == nil {
-			// Fail the process if there doesn't exist an exporter.
-			logger.Errorw("Failed to get a valid metrics config", zap.Error(err))
-		} else {
-			logger.Errorw("Failed to get a valid metrics config; Skip updating the metrics exporter", zap.Error(err))
+	if ops.StackdriverConfigMap != nil {
```
yanweiguo: Tests for these changes?

metrics/exporter.go
```diff
@@ -82,17 +105,48 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
...
+			}
+			return err
+		}
+
+		newConfig = nc
+	} else {
+		// Otherwise, build off copy of the current config.
+		// It is always true that curMetricsConfig is non-nil to reach here.
+		newConfig = new(metricsConfig)
+		*newConfig = *curMetricsConfig
```
yanweiguo: What's the reason for this `if-else` statement. Did you see some issue when `ops.ConfigMap` is null?

stackdriver/config/config.go
```diff
@@ -0,0 +1,128 @@
+/*
```
yanweiguo: See the new files are all about Stackdriver metrics, how about moving them to `metrics/stackdriver`?

anniefu: I have plans to also use the same config for tracing in the future.

stackdriver/config/config.go
anniefu: Yes, that's probably a good idea.

stackdriver/config/config.go
anniefu: On the way!

metrics/stackdriver_exporter.go
anniefu: On the way!

metrics/exporter.go
anniefu: This is the tests that were added to metrics/config_test.go

I placed the tests there instead of in exporter_test.go because they fit the pattern of the existing tests in config_test.go more than exporter_test.go

metrics/exporter.go
anniefu: The new Stackdriver configuration values alone are not enough to create a metrics exporter, so they should always be added on to an existing metricsConfig.

In the case where just the Stackdriver configuration values are changed (someone edits "config-stackdriver" but not "config-observability"), ops.ConfigMap will be null, but we want to overwrite the existing metrics config instead of creating a brand new one with default values for the non-Stackdriver configuration.

metrics/stackdriver_exporter.go

stackdriver/config/config.go
anniefu: stackdriver/config changes moved to #789

metrics/stackdriver_exporter.go
anniefu: added.

anniefu:
/hold 
To let the overlapping PR #789 for stackdriver/config pkg go through first. Please continue reviewing this PR ignoring any files under stackdriver/config.

metrics/exporter.go
```diff
@@ -82,17 +105,48 @@ func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
...
+				logger.Errorw("Failed to get a valid metrics config; Skip updating the metrics exporter", zap.Error(err))
+			}
+			return err
+		}
+
+		newConfig = nc
+	} else {
+		// Otherwise, build off copy of the current config.
+		// It is always true that curMetricsConfig is non-nil to reach here.
+		newConfig = new(metricsConfig)
```
yanweiguo: Is this line needed?

metrics/exporter.go
yanweiguo: How about 
```
newConfig := curMetricsConfig
if ops.ConfigMap != nil {
  // If there's an updated config map, build a new config from scratch.
  ...
}
```

to drop the else statement.

metrics/exporter_test.go
```diff
@@ -101,23 +103,139 @@ func TestMain(m *testing.M) {
...
-func TestMetricsExporter(t *testing.T) {
-	_, err := newMetricsExporter(&metricsConfig{
-		domain:               servingDomain,
-		component:            testComponent,
-		backendDestination:   "unsupported",
-		stackdriverProjectID: ""}, TestLogger(t))
-	if err == nil {
-		t.Errorf("Expected an error for unsupported backend %v", err)
+var (
+	newMetricExporterTests = []struct {
```
yanweiguo: nit: Well actually I prefer not making this var global if this is only used for one test. I know this is how we did in `config_test.go`.

metrics/stackdriver_exporter_test.go
```diff
@@ -132,12 +136,106 @@ var (
 		component:  "broker",
 		metricName: "unsupported",
 	}}
+
+	newStackdriverExporterTests = []struct {
```
yanweiguo: Same here.

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +75,32 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
 func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
-	gm := gcpMetadataFunc()
+	gm := getGCPMetadata(config)
 	mtf := getMetricTypeFunc(config.stackdriverMetricTypePrefix, config.stackdriverCustomMetricTypePrefix)
+
+	proj := config.stackdriverProjectID
+	if config.stackdriverConfig.ProjectID != "" {
+		proj = config.stackdriverConfig.ProjectID
+	}
+
```
yanweiguo: Why not use `gm.project` directly and get rid of this part?

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +75,32 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
+	if config.stackdriverConfig.GCPSecretName != "" {
+		secret, err := getStackdriverSecretFunc(&config.stackdriverConfig)
+
+		if err == nil {
+			co = []option.ClientOption{sdconfig.ConvertSecretToExporterOption(secret)}
+		} else {
+			logger.Errorf("Failed to retrieve Secret for authenticating with Stackdriver: ", zap.Error(err))
+			// Automatically fall back on Google application default credentials
+		}
+	}
```
yanweiguo: Could we move this part to `getGCPMetadata` as well?

metrics/stackdriver_exporter.go
anniefu: The if-clause here is to ensure "config-stackdriver" will take precedent over "config-observability" or "config-logging". (though admittedly it is a little hard to read).

I can move this logic to getGCPMetadata though.

metrics/stackdriver_exporter.go
anniefu: This is more using GCP metadata than getting it. It's also a bit weird to move it into getGCPMetadata because it interacts with Kubernetes. I will move it into its own getExporterClientOption helper function though.

metrics/stackdriver_exporter_test.go
anniefu: Pending

anniefu:
Closing in favor of #794

PR #788, opened 2019-10-22
Add prow-cluster-operation package

chaodaiG:
Add a prow-cluster-operation package as a wrapper of clustermanager/e2e-tests package, which can be invoked from command line. See https://github.com/knative/test-infra/pull/1460 for it's usage

For reviewers:
This PR is copied directly from https://github.com/knative/test-infra/pull/1460, the only changes are importing paths updated since `metahelper` was moved to pkg

/cc @adrcunha 
/cc @chizhg


chaodaiG:
Removed, thanks!

PR #789, opened 2019-10-22
Add pkg stackdriver/config for Stackdriver metadata

anniefu:
Add "knative.dev/pkg/stackdriver/config" pkg that encapsulates metadata required to create a Stackdriver client. This metadata is provided by default on Google Compute Engine (GCE) by the GCE metadata server. 

On other compute platforms, adding this config will allow manually specifying Stackdriver metadata so that metrics & traces can still be sent to Stackdriver by user's choice.

Part of knative/serving#5777

### DIFF
--- a/stackdriver/config/config.go
+++ b/stackdriver/config/config.go
@@ -30,6 +30,9 @@ import (
 
 var (
 	mtx sync.RWMutex
+
+	// kubeclient is the in-cluster Kubernetes kubeclient.
+	kubeclient *kubernetes.Clientset
 )
 
 const (
@@ -43,15 +46,15 @@ const (
 	// service account key to use.
 	secretDataFieldKey = "key.json"
 
-	// ProjectIDKey is the name of the ConfigMap field for ProjectID.
+	// ProjectIDKey is the map key for Config.ProjectID.
 	ProjectIDKey = "project-id"
-	// GcpLocationKey is name of the ConfigMap field for GcpLocation.
+	// GcpLocationKey is the map key for Config.GCPLocation.
 	GcpLocationKey = "gcp-location"
-	// ClusterNameKey is the name of the ConfigMap field for ClusterName.
+	// ClusterNameKey is the map key for Config.ClusterName.
 	ClusterNameKey = "cluster-name"
-	// GcpSecretNameKey is the name of the ConfigMap field for GcpSecretName.
+	// GcpSecretNameKey is the map key for GCPSecretName.
 	GcpSecretNameKey = "gcp-secret-name"
-	// GcpSecretNamespaceKey is the name of the ConfigMap field for GcpSecretNamespace.
+	// GcpSecretNamespaceKey is the map key for GCPSecretNamespace.
 	GcpSecretNamespaceKey = "gcp-secret-namespace"
 )
 
@@ -61,21 +64,21 @@ type Config struct {
 	// This is not necessarily the GCP project ID where the Kubernetes cluster is hosted.
 	// Required when the Kubernetes cluster is not hosted on GCE.
 	ProjectID string
-	// GcpLocation is the GCP region or zone to which data is uploaded.
+	// GCPLocation is the GCP region or zone to which data is uploaded.
 	// This is not necessarily the GCP project ID where the Kubernetes cluster is hosted.
 	// Required when the Kubernetes cluster is not hosted on GCE.
-	GcpLocation string
+	GCPLocation string
 	// ClusterName is the cluster name with which the data will be associated in Stackdriver.
 	// Required when the Kubernetes cluster is not hosted on GCE.
 	ClusterName string
-	// GcpSecretName is the optional GCP service account key which will be used to
+	// GCPSecretName is the optional GCP service account key which will be used to
 	// authenticate with Stackdriver. If not provided, Google Application Default Credentials
 	// will be used (https://cloud.google.com/docs/authentication/production).
-	GcpSecretName string
-	// GcpSecretNamespace is the Kubernetes namespace where GcpSecretName is located.
+	GCPSecretName string
+	// GCPSecretNamespace is the Kubernetes namespace where GcpSecretName is located.
 	// The Kubernetes ServiceAccount used by the pod that is exporting data to
 	// Stackdriver should have access to Secrets in this namespace.
-	GcpSecretNamespace string
+	GCPSecretNamespace string
 }
 
 // NewStackdriverConfigFromConfigMap returns a Config for the given configmap
@@ -95,7 +98,7 @@ func NewStackdriverConfigFromMap(config map[string]string) (*Config, error) {
 	}
 
 	if gl, ok := config[GcpLocationKey]; ok {
-		sc.GcpLocation = gl
+		sc.GCPLocation = gl
 	}
 
 	if cn, ok := config[ClusterNameKey]; ok {
@@ -103,30 +106,25 @@ func NewStackdriverConfigFromMap(config map[string]string) (*Config, error) {
 	}
 
 	if gsn, ok := config[GcpSecretNameKey]; ok {
-		sc.GcpSecretName = gsn
+		sc.GCPSecretName = gsn
 	}
 
 	if gsns, ok := config[GcpSecretNamespaceKey]; ok {
-		sc.GcpSecretNamespace = gsns
+		sc.GCPSecretNamespace = gsns
 	}
 
 	return sc, nil
 }
 
-var (
-	// kubeclient is the in-cluster Kubernetes kubeclient.
-	kubeclient *kubernetes.Clientset
-)
-
 // GetStackdriverSecret returns the Kubernetes Secret specified in the given config.
 func GetStackdriverSecret(config *Config) (*v1.Secret, error) {
 	ensureKubeclient()
 	ns := secretNamespaceDefault
-	if config.GcpSecretNamespace != "" {
-		ns = config.GcpSecretNamespace
+	if config.GCPSecretNamespace != "" {
+		ns = config.GCPSecretNamespace
 	}
 
-	return kubeclient.CoreV1().Secrets(ns).Get(config.GcpSecretName, metav1.GetOptions{})
+	return kubeclient.CoreV1().Secrets(ns).Get(config.GCPSecretName, metav1.GetOptions{})
 }
 
 // ConvertSecretToExporterOption converts a Kubernetes Secret to an OpenCensus Stackdriver Exporter Option.
--- a/stackdriver/config/config.go
+++ b/stackdriver/config/config.go
@@ -29,10 +29,10 @@ import (
 )
 
 var (
-	mtx sync.RWMutex
 
 	// kubeclient is the in-cluster Kubernetes kubeclient.
-	kubeclient *kubernetes.Clientset
+	kubeclient     *kubernetes.Clientset
+	initClientOnce sync.Once
 )
 
 const (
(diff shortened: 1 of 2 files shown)


stackdriver/config/config.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return sc, nil
+}
+
+var (
+	// kubeclient is the in-cluster Kubernetes kubeclient.
+	kubeclient *kubernetes.Clientset
+)
+
+// GetStackdriverSecret returns the Kubernetes Secret specified in the given config.
```
anniefu: These functions cannot be unit tested properly without a Kubernetes cluster.

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,100 @@
...
+				},
+			},
+			expectedConfig: Config{
+				ProjectID:   "project",
+				GcpLocation: "us-west1",
+				ClusterName: "cluster",
+			},
+		},
+		{
+			name: "invalidGcpLocation",
```
yanweiguo: What does this test case mean?

stackdriver/config/config.go
```diff
@@ -0,0 +1,162 @@
...
+	mtx sync.RWMutex
+)
+
+const (
+	// ConfigMapNameEnv is the name of the environment variable that contains the name of
+	// the ConfigMap for stackdriver configuration.
+	ConfigMapNameEnv = "CONFIG_STACKDRIVER_NAME"
+	// secretNamespaceDefault is the namespace in which secrets for authenticating with Stackdriver
+	// should be stored.
+	secretNamespaceDefault = "default"
```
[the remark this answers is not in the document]
anniefu: Yes, I made it configurable because I realized that knative/eventing will not be in the "knative-serving" namespace. When I make changes to knative/serving, I will set it to "knative-serving" there.

Extra configuration will also be useful if users want to add the Secret in their own namespaces instead of "knative-serving"

stackdriver/config/config.go
```diff
@@ -0,0 +1,162 @@
...
+	// secretDataFieldKey is key of the Kubernetes Secret Data field that contains the GCP
+	// service account key to use.
+	secretDataFieldKey = "key.json"
+
+	// ProjectIDKey is the name of the ConfigMap field for ProjectID.
```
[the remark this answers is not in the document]
anniefu: To clarify, you're referring to in the comment?

stackdriver/config/config.go
```diff
@@ -0,0 +1,151 @@
...
+	if gsn, ok := config[GcpSecretNameKey]; ok {
+		sc.GcpSecretName = gsn
+	}
+
+	return sc, nil
+}
+
+var (
+	// kubeclient is the in-cluster Kubernetes kubeclient.
+	kubeclient *kubernetes.Clientset
```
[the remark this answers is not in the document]
anniefu: Pending

stackdriver/config/config.go
```diff
@@ -0,0 +1,162 @@
...
+	if isKubeclientInitialized() {
+		return
+	}
+
+	mtx.Lock()
+	defer mtx.Unlock()
+
+	if kubeclient != nil {
+		return
+	}
```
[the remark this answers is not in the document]
anniefu: Because ensureKubeclient could be called frequently, I wanted to make the nil check as fast as possible and use a RLock() instead of WLock() the whole function.

The double nil check is to avoid this scenario:
1. P1 obtains RLock
2. P2 obtains RLock
3. P1 reads kubeclient as nil
4. P2 reads kubeclient as nil
5. P1 obtains WLock
6. P2 waits on WLock
7. P1 initializes kubeclient & releases WLock
8. P2 obtains WLock & double initalizes kubeclient

stackdriver/config/config_test.go
anniefu: Hmm, it doesn't add much value in this test suite, I'll remove it.

There is the possibility that passing an invalid gcp location causes an error when creating Stackdriver exporter/client, but that codepath is not in this package.

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,100 @@
...
+			if test.expectedConfig != *c {
+				t.Errorf("Incorrect stackdriver config. Expected: [%v], Got: [%v]", test.expectedConfig, *c)
+			}
+		})
+	}
+}
+
+// This test ensures that ensureKubeClient completes. Errors are expected, but ok.
+func TestEnsureKubeClientNoDeadlock(t *testing.T) {
+	for i := 0; i < 10; i++ {
```
garron: Will this leave goroutines running beyond the time this test is executing?

stackdriver/config/config.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// ConvertSecretToExporterOption converts a Kubernetes Secret to an OpenCensus Stackdriver Exporter Option.
+func ConvertSecretToExporterOption(secret *v1.Secret) option.ClientOption {
+	return option.WithCredentialsJSON(secret.Data[secretDataFieldKey])
+}
+
+func isKubeclientInitialized() bool {
+	mtx.RLock()
+	defer mtx.Unlock()
```
garron: I think this should be `.RUnlock()`?

stackdriver/config/config_test.go
anniefu: Technically possible, I've added a check to make sure all the goroutines come back before ending the test.

stackdriver/config/config.go
anniefu: Yeah, good catch! I was wondering why the test was failing when I ran it with a count > 1...

stackdriver/config/config.go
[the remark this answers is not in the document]
yanweiguo: Could you add a comment that why we need this second nil check?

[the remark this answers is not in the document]
anniefu: Done.

stackdriver/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	cs, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
```
JRBANCEL: Do we really want to panic?

stackdriver/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	kubeclient = cs
+}
```
JRBANCEL: `sync.Once` is exactly for this scenario: ensuring thread-safe unique initialization: https://golang.org/src/sync/once.go

stackdriver/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+// GetStackdriverSecret returns the Kubernetes Secret specified in the given config.
+func GetStackdriverSecret(config *Config) (*v1.Secret, error) {
+	ensureKubeclient()
+	ns := secretNamespaceDefault
+	if config.GCPSecretNamespace != "" {
+		ns = config.GCPSecretNamespace
+	}
+
+	return kubeclient.CoreV1().Secrets(ns).Get(config.GCPSecretName, metav1.GetOptions{})
+}
```
JRBANCEL: I am wondering if this should be part of Config.
This probably should be done where the config is used and let Config logic be simple parsing of the ConfigMap.

anniefu: I would say since "stackdriver/config" defines the definition of GCPSecret being a Kubernetes secret name, it should also provide the logic of how to use it. 

It also prevents a consumers that don't need a kubeclient for other purposes than reading this Secret from having to re-implement the kubeclient bit.

stackdriver/config/config.go
anniefu: Probably? It really shouldn't happen on a Kubernetes cluster. The Kubernetes example panics https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go

stackdriver/config/config.go
anniefu: Oh nice, I'll look into it

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,97 @@
...
+			if test.expectedConfig != *c {
+				t.Errorf("Incorrect stackdriver config. Expected: [%v], Got: [%v]", test.expectedConfig, *c)
+			}
+		})
+	}
+}
+
+// This test ensures that ensureKubeClient completes. Errors are expected, but ok.
+func TestEnsureKubeclientNoDeadlock(t *testing.T) {
+	done := make(chan bool)
```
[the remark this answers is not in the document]
anniefu: Pending

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := 0; i < 10; i++ {
+		go testKubeclient(t, done)
+	}
+
+	// Wait for 10 responses
+	for i := 0; i < 10; i++ {
+		<-done
+	}
+}
```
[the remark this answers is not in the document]
anniefu: Pending

stackdriver/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+
+// ConvertSecretToExporterOption converts a Kubernetes Secret to an OpenCensus Stackdriver Exporter Option.
+func ConvertSecretToExporterOption(secret *v1.Secret) option.ClientOption {
+	return option.WithCredentialsJSON(secret.Data[secretDataFieldKey])
+}
+
+// isKubeclientInitialized is a thread-safe check to see if the kubeclient is initialzed.
```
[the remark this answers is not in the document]
anniefu: Pending

stackdriver/config/config.go
```diff
@@ -0,0 +1,168 @@
...
+	// Required when the Kubernetes cluster is not hosted on GCE.
+	GCPLocation string
+	// ClusterName is the cluster name with which the data will be associated in Stackdriver.
+	// Required when the Kubernetes cluster is not hosted on GCE.
+	ClusterName string
+	// GCPSecretName is the optional GCP service account key which will be used to
+	// authenticate with Stackdriver. If not provided, Google Application Default Credentials
+	// will be used (https://cloud.google.com/docs/authentication/production).
+	GCPSecretName string
+	// GCPSecretNamespace is the Kubernetes namespace where GcpSecretName is located.
```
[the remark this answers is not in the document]
anniefu: Pending

stackdriver/config/config.go
anniefu: I've switched to using sync.Once which saves a lot of logic/comments/testing. Good find!

stackdriver/config/config.go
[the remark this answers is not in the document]
anniefu: Function no longer exists.

stackdriver/config/config_test.go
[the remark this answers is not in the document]
anniefu: Test has been simplified to just calling ensureKubeclient once and checking for panic, since otherwise it's just a test of sync.Once. This was a great tip for the future though.

stackdriver/config/config.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			panic(err.Error())
+		}
+
+		cs, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			panic(err.Error())
+		}
+		kubeclient = cs
+
```
JRBANCEL:
```suggestion
```

stackdriver/config/config.go
```diff
@@ -0,0 +1,149 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+var (
+
```
JRBANCEL:
```suggestion
```

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+)
+
+func TestNewStackdriverConfigFromConfigMap(t *testing.T) {
+	tests := []struct {
+		name           string
+		configMap      *corev1.ConfigMap
+		expectedConfig Config
+	}{
+		{
```
JRBANCEL: nit: Coalesce to:
```
}{{
```
see for example: https://github.com/knative/serving/blob/master/pkg/reconciler/route/resources/names/names_test.go#L35

stackdriver/config/config_test.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+			expectedConfig: Config{
+				ProjectID:          "project",
+				GCPLocation:        "us-west1",
+				ClusterName:        "cluster",
+				GCPSecretName:      "secret",
+				GCPSecretNamespace: "non-default",
+			},
+		},
+		{
```
JRBANCEL: nit: coalesce to 
```
}, {
```
see for example https://github.com/knative/serving/blob/master/pkg/reconciler/route/resources/names/names_test.go#L40

anniefu:
Closing in favor of #794

PR #794, opened 2019-10-23
Add field to metrics config for configuring Stackdriver exporter client

anniefu:
Add ability to configure Stackdriver client metadata for the metrics package's Stackdriver exporter.

Currently, Stackdriver metrics rely on the Google Compute Engine metadata server, which is only accessible on GCE, to configure the Stackdriver client. This allows sending metrics to Stackdriver on any compute platform.

The new fields are optional and purely additive and do not break existing usage. If omitted, Stackdriver will fall back on Google Application Credentials defaults (https://cloud.google.com/docs/authentication/production).

Configuration required for knative/serving#5777

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -46,11 +46,11 @@ const (
 	ReportingPeriodKey                  = "metrics.reporting-period-seconds"
 	StackdriverCustomMetricSubDomainKey = "metrics.stackdriver-custom-metrics-subdomain"
 	// Stackdriver client configuration keys
-	StackdriverProjectIDKey          = "metrics.stackdriver-project-id"
-	StackdriverGCPLocationKey        = "metrics.stackdriver-gcp-location"
-	StackdriverClusterNameKey        = "metrics.stackdriver-cluster-name"
-	StackdriverGCPSecretNameKey      = "metrics.stackdriver-gcp-secret-name"
-	StackdriverGCPSecretNamespaceKey = "metrics.stackdriver-gcp-secret-namespace"
+	stackdriverProjectIDKey          = "metrics.stackdriver-project-id"
+	stackdriverGCPLocationKey        = "metrics.stackdriver-gcp-location"
+	stackdriverClusterNameKey        = "metrics.stackdriver-cluster-name"
+	stackdriverGCPSecretNameKey      = "metrics.stackdriver-gcp-secret-name"
+	stackdriverGCPSecretNamespaceKey = "metrics.stackdriver-gcp-secret-namespace"
 
 	// Stackdriver is used for Stackdriver backend
 	Stackdriver metricsBackend = "stackdriver"
@@ -131,17 +131,17 @@ type stackdriverClientConfig struct {
 
 // newStackdriverClientConfigFromMap creates a stackdriverClientConfig from the given map
 func newStackdriverClientConfigFromMap(config map[string]string) *stackdriverClientConfig {
-	sc := &stackdriverClientConfig{}
-	sc.ProjectID = config[StackdriverProjectIDKey]
-	sc.GCPLocation = config[StackdriverGCPLocationKey]
-	sc.ClusterName = config[StackdriverClusterNameKey]
-	sc.GCPSecretName = config[StackdriverGCPSecretNameKey]
-	sc.GCPSecretNamespace = config[StackdriverGCPSecretNamespaceKey]
-	return sc
+	return &stackdriverClientConfig{
+		ProjectID:          config[stackdriverProjectIDKey],
+		GCPLocation:        config[stackdriverGCPLocationKey],
+		ClusterName:        config[stackdriverClusterNameKey],
+		GCPSecretName:      config[stackdriverGCPSecretNameKey],
+		GCPSecretNamespace: config[stackdriverGCPSecretNamespaceKey],
+	}
 }
 
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
-	mc := metricsConfig{stackdriverClientConfig: stackdriverClientConfig{}}
+	var mc metricsConfig
 
 	if ops.Domain == "" {
 		return nil, errors.New("metrics domain cannot be empty")
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -130,30 +130,14 @@ type stackdriverClientConfig struct {
 }
 
 // newStackdriverClientConfigFromMap creates a stackdriverClientConfig from the given map
-func newStackdriverClientConfigFromMap(config map[string]string) (*stackdriverClientConfig, error) {
+func newStackdriverClientConfigFromMap(config map[string]string) *stackdriverClientConfig {
 	sc := &stackdriverClientConfig{}
-
-	if pi, ok := config[StackdriverProjectIDKey]; ok {
-		sc.ProjectID = pi
-	}
-
-	if gl, ok := config[StackdriverGCPLocationKey]; ok {
-		sc.GCPLocation = gl
-	}
-
-	if cn, ok := config[StackdriverClusterNameKey]; ok {
-		sc.ClusterName = cn
-	}
-
-	if gsn, ok := config[StackdriverGCPSecretNameKey]; ok {
-		sc.GCPSecretName = gsn
-	}
-
-	if gsns, ok := config[StackdriverGCPSecretNamespaceKey]; ok {
-		sc.GCPSecretNamespace = gsns
-	}
-
-	return sc, nil
+	sc.ProjectID = config[StackdriverProjectIDKey]
+	sc.GCPLocation = config[StackdriverGCPLocationKey]
+	sc.ClusterName = config[StackdriverClusterNameKey]
+	sc.GCPSecretName = config[StackdriverGCPSecretNameKey]
+	sc.GCPSecretNamespace = config[StackdriverGCPSecretNamespaceKey]
+	return sc
 }
 
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
@@ -206,10 +190,7 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
 	// use the application default credentials. If that is not available, Opencensus would fail to create the
 	// metrics exporter.
 	if mc.backendDestination == Stackdriver {
-		scc, err := newStackdriverClientConfigFromMap(m)
-		if err != nil {
-			return nil, err
-		}
+		scc := newStackdriverClientConfigFromMap(m)
 		mc.stackdriverClientConfig = *scc
 		mc.isStackdriverBackend = true
 		mc.stackdriverMetricTypePrefix = path.Join(mc.domain, mc.component)
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -44,8 +44,13 @@ const (
 	AllowStackdriverCustomMetricsKey    = "metrics.allow-stackdriver-custom-metrics"
 	BackendDestinationKey               = "metrics.backend-destination"
 	ReportingPeriodKey                  = "metrics.reporting-period-seconds"
-	StackdriverProjectIDKey             = "metrics.stackdriver-project-id"
 	StackdriverCustomMetricSubDomainKey = "metrics.stackdriver-custom-metrics-subdomain"
+	// Stackdriver client configuration keys
+	StackdriverProjectIDKey          = "metrics.stackdriver-project-id"
+	StackdriverGCPLocationKey        = "metrics.stackdriver-gcp-location"
+	StackdriverClusterNameKey        = "metrics.stackdriver-cluster-name"
+	StackdriverGCPSecretNameKey      = "metrics.stackdriver-gcp-secret-name"
+	StackdriverGCPSecretNamespaceKey = "metrics.stackdriver-gcp-secret-namespace"
 
 	// Stackdriver is used for Stackdriver backend
 	Stackdriver metricsBackend = "stackdriver"
@@ -76,9 +81,6 @@ type metricsConfig struct {
 	prometheusPort int
 
 	// ---- Stackdriver specific below ----
-	// stackdriverProjectID is the stackdriver project ID where the stats data are
-	// uploaded to. This is not the GCP project ID.
-	stackdriverProjectID string
 	// allowStackdriverCustomMetrics indicates whether it is allowed to send metrics to
 	// Stackdriver using "global" resource type and custom metric type if the
 	// metrics are not supported by the registered monitored resource types. Setting this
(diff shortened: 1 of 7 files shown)


metrics/config.go
```diff
@@ -100,10 +102,62 @@ type metricsConfig struct {
...
+		sc.ClusterName = cn
+	}
+
+	if gsn, ok := config[StackdriverGCPSecretNameKey]; ok {
+		sc.GCPSecretName = gsn
+	}
+
+	if gsns, ok := config[StackdriverGCPSecretNamespaceKey]; ok {
+		sc.GCPSecretNamespace = gsns
+	}
```
JRBANCEL: Since maps return nil value if the key doesn't exist, in this case (strings and no default values),
```
...
sc.GCPSecretNamespace = config[StackdriverGCPSecretNamespaceKey]
...
```
would be enough.

metrics/config.go
```diff
@@ -100,10 +102,62 @@ type metricsConfig struct {
...
+	// will be used (https://cloud.google.com/docs/authentication/production).
+	GCPSecretName string
+	// GCPSecretNamespace is the Kubernetes namespace where GCPSecretName is located.
+	// The Kubernetes ServiceAccount used by the pod that is exporting data to
+	// Stackdriver should have access to Secrets in this namespace.
+	GCPSecretNamespace string
+}
+
+// newStackdriverClientConfigFromMap creates a stackdriverClientConfig from the given map
+func newStackdriverClientConfigFromMap(config map[string]string) (*stackdriverClientConfig, error) {
```
JRBANCEL: This can't return an error, can it?

metrics/exporter_test.go
```diff
@@ -102,22 +103,150 @@ func TestMain(m *testing.M) {
...
+				ProjectID:          "project",
+				GCPLocation:        "us-west1",
+				ClusterName:        "cluster",
+				GCPSecretName:      "secret",
+				GCPSecretNamespace: "secret-ns",
+			},
+		},
+		expectSuccess: true,
+	}, {
+		name: "invalidStackdriverGcpLocation",
```
yanweiguo: Why is the location invalid? What does this test do?

metrics/stackdriver_exporter.go
```diff
@@ -80,6 +116,41 @@ func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (v
 	return e, nil
 }
 
+// getStackdriverExporterOptions creates client options for the opencensus Stackdriver exporter from the given stackdriverClientConfig.
+// On error, an empty array of client options is returned.
+func getStackdriverExporterOptions(sdconfig *stackdriverClientConfig) ([]option.ClientOption, error) {
```
yanweiguo: I think you mean `getStackdriverClientOptions`?

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +91,18 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
+	gm := getGCPMetadata(config)
 	mtf := getMetricTypeFunc(config.stackdriverMetricTypePrefix, config.stackdriverCustomMetricTypePrefix)
+	co, err := getStackdriverExporterOptions(&config.stackdriverClientConfig)
+	if err != nil {
+		logger.Warnw("Issue configuring Stackdriver exporter client options, no additional client options will be used: ", zap.Error(err))
+	}
+	// Automatically fall back on Google application default credentials
 	e, err := newStackdriverExporterFunc(stackdriver.Options{
-		ProjectID:               config.stackdriverProjectID,
+		ProjectID:               config.stackdriverClientConfig.ProjectID,
```
yanweiguo: I think it's better to use `gm.projectId` and `gm.location` to keep consistent in case `gm` has values fetched from GCE metadata service but config.stackdriverClientConfig.ProjectID is empty.

metrics/stackdriver_exporter.go
```diff
@@ -47,13 +58,30 @@ var (
...
+	// In product usage, this is always set to function sdconfig.getStackdriverSecret.
+	// In unit tests this is set to a fake one because an in-cluster kubeclient is required.
+	getStackdriverSecretFunc func(config *stackdriverClientConfig) (*corev1.Secret, error)
+
+	// kubeclient is the in-cluster Kubernetes kubeclient, which is lazy-initialized on first use.
+	kubeclient *kubernetes.Clientset
+	// initClientOnce is the lazy initializer for kubeclient.
+	initClientOnce sync.Once
+	// kubeclientInitErr capture an error during initClientOnce
+	kubeclientInitErr error
```
yanweiguo: I didn't find this is read anywhere after being set. Is it necessary to keep it in memory?

metrics/config.go
anniefu: Yeah, I made it return error for defensive programming reasons. I'll change it to no error since it's internal to this package only and should be easy to change to returning an error if behavior changes in the future.

metrics/exporter_test.go
anniefu: This is to make sure that creating an exporter does not fail if an invalid GCP location is used. GCP locations are restricted to a specified list and Stackdriver will reject requests with an invalid GCP location. 

However, opencensus does not validate and allows the exporter to be created. This behavior is essentially verified by this test. (Stackdriver will send back a descriptive error when the first metric is exported).

metrics/stackdriver_exporter.go
anniefu: If initializing the kubeclient fails, getStackdriverSecret will end up returning this error on every subsequent call, which is logged as a warning whenever stackdriver exporter is updated.

metrics/stackdriver_exporter.go
anniefu: Oh yeah, it is actually a bug not to return gm.projectID & gm.location. Good catch!

metrics/stackdriver_exporter.go

metrics/config.go
anniefu: Nice, will change.

metrics/exporter_test.go
yanweiguo: Could you add comment to explain it?

metrics/stackdriver_exporter.go
yanweiguo: Can't we just make `kubeclientInitErr` a local var in `ensureKubeclient`?

anniefu: Because `initClientOnce` will only run once, after the first error `initClientOnce` will be a no-op. 

If I make the `kubeclientInitErr` local to `ensureKubeclient`, `kubeclientInitErr` will be reset on subsequent calls to `ensureKubeclient` and start to return nil  (or whatever the local initialization value is), even though kubeclient is still not initialized.

metrics/config.go
```diff
@@ -100,10 +102,62 @@ type metricsConfig struct {
...
+	if gsns, ok := config[StackdriverGCPSecretNamespaceKey]; ok {
+		sc.GCPSecretNamespace = gsns
+	}
+
+	return sc, nil
 }
 
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
-	var mc metricsConfig
+	mc := metricsConfig{stackdriverClientConfig: stackdriverClientConfig{}}
```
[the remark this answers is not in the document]
anniefu: Nice, pending

metrics/config.go
```diff
@@ -100,10 +102,62 @@ type metricsConfig struct {
...
+	// The Kubernetes ServiceAccount used by the pod that is exporting data to
+	// Stackdriver should have access to Secrets in this namespace.
+	GCPSecretNamespace string
+}
+
+// newStackdriverClientConfigFromMap creates a stackdriverClientConfig from the given map
+func newStackdriverClientConfigFromMap(config map[string]string) (*stackdriverClientConfig, error) {
+	sc := &stackdriverClientConfig{}
+
+	if pi, ok := config[StackdriverProjectIDKey]; ok {
```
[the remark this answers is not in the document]
anniefu: Pending

metrics/stackdriver_exporter.go
yanweiguo: Got it.

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +65,18 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
 // 	See https://github.com/knative/pkg/issues/608
 func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
-	gm := gcpMetadataFunc()
+	gm := getGCPMetadata(config)
 	mtf := getMetricTypeFunc(config.stackdriverMetricTypePrefix, config.stackdriverCustomMetricTypePrefix)
-	e, err := newStackdriverExporterFunc(stackdriver.Options{
-		ProjectID:               config.stackdriverProjectID,
+	co, err := getStackdriverExporterClientOptions(&config.stackdriverClientConfig)
+	if err != nil {
+		logger.Warnw("Issue configuring Stackdriver exporter client options, no additional client options will be used: ", zap.Error(err))
```
garron: This warning reads to me like a minor issue and I'm not sure everyone will know what this means exactly. If I'm reading the code correctly, it seems that this failure case is more about accessing the configured secret, and not about interacting with the stackdriver client. Thoughts on something like "Failed to read configured Stackdriver secret"? (which might make more sense down in `getStackdriverExporterClientOptions()`)

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +65,18 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
 // TODO should be properly refactored to be able to inject the getMonitoredResourceFunc function.
 // 	See https://github.com/knative/pkg/issues/608
 func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
-	gm := gcpMetadataFunc()
+	gm := getGCPMetadata(config)
```
garron: Reading this code, I think it's easy to conclude that getGCPMetadata is returning the metadata from GCP, and not the result of merging with config. Consider a more descriptive name like `getMergedGCPMetadata()`, or refactoring to something like `mergeGCPMetadata(retrieveGCPMetadata(), config)`.

metrics/stackdriver_exporter.go
anniefu: I added some formatting from the error that gets emitted from getStackdriverSecret to print more info. It will get printed here as part of the `err` field, and should be actionable by the user to fix.

I think it's important to include a bit about the Stackdriver exporter client options so users understand the end result of the Secret failing.

garron: Looks great, thanks!

anniefu:
/hold
Doing some e2e testing on real clusters.

metrics/stackdriver_exporter.go
```diff
@@ -63,10 +65,18 @@ func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
...
 	mtf := getMetricTypeFunc(config.stackdriverMetricTypePrefix, config.stackdriverCustomMetricTypePrefix)
-	e, err := newStackdriverExporterFunc(stackdriver.Options{
-		ProjectID:               config.stackdriverProjectID,
+	co, err := getStackdriverExporterClientOptions(&config.stackdriverClientConfig)
+	if err != nil {
+		logger.Warnw("Issue configuring Stackdriver exporter client options, no additional client options will be used: ", zap.Error(err))
+	}
+	// Automatically fall back on Google application default credentials
+	e, err := newOpencensusSDExporter(stackdriver.Options{
+		ProjectID:               gm.project,
```
evankanderson: Observation: I don't think you need to merge with the GCE metadata server here, since ProjectID says:

```golang
    // ProjectID is the identifier of the Stackdriver
    // project the user is uploading the stats data to.
    // If not set, this will default to your "Application Default Credentials".
    // For details see: https://developers.google.com/accounts/docs/application-default-credentials.
    //
    // It will be used in the project_id label of a Stackdriver monitored
    // resource if the resource does not inherently belong to a specific
    // project, e.g. on-premise resource like k8s_container or generic_task.
```

This would mean that you could simply drop in the result of fetching the key from the confiiguration map, as empty string would mean "use the GCE default".

metrics/stackdriver_exporter.go
```diff
@@ -80,6 +90,41 @@ func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (v
...
+// On error, an empty array of client options is returned.
+func getStackdriverExporterClientOptions(sdconfig *stackdriverClientConfig) ([]option.ClientOption, error) {
+	var co []option.ClientOption
+	if sdconfig.GCPSecretName != "" {
+		secret, err := getStackdriverSecret(sdconfig)
+		if err != nil {
+			return co, err
+		}
+
+		co = []option.ClientOption{convertSecretToExporterOption(secret)}
```
evankanderson: Why not use `append()` here.

metrics/stackdriver_exporter.go
```diff
@@ -80,6 +90,41 @@ func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (v
...
+	}
+
+	return co, nil
+}
+
+// getGCPMetadata returns GCP metadata required to export metrics
+// to Stackdriver. Values explicitly set in the config take the highest precedent.
+func getGCPMetadata(config *metricsConfig) *gcpMetadata {
```
evankanderson: As I mentioned above, both project and location can be defaulted if they are empty strings by the library already:

```golang
    // Location is the identifier of the GCP or AWS cloud region/zone in which
    // the data for a resource is stored.
    // If not set, it will default to the location provided by the metadata server.
    //
    // It will be used in the location label of a Stackdriver monitored resource
    // if the resource does not inherently belong to a specific project, e.g.
    // on-premise resource like k8s_container or generic_task.
```

metrics/stackdriver_exporter.go
```diff
@@ -115,3 +160,44 @@ func getMetricTypeFunc(metricTypePrefix, customMetricTypePrefix string) func(vie
 		return path.Join(customMetricTypePrefix, view.Measure.Name())
 	}
 }
+
+// getStackdriverSecret returns the Kubernetes Secret specified in the given config.
+func getStackdriverSecret(sdconfig *stackdriverClientConfig) (*corev1.Secret, error) {
+	err := ensureKubeclient()
+	if err != nil {
```
evankanderson: Given that `ensureKubeclient` seems to be side-effect only, why not collapse like so:

```golang
if err := ensureKubeclient(); err != nil {
  ...
}
```

metrics/stackdriver_exporter.go
```diff
@@ -115,3 +160,44 @@ func getMetricTypeFunc(metricTypePrefix, customMetricTypePrefix string) func(vie
...
+		return nil, err
+	}
+
+	ns := secretNamespaceDefault
```
evankanderson: Slight preference for writing this in the opposite way:

```golang
ns = sdconfig.GCPSecretNamespace
if ns == "" {
  ns = secretNamespaceDefault
}
```

Mostly because it focuses on the fact that you want `ns` to have a non-empty-string value, rather than that you care about `sdconfig` specifically.

metrics/stackdriver_exporter.go
anniefu: I believe you are correct. However, I am going to be overly cautious and defer it to a separate PR. Because it would be a deviation from existing behavior and can only be properly tested on GCE, I think it deserves extra validation outside of the scope of this change.

metrics/stackdriver_exporter.go

metrics/stackdriver_exporter.go

metrics/stackdriver_exporter.go

PR #797, opened 2019-10-24
Add metrics tags length verification

yanweiguo:
Currently we only check got metrics tags are in the expected tags, but not in reverse. We should verify that got and expected tags are equal.


metrics/metricstest/metricstest.go
```diff
@@ -150,6 +150,10 @@ func checkExactlyOneRow(t *testing.T, name string, wantTags map[string]string) *
 
 func checkRowTags(t *testing.T, row *view.Row, name string, wantTags map[string]string) {
 	t.Helper()
+	gotlen := len(row.Tags)
```
vagababov: just inline
`if got, want := len(row.Tags), len(wantTags); got != want`

yanweiguo: Done.

PR #809, opened 2019-10-25
Add ref normalizing method to Destination

grantr:
`NormalizeDeprecatedObjectReference` updates the Destination to the desired end state post-deprecation. Using this method allows code to ignore the deprecated fields and just operate on Destination's Ref field.

We can use this in sources to first set the Ref, then update its namespace to the current source's namespace as required by URIResolver.

I left the resolver's deprecated fields handling code in place so other projects can continue to use the deprecated fields without normalizing first.

/cc @n3wscott @Harwayne @vaikas-google

### DIFF
--- a/apis/v1alpha1/destination.go
+++ b/apis/v1alpha1/destination.go
@@ -114,6 +114,8 @@ func ValidateDestination(dest Destination, allowDeprecatedFields bool) *apis.Fie
 	return nil
 }
 
+// deprecatedObjectReference returns an ObjectReference composed from the
+// deprecated ObjectReference fields, or nil if those fields are empty.
 func (dest Destination) deprecatedObjectReference() *corev1.ObjectReference {
 	if dest.DeprecatedAPIVersion == "" && dest.DeprecatedKind == "" && dest.DeprecatedName == "" && dest.DeprecatedNamespace == "" {
 		return nil
@@ -126,10 +128,14 @@ func (dest Destination) deprecatedObjectReference() *corev1.ObjectReference {
 	}
 }
 
-// GetRef gets the ObjectReference from this Destination, if one is present. If no ref is present,
-// then nil is returned.
-// Note: this mostly exists to abstract away the deprecated ObjectReference fields. Once they are
-// removed, then this method should probably be removed too.
+// GetRef gets the ObjectReference from this Destination, if one is present. If
+// the Ref field is present, that's returned. If the Ref field isn't present but
+// the deprecated fields are present, then an ObjectReference is built from the
+// deprecated fields and returned, If neither Ref nor deprecated fields are
+// present, then nil is returned.
+// Note: this mostly exists to abstract away the deprecated ObjectReference
+// fields. Once they are removed, then this method should probably be removed
+// too.
 func (dest *Destination) GetRef() *corev1.ObjectReference {
 	if dest == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


apis/v1alpha1/destination.go
```diff
@@ -143,6 +149,25 @@ func (dest *Destination) GetRef() *corev1.ObjectReference {
 	return nil
 }
 
+// NormalizeDeprecatedObjectReference copies the deprecated ObjectReference
+// fields to the Ref field and clears the deprecated fields. It uses the same
+// precedence rules as GetRef(). Use this to normalize code to use the Ref field
+// when deprecated fields may exist.
+// This method should be removed when the deprecated ObjectReference fields are
+// removed.
+func (dest *Destination) NormalizeDeprecatedObjectReference() {
```
n3wscott: i would call this ` NormalizeObjectReference()`

n3wscott:
/lgtm
/approve
/hold
hold for if you agree with the name change.

apis/v1alpha1/destination.go
Harwayne: When do we expect this to be used? My understanding is that we _won't_ be automatically translating anything. Rather, we will just use deprecation warnings in Source's statuses.

grantr: This will be used in source reconcilers. The reconciler needs to set the destination ref's namespace to the source's namespace, but since the ref can come from two places right now, that's more difficult than it should be.

I agree that we don't want to change the given object in Kubernetes storage. I'd expect reconcilers to never attempt to update the spec of the object (by using the status subresource) but I see your point that it may happen if we modify the ref in place.

Does it make sense to refactor this method to take a destination as argument and return a new normalized one? That way the original destination isn't modified.

Harwayne: I see, this is intended to be used only inside a controller's reconcile loop and never have the altered destination be written out to storage.

If that is correct, then I don't think we should alter anything in place. I would probably change the method signature to:

```golang
func (dest Destination) NormalizeObjectReference() Destination
```

Then in controllers, wherever they want the normalized reference, they use:
```golang
source.sink.NormalizeObjectReference()
```

More generally, is this an improvement over just using Destionation's `GetRef()`?

grantr: I updated the method to take a destination as argument and return a copied destination, similar to what you proposed. 

`GetRef()` doesn't work for this case because the reconciler must pass the entire Destination to `URIFromDestination` to get the uri reference resolution behavior.

Harwayne: Can `URIFromDestination` use Destination's `GetRef`?

grantr: That wouldn't help with this situation. If the reconciler always sets `Ref.Namespace`, `GetRef` will ignore the Deprecated fields. If the reconciler always sets `DeprecatedNamespace`, `GetRef` will ignore it if there's already a Ref. 

The behavior I want for the reconciler is to set Namespace in one place (ideally the `Ref` field). I'd rather each reconciler not have to care that there are two possible ways to specify a Ref:

```go
if dest.Ref != nil && dest.Ref.Name != "" {
  dest.Ref.Namespace = source.Namespace
} else if dest.DeprecatedName != "" {
  dest.Ref.DeprecatedNamespace = source.Namespace
}
```

But in the grand scheme of things it's not so bad, and it will only be around for one release, so actually now I'm thinking I should close this PR to avoid needing another pkg update in eventing.

Harwayne: Got it, now I understand. Thank you for explaining it to me (repeatedly...). 

I now understand why this change is desirable, I'll leave it up to you if we want this, #810, or neither.

grantr:
/close

This doesn't seem relevant anymore.

PR #810, opened 2019-10-25
Removing Namespace from Destination

nachocano:
- We do not allow cross namespace references in Destination... Also removing DeprecatedNamespace field. 
- Changing the signature of URIFromDestination resolver to explicitly pass a namespace. This further strengthens the fact that Destination does not have a namespace (for now).
- Using the Validate method from Destination inside the URIFromDestination method. Avoid duplicating code, and potentially corner cases we are not considering there...

### DIFF
--- a/apis/v1alpha1/destination.go
+++ b/apis/v1alpha1/destination.go
@@ -18,7 +18,9 @@ package v1alpha1
 
 import (
 	"context"
+	"reflect"
 
+	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/apis"
 )
@@ -37,9 +39,6 @@ type Destination struct {
 
 	// +optional
 	DeprecatedName string `json:"name,omitempty"`
-
-	// +optional
-	DeprecatedNamespace string `json:"namespace,omitempty"`
 
 	// URI can be an absolute URL(non-empty scheme and non-empty host) pointing to the target or a relative URI. Relative URIs will be resolved using the base URI retrieved from Ref.
 	// +optional
@@ -74,9 +73,6 @@ func ValidateDestination(dest Destination, allowDeprecatedFields bool) *apis.Fie
 		if dest.DeprecatedName != "" {
 			errs = errs.Also(apis.ErrInvalidValue("name is not allowed here, it's a deprecated value", "name"))
 		}
-		if dest.DeprecatedNamespace != "" {
-			errs = errs.Also(apis.ErrInvalidValue("namespace is not allowed here, it's a deprecated value", "namespace"))
-		}
 		if errs != nil {
 			return errs
 		}
@@ -106,23 +102,22 @@ func ValidateDestination(dest Destination, allowDeprecatedFields bool) *apis.Fie
 	}
 	if ref != nil && dest.URI == nil {
 		if dest.Ref != nil {
-			return validateDestinationRef(*ref).ViaField("ref")
+			return IsValidObjectReference(*ref).ViaField("ref")
 		} else {
-			return validateDestinationRef(*ref)
+			return IsValidObjectReference(*ref)
 		}
 	}
 	return nil
 }
 
 func (dest Destination) deprecatedObjectReference() *corev1.ObjectReference {
-	if dest.DeprecatedAPIVersion == "" && dest.DeprecatedKind == "" && dest.DeprecatedName == "" && dest.DeprecatedNamespace == "" {
+	if dest.DeprecatedAPIVersion == "" && dest.DeprecatedKind == "" && dest.DeprecatedName == "" {
 		return nil
 	}
 	return &corev1.ObjectReference{
 		Kind:       dest.DeprecatedKind,
 		APIVersion: dest.DeprecatedAPIVersion,
 		Name:       dest.DeprecatedName,
-		Namespace:  dest.DeprecatedNamespace,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


grantr:
The namespace restriction is policy enforcement, but Destination is a shared API (and URIResolver shared code) that may be used by other objects without the same restriction or with different restrictions.

> We do not allow cross namespace references in Destination... Also removing DeprecatedNamespace field.

Are we sure there aren't any objects with a namespace field? The point of the deprecated fields is to ensure we don't break any existing objects. If they have namespace already, this will break them.

I think Destination should not disallow cross-namespace references. Destination is a shared API, but disallowing cross-namespace references is a policy decision made by the eventing team, and one that we may revisit (e.g. for Broker bridges). I'd prefer to enforce it in object-specific validation webhooks rather than the shared API definition and resolution code.

> Changing the signature of URIFromDestination resolver to explicitly pass a namespace.

This seems unnecessary when we can manipulate the namespace of the ref to enforce policy. A different resource may have a different policy, like enforcing a specific apiVersion and kind.

> Using the Validate method from Destination inside the URIFromDestination method.

Seems like a good idea.

nachocano:
> I think Destination should not disallow cross-namespace references. Destination is a shared API, but disallowing cross-namespace references is a policy decision made by the eventing team, and one that we may revisit (e.g. for Broker bridges). I'd prefer to enforce it in object-specific validation webhooks rather than the shared API definition and resolution code.

I agree we will have to revisit and will probably introduce cross-namespace references in the upcoming releases. It seems that we didn't allow that before, so IMO we shouldn't allow it now, especially since we are ~2 business days away from the release. 

As to where to add the check, I'd rather add it here so that everyone using Destination can use it. It seems more error-prone to have all this scattered throughout the different objects validation logic.

nachocano:
> Are we sure there aren't any objects with a namespace field? The point of the deprecated fields is to ensure we don't break any existing objects. If they have namespace already, this will break them.

If there are, is because we weren't properly validating Sinks throughout the different objects. It's related to the error-prone argument I mentioned above.

nachocano:
> The namespace restriction is policy enforcement, but Destination is a shared API (and URIResolver shared code) that may be used by other objects without the same restriction or with different restrictions.

URIResolver has two public methods. URIFromDestination and URIFromObjectReference. If clients want an ObjectRef with a namespace, I think they can safely use the later.

nachocano:
@grantr I think all are valid points, and perhaps [yours](https://github.com/knative/pkg/pull/809) is cleaner. Either way, I think we should try to get something in before today's pkg cut.

grantr:
> If clients want an ObjectRef with a namespace, I think they can safely use the later.

Not with the current implementation, because clients want the uri reference resolution that is only done by URIFromDestination.

grantr:
> is because we weren't properly validating Sinks throughout the different objects

Correct but those objects will still break if we remove the improperly validated field.

nachocano:
> > If clients want an ObjectRef with a namespace, I think they can safely use the later.
> 
> Not with the current implementation, because clients want the uri reference resolution that is only done by URIFromDestination.

Ohh I see

nachocano:
> Correct but those objects will still break if we remove the improperly validated field.

Yes, I guess I'd rather let them break :) but that is my opinion...

resolver/addressable_resolver.go
```diff
@@ -63,50 +63,31 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
-	}
-	var ref *corev1.ObjectReference
-	if dest.Ref != nil {
-		ref = dest.Ref
-	} else {
-		ref = deprecatedObjectReference
+func (r *URIResolver) URIFromDestination(ctx context.Context, namespace string, dest apisv1alpha1.Destination, parent interface{}) (string, error) {
+	err := dest.Validate(ctx)
+	if err != nil {
+		return "", errors.New(err.Error())
```
Harwayne:
```suggestion
		return "", err
```

resolver/addressable_resolver.go
```diff
@@ -63,50 +63,31 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
 	}
+	ref := dest.GetRef()
 	if ref != nil {
+		// Do not modify original copy w/o namespace.
+		ref = ref.DeepCopy()
```
Harwayne: I don't think this is needed, dest is not a pointer, so it is local to this function.

resolver/addressable_resolver.go
```diff
@@ -63,50 +63,31 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
+		ref.Namespace = namespace
 		url, err := r.URIFromObjectReference(ref, parent)
 		if err != nil {
 			return "", err
 		}
 		if dest.URI != nil {
-			if dest.URI.URL().IsAbs() {
-				return "", fmt.Errorf("absolute URI is not allowed when Ref or [apiVersion, kind, name] exists")
```
Harwayne: Is this no longer true? I think without this check, it means that if `dest.URI` is an absolute URI, then `dest.Ref` is ignored entirely. Is that a correct interpretation?

apis/v1alpha1/destination.go
```diff
@@ -143,19 +138,51 @@ func (dest *Destination) GetRef() *corev1.ObjectReference {
...
+// with the fieldpaths for all those fields.
+func checkDisallowedObjectReferenceFields(f corev1.ObjectReference) *apis.FieldError {
+	disallowedFields := []string{}
+	// See if there are any fields that have been set that should not be.
+	s := reflect.ValueOf(f)
+	typeOf := s.Type()
+	for i := 0; i < s.NumField(); i++ {
+		field := s.Field(i)
+		fieldName := typeOf.Field(i).Name
+		if fieldName == "Name" || fieldName == "Kind" || fieldName == "APIVersion" {
```
Harwayne: (Optional) Switch?

resolver/addressable_resolver.go
nachocano: I removed it as I think is already taken care in the validate of the destination done above

apis/v1alpha1/destination.go
nachocano: yeah, copy/paste from eventing.. can change it, but I think we won't end up merging this

n3wscott:
I would rather put validation on the objects that use destination than limit destination directly. We may have a cluster scoped resource in the future that allows us to use the namespace and we will need to to the same work to add namespace back in and add the validation to the places we do not want namespace. 

I think it would be better to leave namespace in destination and close this PR. And then add validation to eventing objects that confirms namespace is not used for objects we care about.

Also, having namespace allows us to resolve a destination without needing the parent object but only if it is fully filled out. 

An alternate proposal is we could start defaulting the namespace in eventing to the correct namespace and have destinations fully filled out.

nachocano:
This is old stuff. We already agreed on not doing it. Forgot to close it. Closing...

PR #812, opened 2019-10-25
Start moving Stackdriver and Prometheus-specific code to a strategy pattern

evankanderson:
We noticed the other day that there was a lot of packaging-specific logic in `./pkg/metrics` that was product-specific (Google's integration with Stackdriver). This should be factored cleanly into an extension module; eventually this should look like:

```golang

metrics.RegisterHandler("stackdriver", googlemetrics.Stackdriver)
```

Where the `Stackdriver` function might have a signature like `func (map[string]string settings) (exporter, err)`.

Not fixed in this PR: Running `go test ./metrics/...` requires having run `gcloud auth application-default login` on your machine. Will fix that up in time.

/assign @anniefu @mdemirhan


metrics/config.go
```diff
@@ -74,23 +77,17 @@ type metricsConfig struct {
 	// reportingPeriod specifies the interval between reporting aggregated views.
 	// If duration is less than or equal to zero, it enables the default behavior.
 	reportingPeriod time.Duration
+
```
mattmoor-sockpuppet: Remove trailing whitespace:
```suggestion

```

metrics/config.go
```diff
@@ -140,6 +137,14 @@ func newStackdriverClientConfigFromMap(config map[string]string) *stackdriverCli
 	}
 }
 
+func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
+	if mc == nil || mc.recorder == nil {
+		ros = append(ros, stats.WithMeasurements(ms))
```
vagababov: I'd just inline the append

metrics/config.go
```diff
@@ -74,23 +77,17 @@ type metricsConfig struct {
 	// reportingPeriod specifies the interval between reporting aggregated views.
 	// If duration is less than or equal to zero, it enables the default behavior.
 	reportingPeriod time.Duration
+	
+	// recorder provides a hook for performing custom transformations before
+	// writing the metrics to the stats.RecordWithOptions interface.
+	recorder func(context.Context, stats.Measurement, ...stats.Options) error
```
anniefu: Should the recorder func be parented to the metrics pkg itself and be passed the metricsConfig as an input rather than being a part of the metricsConfig? Similar to how kubernetes/client-go/metrics pkg does it (go/blob/03bfb9bdcfe5482795b999f39ca3ed9ad42ce5bb/tools/metrics/metrics.go)?

Limiting metricsConfig to just user-configurable state seems like the ideal state. Stubbing out core functionality of the metrics pkg feels like an overscope of a config struct.

metrics/config.go
```diff
@@ -140,6 +137,14 @@ func newStackdriverClientConfigFromMap(config map[string]string) *stackdriverCli
 	}
 }
 
+func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
```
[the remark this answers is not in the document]
evankanderson: Done.

metrics/config.go
evankanderson: Done.

metrics/config_test.go
```diff
@@ -509,8 +498,8 @@ func TestGetMetricsConfig(t *testing.T) {
 			if err != nil {
 				t.Errorf("Wanted valid config %v, got error %v", test.expectedConfig, err)
 			}
-			if !reflect.DeepEqual(*mc, test.expectedConfig) {
-				t.Errorf("Wanted config %v, got config %v", test.expectedConfig, *mc)
+			if diff := cmp.Diff(test.expectedConfig, *mc, cmp.AllowUnexported(*mc), cmpopts.IgnoreTypes(mc.recorder)); diff != "" {
```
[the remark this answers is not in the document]
evankanderson: Since we're adding an anonymous function here, I think the best we could do would be to validate whether or not `recorder` was set.

I've added a new test, `TestStackdriverRecord`, to the bottom of `config_test.go` for now, but it should be moved to a stackdriver-specific module.

metrics/record.go
```diff
@@ -40,30 +38,7 @@ import (
 func Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) {
```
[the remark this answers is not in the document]
evankanderson: Removed all the complicated algorithm from here.

metrics/config_test.go
```diff
@@ -818,3 +810,76 @@ func TestNewStackdriverConfigFromMap(t *testing.T) {
...
+
+			Record(ctx, statCount.M(1))
+			if data.statCounter != 0 {
+				metricstest.CheckCountData(t, statCount.Name(), emptyTags, data.statCounter)
+			} else {
+				metricstest.CheckStatsNotReported(t, statCount.Name())
+			}
+		})
+	}
+}
```
mattmoor-sockpuppet: Add trailing newline:
```suggestion
}

```

metrics/config.go
```diff
@@ -193,19 +199,34 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
...
+		if ascmStr := m[AllowStackdriverCustomMetricsKey]; ascmStr != "" {
+			allowCustomMetrics, err = strconv.ParseBool(ascmStr)
 			if err != nil {
 				return nil, fmt.Errorf("invalid %s value %q", AllowStackdriverCustomMetricsKey, ascmStr)
 			}
-			mc.allowStackdriverCustomMetrics = ascmBool
+		}
+
+		if !allowCustomMetrics {
+			servingOrEventing := metricskey.KnativeRevisionMetrics.Union(
```
evankanderson: I optimized this a bit, instead of searching two sets on each Record, we store one larger set, which should be slightly more efficient. (One `log(n)` rather than `2*log(n/2)`.)

If we wanted to be even smarter, we could filter this set by `ms.stackdriverMetricTypePrefix`, but that seems less important than factoring this out of `config.go`.

metrics/config.go
evankanderson: I agree with this criticism; this is not the end state, but step 1 of N.

If you want, I can try to do this in bigger chunks. I knew you were working in this general area, and I didn't want to make too many disruptive changes at once (as it was, I got a merge on the rename of `stackdriverClientConfig` -> `StackdriverClientConfig`).

anniefu: Splitting out the changes into smaller sets makes sense to me. Just wanted to clarify where the end state of metricsConfig was going.

Especially when you know I'm about to cause merge conflicts :)

PR #814, opened 2019-10-26
Slack MessageHistory only returns messages sent by the current user

chizhg:
Slack MessageHistory should only return messages sent by the current user.

/cc @vagababov

### DIFF
--- a/test/mako/alerter/slack/message.go
+++ b/test/mako/alerter/slack/message.go
@@ -32,7 +32,7 @@ var minInterval = flag.Duration("min-alert-interval", 24*time.Hour, "The minimum
 
 const (
 	messageTemplate = `
-As of %s, there is a new performance regression detected from test automation for *%s*:
+As of %s, there is a new performance regression detected from test automation for %s:
 %s`
 )
 
@@ -86,16 +86,18 @@ func (smh *MessageHandler) SendAlert(testName, summary string) error {
 			); err != nil {
 				errCh <- fmt.Errorf("failed to retrieve message history in channel %q", channel.Name)
 			}
+			// decorate the test name for more accurate match
+			decoratedTestName := decoratedName(testName)
 			// do not send message again if alert for this test has been sent to
 			// the channel a short while ago
 			for _, message := range messageHistory {
-				if strings.Contains(message, testName) {
+				if strings.Contains(message, decoratedTestName) {
 					return
 				}
 			}
 
 			// send the alert message to the channel
-			message := fmt.Sprintf(messageTemplate, time.Now().UTC(), testName, summary)
+			message := fmt.Sprintf(messageTemplate, time.Now().UTC(), decoratedTestName, summary)
 			if err := helpers.Run(
 				fmt.Sprintf("sending message %q to channel %q", message, channel.Name),
 				func() error {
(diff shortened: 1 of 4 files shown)


test/slackutil/message_read.go
```diff
@@ -53,6 +53,8 @@ func NewReadClient(userName, tokenPath string) (ReadOperations, error) {
 	}, nil
 }
 
+// MessageHistory returns the list of messages sent by the user in the given
+// channel since the given startTime.
 func (c *readClient) MessageHistory(channel string, startTime time.Time) ([]string, error) {
 	u, _ := url.Parse(conversationHistoryURL)
 	q := u.Query()
```
vagababov: does it support query with the user name, so there is no need to filter on the receiving side?

chizhg: According to the API doc - https://api.slack.com/methods/conversations.history#response, it doesn't. I don't remember why I added `username` as one param, but it should not take any effect so just removed it. 
Also changed the alert logic a bit to put the min alert interval restriction for each test.

test/mako/alerter/slack/message.go
```diff
@@ -85,10 +86,14 @@ func (smh *MessageHandler) SendAlert(testName, summary string) error {
 			); err != nil {
 				errCh <- fmt.Errorf("failed to retrieve message history in channel %q", channel.Name)
 			}
-			// do not send message again if messages were sent on the same channel a short while ago
-			if len(messageHistory) != 0 {
-				return
+			// do not send message again if alert for this test has been sent to
+			// the channel a short while ago
+			for _, message := range messageHistory {
+				if strings.Contains(message, testName) {
```
chaodaiG: This doesn't seem like a very reliable way, is there a better way of extracting testName from message?

chizhg: Or we add a function to add brackets around the test name, and also use the result for checking? WDYT?

chaodaiG: That should work, would be nice to:
- have a unit test guard against brackets from somewhere else
- make sure test full name doesn't contain brackets

chizhg: Done.  Test full name won't contain brackets, and it will still work even if it contains.

test/mako/alerter/slack/message.go
```diff
@@ -85,12 +86,18 @@ func (smh *MessageHandler) SendAlert(testName, summary string) error {
...
 			}
-			// do not send message again if messages were sent on the same channel a short while ago
-			if len(messageHistory) != 0 {
-				return
+			// decorate the test name for more accurate match
+			decoratedTestName := decoratedName(testName)
+			// do not send message again if alert for this test has been sent to
+			// the channel a short while ago
+			for _, message := range messageHistory {
+				if strings.Contains(message, decoratedTestName) {
```
chaodaiG: do you need to worry about backward compatible?

chizhg: The last alert was sent **two weeks** ago, and the reason we check the message history is we don't want to sent to many alerts within a short time, which is set to **24 hours** here. So  it won't be impacted by the previously alert messages.

PR #818, opened 2019-10-28
Move hook-specific configuration options out of shared options.

mattmoor:
This builds on https://github.com/knative/pkg/pull/817 and makes further
breaking changes. The options pertinent to each admission controller are
now passed to their respective constructors, which leads to a cleaner
options struct, and better prepares for greater webhook diversity.

This is WIP until the baseline change merges.

### DIFF
--- a/webhook/config_validation_controller.go
+++ b/webhook/config_validation_controller.go
@@ -38,17 +38,21 @@ import (
 
 // ConfigValidationController implements the AdmissionController for ConfigMaps
 type ConfigValidationController struct {
+	// name of the ValidatingWebhookConfiguration
+	name string
+	// path that the webhook should serve on
+	path         string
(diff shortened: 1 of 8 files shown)


mattmoor:
cc @nimakaviani

webhook/config_validation_controller.go
```diff
@@ -38,17 +38,19 @@ import (
 
 // ConfigValidationController implements the AdmissionController for ConfigMaps
 type ConfigValidationController struct {
+	name         string
```
vagababov: Can we comment on what `name` and especially `path` mean here?

mattmoor: Done

nimakaviani:
thanks for cleaning up the options. it was getting really ugly.

PR #819, opened 2019-10-28
Accept a list of AdmissionControllers instead of a map

mattmoor:
WIP until https://github.com/knative/pkg/pull/818 lands

### DIFF
--- a/webhook/config_validation_controller.go
+++ b/webhook/config_validation_controller.go
@@ -62,6 +62,12 @@ func NewConfigValidationController(
 	return cfgValidations
 }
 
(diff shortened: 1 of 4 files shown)


nimakaviani:
neat! should have caught it when creating the interface :P

mattmoor:
Yeah, @nimakaviani I find myself cleaning up a slow creep of stuff that's accumulated over the past 2 years, I'm just trying to break it down into bite-sized reviews.  The bigger chunks are coming up, I'm just clearing out the nits I addressed along the way.

PR #820, opened 2019-10-28
Move WithContext into the resource admission controller.

mattmoor:
WIP until https://github.com/knative/pkg/pull/819 lands

PR #821, opened 2019-10-28
Streamline webhook.New

mattmoor:
This now pulls most of its configuration from context, which brings it a step closer to aligning with our `sharedmain` package.

This is WIP until https://github.com/knative/pkg/pull/820 merges.

### DIFF
--- a/webhook/context.go
+++ b/webhook/context.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


webhook/context.go
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2017 The Knative Authors
```
mattmoor-sockpuppet: Incorrect copyright year:
```suggestion
Copyright 2019 The Knative Authors
```

webhook/context.go
```diff
@@ -0,0 +1,39 @@
...
+
+import "context"
+
+// optionsKey is used as the key for associating information
+// with a context.Context.
+type optionsKey struct{}
+
+// WithOptions associates a set of webhook.Options with
+// the returned context.
+func WithOptions(ctx context.Context, opt Options) context.Context {
```
vagababov: The API seems inconsistent in that it gets by ref, but returns by pointer.

mattmoor: nil cannot be associated, but there may be nothing associated

PR #823, opened 2019-10-28
Switch to using GetCertificate in tls.Config

mattmoor:
GetCertificate allows us to start in TLS mode and dynamically fetch new certificates as they change.  This will eventually allow us to decouple the cert creation process from the core webhook logic, and in a subsequent change service this from a secret lister cache.

WIP until https://github.com/knative/pkg/pull/822 lands

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -142,16 +142,26 @@ func New(
 func (ac *Webhook) Run(stop <-chan struct{}) error {
 	logger := ac.Logger
 	ctx := logging.WithLogger(context.TODO(), logger)
-	tlsConfig, caCert, err := configureCerts(ctx, ac.Client, &ac.Options)
+
+	// TODO(mattmoor): Separate out the certificate creation process and use listers
+	// to fetch this from the secret below.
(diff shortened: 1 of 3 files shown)


webhook/webhook.go
```diff
@@ -148,16 +142,26 @@ func New(
...
 
 	server := &http.Server{
-		Handler:   ac,
-		Addr:      fmt.Sprintf(":%v", ac.Options.Port),
-		TLSConfig: tlsConfig,
+		Handler: ac,
+		Addr:    fmt.Sprintf(":%v", ac.Options.Port),
+		TLSConfig: &tls.Config{
+			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+				cert, err := tls.X509KeyPair(serverCert, serverKey)
```
vagababov: Do we want to cache this ? E.g. with once?

mattmoor: No because eventually this is going to be synthesized from the secret pulled from the informer cache (probably the next change).

PR #824, opened 2019-10-29
Restrict metric pkg Stackdriver Secret to controlled name & namespace

anniefu:
Remove config-observability values for "metrics.stackdriver-secret-name" and "metrics.stackdriver-secret-namespace". 

The functionality for different knative repos & controllers to specify different namespaces and names is left in by calling metrics.SetStackdriverSecretLocation, but the values are no longer user-configurable by ConfigMap.

Eventually when the metrics configurations and Secret-reading permissions are piped through to the user pod for queue-proxy, this will prevent malicious users from abusing Knative controller's admin abilities and using config-observability to give user pods access to Secrets in arbitrary namespaces.


metrics/stackdriver_exporter.go
```diff
@@ -53,8 +55,26 @@ var (
...
+	secretName = DefaultSecretName
+	// secretNamespace is the namespace to search for a k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
+	secretNamespace = DefaultSecretNamespace
 )
 
+// SetStackdriverSecretLocation allows setting non-default values for the name and namespace of
+// the Secret that can be used to authenticate with Stackdriver.
+// The Secret is only used if metricsConfig.stackdriverClientConfig.UseSecret is true.
+// This can only be called once.
+func SetStackdriverSecretLocation(name string, namespace string) {
```
garron: Why is it important that this only happens once?

What happens if this gets called after the stackdriver exporter is created? It appears as though unless the metrics config itself changes, the exporter won't get re-created.

metrics/stackdriver_exporter.go
```diff
@@ -40,8 +40,10 @@ const (
 	// defaultCustomMetricSubDomain is the default subdomain to use for unsupported metrics by monitored resource types.
 	// See: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#MetricDescriptor
 	defaultCustomMetricSubDomain = "knative.dev"
-	// secretNamespaceDefault is the namespace to search for a k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
-	secretNamespaceDefault = "default"
+	// DefaultSecretNamespace is the default namespace to search for a k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
+	DefaultSecretNamespace = "default"
+	// DefaultSecretName is the default name of the k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
+	DefaultSecretName = "stackdriver-service-account-key"
```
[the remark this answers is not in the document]

metrics/stackdriver_exporter.go
anniefu: It's not particularly important, but access to the state does need to be synchronized and thread-safe. So it was between a lock or a sync.Once. 

For security reasons, the values here shouldn't be user configurable and even controllers shouldn't really be modifying them throughout their lifetime, so sync.Once seemed appropriate.

metrics/stackdriver_exporter.go
```diff
@@ -40,8 +40,10 @@ const (
 	// defaultCustomMetricSubDomain is the default subdomain to use for unsupported metrics by monitored resource types.
 	// See: https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#MetricDescriptor
 	defaultCustomMetricSubDomain = "knative.dev"
-	// secretNamespaceDefault is the namespace to search for a k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
-	secretNamespaceDefault = "default"
+	// DefaultSecretNamespace is the default namespace to search for a k8s Secret to pass to Stackdriver client to authenticate with Stackdriver.
+	DefaultSecretNamespace = "default"
```
evankanderson: I'm not sure that "default" is a good option here, but putting this in `pkg` makes it hard to default to something like "knative-serving".

I'm not sure if there's an issue filed to figure out the best setting here and come back to this setting later.

metrics/stackdriver_exporter.go
evankanderson: I'm concerned with @garron's comment about not updating until the metrics config is re-read.

Do we have a way to regenerate the stackdriver client after this Once is done?

metrics/stackdriver_exporter.go
trshafer: Should this be system.Namespace()?
https://github.com/knative/pkg/blob/50398a29a78804bf5e9e64d937849c2747b41dd6/system/names.go#L30

metrics/stackdriver_exporter.go
```diff
@@ -169,15 +189,10 @@ func getStackdriverSecret(sdconfig *stackdriverClientConfig) (*corev1.Secret, er
 		return nil, err
 	}
 
-	ns := sdconfig.GCPSecretNamespace
-	if ns == "" {
-		ns = secretNamespaceDefault
-	}
-
-	sec, secErr := kubeclient.CoreV1().Secrets(ns).Get(sdconfig.GCPSecretName, metav1.GetOptions{})
+	sec, secErr := kubeclient.CoreV1().Secrets(secretNamespace).Get(secretName, metav1.GetOptions{})
```
trshafer: should this do any secret format validation?

metrics/stackdriver_exporter.go
anniefu: Since we're in knative/pkg which should be pretty generic, I don't think panicking is a good option if that environment variable isn't set.

metrics/stackdriver_exporter.go
anniefu: I'm not sure there's much to be done here, since we can't really validate the key itself. There is format validation in that any error getting/using the Secret including missing fields will get logged.

trshafer: Does this perform the validation of the structure of the secret? 

https://github.com/knative/pkg/blob/d62d6b389417178900372cc232c622c4b9c91e93/metrics/stackdriver_exporter.go#L203

metrics/stackdriver_exporter.go
anniefu: The intention here are that these values are essentially hardcoded, with the exception that different Knative consumers may want to set different values (i.e. "knative-serving" vs "knative-eventing"). If non-default options are not sufficient, consumers should set this function before beginning to export metrics, though I can add a comment to clarify that.

The set once is to emphasize that these should effectively be const values, we can't have external agents re-configuring the Secret name/namespace because of the queue-proxy security issue mentioned in the PR description.

metrics/stackdriver_exporter.go
anniefu: Added a descriptive error message of the secretDataFieldKey is missing that will get passed back up and logged.

metrics/stackdriver_exporter.go
anniefu: I switched it to a Lock to give more flexibility. pkg consumers can call UpdateExporter() after setting the secret location if they wish.

Users cannot configure the Secret location.

metrics/stackdriver_exporter.go
anniefu: I think leaving DefaultSecretNamespace as "default" because that should really be the default for any namespace value that can't be predetermined. 

If the consuming package doesn't overwrite the namespace, the pod will just attempt to read a Secret out of the "default" namespace, which will get rejected if the pod doesn't have access to read Secrets anyway.

metrics/stackdriver_exporter.go
```diff
@@ -169,23 +193,21 @@ func getStackdriverSecret(sdconfig *stackdriverClientConfig) (*corev1.Secret, er
...
```
yanweiguo: Don't we need to lock before reading?

anniefu: Yeah, good catch!

metrics/stackdriver_exporter.go
evankanderson: What about defaulting to the empty string and then lazily reading system.Namespace() if this value is empty?

anniefu:
> I'm still concerned that it appears that we only refresh the stackdriver export options on the following path:
> 
> UpdateExporterFromConfigMap
> -> UpdateExporter
> -> newMetricsExporter
> -> newStackdriverExporter
> -> getStackdriverExporterClientOptions
> 
> This means that any change to the Secret won't be picked up until the ConfigMap is updated, IIUC.

Yes, that's correct. The long term solution is to introduce a "SecretWatcher" similar to how we have "ConfigMapWatchers" that could also trigger the update. However, that's a non-trivial change outside of the metrics pkg and seems outside of the scope of the MVP for this feature.

In the meantime, there is a user workaround to toggle a "metrics.stackdriver-*" to trigger an update to the metrics exporter if there is a change to the Secret.

metrics/stackdriver_exporter.go
anniefu: That still leaves the possibility of the metrics pkg causing a panic unnecessarily if that env is missing. Also, I believe empty string is a valid namespace string that conflates to "default" for kubeclient APIs. Consuming packages should be allowed to set the secretNamespace to empty string. Consuming packages can also set the namespace to system.Namespace().

Is the concern here that metrics pkg consumers will accidentally trigger a Secret read in the default namespace without being aware of it?

anniefu: Also, we cannot allow reading an env variable because that is user-configurable which will be a security issue for knative/serving when it starts to propagate the Secret specified here to queue-proxy.

anniefu: Also, we cannot allow reading an env variable because that is user-configurable which will be a security issue for knative/serving when it starts to propagate the Secret specified here to queue-proxy.

anniefu: ok I've added a metrics pkg internal bool for "useStackdriverSecretEnabled" that prevents any attempts to read the Secret until SetStackdriverSecretLocation is called.

More details in the new description for SetStackdriverSecretLocation.

This means that the DefaultSecretName & DefaultSecretNamespace values are never used unless explicitly re-specified by the consuming package. (They are still provided so that packages don't all make up a different Secret name).

evankanderson:
/approve

Using the TODO as a lever to get some cleanup in here eventually, but not block current improvements.

PR #825, opened 2019-10-29
Drive GetCertificate from the lister cache of a secret informer.

mattmoor:


vagababov:
This requires a main.go change as well, right?
/lgtm
/hold

mattmoor:
Yes, there are many changes that will be required once we resume importing PKG.  My plan is to fold everything together into `sharedmain` and kill 100% of the current `main.go` boilerplate

PR #832, opened 2019-10-30
Create a new singleton Reconciler for the webhook secret.

mattmoor:
This change creates a new Reconciler (not yet hooked up) that ensures that the
webhook secret has the appropriate shape.  I call this a "singleton Reconciler"
because this reconciler exists to effectively reconcile a single resource, which
will be a pattern for the webhook's reconcilers.


webhook/certificates/certificates.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		logger.Errorf("error accessing certificate secret %q: %v", r.secretName, err)
+		return err
+	}
+
+	if _, haskey := secret.Data[certresources.ServerKey]; !haskey {
+		logger.Infof("certificate secret %q is missing key %q", r.secretName, certresources.ServerKey)
```
vagababov:
```suggestion
		logger.Infof("Certificate secret %q is missing key %q", r.secretName, certresources.ServerKey)
```
Same below.

webhook/certificates/certificates_test.go
```diff
@@ -0,0 +1,204 @@
...
+*/
+
+package certificates
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	// Fake injection stuff
```
vagababov:
```suggestion
```

mattmoor:
This should be RFAL
/lint

Issue #782, opened 2019-10-20
Allow dynamic reconfiguration of webhooks


mattmoor:
tl;dr I'd like to be able to use our webhook infrastructure to process dynamically discovered implementations of a give duck type.

[Here's an example](https://github.com/mattmoor/boo-maps/blob/5ed68ca4e0f45e01d45ccebaddefecea263b6f9d/cmd/webhook/main.go#L122-L131) that uses a duck type with our webhook (not dynamic), but the last entry clearly hints at why being able to dynamically reconfigure the webhook would be interesting (though it is unclear what it should key off of for discovery).

[Here's another example](https://github.com/mattmoor/foo-binding#extensibility), where the method of dynamic discovery is clearer.  Essentially, as new types are referenced from binding resources, the webhook would adapt to process them.

mattmoor:
This is going to sounds somewhat like a hammer seeing everything as nails, but my initial take on this is that we should push the currently very ad hoc webhook registration process towards the reconciler pattern.  This would naturally give us a pattern that lets the webhook respond and adapt to changes.

Related to https://github.com/knative/pkg/issues/450 (and it's deeper sentiment) we should refactor the webhook to tease apart a few logical parts to break up the monolithic webhook package to try to fit this pattern.

mattmoor:
Note that this gets more interesting (and the logical parts clearer) as the number and kind of webhooks expand.  We now have Mutating for types (defaulting, validation), Validating for configmaps (validation), and should soon have Conversion for types.

vagababov:
How do we do validation and the defaulting the currently a rely on explicitly calling the methods on the types.

nimakaviani:
Defaulting can probably be encoded in the same resource that holds the reference bindings, eg by pointing to a config map or providing key/values?

But to @vagababov 's point, what about validation on fields that fall outside what is validated with the duck type?

n3wscott:
We could bring in the OpenAPI of the duck and use that as a first pass.

Issue #529, opened 2019-07-15
RegistrationDelay in webhook.go


vincent-pli:
About the code in `pkg/webhook/webhook.go`
```
	select {
	case <-time.After(ac.Options.RegistrationDelay):
		cl := ac.Client.AdmissionregistrationV1beta1().MutatingWebhookConfigurations()
		if err := ac.register(ctx, cl, caCert); err != nil {
			logger.Errorw("failed to register webhook", zap.Error(err))
			return err
		}
		logger.Info("Successfully registered webhook")
	case <-stop:
		return nil
	}

	serverBootstrapErrCh := make(chan struct{})
	go func() {
		if err := server.ListenAndServeTLS("", ""); err != nil {
			logger.Errorw("ListenAndServeTLS for admission webhook returned error", zap.Error(err))
			close(serverBootstrapErrCh)
		}
	}()
```

The `time.After(ac.Options.RegistrationDelay)` is to used for delay the  `webhook` registration,
hope to avoid condition race between registration and http server readiness, am i right?

So, I think the ` server.ListenAndServeTLS("", "")` should placed before the `ac.Client.AdmissionregistrationV1beta1().MutatingWebhookConfigurations()`
since the `time.After(ac.Options.RegistrationDelay)` is a block wait.

is this on purpose or I missing something, thanks
## Additional Info
The `delay` can not totally resolve the condition race, any plan to resolve it?

Issue #450, opened 2019-06-08
Design / Implement webhook shared main


mattmoor:
Designed a "sharedmain" method for our webhooks to cut down on boilerplate.

mattmoor:
Ideally this would also enable us to separate certificate provisioning from the actual webhook serving.

vincent-pli:
@mattmoor 
Is this issue resolved, about `webhook`, `injection controller`?

mattmoor:
Nope, we still need a way to share significantly more of what's in `./cmd/webhook/main.go` in each of the repos.

vincent-pli:
Thanks, expect to see the design or discussion

Issue #141, opened 2018-10-26
Refactoring the creating cert process in webhook


zrss:
## Expected Behavior

i can easily add more DNSNames or IP addresses in **Subject Alternate Name values** of Webhook's Cert

currently, we only support the fixed {serviceName}.{namespace} DNSNames values

## Actual Behavior

i expose my webhook server at a IP:PORT address (for fiting the knative to my lab env :neutral_face:), and this forces me to change the all func params in the call chain of creating a cert

## Additional Info

 we will go through this chain when we create a cert,

generateSecret (options) -> 

1. CreateCerts (serviceName, namespace)
1. createCA (serviceName, namespace)
1. createCACertTemplate (serviceName, namespace)
1. createCertTemplate (serviceName, namespace)

and we generate the DNSNames at the final step, i.e. createCertTemplate(serviceName, namespace)

https://github.com/knative/pkg/blob/5c809f170b42301a276080f4fd884a30188ca1a5/webhook/certs.go#L42-L52

we have totally 4 steps to generate the cert, but serviceName and namespace just be paased through in first 3 steps

so i propose **refactoring**

1. CreateCerts/createCA/createCACertTemplate/createCertTemplate  accept a string slice of **DNSName** and a IP slice of **IPAdress** (for my case)
1. Generate DNSNames of webhook svc in `generateSecret` func (as it's the first place that has `options`)

kindly ping @mattmoor , @grantr , @tcnghia for thoughts

PR #833, opened 2019-10-30
This refactors our core webhook logic to be reconciler-based.

mattmoor:
This is the culmination of a large number of changes to refactor our webhook logic, and adopt a reconciler-based approach to make it resilient to unexected system events (e.g. rogue GCs!).

For more details on how this is consumed, see the new `webhook/README.md`.

Fixes: https://github.com/knative/pkg/issues/782
Closes: https://github.com/knative/pkg/issues/529
Fixes: https://github.com/knative/pkg/issues/450
Related: https://github.com/knative/pkg/issues/141

/hold
... as I stage downstream changes.

### DIFF
--- a/webhook/README.md
+++ b/webhook/README.md
@@ -1,0 +1,101 @@
+## Knative Webhooks
+
+Knative provides infrastructure for authoring webhooks under
+`knative.dev/pkg/webhook` and has a few built-in helpers for certain
+common admission control scenarios.  The built-in admission controllers
+are:
+1. Resource validation and defaulting (builds around `apis.Validatable`
+  and `apis.Defaultable` under `knative.dev/pkg/apis`).
+2. ConfigMap validation, which builds around similar patterns from
+  `knative.dev/pkg/configmap` (in particular the `store` concept)
+
+To illustrate standing up the webhook, let's start with one of these
+built-in admission controllers and then talk about how you can write
+your own admission controller.
+
+## Standing up a Webhook from an Admission Controller
+
+We provide facilities in `knative.dev/pkg/injection/sharedmain` to try and
+eliminate much of the boilerplate involved in standing up a webhook. For this
+example we will show how to stand up the webhook using the built-in admission
+controller for validating and defaulting resources.
+
+The code to stand up such a webhook looks roughly like this:
+
+```go
+// Create a function matching this signature to pass into sharedmain.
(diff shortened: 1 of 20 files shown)


webhook/README.md
```diff
@@ -0,0 +1,101 @@
...
+controller for validating and defaulting resources.
+
+The code to stand up such a webhook looks roughly like this:
+
+```
+// Create a function matching this signature to pass into sharedmain.
+func NewResourceAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	return resourcesemantics.NewAdmissionController(ctx,
+		// Name of the resource webhook (created via yaml)
+		fmt.Sprintf("resources.webhook.%s.knative.dev", system.Namespace()),
```
vagababov: perhaps show the example with s "your-name-here" manner?

webhook/README.md
```diff
@@ -0,0 +1,101 @@
...
+		// Whether to disallow unknown fields when parsing the resources' JSON.
+		true,
+	)
+}
+
+func main() {
+	// Set up a signal context with our webhook options.
+	ctx := webhook.WithOptions(signals.NewContext(), webhook.Options{
+        // The name of the Kubernetes service selecting over this deployment's pods.
+		ServiceName: "webhook",
```
vagababov: Formatting seems off here and below.
Also same naming here? Would help to eradicate the comment.

webhook/README.md
mattmoor: this is literally copied our of `knative.dev/sample-controller` 🤷‍♂

webhook/README.md
mattmoor: Fixed, tabs vs. spaces...

webhook/README.md
vagababov: I know, but if this is a walkthrough it's useful to have all the items that need customization to be called out
🤷‍♂

webhook/webhook.go
```diff
@@ -177,34 +168,8 @@ func (ac *Webhook) Run(stop <-chan struct{}) error {
 	}
 
 	logger.Info("Found certificates for webhook...")
-	if ac.Options.RegistrationDelay != 0 {
```
n3wscott: what happened to all the registration delay code?

mattmoor: it'd gone 💥 

A delay makes some sense when it's a one-time thing, but it's effectively continuous now.

PR #836, opened 2019-10-31
Fixed knative activator handshake timeout issue

nanjj:
The original handshake timeout is too short to dial connection
successfully under some situation. So the activator pod can not be
active. And the log message just say "Failed to send ping message to
ws://autoscaler.knative-serving.svc.cluster.local:8080", not easy to
figure out what's the problem underneeth.

The fix switch to websocket default dial, so the default handle shake
time is 45 seconds, and once dial error found, the error will be
reported as error.

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -106,10 +106,11 @@ func NewDurableSendingConnection(target string, logger *zap.SugaredLogger) *Mana
 // go func() {for range messageChan {}}
 func NewDurableConnection(target string, messageChan chan []byte, logger *zap.SugaredLogger) *ManagedConnection {
 	websocketConnectionFactory := func() (rawConnection, error) {
-		dialer := &websocket.Dialer{
-			HandshakeTimeout: 3 * time.Second,
-		}
+		dialer := websocket.DefaultDialer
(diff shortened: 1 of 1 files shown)


nanjj:
@googlebot I signed it!

websocket/connection.go
```diff
@@ -106,10 +106,11 @@ func NewDurableSendingConnection(target string, logger *zap.SugaredLogger) *Mana
 // go func() {for range messageChan {}}
 func NewDurableConnection(target string, messageChan chan []byte, logger *zap.SugaredLogger) *ManagedConnection {
 	websocketConnectionFactory := func() (rawConnection, error) {
-		dialer := &websocket.Dialer{
-			HandshakeTimeout: 3 * time.Second,
-		}
+		dialer := websocket.DefaultDialer
 		conn, _, err := dialer.Dial(target, nil)
+		if err != nil {
+			logger.Error(err)
```
markusthoemmes: Let's be a little more descriptive here, something like

```suggestion
			logger.Errorw("Websocket connection could not be established", zap.Error(err))
```

PR #841, opened 2019-11-01
add EnqueueSentinel to pkg/controller

### LINKED ISSUE #839: webhook reconcilers don't properly deduplicate and throttle
/area API
/kind bug
/kind good-first-issue

The webhook reconcilers use `c.Enqueue` to queue work when the informer alerts them that something has changed, e.g. [here](https://github.com/knative/pkg/blob/d4ce00139499bd3ea11cdb18ef20a41bdb248245/webhook/resourcesemantics/controller.go#L71).

However, done like this we might actually get a queue event from each of the informers enqueuing the MWH name and the secret name, and these keys may get processed in parallel.  In general the workqueue tries to be smart about avoiding the same key being processed on multiple threads.

This issue is still purely theoretical, but for this class of singleton controller we should expose an `EnqueueSingleton` or `EnqueueSentinel` member on `controller.Impl` that enqueues a fixed key that we can use in these cases.

maisem:
This changes introduces Impl.EnqueueSentinel which returns an enqueue function which only ever enqueues a single predefined key.

Fixes #839.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -186,6 +186,14 @@ func (c *Impl) Enqueue(obj interface{}) {
 	c.EnqueueKey(types.NamespacedName{Namespace: object.GetNamespace(), Name: object.GetName()})
 }
 
+// EnqueueSentinel returns a Enqueue method which will always enqueue a
+// predefined key instead of the object key.
+func (c *Impl) EnqueueSentinel(k types.NamespacedName) func(interface{}) {
+	return func(_ interface{}) {
+		c.EnqueueKey(k)
+	}
+}
+
 // EnqueueControllerOf takes a resource, identifies its controller resource,
 // converts it into a namespace/name string, and passes that to EnqueueKey.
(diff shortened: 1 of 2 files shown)


maisem:
@googlebot I signed it!

controller/controller.go
```diff
@@ -186,6 +186,14 @@ func (c *Impl) Enqueue(obj interface{}) {
 	c.EnqueueKey(types.NamespacedName{Namespace: object.GetNamespace(), Name: object.GetName()})
 }
 
+// EnqueueSentinel returns a Enqueue method which will always enqueue a
+// predefined key instead of the object key.
+func (c *Impl) EnqueueSentinel(k types.NamespacedName) func(interface{}) {
+	return func(_ interface{}) {
```
vagababov:
```suggestion
	return func(interface{}) {
```
I think this should work as wll.

maisem: Yep. Done.

PR #843, opened 2019-11-04
Avoid race on "server-key.pem" in `webhook.Run`.

evankanderson:
@houshengbo reported webhook failures while running the operator test of the form:

```json
{
  "level":"error",
  "ts":"2019-11-04T17:52:51.002Z",
  "logger":"webhook",
  "caller":"webhook/webhook.go:160",
  "msg":"could not configure admission webhook certs",
  "commit":"33e4197",
  "knative.dev/controller":"webhook",
  "error":"server key missing",
  "stacktrace":"knative.dev/serving/vendor/knative.dev/pkg/webhook.(*Webhook).Run
      /Users/vincent/go/src/knative.dev/serving/vendor/knative.dev/pkg/webhook/webhook.go:160
main.main.func2
     /Users/vincent/go/src/knative.dev/serving/cmd/webhook/main.go:198
knative.dev/serving/vendor/golang.org/x/sync/errgroup.(*Group).Go.func1
     Users/vincent/go/src/knative.dev/serving/vendor/golang.org/x/sync/errgroup/errgroup.go:58"
}
{
  "level":"error",
  "ts":"2019-11-04T17:52:51.003Z",
  "logger":"webhook",
  "caller":"webhook/main.go:209",
  "msg":"Error while running server",
  "commit":"33e4197",
  "knative.dev/controller":"webhook",
  "error":"server key missing",
  "stacktrace":"main.main
     /Users/vincent/go/src/knative.dev/serving/cmd/webhook/main.go:209
runtime.main
     /usr/local/go/src/runtime/proc.go:203"
}
```

Debugging, it looks like knative-serving defines the `webhook-certs` Secret as:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: webhook-certs
  namespace: knative-serving
  labels:
# The data is populated at install time.
```

Unfortunately, the "install time" population is done by the webhook controller by calling `certificates.NewController`:

https://github.com/knative/serving/tree/master/cmd/webhook/main.go#L138

Which is then batched with other controller startup in `controller.StartAll`:

https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#L186

Code then proceeds to start the Webhooks with `Webhook.Run()` while the controllers start to initialize. Unfortunately, if the Webhooks return an error, we call `logger.Fatalw()` with the error:

https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#L201

And the Webhook code will return an error if the secret doesn't exist in the Lister or if it doesn't contain the server key or server cert created by the `certificates.NewController`'s `Reconcile()` method:

https://github.com/knative/pkg/blob/master/webhook/webhook.go#L144

This change causes construction of the new controllers on [lines 154-162 of sharedmain/main.go](https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#154) to initialize a (package-shared global) `chan struct{}` which will be closed *after* the first reconcile loop of `certificates.reconciler.Reconcile()`.

I looked at the current webhook test and wasn't easily able to figure out how to test the await functionality there -- suggestions welcome.

### DIFF
--- a/webhook/certificates/certificates.go
+++ b/webhook/certificates/certificates.go
@@ -25,6 +25,7 @@ import (
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/system"
+	"knative.dev/pkg/webhook"
 	certresources "knative.dev/pkg/webhook/certificates/resources"
 )
 
@@ -39,6 +40,13 @@ var _ controller.Reconciler = (*reconciler)(nil)
 
 // Reconcile implements controller.Reconciler
 func (r *reconciler) Reconcile(ctx context.Context, key string) error {
+	defer func() {
+		s := webhook.SecretsReconciled
+		webhook.SecretsReconciled = nil
+		if s != nil {
+			close(s)  // Notify current listeners that the channel is closed.
+		}
+	}()
 	return r.reconcileCertificate(ctx)
 }
 
--- a/webhook/certificates/certificates_test.go
+++ b/webhook/certificates/certificates_test.go
@@ -200,4 +200,8 @@ func TestNew(t *testing.T) {
 	if c == nil {
 		t.Fatal("Expected NewController to return a non-nil value")
 	}
+
+	if webhook.SecretsReconciled == nil {
+		t.Fatal("Expected NewController to initialize webhook.SecretsReconciled")
+	}
 }
--- a/webhook/certificates/certificates_test.go
+++ b/webhook/certificates/certificates_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"testing"
+	"time"
 
 	kubeclient "knative.dev/pkg/client/injection/kube/client/fake"
 	_ "knative.dev/pkg/client/injection/kube/informers/core/v1/secret/fake"
@@ -37,6 +38,21 @@ import (
 	. "knative.dev/pkg/reconciler/testing"
 	. "knative.dev/pkg/webhook/testing"
 )
+
+var blockingFinished = []func(*testing.T, *TableRow){
+	func(t *testing.T, _ *TableRow) {
+		if s := webhook.SecretsReconciled; s != nil {
+			// We expect webhook.SecretsReconciled to be set to `nil` after
+			// reconciliation, so this should be the main case.
+			select {
+			case _ = <- webhook.SecretsReconciled:
+				// Expected, webhook.SecretsReconciled is closed or nil.
+			case <-time.After(10 * time.Millisecond):
+				t.Errorf("webhook.SecretsReconciled blocked unexpectedly: %+v", webhook.SecretsReconciled)
+			}
+		}
+	},
+}
 
 func TestReconcile(t *testing.T) {
 	secretName, serviceName := "webhook-secret", "webhook-service"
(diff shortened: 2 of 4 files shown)


webhook/certificates/certificates_test.go
```diff
@@ -38,6 +39,21 @@ import (
 	. "knative.dev/pkg/webhook/testing"
 )
 
+var blockingFinished = []func(*testing.T, *TableRow){
+	func(t *testing.T, _ *TableRow) {
+		if s := webhook.SecretsReconciled; s != nil {
+			// We expect webhook.SecretsReconciled to be set to `nil` after
+			// reconciliation, so this should be the main case.
+			select {
+			case _ = <- webhook.SecretsReconciled:
```
vagababov:
```suggestion
			case <- s:
```
I guess the channel might become nil, but it seems they're in the same func, so 🤷‍♂ .
But definitely don't need `_ := `.

webhook/webhook.go
```diff
@@ -136,6 +142,14 @@ func (ac *Webhook) Run(stop <-chan struct{}) error {
 	logger := ac.Logger
 	ctx := logging.WithLogger(context.Background(), logger)
 
+	// If certificates are being populated, make sure that the certificate
+	// controller has had at least one chance to reconcile before starting the
+	// HTTP server with a cert that might not yet exist.
+	// Note that we use the assignment version of `if` to avoid a double-read.
+	if s := SecretsReconciled; s != nil {
+		_ = <- s
```
vagababov:
```suggestion
		<- s
```

webhook/certificates/certificates.go
```diff
@@ -39,6 +40,13 @@ var _ controller.Reconciler = (*reconciler)(nil)
 
 // Reconcile implements controller.Reconciler
 func (r *reconciler) Reconcile(ctx context.Context, key string) error {
+	defer func() {
```
vagababov: This should probably be `sync.Once` instead.

evankanderson:
Unfortunately @houshengbo indicated this didn't actually solve the problem.

I have 2 kubecon day zero presentations to work on today, so I probably wont't get a chance to poke at this before Monday.

My next thought is to add a 2-5 second wait in the webhook start for the secret to become available (by using a `select` between a channel that will notify when ready and a timeout, not a sleep).

houshengbo:
Thanks to @evankanderson for working on this.

To reproduce this issue, you have to go thru serving-operator:
1. Download the source code of serving-operator.
2. Use ./hack/generate.yaml.sh in serving to generate the latest serving/yaml. Then replace the yaml files under cmd/manager/kodata/knative-serving in serving-operator. The name of the yaml file does not matter.
3. After that, launch the operator from source code, with `ko apply -f config/` to deploy the CR as in here: 
```
cat <<-EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
 name: knative-serving
---
apiVersion: operator.knative.dev/v1alpha1
kind: KnativeServing
metadata:
  name: knative-serving
  namespace: knative-serving
spec:
  config:
    defaults:
      revision-timeout-seconds: "300"  # 5 minutes
    autoscaler:
      stable-window: "60s"
    deployment:
      registriesSkippingTagResolving: "ko.local,dev.local"
    logging:
      loglevel.controller: "debug"
EOF
```

evankanderson:
It turns out that this was missing functionality in the serving-operator, which was installing the webhook into a namespace other than `knative-serving`, but the webhook YAML was pointing at a service reference in `knative-serving`. https://github.com/jcrossley3/manifestival/pull/22 should fix this by overriding the namespace in the webhook configuration when installing.

PR #846, opened 2019-11-05
Eventing StackDriver internal metrics

nachocano:
- Making eventing metrics internal for stackdriver as we will tag them with container and pod names to avoid problems when reporting the same metric from different pods/containers. Doing so for KnativeBroker and KnativeTrigger only.
- Added some constants.
- Note that this shouldn't be here. There are TODOs in the code with issue numbers to do a proper refactor and move eventing and serving specific metrics out of pkg.
- Removing eventSource and filterSource tags from broker/trigger stuff to keep the cardinality manageable. Leaving it for KnativeSource for now.

- We should cherry-pick this for 0.10.


metrics/metricskey/constants_eventing.go
```diff
@@ -59,16 +59,16 @@ var (
 		LabelLocation,
 		LabelClusterName,
 		LabelNamespaceName,
-		LabelName,
```
yanweiguo: Is `LabelName` still used somewhere? Could we remove its definition?

metrics/stackdriver_exporter_test.go
```diff
@@ -185,15 +185,11 @@ func TestGetMonitoredResourceFunc_UseKnativeBroker(t *testing.T) {
 		if got != testEventType {
 			t.Errorf("expected new tag: %v, got: %v", eventTypeKey, newTags)
 		}
-		got = getResourceLabelValue(metricskey.LabelEventSource, newTags)
```
yanweiguo: Do you want to check `LabelSourceName` instead?

metrics/metricskey/constants_eventing.go
nachocano: I'll rollback the change on sources. If I do so, then nothing should break in other repos, and will make it easy for cherry-picking to 0.10.
So, yeah, this Name will be used.

metrics/stackdriver_exporter_test.go
nachocano: Yeah, but I'll change it to Name.

Thanks for taking a look! Will ping you tmrw for another round :)

nachocano: this label has been dropped from KnativeBroker, so it's fine not to check this

PR #850, opened 2019-11-05
Add fake functions for GCP metadata and Stackdriver back

yanweiguo:
The fake functions were removed in #794. Sorry I didn't catch that. We need those fake metrics for unit tests instead of connecting to real GCP service.


metrics/exporter_test.go
```diff
@@ -95,6 +95,9 @@ func getResourceLabelValue(key string, tags []tag.Tag) string {
 
 func TestMain(m *testing.M) {
 	resetCurPromSrv()
+	// Set gcpMetadataFunc and newStackdriverExporterFunc for testing
```
anniefu: Since these are really only for Stackdriver exporter, could you move this to stackdriver_exporter_test.go?

metrics/stackdriver_exporter_test.go
```diff
@@ -134,6 +135,19 @@ var (
 	}}
 )
 
+func fakeGcpMetadataFun() *gcpMetadata {
```
anniefu: [nit] Missing "c" in Func

metrics/exporter_test.go
yanweiguo: We can't have multiple `TestMain` func.

metrics/stackdriver_exporter_test.go
yanweiguo: Done

metrics/exporter_test.go
anniefu: oh, darn :(

PR #853, opened 2019-11-07
Register the dynamic informer factory

lionelvillard:
See https://github.com/knative/eventing/issues/2154 for context.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -971,12 +971,14 @@
   version = "kubernetes-1.15.3"
 
 [[projects]]
-  digest = "1:f3d53e8c6ee1c0312bbf66e5a98aee44288388a19e8dc164d8beeeae0b32a925"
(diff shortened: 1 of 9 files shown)


lionelvillard:
@Harwayne @nachocano @aslom

vaikas:
Can we not use the existing Duck based stuff to achieve what you're trying to do? Based on the issue we should be able to use the PodSpecable with the existing duck machinery. Or, are you trying to then later on add the concrete types as just go structs without vendoring the serving tree? If so, what's the plan for actually creating the resources?

Just trying to understand the plan here.

n3wscott:
Checkout `+genduck` https://github.com/knative/pkg/blob/43ca049cdbdad4892b0c926e7c6536d064efde72/apis/duck/v1/addressable_types.go#L30

It might be what you really want.

lionelvillard:
Ah yes genduck. I'll look at it, thanks.

PR #856, opened 2019-11-08
Change test log to avoid it being highlighted

chizhg:
I had been asked a few times "why the cluster creation failed in the test", see https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_pkg/854/pull-knative-pkg-unit-tests/1192510387593940992.

Change the test log message a bit to avoid it being highlighted..

/cc @chaodaiG


test/gke/wait.go
```diff
@@ -49,7 +49,7 @@ func Wait(gsc SDKOperations, project, region, zone, opName string, wait time.Dur
 		select {
 		// Got a timeout! fail with a timeout error
 		case <-timeout:
-			return errors.New("timed out waiting")
+			return errors.New("got time out waiting")
```
vagababov: What is "highlighted" and how does this help?

chaodaiG: there are cases where we do want to see this error when we use this package for cluster creation, if not we should reconsider whether to surface this error at all

chaodaiG: @vagababov , "highlighted" is a term of spyglass, it highlights certain error patterns as defined here: https://github.com/knative/test-infra/blob/b6636d634ad40bb2bc99449b3f6c76ab70d4ae67/ci/prow/config.yaml#L57

vagababov: I see. Thanks.

chizhg: > there are cases where we do want to see this error when we use this package for cluster creation, if not we should reconsider whether to surface this error at all

IMO this kind of error should be matched by a specific regex we set in Spyglass config, like `failed creating test cluster`. Making the error message matched with a generic regex and highlighted is somewhat confusing to the users, since it looks like a real "cluster error".

coryrc: /cc @coryrc 
Chi says he wants to keep "timed out" as a Spyglass term because integration tests sometimes have it.
So I suggested we add the new, more specific regex to Spyglass first, then change this error message like this. What do you think?

chaodaiG: IMO "timed out" is not a strict enough regex for error matching, and not strict enough error matching was part of the reason why we had > 200 highlighted lines on a succeeded run before spyglass. I would like to see more strict regex matching for all possible serving/eventing tests (i.e. "timed out waiting for service") instead of changing the other error messages to not using it, what if there is a 3rg party code that we can't change uses "timed out"?

chizhg: The main reason that I want to change this log message is it shows `timed out` in cluster creation, which is adding more confusion, other `timed out` log should not be that confusing.

But I'm okay with more strict regex matching for it instead of making changes here, https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_serving/5760/pull-knative-serving-integration-tests/1184527739000983552 is one example we discussed before, how about changing to `timed out waiting for the condition`?

chaodaiG: I think `timed out waiting for the condition` is better than `timed out`, still pretty generic though

chizhg: That is true, but I don't think we will be able to have even more strict regex matching, unless we define a logging standard that forces everyone to follow. The similar case also applies to the monitoring configuration.

chizhg: Updated in https://github.com/knative/test-infra/pull/1550.

/close

PR #857, opened 2019-11-08
Add Namespace index to listers in tests.

markusthoemmes:
Not having this causes the listers to log an annoying log message like

```
W1108 02:48:54.848573  211995 listers.go:77] can not retrieve list of objects using index : Index with name namespace does not exist
```

That message doesn't cause the tests to fail because the cache logic works around by doing a "slow indexing". I was very confused by this log though and sunk a bit of time into debugging it, so here we go.


markusthoemmes:
Unrelated.

vagababov:
🎉 
/lgtm
/approve

Issue #859, opened 2019-11-09
Extend the tracker to support label selectors


mattmoor:
Today the "tracker" supports having an object `Track()` another object given a `corev1.ObjectReference` with its exact coordinates (of sorts).

Another interesting use-case is to track a label-selected collection, e.g.
```
tracker.Track(tracker.Reference{
   APIVersion: "v1",
   Kind: "Pod",
   Namespace: "default",
   Selector: map[string]string{
       "foo": "bar",
   },
})
```

I'd propose the following breakdown of work to stage this change:
1. Introduce `tracker.Reference` that is a derivative of `corev1.ObjectReference` capturing the fields relevant to today's `tracker.Track` semantics, and migrating downstream repositories to this new variant.

2. Introduce `Selector *metav1.LabelSelector` to `tracker.Reference` to capture inexact matching semantics, and implement the semantics in the provided `tracker.Interface` implementations.

/assign

PR #860, opened 2019-11-09
Expand the tracker interface to include a variant with its own type.

mattmoor:
`TrackReference` is the same as `Track`, but takes a `tracker.Reference` instead.  This type has been seeded with the subset of `corev1.ObjectReference` that the tracker currently consumes / supports, but the intention is to expand this type to allow inexact references that (optionally) use label selectors in place of name to reference objects.

See also: https://github.com/knative/pkg/issues/859

/hold
for the downstream changes to be staged.

### DIFF
--- a/reconciler/testing/tracker.go
+++ b/reconciler/testing/tracker.go
@@ -20,7 +20,6 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
-
 	"knative.dev/pkg/tracker"
 )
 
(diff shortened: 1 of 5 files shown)


mattmoor:
I think the current use of `ObjectReference` is a bit too entrenched in knative/eventing for the plan I had originally outlined.  I think instead I'm going to expand the API with a new method that takes `tracker.Reference`, and leave the current method alone.

mattmoor:
/hold cancel

I don't think this is any longer breaking for downstream repos, so no need to stage things.

PR #862, opened 2019-11-11
Avoid callback calls while lock is held.

mattmoor:
This came up here: https://github.com/knative/pkg/pull/861#discussion_r344574547

/assign @vaikas

### DIFF
--- a/tracker/enqueue.go
+++ b/tracker/enqueue.go
@@ -131,6 +131,13 @@ func (i *impl) TrackReference(ref Reference, obj interface{}) error {
 	key := types.NamespacedName{Namespace: object.GetNamespace(), Name: object.GetName()}
 
 	i.m.Lock()
+	// Call the callback without the lock held.
+	var keys []types.NamespacedName
+	defer func(cb func(types.NamespacedName)) {
+		for _, key := range keys {
+			cb(key)
+		}
+	}(i.cb) // read i.cb with the lock held
 	defer i.m.Unlock()
 	if i.exact == nil {
 		i.exact = make(map[Reference]set)
@@ -158,7 +165,7 @@ func (i *impl) TrackReference(ref Reference, obj interface{}) error {
 			// The simplest way of eliminating such a window is to call the
 			// callback to "catch up" immediately following new
 			// registrations.
-			i.cb(key)
+			keys = append(keys, key)
 		}
 		// Overwrite the key with a new expiration.
 		l[key] = time.Now().Add(i.leaseDuration)
@@ -191,7 +198,7 @@ func (i *impl) TrackReference(ref Reference, obj interface{}) error {
 		// The simplest way of eliminating such a window is to call the
 		// callback to "catch up" immediately following new
 		// registrations.
-		i.cb(key)
+		keys = append(keys, key)
 	}
 	// Overwrite the key with a new expiration.
 	l[key] = matcher{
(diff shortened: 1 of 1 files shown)


tracker/enqueue_test.go
```diff
@@ -70,7 +71,7 @@ func TestHappyPathsExact(t *testing.T) {
 	// Not tracked yet
 	{
 		trk.OnChanged(thing1)
-		if got, want := calls, 0; got != want {
+		if got, want := int(calls), 0; got != want {
```
vagababov: You can have want=int32(0) and increment it along the way, this avoiding the  ugly casts in every check

tracker/enqueue.go
```diff
@@ -158,12 +158,13 @@ func (i *impl) TrackReference(ref Reference, obj interface{}) error {
 			// The simplest way of eliminating such a window is to call the
 			// callback to "catch up" immediately following new
 			// registrations.
-			i.cb(key)
+			defer i.cb(key)
```
vagababov: Would it be better to have a "locked" get func that reruns nil or cb. Then the code here can be simplified and the unlocks won't be spread around

mattmoor: I don't know what you mean?

tracker/enqueue_test.go
mattmoor: I am trying a different approach now.

tracker/enqueue.go
vaikas: well, first of all, @vagababov do you agree that we should not be holding the lock while making callbacks. If I'm just making up dumb things, then this change is moot.

vagababov: As a general approach, I do agree with that, since god knows what the callback might do.
What I meant here is that the unlock and lock are now far from each other. I'd prefer we have a "locked" function that returns the callback and then the code is
```
lock
defer getCB()
unlock
```

mattmoor: 🤔 I would expect that only the invocation (and not the read of `i.cb`) happens outside of the lock, but that is perhaps incorrect...

mattmoor: Are you suggesting that this is why `race` is freaking out?

vagababov: I haven't looked at the race errors, let me check.
But in general the further the unlock is from lock, the bigger the chance some path will forget to unlock it, or some return/panic will escape unlocking it.

Fun story: on one of my GCP projects we were seeing runaway latencies on some instances.
Turned out that we were using LRU map whose `Get` was not readlock safe (which is kind of reasonable, given you have to update the last used timestamp).
And we were using `lock() x unlock()` pattern, where x would include map read. Which would sometimes panic and keep the mutex locked and the whole system would then grind to a halt. 

Here you have 71 lines of separation between lock and unlock 133-204 (well each path is shorter, but I'll exaggerate) or 227-274, each of the code lines in theory can panic.
So I'd make the code along the lines of:
```
func acquireCB() func(namespacedname){
  lock()
  defer unlock()
  if selector == nil {...return x}
  ...
 return y
}

func (i *impl) OnChanged(obj interface{}) {
  ...
  cb := acquireCB(...)
  cb()
}
```
Same for track reference.

vaikas:
@mattmoor should we try to get this in before the next release?

mattmoor:
😅 Well the pkg release branch is already cut, so too late 😬

n3wscott:
Ping on this, still want this to land?

tracker/enqueue.go
mattmoor: I've rewritten this as something much simpler.

tracker/enqueue.go
```diff
@@ -131,6 +131,13 @@ func (i *impl) TrackReference(ref Reference, obj interface{}) error {
...
 
 	i.m.Lock()
+	// Call the callback without the lock held.
+	var keys []types.NamespacedName
+	defer func(cb func(types.NamespacedName)) {
+		for _, key := range keys {
+			cb(key)
+		}
+	}(i.cb) // read i.cb with the lock held
 	defer i.m.Unlock()
```
vagababov: since you're still defering, move next to the lock
same below

vagababov:
/lgtm
😬

PR #864, opened 2019-11-11
Allow passing options to NewSDKClient()

adrcunha:
/cc @chaodaiG

### DIFF
--- a/test/gke/client_test.go
+++ b/test/gke/client_test.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


chaodaiG:
/approve
lgtm to me, will let Chi taking a look since he has done quite some work on this package

/cc @chizhg

test/gke/client_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"testing"
+	option "google.golang.org/api/option"
+)
+
+// func NewSDKClient(opts ...option.ClientOption) (SDKOperations, error) {
+func TestNewSDKClient(t *testing.T) {
+	datas := []struct {
+		req           option.ClientOption
```
chizhg: Nit. 
I think you can define it as an array like `opts` and use `opts...` to unpack it, when you call the `NewSDKClient()` function, then you won't need the `if` check below.

adrcunha: I do want to test the call without passing options, because this change modified the function signature. If this is silly, LMK and I'll simply remove it.

chizhg: I see. It makes sense to me. Thanks for explaining.

/lgtm

PR #867, opened 2019-11-12
Add QueryParams field in HTTPMatchRequest to istio CRDs

drpmma:
The istio VirtualService's subfield [HTTPMatchRequest](https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#HTTPMatchRequest) has a field queryParams, which is missing in knative istio api.  

[The code of istio api](https://github.com/istio/api/blob/e0714535229bd6078e2ab57f8b9a5cf7f66b0b62/networking/v1alpha3/virtual_service.pb.go#L1160-L1172 ).

For now, I need to use the queryParams field. Therefore, I propose to add this field in this PR.

### DIFF
--- a/apis/istio/v1alpha3/virtualservice_types.go
+++ b/apis/istio/v1alpha3/virtualservice_types.go
@@ -350,6 +350,20 @@ type HTTPMatchRequest struct {
 	// at the top of the VirtualService (if any) are overridden. The gateway match is
 	// independent of sourceLabels.
 	Gateways []string `json:"gateways,omitempty"`
+
+	// Query parameters for matching.
+	//
+	// Ex:
+	// - For a query parameter like "?key=true", the map key would be "key" and
+	//   the string match could be defined as `exact: "true"`.
+	// - For a query parameter like "?key", the map key would be "key" and the
+	//   string match could be defined as `exact: ""`.
+	// - For a query parameter like "?key=123", the map key would be "key" and the
+	//   string match could be defined as `regex: "\d+$"`. Note that this
+	//   configuration will only match values like "123" but not "a123" or "123a".
+	//
+	// **Note:** `prefix` matching is currently not supported.
+	QueryParams map[string]v1alpha1.StringMatch `json:"queryParams,omitempty"`
 }
 
 type HTTPRouteDestination struct {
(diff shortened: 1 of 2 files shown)


ZhiminXiang:
I am fine with this PR.
@drpmma Can you elaborate more about how this field will be used within Knative?

drpmma:
> I am fine with this PR.
> @drpmma Can you elaborate more about how this field will be used within Knative?

With this field, We can route the HTTP request with specific queryParams to a specific revision by exporting and abstracting the HTTPMatchRequest of Istio VirtualService to Knative Route. 

With this feature, we can route, for example, the traffic with query region=asia-northeast to the revision we specify.

The current Knative release strategy is simply splitting traffic, which is not enough. Introducing this field can help us to route the release version to a specific range of users by queryParams, HTTP headers, etc, which is a more flexible release strategy.

ZhiminXiang:
> > I am fine with this PR.
> > @drpmma Can you elaborate more about how this field will be used within Knative?
> 
> With this field, We can route the HTTP request with specific queryParams to a specific revision by exporting and abstracting the HTTPMatchRequest of Istio VirtualService to Knative Route.
> 
> With this feature, we can route, for example, the traffic with query region=asia-northeast to the revision we specify.

I am not sure if Knative should support this feature as in general Knative itself is not a reverse proxy or L7 load balancer. At least, more discussion and thoughts are needed about supporting this feature.

For the short term, you can follow the pattern in the sample https://github.com/knative/docs/tree/master/docs/serving/samples/knative-routing-go to create your own VirtualService with queryParam and define your own routing rule.

drpmma:
> I am not sure if Knative should support this feature as in general Knative itself is not a reverse proxy or L7 load balancer. At least, more discussion and thoughts are needed about supporting this feature.

Thanks for your option. I agree that more discussion is needed for this feature. 
However, the action of adding the QueryParams field to Knative VirtualService to make it consistent with Istio VirtualService CRD won't break anything. Why not just add it for the convenience for future extensions.

drpmma:
There are some discussions about the so-called "feature flag routing" https://github.com/knative/serving/issues/4736.

And this pr https://github.com/knative/serving/pull/6036 also mentioned the idea.

tcnghia:
We moved to using istio/client-go and no longer maintain this API client ourselves.

drpmma:
> We moved to using istio/client-go and no longer maintain this API client ourselves.

Sounds good.

Issue #799, opened 2019-10-24
Mako stub log doesn't report all columns


slinkydeveloper:
/area test-and-release
/kind feature
/kind good-first-issue

## Expected Behavior
Mako stub should print all columns always in same order (that could be the defined order in the .config file), including the one without data points.

This simplifies the scripting for result processing

## Actual Behavior
Mako stub doesn't print all columns. e.g. when no error occurs, no failure column appears

grantr:
:+1: This would be a useful improvement.

slinkydeveloper:
@grantr there is any way from mako apis to retrieve all metrics described in `dev.config` file? Because I can't find it :)

grantr:
Good idea! `dev.config` is a proto file in text format defined at https://github.com/google/mako/blob/master/spec/proto/mako.proto#L255. knative/pkg parses it to get the benckmark key: https://github.com/knative/pkg/blob/master/test/mako/config/benchmark.go#L54

Issue #728, opened 2019-09-30
Mako-stub logs interrupted abruptly


slinkydeveloper:
## Expected Behavior
Write all points published to mako to the stdout of the container

## Actual Behavior
After a bit the container is closed and the output is truncated, without writing all output lines

## Steps to Reproduce the Problem

Run a benchmark in eventing with a long duration configuration (e.g. more than 300 seconds) and then pipe the mako stub logs to a file **from the container creation**. You'll see that at some point the log is abruptly terminated and not all results are written.

## Additional Info
My hypothesis is that the mako-stub doesn't wait for stdout to be completely flushed before exit 

## Alternative proposal
We could refactor the stub to expose an http server that expects one single http request and replies with all metrics. When the response is complete, the mako-stub exits. 

/area test-and-release
/kind bug

grantr:
I looked into whether stdout needs to be synced on exit. Apparently trying to sync stdout on linux returns `EINVAL` because stdout can't be synced: https://github.com/uber-go/zap/issues/328#issuecomment-284337436.

The stub code calls https://golang.org/pkg/encoding/csv/#Writer.Flush which should take care of flushing the csv buffers to stdout. If the container is exiting before that call occurs, that could cause some rows to be missing from the output.

Does the log always print complete rows, or does it print partial rows?
Is there any indication of the exit status of the pod?

slinkydeveloper:
@grantr 
> Does the log always print complete rows, or does it print partial rows?
It prints partial rows sometimes

> Is there any indication of the exit status of the pod?
Success exit status (0)

slinkydeveloper:
Closed by https://github.com/knative/pkg/pull/868 . Now for big logs devs should use the http mode of the stub and collect results with the provided script

PR #868, opened 2019-11-12
Mako stub improvements

slinkydeveloper:
This is an attempt to mitigate https://github.com/knative/pkg/issues/728 and fix https://github.com/knative/pkg/issues/799 

* mako-stub now writes columns before the message output
* mako-stub now REQUIRES the dev.config in the config-mako ConfigMap
* mako-stub now serves results with http (I included a script to use it)

/assign grantr


test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
...
+  echo "Succesfully transfered results into $OUTPUT_FILE and closed the mako-stub"
+else
+  echo "Cannot retrieve results, curl exit status code $curl_exit_status"
+  out_code=1
+fi
+
+kill $PORT_FORWARD_PID
+wait $PORT_FORWARD_PID 2>/dev/null
+
+exit $out_code
```
mattmoor: Add trailing newline:
```suggestion
exit $out_code

```

test/mako/config/benchmark.go
```diff
@@ -25,39 +25,44 @@ import (
...
 	env, err := getEnvironment()
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	// Read the Mako config file for this environment.
 	data, err := readFileFromKoData(env + ".config")
 	if err != nil {
-		return nil, nil, err
+		data, err = ioutil.ReadFile(filepath.Join(configMako, env+".config"))
```
chizhg: If we support adding the benchmark config in the configmap, then `SetupWithBenchmarkConfig` in https://github.com/knative/pkg/blob/d9e3e244a6c044379b71d90b7440daad5b7910f3/test/mako/sidecar.go#L183 will not be needed.
@grantr do you think we can change to this implementation?

test/mako/stub-sidecar/main.go
```diff
@@ -19,104 +19,151 @@ package main
...
+	mako "github.com/google/mako/spec/proto/mako_go_proto"
+
+	"log"
+	"net"
+	"sync"
 
 	"google.golang.org/grpc"
+
+	"knative.dev/pkg/test/mako/config"
 	qspb "knative.dev/pkg/third_party/mako/proto/quickstore_go_proto"
```
chizhg: It looks `quickstore_go_proto` is vendored, can we change it to `github.com/google/mako/proto/quickstore/quickstore_go_proto` and remove the `third_party`?

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+# Copyright 2018 The Knative Authors
```
chizhg:
```suggestion
# Copyright 2019 The Knative Authors
```

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This scripts helps to read results of mako-stub from http
+
+check_command_exists() {
+  CMD_NAME=$1
+  CMD_INSTALL_WITH=$([ -z "$2" ] && echo "" || printf "\nInstall using '%s'" "$2")
+  command -v "$CMD_NAME" > /dev/null || {
+    echo "Command $CMD_NAME not exists$CMD_INSTALL_WITH"
```
chizhg:
```suggestion
    echo "Command $CMD_NAME does not exist"
```

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
...
+    exit 1
+  }
+}
+
+check_command_exists kubectl
+check_command_exists curl
+
+if [[ $# -lt 3 ]]
```
chizhg: It looks `OUTPUT_FILE` is required in the following command, shall we change to `if [[ $# -lt 7 ]]`?

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
...
+then
+  echo "Usage: $0 <mako_stub_pod_name> <mako_stub_namespace> <mako_stub_port> <timeout> <retries> <retries_interval> <out_file>"
+  exit 1
+fi
+
+MAKO_STUB_POD_NAME="$1"
+MAKO_STUB_NAMESPACE="$2"
+MAKO_STUB_PORT="$3"
+TIMEOUT="$4"
```
chizhg: Similar as the above, will there be any error if the script runner does not provide these three args?

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,85 @@
...
+curl --connect-timeout $TIMEOUT \
+    --max-time $TIMEOUT \
+    --retry $RETRIES \
+    --retry-connrefused \
+    --retry-delay $RETRIES_INTERVAL \
+    "http://localhost:$port/results" > $OUTPUT_FILE
+
+curl_exit_status=$?
+
+out_code=0
```
chizhg: It look this var is not needed, can we just use `curl_exit_status` below?

test/mako/stub-sidecar/main.go
slinkydeveloper: quickstore_go_proto doesn't contain all stuff needed

test/mako/config/benchmark.go
```diff
@@ -25,39 +25,44 @@ import (
...
+		if err != nil {
+			return nil, errors.Wrap(err, "cannot load both from kodata and from config mako config map")
```
grantr: Maybe this should return both the kodata `err` and the configmap `err`. Otherwise the message of the kodata `err` will always be hidden.

test/mako/config/benchmark.go
grantr: Yes, I think we can change to this implementation. The current solution isn't sufficient to enable different configs in prod/dev environments, and benchmark config in ConfigMaps fixes that.

We'll need to add "also update the ConfigMaps" to the instructions for updating benchmark config files.

test/mako/config/benchmark.go
```diff
@@ -25,39 +25,44 @@ import (
...
+	"github.com/pkg/errors"
 )
 
 const koDataPathEnvName = "KO_DATA_PATH"
+const configMako = "/etc/config-mako"
 
 // MustGetBenchmark wraps getBenchmark in log.Fatalf
-func MustGetBenchmark() (*string, *string) {
-	benchmarkKey, benchmarkName, err := getBenchmark()
+func MustGetBenchmark() *mpb.BenchmarkInfo {
```
grantr: We'll need to stage a PR in serving to pick up this signature change.

test/mako/stub-sidecar/main.go
grantr: There are two quickstore_go_proto packages that are different:

https://github.com/knative/pkg/blob/master/vendor/github.com/google/mako/internal/quickstore_microservice/proto/quickstore_go_proto/quickstore.pb.go
https://github.com/knative/pkg/blob/master/vendor/github.com/google/mako/proto/quickstore/quickstore_go_proto/quickstore.pb.go

The one in third_party is a copy of the first one with `internal` in the path. Go won't allow code in a different path to import internal packages, so we need to copy it to third_party.

test/mako/stub-sidecar/read_results.sh
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This scripts helps to read results of mako-stub from http
+
+check_command_exists() {
+  CMD_NAME=$1
+  command -v "$CMD_NAME" > /dev/null || {
+    echo "Command $CMD_NAME does not exists"
```
grantr: grammar nit: `Command $CMD_NAME does not exist`

test/mako/config/benchmark.go
grantr: And yes I think we can remove `SetupWithBenchmarkConfig` now.

test/mako/config/benchmark.go
chizhg: serving is not calling this function directly, so it's not a breaking change.

coryrc:
Part of #728 is glog having bad semantics. I'm going to ask Mako to change to klog (it's nearly source-compatible).

chizhg:
> Part of #728 is glog having bad semantics. I'm going to ask Mako to change to klog (it's nearly source-compatible).

@coryrc this fake Mako stub is the temporary solution for the issue that the real Mako cannot accept data from non-google developers - https://github.com/google/mako. It does not use `glog`.

test/mako/stub-sidecar/main.go
slinkydeveloper: So I can leave as it is now?

chizhg: Sure, let's keep it.

test/mako/config/benchmark.go
slinkydeveloper: Yep just checked serving uses `Setup` method

slinkydeveloper:
@grantr I've done a PR to remove usage of `SetupWithBenchmarkConfig` from eventing https://github.com/knative/eventing/pull/2202. I think we can remove the hold since there is no usage in serving

grantr:
/hold

Question for @chizhg: if the Mako config is embedded in the same ConfigMap as the environment name, how will we set the environment to `prod` in CI clusters?

In https://github.com/knative/test-infra/blob/master/scripts/performance-tests.sh#L54, we create the `config-mako` ConfigMap with the environment key, but that won't work with this change because the configmap will have the `dev.config` keys as well.

Maybe the ConfigMap specifying the benchmarking environment should be a separate ConfigMap from the one specifying the Mako configs. We could do this with [projected volumes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/).

grantr:
/hold cancel

Canceling the hold because @chizhg has to approve this anyway :)

chizhg:
> /hold
> 
> Question for @chizhg: if the Mako config is embedded in the same ConfigMap as the environment name, how will we set the environment to `prod` in CI clusters?
> 
> In https://github.com/knative/test-infra/blob/master/scripts/performance-tests.sh#L54, we create the `config-mako` ConfigMap with the environment key, but that won't work with this change because the configmap will have the `dev.config` keys as well.
> 
> Maybe the ConfigMap specifying the benchmarking environment should be a separate ConfigMap from the one specifying the Mako configs. We could do this with [projected volumes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-projected-volume-storage/).

I think we can add the configmap in the xxx-setup.yaml for continuous runs, or patch the configmap in the `update_benchmark` function in https://github.com/knative/eventing/blob/e46fbaf28665e0774779ac8a37621819c3e73031/test/performance/performance-tests.sh#L47.

I prefer the latter one. But either way, I think we can still reuse the configmap.

grantr:
cc @chaodaiG for approval

chizhg:
Since Chao is OOO
/cc @adrcunha

PR #869, opened 2019-11-12
added Equal methods for TestCase, TestProperty, TestProperties 

yt3liu:
added Equal methods for TestCase, TestProperty, TestProperties struct for simpler junit test comparison with https://github.com/google/go-cmp


yt3liu:
/close 
reflect.DeepEqual compares pointers as expected. No custom Equal logic needed.

PR #870, opened 2019-11-12
Address serving#5949 by putting flag initialization into pkg.

coryrc:
Created a unit test for flag setup.

Moved from https://github.com/knative/serving/pull/5984

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -25,11 +25,27 @@ import (
 	"os"
 	"os/user"
 	"path"
+	"sync"
+
+	"k8s.io/klog"
+	"knative.dev/pkg/test/logging"
 )
 
-// Flags holds the command line flags or defaults for settings in the user's environment.
-// See EnvironmentFlags for a list of supported fields.
-var Flags = initializeFlags()
+const (
+	// e2eMetricExporter is the name for the metrics exporter logger
+	e2eMetricExporter = "e2e-metrics"
+
+	// The recommended default log level https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md
+	klogDefaultLogLevel = "2"
+)
+
+var (
+	flagsSetupOnce = &sync.Once{}
+
+	// Flags holds the command line flags or defaults for settings in the user's environment.
+	// See EnvironmentFlags for a list of supported fields.
+	Flags = initializeFlags()
+)
 
 // EnvironmentFlags define the flags that are needed to run the e2e tests.
 type EnvironmentFlags struct {
@@ -73,7 +89,40 @@ func initializeFlags() *EnvironmentFlags {
 
 	flag.StringVar(&f.Tag, "tag", "latest", "Provide the version tag for the test images.")
 
+	klog.InitFlags(nil)
+	flag.Set("v", klogDefaultLogLevel)
+	flag.Set("alsologtostderr", "true")
+
 	return &f
+}
+
+func printFlags() {
+	fmt.Print("Test Flags: {")
+	flag.CommandLine.VisitAll(func(f *flag.Flag) {
+		fmt.Printf("'%s': '%s', ", f.Name, f.Value.String())
+	})
+	fmt.Println("}")
+}
+
+// SetupLoggingFlags initializes the logging libraries at runtime
+func SetupLoggingFlags() {
+	flagsSetupOnce.Do(func() {
+		if Flags.LogVerbose {
+			// If klog verbosity is not set to a non-default value (via "-args -v=X"),
+			if flag.CommandLine.Lookup("v").Value.String() == klogDefaultLogLevel {
+				// set up verbosity for klog so round_trippers.go prints:
+				//   URL, request headers, response headers, and partial response body
+				// See levels in vendor/k8s.io/client-go/transport/round_trippers.go:DebugWrappers for other options
+				flag.Set("v", "8")
+			}
+			printFlags()
+		}
+		logging.InitializeLogger(Flags.LogVerbose)
+
+		if Flags.EmitMetrics {
+			logging.InitializeMetricExporter(e2eMetricExporter)
+		}
+	})
 }
 
 // ImagePath is a helper function to prefix image name with repo and suffix with tag
--- a/test/e2e_flags_test.go
+++ b/test/e2e_flags_test.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"flag"
+	"sync"
+	"testing"
+)
+
+func TestE2eFlags(t *testing.T) {
+	v := flag.Lookup("v")
+	if v == nil {
+		t.Fatal("Could not find 'v' flag; klog was not init-ed")
+	}
+	if v.Value.String() != klogDefaultLogLevel {
+		t.Fatalf("Either '%s' was passed in for v or v is not being initialized properly. The former is a test limitation, but the latter is an actual problem", v.Value.String())
+	}
+
+	flagsAreUninitialized := false
+	flagsSetupOnce.Do(func() {
+		flagsAreUninitialized = true
+	})
+	if !flagsAreUninitialized {
+		t.Error("SetupLoggingFlags should not have been called at package initialization stage")
+	}
+
+	flagTests := []struct {
+		testCase    string
+		logVerbose  bool
+		nonDefaultV string
+		expectedV   string
+	}{
+		{
+			testCase:    "Check default case with logverbose off",
+			logVerbose:  false,
+			nonDefaultV: "",
+			expectedV:   klogDefaultLogLevel,
+		},
+		{
+			testCase:    "Check default case with logverbose on",
+			logVerbose:  true,
+			nonDefaultV: "",
+			expectedV:   "8",
+		},
+		{
+			testCase:    "Check default case with logverbose on and v set to non-default value",
+			logVerbose:  true,
+			nonDefaultV: "5",
+			expectedV:   "5",
+		},
+	}
+
+	alsologtostderr := flag.Lookup("alsologtostderr").Value.String()
+	if "true" != alsologtostderr {
+		t.Errorf("alsologtostderr = '%s', want: 'true'\n", alsologtostderr)
+	}
+
+	for _, tc := range flagTests {
+		t.Run(tc.testCase, func(t *testing.T) {
+			flagsSetupOnce = &sync.Once{}
+			Flags.LogVerbose = tc.logVerbose
+			if tc.nonDefaultV != "" {
+				flag.Set("v", tc.nonDefaultV)
+			}
+			SetupLoggingFlags()
+			v := flag.Lookup("v").Value.String()
+			if tc.expectedV != v {
+				t.Errorf("v = '%s', want: '%s'\n", v, tc.expectedV)
+			}
+		})
+	}
+}


test/e2e_flags.go
```diff
@@ -73,9 +82,48 @@ func initializeFlags() *EnvironmentFlags {
 
 	flag.StringVar(&f.Tag, "tag", "latest", "Provide the version tag for the test images.")
 
+	klog.InitFlags(nil)
+	flag.Set("v", "2") // A normal level of logging for testing
```
chaodaiG: I thought this would silently fail

test/e2e_flags.go
```diff
@@ -73,9 +82,48 @@ func initializeFlags() *EnvironmentFlags {
...
+
 	return &f
 }
 
+func printFlags() {
+	fmt.Print("Test Flags: {")
+	flag.CommandLine.VisitAll(func(f *flag.Flag) {
+		fmt.Printf("'%s': '%s', ", f.Name, f.Value.String())
```
chaodaiG: %q is equivalent to '%s', and slightly better than it in some cases
not sure what `"%+v", f` would do

test/e2e_flags_test.go
```diff
@@ -0,0 +1,69 @@
...
+		t.Fatal("Could not find 'v' flag; klog was not init-ed")
+	}
+	if v.Value.String() != "2" {
+		t.Fatalf("Either '%s' was passed in for v or v is not being initialized properly. The former is a test limitation, but the latter is an actual problem", v.Value.String())
+	}
+	if flagsSetup {
+		t.Error("SetupLoggingFlags should not have been called by the unit tests")
+	}
+
+	t.Log("Check default case with logverbose off")
```
chaodaiG: this can be handled by table drive test
https://github.com/golang/go/wiki/TableDrivenTests

test/e2e_flags.go
coryrc: I mistaken why changes weren't having any effect; it wasn't having effect because of the level (v) not because of a flag being Set here. I'll move alsologtostderr up.

test/e2e_flags.go
coryrc: %q is "escaped with Go syntax" and I don't wish to alter the values as they aren't limited by Go syntax.

*Flag has several irrelevant children: usage, DefValue, and Value has two functions at children instead of actually printing the value: https://godoc.corp.google.com/pkg/flag#Flag

test/e2e_flags_test.go
coryrc: It takes an additional line of code to do that; still want me to?

diff --git a/test/e2e_flags_test.go b/test/e2e_flags_test.go
index 1fca733..c838511 100644
--- a/test/e2e_flags_test.go
+++ b/test/e2e_flags_test.go
@@ -45,25 +45,26 @@ func TestE2eFlags(t *testing.T) {
 		t.Error("SetupLoggingFlags should not have been called by the unit tests")
 	}
 
-	t.Log("Check default case with logverbose off")
-	flagsSetup = false
-	Flags.LogVerbose = false
-	flag.Set("alsologtostderr", "false")
-	SetupLoggingFlags()
-	assertAlsoAndV(t, "2")
-
-	t.Log("Check default case with logverbose on")
-	flagsSetup = false
-	Flags.LogVerbose = true
-	flag.Set("alsologtostderr", "false")
-	SetupLoggingFlags()
-	assertAlsoAndV(t, "8")
+	var flagTests = []struct {
+		log         string
+		logverbose  bool
+		nonDefaultV string
+		expectedV   string
+	}{
+		{"Check default case with logverbose off", false, "", "2"},
+		{"Check default case with logverbose on", true, "", "8"},
+		{"Check default case with logverbose on and v set to non-default value", true, "5", "5"},
+	}
 
-	t.Log("Check default case with logverbose on and v set to non-default value")
-	flagsSetup = false
-	Flags.LogVerbose = true
-	flag.Set("alsologtostderr", "false")
-	flag.Set("v", "5")
-	SetupLoggingFlags()
-	assertAlsoAndV(t, "5")
+	for _, tt := range flagTests {
+		t.Log(tt.log)
+		flagsSetup = false
+		Flags.LogVerbose = tt.logverbose
+		flag.Set("alsologtostderr", "false")
+		if tt.nonDefaultV != "" {
+			flag.Set("v", tt.nonDefaultV)
+		}
+		SetupLoggingFlags()
+		assertAlsoAndV(t, tt.expectedV)
+	}
 }

test/e2e_flags_test.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2019 The Knative Authors
```

test/e2e_flags.go
chaodaiG: If that's the case I would suggest move alsologtostderr back to serving

test/e2e_flags_test.go
chaodaiG: It's much easier to read from my point of view. It also makes expanding test coverage easier to me. Btw this test only covers logverbose, any plan to cover all other flags defined here? I think one more testcase with all flags set would be good enough

coryrc: I don't think settings should be tested, just logic.

test/e2e_flags_test.go
```diff
@@ -0,0 +1,70 @@
...
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2019 The Knative Authors
```

chaodaiG:
Sugges testing this in serving first

coryrc:
> Suggest testing this in serving first

Tested and works, set Gopkg.toml to my fork and serving-issue5949 branch, and things worked as expected.

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
...
+	})
+	fmt.Println("}")
+}
+
+var flagsSetup = false
+var flagsMutex = &sync.Mutex{}
+
+func SetupLoggingFlags() {
```
dgerd: nit: Comment for exported functions

test/e2e_flags_test.go
```diff
@@ -0,0 +1,66 @@
...
+		nonDefaultV string
+		expectedV   string
+	}{
+		{"Check default case with logverbose off", false, "", "2"},
+		{"Check default case with logverbose on", true, "", "8"},
+		{"Check default case with logverbose on and v set to non-default value", true, "5", "5"},
+	}
+
+	alsologtostderr := flag.Lookup("alsologtostderr").Value.String()
+	if fmt.Sprintf("%t", true) != alsologtostderr {
```
dgerd: Why the fmt.Sprintf() instead of just "true"?

test/e2e_flags_test.go
```diff
@@ -0,0 +1,66 @@
...
+		t.Errorf("alsologtostderr was '%s', expected '%t'\n", alsologtostderr, true)
```
dgerd: expected type should be the same as the got type.

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
...
+	flagsMutex.Lock()
+	if !flagsSetup {
+		if Flags.LogVerbose {
+			// If klog verbosity is not set to a non-default value (via "-args -v=X"),
+			// set up verbosity for klog so round_trippers.go prints:
+			//   URL, request headers, response headers, and partial response body
+			// Definition in vendor/k8s.io/client-go/transport/round_trippers.go:DebugWrappers
+			if flag.CommandLine.Lookup("v").Value.String() == "2" {
+				flag.Set("v", "8")
```
dgerd: Why 8? Where does this constant come from?

test/e2e_flags.go
vagababov: I think `%q` is completely appropriate here.

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
 
 	flag.StringVar(&f.Tag, "tag", "latest", "Provide the version tag for the test images.")
 
+	klog.InitFlags(nil)
+	flag.Set("v", "2") // A normal level of logging for testing
```
dgerd: Why 2? Why not make this a constant

`DefaultLogLevel = 2`

Then you can use this in the unit test below, test that default is getting set properly, and be able to change the default without updating the test.

test/e2e_flags_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if v.Value.String() != "2" {
+		t.Fatalf("Either '%s' was passed in for v or v is not being initialized properly. The former is a test limitation, but the latter is an actual problem", v.Value.String())
+	}
+	if flagsSetup {
+		t.Error("SetupLoggingFlags should not have been called by the unit tests")
+	}
+
+	var flagTests = []struct {
+		log         string
```
dgerd:
```suggestion
		name         string
```

Isn't this just the name or description of the test case?

test/e2e_flags_test.go
```diff
@@ -0,0 +1,66 @@
...
+		logverbose  bool
+		nonDefaultV string
+		expectedV   string
+	}{
+		{"Check default case with logverbose off", false, "", "2"},
```
dgerd: nit: I find it easier to read with the values on their own lines and prefixed with the type.

test/e2e_flags_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if v.Value.String() != "2" {
+		t.Fatalf("Either '%s' was passed in for v or v is not being initialized properly. The former is a test limitation, but the latter is an actual problem", v.Value.String())
+	}
+	if flagsSetup {
+		t.Error("SetupLoggingFlags should not have been called by the unit tests")
+	}
+
+	var flagTests = []struct {
+		log         string
+		logverbose  bool
```
dgerd:
```suggestion
		logVerbose  bool
```

test/e2e_flags.go
vagababov: `defaultLogLevel`, I guess?

test/e2e_flags.go
```diff
@@ -25,6 +25,15 @@ import (
...
 	"path"
+	"sync"
+
+	"k8s.io/klog"
+	"knative.dev/pkg/test/logging"
+)
+
+const (
+	// E2EMetricExporter is the name for the metrics exporter logger
+	E2EMetricExporter = "e2e-metrics"
```
vagababov: I presume this is used somewhere?

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
...
 
+func printFlags() {
+	fmt.Print("Test Flags: {")
+	flag.CommandLine.VisitAll(func(f *flag.Flag) {
+		fmt.Printf("'%s': '%s', ", f.Name, f.Value.String())
+	})
+	fmt.Println("}")
+}
+
+var flagsSetup = false
```
vagababov:
```suggestion
var(
...
)
```

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
...
+			printFlags()
+		}
+		logging.InitializeLogger(Flags.LogVerbose)
+
+		if Flags.EmitMetrics {
+			logging.InitializeMetricExporter(E2EMetricExporter)
+		}
+		flagsSetup = true
+	}
+	flagsMutex.Unlock()
```
vagababov: mode unlock to the lock `defer m.Unlock()`.

test/e2e_flags.go
```diff
@@ -73,9 +82,47 @@ func initializeFlags() *EnvironmentFlags {
...
+func printFlags() {
+	fmt.Print("Test Flags: {")
+	flag.CommandLine.VisitAll(func(f *flag.Flag) {
+		fmt.Printf("'%s': '%s', ", f.Name, f.Value.String())
+	})
+	fmt.Println("}")
+}
+
+var flagsSetup = false
+var flagsMutex = &sync.Mutex{}
```
vagababov: Are we doing this more than once? Should it be just `sync.Once`?

test/e2e_flags_test.go
vagababov:
```suggestion
		t.Errorf("alsologtostderr was %q, expected true", alsologtostderr)
```

test/e2e_flags_test.go
coryrc: Was a parameter and made it not.

test/e2e_flags.go
coryrc: It's described in the comment above?

test/e2e_flags.go
coryrc: Adding comment why "2", and using your suggestion

test/e2e_flags_test.go
coryrc: It's a description of the testcase; how about testCase?

test/e2e_flags_test.go
coryrc: I feel it should be either "LogVerbose", to match the flag variable name, or "logverbose" to match the name of the option on the command line. If you still disagree, please say so.

test/e2e_flags.go
coryrc: roughly line 123

test/e2e_flags.go
coryrc: I would like to verify the function doesn't get called during package initialization and I can't seem to do that with sync.Once() without introducing a second variable and thus having essentially the same code now + boilerplate func calls, so I'll use your other suggestions to make it cleaner.

test/e2e_flags.go
coryrc: done

test/e2e_flags_test.go
coryrc: I don't like %q because it adds Go escaping, which could potentially mangle the value

test/e2e_flags.go
coryrc: done

test/e2e_flags_test.go
coryrc: I don't know what "prefixed with the type" means, but they're on separate lines

dgerd:
```
{
   name: "Check default case with logverbose off",
   logVerbose: false,
   expectedV: "2",
},
{
   name: "Check ..."
   ...
 }
```

test/e2e_flags_test.go
vagababov: `label` is the most common

test/e2e_flags.go
vagababov: Sorry I wasn't clear.
It is not used outside the package, thus the constant must be lowecased.

test/e2e_flags_test.go
dgerd: Camel case is convention: https://golang.org/doc/effective_go.html#mixed-caps

Capitalizing the first letter exports it and this does not need to be accessible outside of the package.

test/e2e_flags_test.go
dgerd: desc, description, test, testCase, name are work for me and make more sense here.

test/e2e_flags.go
vagababov: Please don't resolve until we settle on a decision.
I still don't understand the `%q` problem. Please bring an example that you think will make this worse? (

test/e2e_flags.go
vagababov: Not really. It just refences something. It would make it more useful to rewrite as `...v=8 prints...`

test/e2e_flags_test.go
vagababov: The way go does capitalcasing is `logVerbose`. I concur with Dan here.

test/e2e_flags_test.go
coryrc: done

test/e2e_flags_test.go
coryrc: done

test/e2e_flags.go
coryrc: Because '%s' is obvious to anybody reading the code, has ever used printf strings, and will do exactly what I want it to do. It's not easy to find what %q is going to do besides "escaped with Go syntax" -- whatever that means -- and uses double-quotes instead.

test/e2e_flags.go
vagababov: Still here.

test/e2e_flags.go
vagababov: Double quotes is not really a problem.
`%q` is part of the basic `fmt` package which is usually covered in the go walkthrough early pages.

test/e2e_flags.go
vagababov: Still here.

test/e2e_flags.go
coryrc: It has to be called after klog.InitFlags, so I wouldn't trust it anywhere else.

test/e2e_flags.go
coryrc: The comments describe Why, not How. Rearranged them to make it more obvious.

test/e2e_flags.go
coryrc: Oh, good point, no reason to export.

test/e2e_flags.go
coryrc: (I had to refresh because comments weren't coming up)
I copied verbatim from serving, but `grep -r E2EMetricExporter` didn't come up with anything so I'll assume it doesn't need to be

test/e2e_flags.go
coryrc: I have never encountered it before this comment and we're gaining nothing with %q that '%s' doesn't do just fine.

vagababov: It is at least 2 characters longer... :)

test/e2e_flags.go
```diff
@@ -25,11 +25,28 @@ import (
...
 	"path"
+	"sync"
+
+	"k8s.io/klog"
+	"knative.dev/pkg/test/logging"
+)
+
+const (
+	// e2EMetricExporter is the name for the metrics exporter logger
+	e2EMetricExporter = "e2e-metrics"
```
vagababov:
```suggestion
	e2eMetricExporter = "e2e-metrics"
```

test/e2e_flags.go
```diff
@@ -73,9 +90,45 @@ func initializeFlags() *EnvironmentFlags {
...
+	flag.CommandLine.VisitAll(func(f *flag.Flag) {
+		fmt.Printf("'%s': '%s', ", f.Name, f.Value.String())
+	})
+	fmt.Println("}")
+}
+
+// SetupLoggingFlags initializes the logging libraries at runtime
+func SetupLoggingFlags() {
+	flagsMutex.Lock()
+	defer flagsMutex.Unlock()
```
vagababov: As I mentioned in the other PR ( I am still not sure how they relate to each other), should we use `sync.Once` here, rather than lock/unlock? This is supposed to be called once anyway, right?

coryrc: I can't test that it doesn't get called during package initialization then without adding much more variables that just end up being the same thing + extra boilerplate

test/e2e_flags.go
coryrc: done

test/e2e_flags.go
vagababov: Sorry, I don't really parse your comment.
sync.Once will verify it is called once. _do we need_ it to be called more than once?

coryrc: Yes, we do for the unit test, but I could just override the sync.Once variable to make that happen.

But I want to verify this function does NOT get called during package initialization, and sync.Once does not provide any way to see if it has been triggered. And if you add in another variable to detect if it's been triggered, it ends up looking nearly the same as this but worse.

coryrc: (because flag parsing happening at package initialization time is why it broke with Go 1.13 and the fix didn't actually fix it)

vagababov: HOw would you do it, if your `sync.Once` variable is package private?

coryrc: The unit test is in the same package

vagababov: Ah, yeah, but I care more about the code that goes to the library, than the bloat in the unit test.

coryrc: The unit test won't get any worse, the library will be much worse, because it will still need to set a value saying it ran, like it does now, then define the sync struct and wrap all work into an anonymous function into syncVariable.Do(func () {...
So it ends up being more lines and indentation to achieve the same thing.

test/e2e_flags.go
coryrc: I don't know when I'm supposed to resolve this.

test/e2e_flags.go
vagababov: OK.

test/e2e_flags.go
vagababov: I think it's fine, I'd love to see that 8 is what logs that info, but I am not going to beat the bush around this.

test/e2e_flags_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestE2eFlags(t *testing.T) {
+	v := flag.Lookup("v")
+	if v == nil {
+		t.Fatal("Could not find 'v' flag; klog was not init-ed")
+	}
+	if v.Value.String() != klogDefaultLogLevel {
+		t.Fatalf("Either '%s' was passed in for v or v is not being initialized properly. The former is a test limitation, but the latter is an actual problem", v.Value.String())
+	}
+	if flagsSetup {
+		t.Error("SetupLoggingFlags should not have been called by the unit tests")
```
vagababov: Should this be fatal

test/e2e_flags_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	var flagTests = []struct {
```
vagababov:
```suggestion
	flagTests := []struct {
```

test/e2e_flags_test.go
```diff
@@ -0,0 +1,80 @@
...
+			expectedV:   "5",
+		},
+	}
+
+	alsologtostderr := flag.Lookup("alsologtostderr").Value.String()
+	if "true" != alsologtostderr {
+		t.Errorf("alsologtostderr was '%s', expected 'true'\n", alsologtostderr)
+	}
+
+	for _, tt := range flagTests {
```
vagababov: Rather than logging `tt.testCase`, the standard way is to do
```
for _, tc := range tests {
  t.Run(tc.testCase, func(t *testing.T) {
    ...
  })
}
```

test/e2e_flags_test.go
```diff
@@ -0,0 +1,80 @@
...
+			testCase:    "Check default case with logverbose on and v set to non-default value",
+			logVerbose:  true,
+			nonDefaultV: "5",
+			expectedV:   "5",
+		},
+	}
+
+	alsologtostderr := flag.Lookup("alsologtostderr").Value.String()
+	if "true" != alsologtostderr {
+		t.Errorf("alsologtostderr was '%s', expected 'true'\n", alsologtostderr)
```
vagababov: I'd recommend the following:
`alsologtostderr = %q, want: true`.
Same below.

test/e2e_flags_test.go
coryrc: The rest of the test can still run and give useful information, so Fatal isn't necessary.

test/e2e_flags_test.go
coryrc: done, thanks.

test/e2e_flags_test.go
coryrc: done, thanks.

test/e2e_flags.go
coryrc: If sync.Once said whether it had been run before, it would be cleaner, but it doesn't ....
Oh, wait, it doesn't matter what function you call, so I could call it with anything...
hold on...

test/e2e_flags_test.go
coryrc: coryrc@coryrc:~/go/src/knative.dev/pkg$ grep -r "'%s'" | wc -l
74

test/e2e_flags.go
coryrc: coryrc@coryrc:~/go/src/knative.dev/pkg$ grep -r "'%s'" | wc -l
74

test/e2e_flags.go
coryrc: okay, done.

test/e2e_flags.go
vagababov: A more proper command is `grep -r "'%s'" * | grep -v "vendor" | wc -l`, but it still returns 33 instances, which I guess I will need to fix :-)

test/e2e_flags_test.go
vagababov: yeah, still `alsologtostderr = %<>, want: true`.

coryrc: Changing the other statement on line 83 to match too

test/e2e_flags.go
coryrc: I hope you don't! :-)

chizhg:
This PR unfortunately broke all performance benchmarks - `/ko-app/continuous flag redefined: log_dir`, as Mako is using `glog`. Why are we calling `klog.InitFlags` here?

coryrc:
I'm restoring the functionality broken when we change flag parsing in Go 1.13.

Addressing your issue: https://github.com/knative/pkg/pull/881

PR #872, opened 2019-11-13
Skip watching the ConfigMaps if they are missing

houshengbo:
This PR should be the most straightforward fix by NOT adding the ConfigMaps into the config map watcher, if config-logging or config-observability is missing.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -167,13 +167,18 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(),
 		metav1.GetOptions{}); err == nil {
 		cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMap: " + logging.ConfigMapName(), zap.Error(err))
 	}
+
 	// Watch the observability config map
 	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(metrics.ConfigMapName(),
 		metav1.GetOptions{}); err == nil {
 		cmw.Watch(metrics.ConfigMapName(),
 			metrics.UpdateExporterFromConfigMap(component, logger),
 			profilingHandler.UpdateFromConfigMap)
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMap: " + metrics.ConfigMapName(), zap.Error(err))
 	}
 
 	if err := cmw.Start(ctx.Done()); err != nil {
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -164,12 +164,17 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	profilingHandler := profiling.NewHandler(logger, false)
 
 	// Watch the logging config map and dynamically update logging levels.
-	cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
-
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -164,12 +164,17 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	profilingHandler := profiling.NewHandler(logger, false)
 
 	// Watch the logging config map and dynamically update logging levels.
-	cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
-
+	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(),
```
vagababov: you're swallowing the error here, which is not nice. At least it should be logged.

injection/sharedmain/main.go
```diff
@@ -167,13 +167,18 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(),
 		metav1.GetOptions{}); err == nil {
 		cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else if !apierrors.IsNotFound(err) {
```
vagababov: Do we not care if it is NF?

injection/sharedmain/main.go
```diff
@@ -164,12 +164,22 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	profilingHandler := profiling.NewHandler(logger, false)
 
 	// Watch the logging config map and dynamically update logging levels.
-	cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(),
+		metav1.GetOptions{}); err == nil {
+		cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMaps:", err)
```
vagababov:
```suggestion
		logger.Fatalw("Error reading ConfigMap: " + logging.ConfigMapName(), zap.Error(err))
```

injection/sharedmain/main.go
```diff
@@ -164,12 +164,22 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
...
-	cmw.Watch(metrics.ConfigMapName(),
-		metrics.UpdateExporterFromConfigMap(component, logger),
-		profilingHandler.UpdateFromConfigMap)
+	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(metrics.ConfigMapName(),
+		metav1.GetOptions{}); err == nil {
+		cmw.Watch(metrics.ConfigMapName(),
+			metrics.UpdateExporterFromConfigMap(component, logger),
+			profilingHandler.UpdateFromConfigMap)
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMaps:", err)
```
vagababov:
```suggestion
		logger.Fatalw("Error reading ConfigMap: " + metrics.ConfigMapName(), zap.Error(err))
```

injection/sharedmain/main.go
houshengbo: If it is not found, we just bypass watching it in cmw. I do not think that case should block the process.

injection/sharedmain/main.go
houshengbo: done.

injection/sharedmain/main.go
houshengbo: done.

injection/sharedmain/main.go
vagababov: I'd log it, rather than kill, but OK.

PR #880, opened 2019-11-15
Add check in informer to make controllers work when process invalid resources

grac3gao:
Related on issue in eventing repo
[#2047](https://github.com/knative/eventing/issues/2047) sources-controller can't recover from faulty sink 
If there has an invalid resource, the controller will keep showing error and not process new sources until the controller pod is deleted.

The root cause is explained in this [comment](https://github.com/knative/eventing/issues/2047#issuecomment-554084211)

### DIFF
--- a/apis/duck/typed.go
+++ b/apis/duck/typed.go
@@ -46,6 +46,11 @@ var _ InformerFactory = (*TypedInformerFactory)(nil)
 
 // Get implements InformerFactory.
 func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.SharedIndexInformer, cache.GenericLister, error) {
+	// Avoid error cases, like resource not exists
+	if _, err := dif.Client.Resource(gvr).List(metav1.ListOptions{}); err != nil {
+		return nil, nil, err
+	}
+
 	listObj := dif.Type.GetListType()
 	lw := &cache.ListWatch{
 		ListFunc:  asStructuredLister(dif.Client.Resource(gvr).List, listObj),
@@ -57,43 +62,13 @@ func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.Sha
 
 	lister := cache.NewGenericLister(inf.GetIndexer(), gvr.GroupResource())
 
-	infChannel := make(chan struct{})
-	errorChannel := make(chan error)
+	go inf.Run(dif.StopChannel)
 
-	go inf.Run(infChannel)
-
-	go func() {
-		if ok := cache.WaitForCacheSync(infChannel, inf.HasSynced); !ok {
-			errorChannel <- fmt.Errorf("failed starting shared index informer for %v with type %T", gvr, dif.Type)
-		} else {
-			errorChannel <- nil
-		}
-	}()
-
-	timer := time.NewTimer(time.Second * time.Duration(30))
-	defer timer.Stop()
-
-	go func() {
-		<-timer.C
-		// Close infChannel if the informer hasn't synced
-		if !inf.HasSynced() {
-			close(infChannel)
-		}
-	}()
-
-	// If StopChannel closed, infChannel should close too
-	go func() {
-		<-dif.StopChannel
-		if _, ok := <-infChannel; !ok {
-			close(infChannel)
-		}
-	}()
-
-	err := <-errorChannel
-	if err != nil {
-		return nil, nil, err
+	if ok := cache.WaitForCacheSync(dif.StopChannel, inf.HasSynced); !ok {
+		return nil, nil, fmt.Errorf("failed starting shared index informer for %v with type %T", gvr, dif.Type)
 	}
-	return inf, lister, err
+
+	return inf, lister, nil
 }
 
 type unstructuredLister func(metav1.ListOptions) (*unstructured.UnstructuredList, error)
--- a/apis/duck/typed.go
+++ b/apis/duck/typed.go
@@ -46,7 +46,8 @@ var _ InformerFactory = (*TypedInformerFactory)(nil)
 
 // Get implements InformerFactory.
 func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.SharedIndexInformer, cache.GenericLister, error) {
-	// Avoid error cases, like resource not exists.
+	// Avoid error cases, like the GVR does not exist.
+	// It is not a full check. Some RBACs might sneak by, but the window is very small.
 	if _, err := dif.Client.Resource(gvr).List(metav1.ListOptions{}); err != nil {
 		return nil, nil, err
 	}
--- a/apis/duck/typed.go
+++ b/apis/duck/typed.go
@@ -57,13 +57,43 @@ func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.Sha
 
 	lister := cache.NewGenericLister(inf.GetIndexer(), gvr.GroupResource())
 
-	go inf.Run(dif.StopChannel)
+	infChannel := make(chan struct{})
+	errorChannel := make(chan error)
 
-	if ok := cache.WaitForCacheSync(dif.StopChannel, inf.HasSynced); !ok {
-		return nil, nil, fmt.Errorf("failed starting shared index informer for %v with type %T", gvr, dif.Type)
+	go inf.Run(infChannel)
+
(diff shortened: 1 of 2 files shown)


apis/duck/typed.go
```diff
@@ -57,13 +57,43 @@ func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.Sha
...
+	go func() {
+		if ok := cache.WaitForCacheSync(infChannel, inf.HasSynced); !ok {
+			errorChannel <- fmt.Errorf("failed starting shared index informer for %v with type %T", gvr, dif.Type)
+		} else {
+			errorChannel <- nil
+		}
+	}()
 
-	return inf, lister, nil
+	timer := time.NewTimer(time.Second * time.Duration(30))
```
Harwayne: Add a comment that this specific time wasn't chosen for any particular reason, it just 'feels' right. And that it bounds how long a single threaded reconciler may be stuck processing this particular resource.

apis/duck/typed.go
```diff
@@ -57,13 +57,43 @@ func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.Sha
 
 	lister := cache.NewGenericLister(inf.GetIndexer(), gvr.GroupResource())
 
-	go inf.Run(dif.StopChannel)
+	infChannel := make(chan struct{})
+	errorChannel := make(chan error)
```
mattmoor: I have a number of concerns with this approach, one of which is the complexity (channel stuff like this is incredibly hard to follow).

I suspect that when this happens, the controller will typically requeue the key, meaning that a bad key can effectively DoS the workqueue by inserting these 30s bubbles that we sit reconciling in perpetuity.

I think we need to find a way to either fail fast, or not fail here...

Part of me things that we should WaitForCacheSync in a go routine and not block on its completion, but that has its own downsides.  I'm happy to jump on a call to discuss options here, if that's easiest?

vagababov: Yes, this look incredibly more complex than what we have discussed in person.

vaikas: +1 to failing fast as a preference. Could we hit the API server for list (not get) to see if this gvk even exists and if not, fail on that? It would still have the problem that if it existed when this was created and later went away would have problems (or, I at least assume so, but not sure if that would stop progress).

mattmoor: > Could we hit the API server for list (not get) to see if this gvk even exists and if not, fail on that?
Yeah, I was thinking something similar.  I think that gives us the happiest middle ground that fails fast accurately.

vagababov: I think the client runs a background thread of sorts and pulls all the types anyway. So this might not even require call to the API server.

Harwayne: > Could we hit the API server for list (not get) to see if this gvk even exists and if not, fail on that?

That would fix the primary case, where the GVK just doesn't exist. But does not fix other cases, such as this service account has permission to Get and List, but not Watch. So the List succeeds, but then, by calling `WaitForCacheSync`, would still block this thread and therefore reconciler forever.

Calling the Watch API is probably a happy middle ground. It should solve 99% of our problems without much additional code complexity. At the cost of slightly increased function run time. But any problem that isn't caught by that would cause the reconcilers to get stuck forever. I can only think of fairly esoteric problems that would still get caught (e.g. the Watch permission is revoked between the call to the Watch API and the call to `WaitForCacheSync`).

My personal vote is for what we think is the most correct fix, even though it is more complex. Because when this fails, it is not easy to locate the issue to this code and has extermely bad impacts on the user's cluster.

---

The logic that we went through to make this change:
1. Try not to change the interface for callers of the method, as there are many, including some we are unaware of.
1. Regardless of the problem, we want to bound the amount of time this blocks a reconciler. As well as not spewing errors to the logs every second.
1. To bound the time, because `WaitForCacheSync` doesn't give us a way to provide a timeout, we need a separate goroutine to check the timeout and force things to return. We happened to decide to put the `WaitForCacheSync` in the background goroutine, as well as the timer. But, if needed, we could probably move one of them back to the main thread.
1. To stop the error logs, we need to stop the informer. Which we do by closing its stop channel. It shares a stop channel with _all_ the informers created by this factory, so first create a new, unique stop channel just for it, `infChannel`. 
1. If the factory's stop channel is closed, then we need to stop the informer, so have a goroutine that watches the factory's stop channel and if it is closed, close `infChannel`.

mattmoor: > the most correct fix

It is extremely uncommon in my experience for this much use of go routines and channels to be completely correct on its first try, and the probability of later bugs is significantly higher because the code is so much harder to understand.  There are almost certainly higher level synchronization primitives we could use here (e.g. errgroup, sync packages).

The "watch" hole is why I hadn't suggested this originally, but honestly the list/watch delta and the esoteric issues you mention will be exceedingly uncommon.  The two common failure modes will inevitably be: typo in the TypeMeta, forgetting to grant permissions.  For the more niche failure modes, as you say there's log spew, so it's not subtle when folks go looking.

I also don't think that even if the channel stuff is as-intended, that things will be peachy since returning an error here will probably percolate back through the Reconciler and resulting in requeuing, so:
1. At least one of two execution slots on the controller will be stalled for 30s each time this key is processed,
2. For that 30s there will still be the log spam.

Harwayne: > It is extremely uncommon in my experience for this much use of go routines and channels to be completely correct on its first try, and the probability of later bugs is significantly higher because the code is so much harder to understand. There are almost certainly higher level synchronization primitives we could use here (e.g. errgroup, sync packages).

Almost certainly true.

> For the more niche failure modes, as you say there's log spew, so it's not subtle when folks go looking.

While that is true, it takes a while to recognize no Foos are getting reconciled, to look at the logs, and track the error in the logs back to a Foo is trying to track something that doesn't exist. Now that we have done it and recognize the pattern, it is fairly quick, but that won't be true of users seeing the problem for the first time.

> I also don't think that even if the channel stuff is as-intended, that things will be peachy since returning an error here will probably percolate back through the Reconciler and resulting in requeuing, so:
> 
> 1. At least one of two execution slots on the controller will be stalled for 30s each time this key is processed,
> 2. For that 30s there will still be the log spam.

I agree.

So do you think the 99%, try Watching before calling `WaitForCacheSync` is the fix we should use?

apis/duck/typed.go
```diff
@@ -46,6 +46,11 @@ var _ InformerFactory = (*TypedInformerFactory)(nil)
 
 // Get implements InformerFactory.
 func (dif *TypedInformerFactory) Get(gvr schema.GroupVersionResource) (cache.SharedIndexInformer, cache.GenericLister, error) {
+	// Avoid error cases, like resource not exists.
```
vagababov:
```suggestion
	// Avoid error cases, like the GVR does not exist.
```
I'd also expand the comment to say that it's not a full check, that some RBACs might sneak by, but the window is very small.

apis/duck/typed_test.go
```diff
@@ -94,6 +95,26 @@ func TestSimpleList(t *testing.T) {
...
+		Client:       client,
+		Type:         &duckv1alpha1.AddressableType{},
+		ResyncPeriod: 1 * time.Second,
+		StopChannel:  stopCh,
+	}
+
+	want := errors.New("failed to get list")
+	_, _, got := tif.Get(SchemeGroupVersion.WithResource("resources"))
+
+	if got == nil || got.Error() != want.Error() {
```
vagababov: you can do this easier, by using a variable (global to the test, or passing it to the invalid resource, up to you), so that it is returned from the call. 
That way you can just do
```
var testErr = errors.New("...")
...
if got != testErr {
  t.Errorf(....)
}
```

apis/duck/typed_test.go
```diff
@@ -94,6 +95,25 @@ func TestSimpleList(t *testing.T) {
...
+		Client:       client,
+		Type:         &duckv1alpha1.AddressableType{},
+		ResyncPeriod: 1 * time.Second,
+		StopChannel:  stopCh,
+	}
+
+	_, _, got := tif.Get(SchemeGroupVersion.WithResource("resources"))
+
+	if got != testErr {
+		t.Errorf("Didn't see expected error message. Got: %v, wanted: %v", got, testErr)
```
vagababov:
```suggestion
		t.Errorf("Error = %v, want: %v", got, testErr)
```

PR #881, opened 2019-11-15
Allow glog to coexist with klog

coryrc:
glog isn't used in our repos with good reason, but performance tests
use it, for now. This commit removes the ability to set klog flags
that aren't shared with glog, in return for the ability of glog to be imported.

/assign @chizhg


chizhg:
`Please run hack/update-codegen.sh`

chizhg:
/cc @vagababov 
since he also reviewed your last PR.

chizhg:
I have verified that the benchmarks will be working with the change.

PR #896, opened 2019-11-24
Reduce the boilerplate for setting up duck.InformerFactory's.

mattmoor:
This augments the injection codegen with the capability to produce a duck.InformerFactory
attached to context for each type that we process.

Now a `duck.InformerFactory` for "Addressable" can be produced by "Get"ing it from the context.

This is triggered by placing `// +genduck` on the type that implements `duck.Implementable`.

/hold

I am going to stage this into the downstream repos (partially to make sure everything WAI).


n3wscott:
This is going to be so awesome. 👏 😎 

/lgtm

PR #899, opened 2019-11-25
Fix the serialized name of the spec field (oops)

mattmoor:


apis/duck/v1alpha1/binding_types.go
```diff
@@ -34,7 +34,7 @@ type Binding struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec BindingSpec `json:"status"`
```
n3wscott: checks out.

PR #903, opened 2019-11-26
Add an -imagetemplate option to allow for more flexible test image locations.

markusthoemmes:
Some systems do weird things and or I could imagine somebody does not want to "pollute" their repository with test images. I for instance wanted to prefix images with `knative-serving-test`, but the current code doesn't allow for that.

This adds a templating option which gives a little more flexibility for running the tests.

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -20,12 +20,14 @@ limitations under the License.
 package test
 
 import (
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Tested for non-breakage here: https://github.com/knative/serving/pull/6102

PR #905, opened 2019-11-26
Resolvev1destination

### LINKED ISSUE #890: URIResolver uses v1beta1.Destination instead of v1
/area API
/kind cleanup

## Expected Behavior

URIResolver uses v1.Destination

## Actual Behavior

URIResolver uses v1beta1.Destination, requiring code using it to translate between the two types.

## Additional Info
Alternately, if v1beta1.Destination is the right type for URIResolver, add method(s) to Destination types that translate between them.

vaikas:
Fixes #890 

Add a new method URIFromDestinationV1 for resolving v1.Destination
Since v1beta1 and v1 Addressable are the same, watch for v1 duck addressable.

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -50,15 +50,7 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
 	ret.tracker = tracker.New(callback, controller.GetTrackerLease(ctx))
 	ret.informerFactory = &pkgapisduck.CachedInformerFactory{
 		Delegate: &pkgapisduck.EnqueueInformerFactory{
-			Delegate: addressable.Get(ctx),
-			/*
-				Delegate: &pkgapisduck.TypedInformerFactory{
-					Client:       dynamicclient.Get(ctx),
-					Type:         &duckv1.AddressableType{},
-					ResyncPeriod: controller.GetResyncPeriod(ctx),
-					StopChannel:  ctx.Done(),
-				},
-			*/
+			Delegate:     addressable.Get(ctx),
 			EventHandler: controller.HandleAll(ret.tracker.OnChanged),
 		},
 	}
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -27,6 +27,7 @@ import (
 
 	"knative.dev/pkg/apis"
 	pkgapisduck "knative.dev/pkg/apis/duck"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
 	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/network"
@@ -41,14 +42,23 @@ type URIResolver struct {
 	informerFactory pkgapisduck.InformerFactory
 }
 
-// NewURIResolver constructs a new URIResolver with context and a callback passed to the URIResolver's tracker.
+// NewURIResolver constructs a new URIResolver with context and a callback
+// for a given listableType (Listable) passed to the URIResolver's tracker.
 func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *URIResolver {
 	ret := &URIResolver{}
 
 	ret.tracker = tracker.New(callback, controller.GetTrackerLease(ctx))
 	ret.informerFactory = &pkgapisduck.CachedInformerFactory{
 		Delegate: &pkgapisduck.EnqueueInformerFactory{
-			Delegate:     addressable.Get(ctx),
+			Delegate: addressable.Get(ctx),
+			/*
+				Delegate: &pkgapisduck.TypedInformerFactory{
+					Client:       dynamicclient.Get(ctx),
+					Type:         &duckv1.AddressableType{},
+					ResyncPeriod: controller.GetResyncPeriod(ctx),
+					StopChannel:  ctx.Done(),
+				},
+			*/
 			EventHandler: controller.HandleAll(ret.tracker.OnChanged),
 		},
 	}
@@ -56,7 +66,7 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
 	return ret
 }
 
-// URIFromDestination resolves a Destination into a URI string.
+// URIFromDestination resolves a v1beta1.Destination into a URI string.
 func (r *URIResolver) URIFromDestination(dest duckv1beta1.Destination, parent interface{}) (string, error) {
 	var deprecatedObjectReference *corev1.ObjectReference
 	if dest.DeprecatedAPIVersion == "" && dest.DeprecatedKind == "" && dest.DeprecatedName == "" && dest.DeprecatedNamespace == "" {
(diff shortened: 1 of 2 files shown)


resolver/addressable_resolver.go
```diff
@@ -103,6 +113,33 @@ func (r *URIResolver) URIFromDestination(dest duckv1beta1.Destination, parent in
...
 
+// URIFromDestinationV1 resolves a v1.Destination into a URI string.
+func (r *URIResolver) URIFromDestinationV1(dest duckv1.Destination, parent interface{}) (string, error) {
+	if dest.Ref != nil {
+		url, err := r.URIFromObjectReference(dest.Ref, parent)
+		if err != nil {
+			return "", err
+		}
+		if dest.URI != nil {
+			if dest.URI.URL().IsAbs() {
```
n3wscott: I thought we loosened this to just use `ResolveReference` when not nil and if the uri is a full rewrite, so be it.

vaikas: The old one still uses it above, so I didn't think so? I'd rather decouple that change from this though if that's cool?

resolver/addressable_resolver.go
```diff
@@ -103,6 +105,33 @@ func (r *URIResolver) URIFromDestination(dest duckv1beta1.Destination, parent in
...
+				return "", fmt.Errorf("absolute URI is not allowed when Ref or [apiVersion, kind, name] exists")
```
vagababov: nit: no formatting, so just `return "", errors.New("...")`.

resolver/addressable_resolver.go
```diff
@@ -103,6 +105,33 @@ func (r *URIResolver) URIFromDestination(dest duckv1beta1.Destination, parent in
...
+			}
+			return url.URL().ResolveReference(dest.URI.URL()).String(), nil
+		}
+		return url.URL().String(), nil
+	}
+
+	if dest.URI != nil {
+		// IsAbs check whether the URL has a non-empty scheme. Besides the non non-empty scheme, we also require dest.URI has a non-empty host
+		if !dest.URI.URL().IsAbs() || dest.URI.Host == "" {
+			return "", fmt.Errorf("URI is not absolute(both scheme and host should be non-empty): %v", dest.URI.String())
```
vagababov:
```suggestion
			return "", fmt.Errorf("URI is not absolute(both scheme and host should be non-empty): %q", dest.URI.String())
```

resolver/addressable_resolver.go
```diff
@@ -103,6 +105,33 @@ func (r *URIResolver) URIFromDestination(dest duckv1beta1.Destination, parent in
...
+		}
+		return dest.URI.String(), nil
+	}
+
+	return "", fmt.Errorf("destination missing Ref and URI, expected at least one")
```
vagababov: ditto here re: `errors.New`

resolver/addressable_resolver_test.go
```diff
@@ -360,6 +361,233 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
+			r := resolver.NewURIResolver(ctx, func(types.NamespacedName) {})
+
+			// Run it twice since this should be idempotent. URI Resolver should
+			// not modify the cache's copy.
+			_, _ = r.URIFromDestinationV1(tc.dest, getAddressable())
+			uri, gotErr := r.URIFromDestinationV1(tc.dest, getAddressable())
+
+			if gotErr != nil {
+				if tc.wantErr != nil {
+					if diff := cmp.Diff(tc.wantErr.Error(), gotErr.Error()); diff != "" {
```
vagababov: `.Error()` is a string, just compare with `got != want`.

resolver/addressable_resolver_test.go
```diff
@@ -360,6 +361,233 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
+						t.Errorf("%s: unexpected error (-want, +got) = %v", n, diff)
+					}
+				} else {
+					t.Errorf("%s: unexpected error: %v", n, gotErr.Error())
```
vagababov: `t.Run` already specifies the test name; `%s: ` in the beginning is redundant.

resolver/addressable_resolver_test.go
```diff
@@ -360,6 +361,233 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
+				}
+			}
+			if gotErr == nil {
```
vagababov: do a continue in the `if gotErr != nil` above, then remove if here and unindent.

resolver/addressable_resolver_test.go
```diff
@@ -360,6 +361,233 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
+				got := uri
+				if diff := cmp.Diff(tc.wantURI, got); diff != "" {
```
vagababov: same here: strings just `!=` and no need to print the test name.

resolver/addressable_resolver_test.go
```diff
@@ -360,6 +361,233 @@ func TestGetURI_ObjectReference(t *testing.T) {
 	}
 }
 
+func TestGetURI_DestinationV1(t *testing.T) {
+	tests := map[string]struct {
+		objects []runtime.Object
+		dest    duckv1.Destination
+		wantURI string
+		wantErr error
```
vagababov: since you're comparing error strings, just make `wantErr` a string.

n3wscott:
/lgtm
/hold
hold for if you want to follow up on the victor comments 👍

resolver/addressable_resolver_test.go
vaikas: Sorry, are you saying get rid of cmp.Diff? I like the fact the diff actually makes it easier to understand what the diffs are and not just that they are different. Or maybe I misunderstood what you meant here :)

vaikas:
Thanks, @vagababov PTAL.

resolver/addressable_resolver_test.go
```diff
@@ -58,14 +59,14 @@ func init() {
...
+func TestGetURI_DestinationV1Beta1(t *testing.T) {
 	tests := map[string]struct {
 		objects []runtime.Object
 		dest    duckv1beta1.Destination
 		wantURI string
-		wantErr error
+		wantErr string
 	}{"nil everything": {
-		wantErr: fmt.Errorf("destination missing Ref, [apiVersion, kind, name] and URI, expected at least one"),
+		wantErr: fmt.Sprintf("destination missing Ref, [apiVersion, kind, name] and URI, expected at least one"),
```
vagababov:
```suggestion
		wantErr: "destination missing Ref, [apiVersion, kind, name] and URI, expected at least one",
```

resolver/addressable_resolver_test.go
```diff
@@ -100,7 +101,7 @@ func TestGetURI_ObjectReference(t *testing.T) {
 				DeprecatedAPIVersion: addressableAPIVersion,
 				DeprecatedNamespace:  testNS,
 			},
-			wantErr: fmt.Errorf("ref and [apiVersion, kind, name] can't be both present"),
+			wantErr: fmt.Sprintf("ref and [apiVersion, kind, name] can't be both present"),
```
vagababov:
```suggestion
			wantErr: "ref and [apiVersion, kind, name] can't be both present",
```
And same below where there's no formatting.

resolver/addressable_resolver_test.go
```diff
@@ -298,36 +299,36 @@ func TestGetURI_ObjectReference(t *testing.T) {
...
 			objects: []runtime.Object{
 				getAddressableNoStatus(),
 			},
 			dest:    duckv1beta1.Destination{Ref: getUnaddressableRef()},
-			wantErr: fmt.Errorf(`address not set for %+v`, getUnaddressableRef()),
+			wantErr: fmt.Sprintf(`address not set for %+v`, getUnaddressableRef()),
 		}, "notFound": {
 			dest:    duckv1beta1.Destination{Ref: getUnaddressableRef()},
-			wantErr: fmt.Errorf(`failed to get ref %+v: %s "%s" not found`, getUnaddressableRef(), unaddressableResource, unaddressableName),
+			wantErr: fmt.Sprintf(`failed to get ref %+v: %s "%s" not found`, getUnaddressableRef(), unaddressableResource, unaddressableName),
```
vagababov: here instead of `"%s"`   you can just do `%q`.

PR #916, opened 2019-11-29
Use duck_test package to avoid dependency cycles.

mattmoor:
As seen in: https://github.com/knative/pkg/pull/915 :(


mattmoor:
Looks like the tracker tests have similar problems, so it may be simpler to move `testbindable` into a subpackage of `pkg/testing` :(

PR #921, opened 2019-12-03
golang format tools

mattmoor:
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -type f -name '*.go' -print))`
  `goimports -w $(find -name '*.go' | grep -v vendor)`
/assign n3wscott


vaikas:
Why are we modifying vendor files?

mattmoor:
@vaikas I'm moving the knobots over to Tekton and there must be a bug in the gotool tasks, need to debug...

vaikas:
aha, ok :) Was just confused :)

Issue #842, opened 2019-11-04
Make Stackdriver metrics exporter watch Stackdriver Secret


anniefu:
Follow-up to enhance #824

Goal is to have this in by Nov 15, 2019.

## Expected Behavior
The metrics Stackdriver exporter can support reading a k8s Secret to configure the Stackdriver client with for authentication.

When the specified Stackdriver Secret for Stackdriver metrics exporter changes (same name/namespace, but is modified in-place), the Stackdriver exporter should be re-created with the new Secret for authenticating to Stackdriver.

## Actual Behavior
Stackdriver exporter continues on with the original Secret until the next metrics ConfigMap update that triggers an exporter update.

## Steps to Reproduce the Problem

1. Run Knative on GKE
2. Add a GCP service account credential as a k8s Secret named "stackdriver-service-account-key" to "knative-serving" namespace
3. Set "metrics.backend-destination" in "config-observability" to "stackdriver"
4. Set "metrics.stackdriver-use-secret" in "config-observability" to "true" 
5. Check the container logs for a "knative-serving" namespace Knative controller, see that the metrics exporter has updated and the config includes "stackdriver" and "true"
6. Update the k8s Secret however you feel
7. Check the same container logs, see there is no new metrics exporter update.

anniefu:
Closing this issue since it will become obsolete with the new metrics plan to use the opencensus agent exporter to export to the OpenCensus collector service: https://github.com/knative/pkg/blob/master/metrics/README.md

PR #925, opened 2019-12-05
Add SecretWatcher for Stackdriver Secret to internal metrics package

anniefu:
Add a SecretWatcher that can be used to update the metrics exporter when the Stackdriver Secret is changed.

Changes 
- Add "knative.dev/metrics/internal/stackdriver" pkg
- Add SecretWatcher interface and implementation that can be used to watch Secret

Part of #842, using SecretWatcher in stackdriver_exporter.go will come in a follow-up PR.

Because we would like to be able to extract custom metrics backend code out of Knative at some point, this explicitly does not take a dependency on "knative.dev/pkg/injection".

### DIFF
--- a/metrics/internal/stackdriver/secret_watcher.go
+++ b/metrics/internal/stackdriver/secret_watcher.go
@@ -1,0 +1,279 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package stackdriver
+
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+// Observer is the interface for callbacks that notify an Observer of the latest
+// state of a secret being watched by a secretWatcher.  An Observer should not modify the provided
+// Secrets, and should `.DeepCopy()` it for persistence (or otherwise process its
+// contents).
+type Observer interface {
+	OnAdd(*corev1.Secret)
+	OnUpdate(*corev1.Secret, *corev1.Secret)
+	OnDelete(*corev1.Secret)
+}
+
+// ObserverFuncs implements the observer interface.
+type ObserverFuncs struct {
+	AddFunc    func(*corev1.Secret)
+	UpdateFunc func(*corev1.Secret, *corev1.Secret)
+	DeleteFunc func(*corev1.Secret)
+}
+
+// OnAdd is the OnAdd of the observer interface.
+func (of *ObserverFuncs) OnAdd(s *corev1.Secret) {
+	// if of.AddFunc == nil {
+	// 	return
+	// }
+	of.AddFunc(s)
+}
+
+// OnUpdate is the OnUpdate of the observer interface.
+func (of *ObserverFuncs) OnUpdate(sOld *corev1.Secret, sNew *corev1.Secret) {
+	// if of.UpdateFunc == nil {
+	// 	return
+	// }
+	of.UpdateFunc(sOld, sNew)
+}
+
+// OnDelete is the OnDelete of the observer interface.
+func (of *ObserverFuncs) OnDelete(s *corev1.Secret) {
+	// if of.DeleteFunc == nil {
+	// 	return
+	// }
+	of.DeleteFunc(s)
+}
+
+// SecretWatcher defines the interface that a secret watcher must implement.
+type SecretWatcher interface {
+	// StartWatch starts the secret watch allowing observer callbacks to be triggered.
+	StartWatch() error
+
+	// StopWatch stops the secret watch.
+	StopWatch()
+
+	// GetSecret gets the value of the specified secret.
+	GetSecret(namespace string, name string) (*corev1.Secret, error)
+
+	// GetNamespaceWatched returns the namespace of the Secret being watched;
+	// returns empty string if more than one namespace is being watched.
+	GetNamespaceWatched() string
+
+	// GetNameWatched returns the name of the Secret being watched;
+	// returns empty string if more than one Secret is being watched.
+	GetNameWatched() string
+}
+
+var (
+	// createStackdriverKubeclientFunc provides a hook for setting up the kubeclient for testing.
+	// It is set to createKubeclient otherwise.
+	createStackdriverKubeclientFunc func() (kubernetes.Interface, error)
+)
+
+func init() {
+	createStackdriverKubeclientFunc = createKubeclient
+}
+
+// NewSecretWatcher constructs a secretWatcher that watches all Secrets.
+func NewSecretWatcher(observers ...Observer) (SecretWatcher, error) {
+	return newSecretWatcherInternal("", "", observers...)
+}
+
+// NewSecretWatcherSingleNamespace constructs a secretWatcher that watches all Secrets in a namespace.
+func NewSecretWatcherSingleNamespace(namespace string, observers ...Observer) (SecretWatcher, error) {
+	return newSecretWatcherInternal(namespace, "", observers...)
+}
+
+// NewSecretWatcherSingleSecret constructs a secretWatcher that watches all a single Secret.
+func NewSecretWatcherSingleSecret(namespace string, name string, observers ...Observer) (SecretWatcher, error) {
+	return newSecretWatcherInternal(namespace, name, observers...)
+}
+
+// newSecretWatcherInternal constructs a secretWatcher.
+func newSecretWatcherInternal(namespace string, name string, observers ...Observer) (SecretWatcher, error) {
+	impl := &secretWatcherImpl{
+		SecretNamespace: namespace,
+		SecretName:      name,
+		observers:       observers,
+	}
+
+	if clientErr := impl.setupKubeclient(); clientErr != nil {
+		return nil, clientErr
+	}
+
+	return impl, nil
+}
+
+// secretWatherImpl implmements the secretWatcher interface.
+type secretWatcherImpl struct {
+	// kubeclient is the in-cluster Kubernetes kubeclient, which is lazy-initialized on first use.
+	kubeclient kubernetes.Interface
+
+	// watchLock ensures that only one watch is being run at a time.
+	watchLock sync.Mutex
+	// informer watches the Stackdriver secret if useStackdriverSecretEnabled is true.
+	informer cache.SharedIndexInformer
+	// stopCh stops the secretWatcherInformer when required.
+	stopCh chan struct{}
+	// watchStarted is whether or not the informer is running for the secret being watched.
+	watchStarted bool
+
+	// SecretName is the name of the secret being watched.
+	SecretName string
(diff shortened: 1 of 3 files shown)


metrics/internal/stackdriver/secret_watcher.go
```diff
@@ -0,0 +1,279 @@
...
+	if !s.watchStarted {
+		return
+	}
+	s.watchStarted = false
+
+	close(s.stopCh)
+}
+
+// GetSecret implements GetSecret of secretWatcher interface.
+func (s *secretWatcherImpl) GetSecret(namespace string, name string) (*corev1.Secret, error) {
```
trshafer: Is GetSecret necessary? Should the only interface be on the observer?

metrics/internal/stackdriver/secret_watcher.go
```diff
@@ -0,0 +1,279 @@
...
+	if namespace == "" || name == "" {
+		return nil, fmt.Errorf("Must specify a non-empty Secret namespace and name to get, namespace specified: %v, name specified: %v", namespace, name)
+	}
```
trshafer: Should this check also live on the NewSecretWatcher initializers?

metrics/internal/stackdriver/secret_watcher.go
anniefu: It is not strictly necessary for a SecretWatcher, but it is convenient to include here to avoid duping the kubeclient initialization because it can reuse the same kubeclient as the SecretWatcher. 

Could you clarify what you mean by the only interface on the observer?

metrics/internal/stackdriver/secret_watcher.go
anniefu: Yes, that would be a good check, I'll add it.

metrics/internal/stackdriver/secret_watcher.go
trshafer: Should the only API to receive the secret be the observer? Meaning the API does not support a non-observable way to fetch the secret. It ensures that consumers must respect that the secret value can change and build their usage accordingly.

anniefu: Ah, I see. That's an interesting idea, but it may be a bit limiting in usage. (Also, consumers can always cache the last Secret they saw from an observer which is potentially less desirable than allowing getting the most up-to-date value). 

In the case of watching the Stackdriver Secret, I would like check the Secret's existence with a Get before calling StartWatch. That way I can ensure (stackdriver secret configured && stackdriver secret exists) before burning any extra CPU to watch the Secret in the background.

trshafer: I understand the usecase of checking the configmap if the secret type is desired before creating this observable. But if the user specifies the stackdriver secret as desired, then an observer will need to be created eventually. Does the observer emit the current value set before the observer was created, or only on updates?

anniefu: The observer does get "create" notifications for all pre-existing secrets when the watch starts. I see your point that we should start the SecretWatch as long as "use-secret" configmap field is set. 

I'll remove the GetSecret func from the interface. It might create some disconnect for the user between when they update config-observability & when they actually see the metrics exporter get updated, but I think we can make that acceptable with some extra logging.

metrics/internal/stackdriver/secret_watcher_test.go
```diff
@@ -0,0 +1,569 @@
...
+	oFuncs         *ObserverFuncs
+	onAddCalled    bool
+	onUpdateCalled bool
+	onDeleteCalled bool
+	mux            sync.Mutex
+}
+
+// newTestObserver constructs a testObserver from observerFuncs.
+func newTestObserver(inputObsFuncs *ObserverFuncs) *testObserver {
+	// If a callback is nil, consider it to be called by default.
```
yanweiguo: Why default to true when a callback is nil?

metrics/internal/stackdriver/secret_watcher_test.go
```diff
@@ -0,0 +1,569 @@
...
+
+func (tObs *testObserver) markCallbackCalled(callbackSentinel *bool) {
+	tObs.mux.Lock()
+	defer tObs.mux.Unlock()
+	*callbackSentinel = true
+}
+
+func (tObs *testObserver) OnAdd(s *corev1.Secret) {
+	tObs.oFuncs.OnAdd(s)
+	tObs.markCallbackCalled(&tObs.onAddCalled)
```
yanweiguo: I prefer to just use `tObs.onAddCalled = true` and add the lock around and get rid of  `markCallbackCalled`.

metrics/internal/stackdriver/secret_watcher_test.go
```diff
@@ -0,0 +1,569 @@
...
+	// The testObserver wrapper struct will invoke the funcs from "o" and track whether they were called
+	tObs := newTestObserver(o)
+	watcher := defaultGlobalSecretWatcherConstructor(t, tObs)
+
+	watcher.StartWatch()
+	defer watcher.StopWatch()
+
+	kubeclientForTest.CoreV1().Secrets(dsft.namespace).Create(&dsft.secret)
+	waitForCondition(t, func() bool {
+		return tObs.CheckOnAddCalled()
```
yanweiguo: I don't think it make sense to test it has been called as you just set `called` to true. I would just do the add/update/delete operations and make sure they don't introduce any error when the callback funcs are nil.

metrics/internal/stackdriver/secret_watcher_test.go
```diff
@@ -0,0 +1,569 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if test.testSetupFunc != nil {
+				test.testSetupFunc()
+			} else {
+				testSetup()
+			}
+
+			obsFuncs := &ObserverFuncs{
```
yanweiguo: What's the difference between this one and `defaultObserverFuncs`?

metrics/internal/stackdriver/secret_watcher_test.go
anniefu: It is for the conveinence of the AllCallbacksCalled() helper function. If only the AddFunc is defined, I want that to still return "true".

metrics/internal/stackdriver/secret_watcher_test.go
anniefu: markCallbackCalled exists to prevent code duplication of the lock/unlock call. Otherwise each of OnAdd, OnUpdate, and OnDelete would have to reimplement that part. (Without the lock, gotest will report a Data Race when trying to mark the callback as called).

metrics/internal/stackdriver/secret_watcher_test.go
anniefu: Oh nothing, I'll use defaultObserverFuncs.

metrics/internal/stackdriver/secret_watcher.go
```diff
@@ -0,0 +1,265 @@
...
+type SecretWatcher interface {
+	// StartWatch starts the secret watch allowing observer callbacks to be triggered.
+	StartWatch() error
+
+	// StopWatch stops the secret watch.
+	StopWatch()
+
+	// GetNamespaceWatched returns the namespace of the Secret being watched;
+	// returns empty string if more than one namespace is being watched.
+	GetNamespaceWatched() string
```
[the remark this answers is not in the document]
anniefu: It is purely for convenience. I imagine it could be useful for logging. For example, something like "Stopped SecretWatcher watching {watcher.GetNamespace()/watcher.GetNameWatched()}".

metrics/internal/stackdriver/secret_watcher_test.go
anniefu: Hmm good catch... I cannot call add/update/delete on a SecretWatcher directly as that is not part of the interface. I will refactor testObs to not consider nil funcs to be "called" by default.

metrics/internal/stackdriver/secret_watcher.go
```diff
@@ -0,0 +1,265 @@
...
+	watchLock sync.Mutex
+	// informer watches the Stackdriver secret if useStackdriverSecretEnabled is true.
+	informer cache.SharedIndexInformer
+	// stopCh stops the secretWatcherInformer when required.
+	stopCh chan struct{}
+	// watchStarted is whether or not the informer is running for the secret being watched.
+	watchStarted bool
+
+	// SecretName is the name of the secret being watched.
+	SecretName string
```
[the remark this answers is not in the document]
anniefu: Yeah, that makes sense since they're exposed through SecretWatcher API.

metrics/internal/stackdriver/secret_watcher.go
```diff
@@ -0,0 +1,265 @@
...
+	s.stopCh = make(chan struct{})
+	go s.informer.Run(s.stopCh)
+
+	return nil
+}
+
+// StopWatch implements StopWatch secretWatcher interface.
+func (s *secretWatcherImpl) StopWatch() {
+	s.watchLock.Lock()
+	defer s.watchLock.Unlock()
```
mattmoor: Remove trailing whitespace:
```suggestion
	defer s.watchLock.Unlock()
```

metrics/internal/stackdriver/secret_watcher_test.go
anniefu: Turns out I stopped using AllCallbacksCalled() at some point, so I've removed that function all together and removed the "true" default for nil funcs.

mattmoor:
If the plan is to link this into `queue-proxy` then what you are doing here really worries me and/or won't work.  The QP runs as the same SA as user code, so anything this logic can do while in the QP is also available to user code, and it seems that extends to being able to read secrets from a namespace that runs highly privileged services.

You should have a privileged component copy the secret into user namespaces and leverage a binding to link it in as a volume (this requires zero additional capabilities granted to the user-container).  It should be trivial to create a binding that applies to all Knative services in a namespace that mounts a volume into the queue-proxy and sets `GOOGLE_APPLICATION_DEFAULT` to point there.  I would be happy to sit down with someone to explain this or work out some alternatives.

We really need to stop building debt in the observability space. 🙏

anniefu:
The end goal is to enable Stackdriver metrics for Knative outside of running on GKE. 

The plan to propagate the "Stackdriver Secret" is to use a RoleBinding to give all ServiceAccounts permission to read just that one Secret in the "knative-serving" namespace, where the name  of the "Stackdriver Secret" is hardcoded and not configurable.

mattmoor:
I know what the goal is 😉

Do you have the example role and rolebinding you plan to use?

-M(obile)

anniefu:
Something like:
```
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: knative-serving
  name: knative-serving-stackdriver-secret-user
rules:
- apiGroups: [""]
  resources: ["secrets"]
  resourceNames: ["stackdriver-service-account-key"]
  verbs: ["get", "list", "watch"]
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: knative-serving
  name: use-stackdriver-secret
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: knative-serving-stackdriver-secret-user
subjects:
- kind: Group
  name: system:serviceaccounts
```

mattmoor:
Don't you need a RoleBinding like that in every namespace?

anniefu:
No, just the one is sufficient. The namespace on the RoleBinding refers to the namespace of the resource being granted access to instead of the namespace of the subjects. For each subject, you can also scope by namespace, but here it is just all "system:serviceaccounts".

The docs around this are a bit confusing IMO, but there's a full description here https://kubernetes.io/docs/reference/access-authn-authz/rbac/

anniefu:
@mattmoor any thoughts here? Also, do you see making a decision on how best to propagate the Secret as a blocker for this PR? 

I view this PR as independent of what changes need to be done in knative/serving. The metrics pkg appropriately assumes it can read the Secret or returns an error if it can't.

mattmoor:
I'll add this to the TOC agenda today.  In general, linking in vendor logic like this doesn't scale, and we need to start addressing that.

We've been operating in a fairly tactical short-term mode where this debt builds without a long-term plan to address it, and that worries me.  We need to start planning how we address this.

I apologize that this meta issue that's much broader than your change is jamming this up, but we need to start having this conversation.

anniefu:
Knative's embedded Stackdriver has become obsolete with https://github.com/knative/pkg/blob/master/metrics/README.md

PR #926, opened 2019-12-06
Add a junithelper for creating a junit result from command

chaodaiG:
Add this util so that we can create a custom junit result from shell

/cc @chizhg

### DIFF
--- a/testutils/junithelper/main.go
+++ b/testutils/junithelper/main.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


testutils/junithelper/main.go
```diff
@@ -0,0 +1,57 @@
...
+func main() {
+	flag.StringVar(&suite, "suite", "", "Name of suite")
+	flag.StringVar(&name, "name", "", "Name of test")
+	flag.StringVar(&errMsg, "err-msg", "", "Error message, empty means test passed, default empty")
+	flag.StringVar(&dest, "dest", "junit_result.xml", "Where junit xml writes to")
+	flag.Parse()
+
+	suites := junit.TestSuites{}
+	suite := junit.TestSuite{Name: suite}
+	var errP *string
```
chizhg: Do we need this? Cannot we just use `&errMsg`?

chaodaiG: Unfortunately we can't, it has to be null pointer when test passed, but errMsg always got assigned the default value and is never null

Issue #883, opened 2019-11-16
Unite underlying log libraries into Zap


coryrc:
Our project has chosen to use Zap for structured logging. Some of our dependencies (Mako, client-go) use other logging libraries: glog, klog. Those two don't coexist well and we shouldn't have several knobs to control logging.

Istio team had this same issue and has a very neat solution of writing API wrapper with calls to Zap, and updating go.mod to redirect imports for github.com/golang/glog and k8s.io/klog to the wrappers. We can do the same thing with Go dep and reuse Istio's wrappers.

Then all logging can be controlled with one package. This needs to be done in all repos.

Istio wrappers:
https://github.com/istio/klog/
https://github.com/istio/glog/

coryrc:
Somewhat related: #441

coryrc:
klog provides a way: https://github.com/kubernetes/klog/blob/8422fac62d1e961e89426ffb5ae3a07f2d0bcca2/klog.go#L775

+ the glog->klog wrapper should do the job

grantr:
@coryrc do you think this would be a good first issue?

coryrc:
No.

I've sent PRs to the mentioned repos. Maybe we need to fork them or I need to directly contact people to get the changes through.

grantr:
Ok, thanks for explaining. I didn't realize this was about getting repos owned by other organizations to change their logging libraries. That doesn't seem like a good first issue at all.

coryrc:
knative/serving#6976 could be completed if the repos at [glog](https://github.com/coryrc/glog) and [klog](https://github.com/coryrc/klog) were brought into knative-sandbox.
/unassign

PR #937, opened 2019-12-10
Add glog/klog->Zap redirect designs to repo

coryrc:
This is progress toward #883.

It is not possible to test updating the Gopkg.toml to override usage
of glog or klog to these libraries until they are already in the repo.

logging.go has been updated to supply the value logging.Verbosity, but
it's not intended to be the final way it is setup.

Can't use istio/[gk]log because:

1. It doesn't compile
2. It doesn't use verbosity levels, and I don't want istio to need our changes.
3a. We usually don't need glog, so I don't want to force-include it just to set verbosity somehow
3. I don't want to be dependent on istio to make changes if any are needed (given the file is so small and simple)

### DIFF
--- a/test/logging/glog/glog.go
+++ b/test/logging/glog/glog.go
@@ -42,7 +42,7 @@ func Flush() {
 
 // V is a shim
 func V(level Level) Verbose {
-	return Verbose(level <= logging.Verbosity)
+	return Verbose(int(level) <= logging.Verbosity)
 }
 
 // Info is a shim
--- a/test/logging/glog/glog.go
+++ b/test/logging/glog/glog.go
@@ -1,0 +1,181 @@
+// Copyright 2019 Knative Authors
+// Copyright 2017 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package glog exposes an API subset of the [glog](https://github.com/golang/glog) package.
+// All logging state delivered to this package is shunted to the global [zap logger](https://github.com/uber-go/zap).
+//
+// Istio is built on top of zap logger. We depend on some downstream components that use glog for logging.
+// This package makes it so we can intercept the calls to glog and redirect them to zap and thus produce
+// a consistent log for our processes.
+package glog
+
+import (
+	"fmt"
+	"os"
+
+	"go.uber.org/zap"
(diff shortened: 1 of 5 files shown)


chaodaiG:
Sorry but I don't quite get the idea behind this PR, could you either update the PR body to reflect more on why making this change, what this change is doing, or you can include several README or `doc.go` for each package you added to reflect what they do, or link a bug associate with it, so that it's easier to review the code. Thanks!

coryrc:
> Sorry but I don't quite get the idea behind this PR

Whoops, sorry, I wrote it in the branch name but didn't actually say anything. This is for #883 

The istio libraries mentioned there don't build (though the main repo istio/istio has the override in their go.mod!?) and don't have the proper input for controlling Verbosity, so I forked them into this repo. They're only single file and 2 pages long, so the cost seems trivial to manage. glog is abandoned so its API won't change & klog is swore to be API compatible until v2, at which time we can just use v2 because it allows logging to Zap directly on its own. So these should not have any maintenance concerns.

coryrc:
Can't use istio/[gk]log because:

1. It doesn't compile
2. It doesn't use verbosity levels, and I don't want istio to need our changes.
3a. We usually don't need glog, so I don't want to force-include it just to set verbosity somehow
3. I don't want to be dependent on istio to make changes if any are needed (given the file is so small and simple)

Chao to lgtm, but want @mattmoor to check the copyright stuff.

/assign @mattmoor

chaodaiG:
lgtm from my side so that Cory can be unblocked testing how these changes work. However this PR contains some code that's almost 100% copy/pasted from istio repos, see `// Copyright 201X Istio Authors` on top of the files added in this PR, and Cory had explained why not use the packages from istio straight in PR description above. @mattmoor , what do you think about the copyright, and how about the boilerplate?

chizhg:
> This is progress toward #883.
> 
> It is not possible to test updating the Gopkg.toml to override usage
> of glog or klog to these libraries until they are already in the repo.

I'm not sure if I understand this, but there should be a way to test it without merging this PR. Let's sync up offline tomorrow.

coryrc:
> > This is progress toward #883.
> > It is not possible to test updating the Gopkg.toml to override usage
> > of glog or klog to these libraries until they are already in the repo.
> 
> I'm not sure if I understand this, but there should be a way to test it without merging this PR. Let's sync up offline tomorrow.

I can test the library usage, but not how Gopkg.toml is supposed to do the override.

coryrc:
Because some people are confused: This has to be checked in before running `dep ensure -update` with a Gopkg.toml override source pointing to a package path "knative.dev/pkg/test/logging/[gk]log" and creating the PR that will test it.

vaikas:
@mattmoor 
@evankanderson 
Thoughts on the copyright?

vaikas:
@pmorie @rgregg @isdal As this seems a bit more on the Steering corner? Could you please provide guidance on this, please?

isdal:
Cory mentioned that he spoke to product counsel who approved the approach. That is good enough for me.

coryrc:
It turns out `dep` can't do this: https://github.com/golang/dep/issues/982

I think Go modules can, so this is on hold until then :(.

PR #946, opened 2019-12-16
This should be done directly in Serving, not in knative/pkg.

mattmoor:
/hold

While I get the serving side decoupled from this.


mattmoor:
/hold cancel

Ok the serving PR is out.

PR #948, opened 2019-12-16
Feature to serve the version using /version endpoint

waveywaves:
Using this feature it should be possible for knative products
to be able to leverage this featureto serve the curent version
of product in use. This would help in debugging based on the
version as well as facilitate checking cli + controller
compatibility.

Related to 
https://github.com/tektoncd/pipeline/pull/1650
https://github.com/tektoncd/cli/pull/503

cc @vdemeester

Signed-off-by: Vibhav Bobade <vibhav.bobde@gmail.com>

### DIFF
--- a/version/version_exporter.go
+++ b/version/version_exporter.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type versionRequestHandler struct {
+	version string
+}
+
+func (vrh *versionRequestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+  w.Write([]byte(fmt.Sprintf("%+v", vrh.version)))
+}
+
+func StartVersionSrv(version string) {
+	vs := http.NewServeMux()
+
+	vrh := &versionRequestHandler{version : version}
+	vs.Handle("/version", vrh)
+
+	http.ListenAndServe(":5678", vs)
+}


waveywaves:
Basically, I need to be able to serve the release version of the controller in use in tektoncd/pipline so that it can be consumed by the cli for showing the version.

version/version_exporter.go
```diff
@@ -0,0 +1,39 @@
...
+package version
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type versionRequestHandler struct {
+	version string
+}
```
mattmoor: This type is impossible to create, and the function below is impossible to call 🤔 

We have a convention where we have resources labeled with `serving.knative.dev/release: devel` and rewrite `devel` to a version when we `./hack/release.sh`.  Perhaps we should have this projected into an EnvVar or volume that this serves?

waveywaves: My bad, the StartVersionSrv is supposed to be public so that it can be called from tektoncd/pipeline. As `pkg` is used there as well for serving metrics. I have taken inspiration from there wrt serving the version.

waveywaves: If the version is reflected in the endpoints it would be easier for `cli` to query it for the release version.

waveywaves: The type is only consumed by StartVersionSrv, so it should be alright I suppose.

mattmoor: Hmm, not unless there's a publicly exposed network endpoint.  Perhaps it could key off of the label I gave above on a fixed resource (e.g. the tekton namespace)?

waveywaves: @mattmoor To clarify, you are suggesting that the tekton release be given by a label on a namespace or as a volume correct ? If that is so, should I make changes here or open a PR on tektoncd for this ?

waveywaves: Okay @mattmoor It makes more sense to do this directly in tekton then in terms of the label on the resource.
https://github.com/knative/serving/blob/916ab9f5ac54c9cd041ad7c502e02b3bc8f6ea17/config/500-webhook-configuration.yaml#L20

mattmoor: I could see value in having `/version` that keyed off of a label like what our convention is, but I doubt that is adequate for your use case since unless there's an externally accessible endpoint you can't hit it without tunneling into the cluster (unless you are running on-cluster).

If the CLI just looked at labels in a fixed place (e.g. on the Tekton namespace of CR defintion) then you could piggyback on the API server as your endpoint.

waveywaves: Was planning to have the release show up on the pods running the `controller` and the `webhook`. In that case I would be able to use the cli to pick the labels up from there.

waveywaves: @mattmoor I have opened up a PR directly in tektoncd/pipeline for setting annotation on the controller pod for the release. If I am not wrong it should work with CLI for querying the annotation off of the pod for getting the version.

PR #947, opened 2019-12-16
Generate Base Reconciler Code for Injection Based Controllers.

n3wscott:
There are a few cases where the setup of the controller and the wrapper around `Reconciler` are all the same. This aims to help reduce the amount of copy/paste code required to get a correct reconciler up and running. 

To enable generation of the Reconciler code, it is gated by a flag (**+genreconciler**) in the API structure:

```go

// +genclient
// +genreconciler
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object

// AddressableService is a Knative abstraction that encapsulates the interface by which Knative
// components express a desire to have a particular image cached.
type AddressableService struct {
...
```

This generates two things: a Reconciler/Controller pair that are intended to be imported and used, and the Stub of a ReconcilerImpl/ControllerImpl that import use the Reconciler/Controller with correct paths, intended to be copied out of the `./stub` directory.

See https://github.com/knative/sample-controller/pull/167 for full code changes.

Look at ApiServerSource for a more complex example: https://github.com/knative/eventing/pull/2487

PR #952, opened 2019-12-19
Creating an Event interface and ReconcilerEvent type to be used in ReconcileKind

n3wscott:
This will allow us to propagate events from the reconcilers easier.

Create wrapper functions like:

```go
func WarnSomeReason(msgf string, args ...interface{}) reconciler.Event {
	reconciler.NewEvent(corev1.EventTypeWarning, "SomeReason", msgf, args...)
}
```

Use from a reconciler like:

```go
if somethingBad {
  return WarnSomeReason("this was bad")
}
```

And then use the returned error inside the bigR reconciler like:

```go
var event *ReconcilerEvent
if reconciler.EventAs(e, &event) {
	recorder.Event(resource, event.EventType, event.Reason, event.Format, event.Args...)
}
```

Bigger picture, this ends up being used like:

```go
// Interface defines the strongly typed interfaces to be implemented by a
// controller reconciling v1alpha1.ContainerSource.
type Interface interface {
	// ReconcileKind implements custom logic to reconcile v1alpha1.ContainerSource. Any changes
	// to the objects .Status or .Finalizers will be propagated to the stored
	// object. It is recommended that implementors do not call any update calls
	// for the Kind inside of ReconcileKind, it is the responsibility of the core
	// controller to propagate those properties.
	ReconcileKind(ctx context.Context, o *v1alpha1.ContainerSource) reconciler.Event
}
```

Relates to https://github.com/knative/pkg/pull/947/

### DIFF
--- a/reconciler/errors.go
+++ b/reconciler/errors.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Veroute.on 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	"fmt"
+)
+
+// New returns a ReconcilerEvent fully populated.
+func New(eventtype, reason, messageFmt string, args ...interface{}) error {
+	return &ReconcilerEvent{
+		EventType:  eventtype,
+		Reason:     reason,
+		Format: messageFmt,
+		Args:       args,
+	}
+}
+
+// ReconcilerEvent wraps the fields required for recorders to create an Event.
+type ReconcilerEvent struct {
+	EventType  string
+	Reason     string
+	Format string
+	Args       []interface{}
+}
+
+// make sure ReconcilerEvent implements error.
+var _ error = (*ReconcilerEvent)(nil)
+
+// Is returns if the target error is a ReconcilerEvent type checking that
+// EventType and Reason match.
+func (e *ReconcilerEvent) Is(target error) bool {
+	if t, ok := target.(*ReconcilerEvent); ok {
+		if t != nil && t.EventType == e.EventType && t.Reason == e.Reason {
+			return true
+		}
+	}
+	return false
+}
+
+// Error returns the string that is formed by using the format string with the
+// provided args.
+func (e *ReconcilerEvent) Error() string {
+	return fmt.Sprintf(e.Format, e.Args...)
+}
--- a/reconciler/errors.go
+++ b/reconciler/errors.go
@@ -17,6 +17,7 @@ limitations under the License.
 package reconciler
 
 import (
+	"errors"
 	"fmt"
 )
 
@@ -48,6 +49,10 @@ func (e *ReconcilerEvent) Is(target error) bool {
 		if t != nil && t.EventType == e.EventType && t.Reason == e.Reason {
 			return true
 		}
+	} else {
+		// Allow for wrapped errors.
+		err := fmt.Errorf(e.Format, e.Args...)
+		return errors.Is(err, target)
 	}
 	return false
 }
--- a/reconciler/errors.go
+++ b/reconciler/errors.go
@@ -20,22 +20,22 @@ import (
 	"fmt"
 )
 
-// New returns a ReconcilerEvent fully populated.
-func New(eventtype, reason, messageFmt string, args ...interface{}) error {
+// NewReconcilerEvent returns a ReconcilerEvent fully populated.
+func NewReconcilerEvent(eventtype, reason, messageFmt string, args ...interface{}) error {
 	return &ReconcilerEvent{
-		EventType:  eventtype,
-		Reason:     reason,
-		Format: messageFmt,
-		Args:       args,
+		EventType: eventtype,
+		Reason:    reason,
+		Format:    messageFmt,
+		Args:      args,
 	}
 }
 
 // ReconcilerEvent wraps the fields required for recorders to create an Event.
 type ReconcilerEvent struct {
-	EventType  string
-	Reason     string
-	Format string
-	Args       []interface{}
+	EventType string
+	Reason    string
+	Format    string
+	Args      []interface{}
 }
 
 // make sure ReconcilerEvent implements error.
--- a/reconciler/errors_test.go
+++ b/reconciler/errors_test.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Veroute.on 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	"errors"
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	"testing"
+)
+
+var exampleStatusFailed = "ExampleStatusFailed"
+
+
+func TestNil_Is(t *testing.T) {
+	var err error
+	if errors.Is(err, New(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
+	}
+}
+
+func TestError_Is(t *testing.T) {
+	err := errors.New("some other error")
+	if errors.Is(err, New(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
+	}
+}
+
+func TestNew_Is(t *testing.T) {
+	err := New(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+	if !errors.Is(err, New(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
+	}
+}
+
+func TestNewOtherType_Is(t *testing.T) {
+	err := New(corev1.EventTypeNormal, exampleStatusFailed, "this is an example error, %s", "yep")
+	if errors.Is(err, New(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a Warn, ExampleStatusFailed")
+	}
+}
+
+func TestNewOtherReason_Is(t *testing.T) {
+	err := New(corev1.EventTypeWarning, "otherReason", "this is an example error, %s", "yep")
+	if errors.Is(err, New(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a Warn, ExampleStatusFailed")
+	}
+}
+
+func TestNew_As(t *testing.T) {
+	err := New(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+
+	var event *ReconcilerEvent
+	if errors.As(err, &event) {
+		if event.EventType != "Warning" {
+			t.Error("mismatched reason")
+		}
+		if event.Reason != exampleStatusFailed {
+			t.Error("mismatched eventtype")
+		}
+	} else {
+		t.Error("not a ReconcilerEvent")
+	}
+}
+
+func TestNil_As(t *testing.T) {
+	var err error
+
+	var event *ReconcilerEvent
+	if errors.As(err, &event) {
+		t.Error("not a ReconcilerEvent")
+	}
+}
+
+func TestNew_Error(t *testing.T) {
+	err := New(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+
+	want := "this is an example error, yep"
+	got := err.Error()
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("unexpected diff (-want, +got) = %v", diff)
+	}
+}
--- a/reconciler/errors_test.go
+++ b/reconciler/errors_test.go
@@ -18,6 +18,7 @@ package reconciler
 
 import (
 	"errors"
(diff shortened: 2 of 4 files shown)


n3wscott:
/hold
still validating

reconciler/errors_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestNew_Error(t *testing.T) {
+	err := New(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+
+	want := "this is an example error, yep"
+	got := err.Error()
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("unexpected diff (-want, +got) = %v", diff)
+	}
+}
```
mattmoor: Add trailing newline:
```suggestion
}

```

reconciler/errors.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"fmt"
+)
+
+// New returns a ReconcilerEvent fully populated.
+func New(eventtype, reason, messageFmt string, args ...interface{}) error {
+	return &ReconcilerEvent{
+		EventType:  eventtype,
+		Reason:     reason,
+		Format: messageFmt,
```
dprotaso: `go fmt`

reconciler/errors.go
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+package reconciler
+
+import (
+	"fmt"
+)
+
+// New returns a ReconcilerEvent fully populated.
+func New(eventtype, reason, messageFmt string, args ...interface{}) error {
```
dprotaso: `reconciler.New` seems misleading - since I'd expect it to return a reconciler

vagababov: Also why is the file `errors.go`?!

n3wscott: oh yeah, sorry I moved this from an error dir in my PoC, will rename.

n3wscott:
> What confuses me is that the corev1 events types are warnings & info - so returning them as errors is interesting.

Yeah it is a little bit of an abuse, but it works really well I think for this usecase.

> Secondly, should the `ReconcilerEvent` support wrapping errors?

Wrapping is supported by using `%w` in the format and passing an error. Ah but I see... Let me fix...

n3wscott:
Wrapped errors now supported and test added.

n3wscott:
/hold
Gonna work this to be returning an interface `ReconcilerResult` rather than an `error`

n3wscott:
ok reworked, Please take another look.

/hold cancel

reconciler/events.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Event leverages go's 1.13 error wrapping.
+type Event error
+
+// Is reports whether any error in err's chain matches target.
+//
+// The chain consists of err itself followed by the sequence of errors obtained by
+// repeatedly calling Unwrap.
+//
+// An error is considered to match a target if it is equal to that target or if
```
antoineco: TIL, thanks for the useful comments.

reconciler/events.go
```diff
@@ -0,0 +1,94 @@
...
+	if t, ok := target.(*ReconcilerEvent); ok {
+		if t != nil && t.EventType == e.EventType && t.Reason == e.Reason {
+			return true
+		}
+	} else {
+		// Allow for wrapped errors.
+		err := fmt.Errorf(e.Format, e.Args...)
+		return errors.Is(err, target)
+	}
+	return false
```
antoineco: I think you could simply get rid of the else instead of ending with that dummy return.

reconciler/events_test.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var exampleStatusFailed = "ExampleStatusFailed"
```
antoineco: You could use a constant.

reconciler/events_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func TestNil_Is(t *testing.T) {
+	var err error
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
```
antoineco: I find the failure message confusing. Did you mean "expected nil error not to be a ReconcilerError"? Future devs will thank you for the clear logs :)

reconciler/events_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func TestError_Is(t *testing.T) {
+	err := errors.New("some other error")
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
```
antoineco: See comment above.

reconciler/events_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func TestNew_Is(t *testing.T) {
+	err := NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+	if !EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a ReconcilerEvent")
```
antoineco: "expected a ReconcilerError" (I think by now it should be clear what I mean)

antoineco:
How about implementing the change in `pkg/controller` already so we can see the bigger picture?

vagababov:
LG in general, but I think you want to tweak the title a bit. It makes no sense to me now :)

reconciler/events.go
n3wscott: I still need a false return inside the type test if, but I can get rid of the else.

reconciler/events_test.go
n3wscott: > ady so we can see the bigger picture?

It is not for controller. This is supporting work for https://github.com/knative/pkg/pull/947

This ends up being used like:

```go
// Interface defines the strongly typed interfaces to be implemented by a
// controller reconciling v1alpha1.ContainerSource.
type Interface interface {
	// ReconcileKind implements custom logic to reconcile v1alpha1.ContainerSource. Any changes
	// to the objects .Status or .Finalizers will be propagated to the stored
	// object. It is recommended that implementors do not call any update calls
	// for the Kind inside of ReconcileKind, it is the responsibility of the core
	// controller to propagate those properties.
	ReconcileKind(ctx context.Context, o *v1alpha1.ContainerSource) reconciler.Event
}

```

n3wscott:
> LG in general, but I think you want to tweak the title a bit. It makes no sense to me now :)

done.

reconciler/events_test.go
n3wscott: Sorry for lazy error strings... Updated.

reconciler/events.go
n3wscott: Reworked...

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const (
+	exampleStatusFailed = "ExampleStatusFailed"
+)
+
+func TestNil_Is(t *testing.T) {
+	var err error
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("expected error to be a ReconcilerEvent, is not")
```
antoineco: The other way around.

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestError_Is(t *testing.T) {
+	err := errors.New("some other error")
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("expected error to be a ReconcilerEvent, is not")
```
antoineco: The other way around.

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestNil_As(t *testing.T) {
+	var err error
+
+	var event *ReconcilerEvent
+	if EventAs(err, &event) {
+		t.Error("expected error to be a ReconcilerEvent, is not")
```
antoineco: The other way around.

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	err := NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+	if !EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("expected error to be a ReconcilerEvent, is not")
+	}
+}
+
+func TestNewOtherType_Is(t *testing.T) {
+	err := NewEvent(corev1.EventTypeNormal, exampleStatusFailed, "this is an example error, %s", "yep")
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a Warn, ExampleStatusFailed")
```
antoineco: I'd be very puzzled if I saw that error without looking at the code. How about "expected errors to not be equivalent"?

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestNewWrappedErrors_Is(t *testing.T) {
+	err := NewEvent(corev1.EventTypeNormal, exampleStatusFailed, "this is a wrapped error, %w", io.ErrUnexpectedEOF)
+	if !EventIs(err, io.ErrUnexpectedEOF) {
+		t.Error("wrapping did not work")
```
antoineco: "expected wrapped and non-wrapped errors to be equivalent"?

reconciler/events_test.go
antoineco: @n3wscott leftover. Should be `did not expect event to be [Warn, ""]` to stick to the pattern you started.

reconciler/events.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// make sure ReconcilerEvent implements error.
+var _ error = (*ReconcilerEvent)(nil)
+
+// Is returns if the target error is a ReconcilerEvent type checking that
+// EventType and Reason match.
+func (e *ReconcilerEvent) Is(target error) bool {
+	if t, ok := target.(*ReconcilerEvent); ok {
+		if t != nil && t.EventType == e.EventType && t.Reason == e.Reason {
```
vagababov: can `t` here actually be nil?

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const (
+	exampleStatusFailed = "ExampleStatusFailed"
+)
+
+func TestNil_Is(t *testing.T) {
+	var err error
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("did not expect error to be a ReconcilerEvent")
```
vagababov:
```suggestion
		t.Error("Did not expect error to be a ReconcilerEvent")
```

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestError_Is(t *testing.T) {
+	err := errors.New("some other error")
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("did not expect error to be a ReconcilerEvent")
```
vagababov:
```suggestion
		t.Error("Did not expect error to be a ReconcilerEvent")
```

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestNew_Is(t *testing.T) {
+	err := NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+	if !EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("expected error to be a ReconcilerEvent, is not")
```
vagababov:
```suggestion
		t.Error("Expected error to be a ReconcilerEvent, is not")
```

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+func TestNewOtherType_Is(t *testing.T) {
+	err := NewEvent(corev1.EventTypeNormal, exampleStatusFailed, "this is an example error, %s", "yep")
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("not a Warn, ExampleStatusFailed")
```
vagababov:
```suggestion
		t.Error("Not a Warn, ExampleStatusFailed")
```
etc, won't comment on the rest :)

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+	if EventIs(err, NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "")) {
+		t.Error("did not expect event to be [Warn, ExampleStatusFailed]")
+	}
+}
+
+func TestNew_As(t *testing.T) {
+	err := NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+
+	var event *ReconcilerEvent
+	if EventAs(err, &event) {
```
vagababov: I'd reflow as
```
if !EventAs() {
  t.Fatal()
}
if event.EventType...
```

reconciler/events_test.go
```diff
@@ -0,0 +1,107 @@
...
+		t.Error("did not expect error to be a ReconcilerEvent")
+	}
+}
+
+func TestNew_Error(t *testing.T) {
+	err := NewEvent(corev1.EventTypeWarning, exampleStatusFailed, "this is an example error, %s", "yep")
+
+	want := "this is an example error, yep"
```
vagababov:
```suggestion
const	want = "this is an example error, yep"
```

reconciler/events_test.go
n3wscott: you don't capital errors.

reconciler/events.go
n3wscott: yes, target could be a nil pointer to a ReconcilerEvent

reconciler/events_test.go
n3wscott: meh.. it is a test

reconciler/events_test.go
n3wscott: for a test? 🙄

vagababov: Test _is_ code :)

reconciler/events_test.go
vagababov: Test _is_ code :)

reconciler/events_test.go
antoineco: But you capitalize logs :)

vagababov: You do capital errors in the unit tests. Those are _not_ errors.

n3wscott:
ok @vagababov fixed it for us.

PR #953, opened 2019-12-20
Document the plan for metrics

evankanderson:
/assign @anniefu @mattmoor @grantr


metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+OpenTelemetry agent running on a localhost port as part of the runtime
+contract.**
+
+We need to make sure that the OpenCensus library does not block, fail, or queue
+metrics in-process excessively in the case where the OpenCensus Agent is not
+present on the cluster. This will allow us to ship Knative components which
+attempt to reach out the Agent if present, and which simply retain local
+statistics for a short period of time if not.
+
+### Concerns
```
mattmoor: Can we open issues for the AIs and link them here?

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+    `fluentd` now).
+- Running as a DaemonSet may make it more difficult to ensure that metrics for
+  each namespace end up in the right place.
+  - We have this problem with the built-in configurations today, so this doesn't
+    make the problem substantially worse.
+  - May want/need some connection between the Agent and the Kubelet to verify
+    sender identities eventually.
+  - Only expose OpenCensus Agent on localhost, not outside the node.
+
+### Steps to reach the goal
```
mattmoor: We discussed putting a deadline on this to prevent an infinite creep, which will influence how the AI issues tile into milestones.  Do we want to incorporate that here at all?

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+**Additionally, once OpenTelemetry agent is stable, we will propose adding the
+OpenTelemetry agent running on a localhost port as part of the runtime
+contract.**
+
+We need to make sure that the OpenCensus library does not block, fail, or queue
+metrics in-process excessively in the case where the OpenCensus Agent is not
+present on the cluster. This will allow us to ship Knative components which
+attempt to reach out the Agent if present, and which simply retain local
+statistics for a short period of time if not.
```
mattmoor: I think that buried in this comment is a subtle piece: Once we move to out-of-tree providers, you will have to either operate a DaemonSet that performs this role, or leverage some sort of Binding to inject the sidecar (and that is how vendors can provide arbitrary configuration to their exporter).

metrics/README.md
```diff
@@ -0,0 +1,90 @@
+# Common metrics export interfaces for Knative
```
mattmoor: We didn't discuss this, but I think that everything we discussed for metrics should apply to tracing as well.  

Are folks ok with that?

cc @mdemirhan @anniefu @evankanderson @grantr

metrics/README.md
mattmoor: This also doesn't surface our level of tolerance for featurework in the short-term, now that we have this plan.

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+We need to make sure that the OpenCensus library does not block, fail, or queue
+metrics in-process excessively in the case where the OpenCensus Agent is not
+present on the cluster. This will allow us to ship Knative components which
+attempt to reach out the Agent if present, and which simply retain local
+statistics for a short period of time if not.
+
+### Concerns
+
+- Unsure about the stability of the OpenCensus Agent (or successor). We're
+  currently investingating this, but the OpenCensus agent seems to have been
```
anniefu: "investigating"

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## The plan
+
+OpenCensus (and eventually OpenTelemetry) offers an sidecar or host-level agent
+with speaks the OpenCensus protocol and can proxy from this protocol to multiple
+backends.
+
+![OpenCensus Agent configuration](https://github.com/census-instrumentation/opencensus-service/raw/master/images/opencensus-service-deployment-models.png)
+
+**We will standardize on export to the OpenCensus agent, and encourage vendors
```
anniefu: I looked into the OpenCensus Agent vs OpenCensus Collector Service more. I believe we do **not** need the OpenCensus Agent as a sidecar as long as we instrument using the OpenCensus library.

I believe the agent's purpose is to allow code that is already implemented using Prometheus or Jaeger to use OpenCensus without modification.

Design doc for collector service: https://github.com/census-instrumentation/opencensus-service/blob/master/DESIGN.md#opencensus-collector

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+to implement their own OpenCensus DaemonSet which connects to their desired
+monitoring environment.**
+
+**Additionally, once OpenTelemetry agent is stable, we will propose adding the
```
anniefu: Would Knative be willing to migrate to OpenTelemetry when they reach Beta, or would we wait for GA? I believe they are currently in Alpha with Beta around the corner.

On the collector service side, the OpenTelemetry collector service will actually support metrics instrumented using the OpenCensus library as well as the OpenTelemetry library. This gives us the nice benefit that we can switch from OpenCensus collector service -> OpenTelemetry collector service without modifying Knative code.

We'll need to migrate the Knative code to the OpenTelemetry library eventually, but we can consider the collector service migration and the library migration to be orthogonal issues.

metrics/README.md
anniefu: Yep! Though in order of implementation, I would say metrics should definitely go first.

metrics/README.md
```diff
@@ -0,0 +1,90 @@
...
+[this doc](https://docs.google.com/document/d/1t-aov3XrhobjCKW4kwScY44QAoahiwxoyXXFtZyL8jw/edit),
+where the plan was worked out.
+
+## The plan
+
+OpenCensus (and eventually OpenTelemetry) offers an sidecar or host-level agent
+with speaks the OpenCensus protocol and can proxy from this protocol to multiple
+backends.
+
+![OpenCensus Agent configuration](https://github.com/census-instrumentation/opencensus-service/raw/master/images/opencensus-service-deployment-models.png)
```
anniefu: [nit] are we allowed to take diagrams from other sources w/o explicit reference to original webpage?

metrics/README.md
anniefu: So instead of using OpenCensus agent, we could do something like:
- Knative instruments using plain OpenCensus library, no exporters
- Knative always provides an OpenCensus Collector Service that supports Prometheus & Stackdriver out-of-the-box
- Some Knative controller converts 'config-observability' metrics fields into Collector service config?

n3wscott:
Please fix:

```bash
--- FAIL: TestNewSDKClient (0.07s)
    client_test.go:46: Expected no error from request '<nil>', but got 'failed to create container service: 'google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.''
    client_test.go:49: Expected a valid client, but got nil
FAIL
FAIL	knative.dev/pkg/test/gke	1.268s
```

metrics/README.md
evankanderson: Sure, done.

metrics/README.md
evankanderson: I put a cutoff in at 0.13, since the next two weeks are probably out for most people.

metrics/README.md
evankanderson: I think the *bold* paragraph on line 53-55 covers this?

metrics/README.md
evankanderson: I think we can replace `config-observability` with a ConfigMap that configures the OpenCensus component.

I'd prefer to be able to use the default `localhost:55678` address rather than needing to configure a separate address for OpenCensus collection.

metrics/README.md
evankanderson: Rather than using an arbitrary cutoff like "Beta" or "GA" for OpenTelemetry for the agent, it seems like vendors could choose which they want to ship. As for the runtime contract, that seems like a nice addition, but it will be a "MAY" or eventually "SHOULD" but not a "MUST" for at least 6-9 months after introduction (and I'd only move to "MUST" if everyone participating agrees that they are already providing that service in a compatible way).

metrics/README.md
evankanderson: I can add a comment, but this is a link to an asset on GitHub, so I think it counts as an explicit reference to the original.

metrics/README.md
evankanderson: Done.

n3wscott:
Fixed the GOOGLE CREDS issue: https://github.com/knative/pkg/pull/959

evankanderson:
Ready for another look

metrics/README.md
```diff
@@ -0,0 +1,99 @@
...
+(From OpenCensus Documentation)
+
+**We will standardize on export to the OpenCensus export protocol, and encourage
+vendors to implement their own OpenCensus Agent or Collector DaemonSet, Sidecar,
+or other `localhost:55678` service which connects to their desired monitoring
```
anniefu: Hmm, I'm not sure I understand why Knative would want to support the OpenCensus Agent. It seems like Knative should force metrics to go through the centralized collector service, and do a bring-your-own-collector-service model. Otherwise, we'll have the issues of short-lived queue-proxy pods dropping metrics if we allow OpenCensus Agent sidecars to export metrics directly.

Wouldn't the ideal state be for Knative controllers to be like the task + Open Census library (embedded) component of this diagram? https://github.com/census-instrumentation/opencensus-service/blob/master/DESIGN.md#architecture-overview-1

metrics/README.md
anniefu: For agents/collector services vendors can choose independently, but for the actual instrumentation libraries Knative will have to choose when to migrate from OpenCensus or OpenTelemetry.

metrics/README.md
```diff
@@ -0,0 +1,99 @@
...
+## The plan
+
+OpenCensus (and eventually OpenTelemetry) offers an sidecar or host-level agent
+with speaks the OpenCensus protocol and can proxy from this protocol to multiple
+backends.
+
+![OpenCensus Agent configuration](https://github.com/census-instrumentation/opencensus-service/raw/master/images/opencensus-service-deployment-models.png)
+(From OpenCensus Documentation)
+
+**We will standardize on export to the OpenCensus export protocol, and encourage
```
anniefu: I think this is more accurately standardizing on the OpenCensus client library? Exporting to be handled by some other OpenCensus component, either agent or collector service (though IMO, we should not support agent).

metrics/README.md
evankanderson: I don't want to specify what the ideal state might be, particularly for the case of things like Cloud Run (fully managed) that might end up running their own implementation of an OpenCensus receiver.

Assuming that you're using Prometheus or Stackdriver, I would imagine implementing this as a node-level service (e.g. a DaemonSet, rather than a sidecar) such that:

1. All the containers can simply talk to `localhost` and don't need to agree on a name. (Also, they can just use the default from the OpenCensus library.)
2. There's only a single network hop for collection -- for Prometheus, this would be a fixed list of all the nodes in the system; for Stackdriver, this would be localhost -> agent -> off-cluster.
3. It would be fine to run the collector instead of the agent if you needed something that the collector did. It seemed like the collector might be a little more complicated, but some operators might need that power.

metrics/README.md
evankanderson: No, the standardization is on the gRPC and/or REST/JSON API exposed by the the OpenCensus Agent and Collector. The API is documented here: https://github.com/census-instrumentation/opencensus-service/blob/master/DESIGN.md#agent-communication and here: https://github.com/census-instrumentation/opencensus-proto/tree/master/src/opencensus/proto/agent

metrics/README.md
evankanderson: Are there different words I should put in here now?

I don't expect that we'll make a decision now about when Knative will switch to OpenTelemetry in the future -- I expect at some point we'll say "this seems stable enough, let's switch" as opposed to something date-driven.

metrics/README.md
anniefu: Nvm, found some code samples, OpenCensus exporter is the way to go.

metrics/README.md
anniefu: No, no need for wording change, just a general question :) 

Just trying to get a feel for when we would start investing in the migration work. I believe starting with OpenTelemetry Collector Service Beta sometime in the next couple months, OpenTelemetry will start recommending OpenTelemetry Collector Service over OpenCensus. That might be the turning point for starting to strip out vendor-specific code.

metrics/README.md
anniefu: Ah I was late to see your response too :)

metrics/README.md
anniefu: Hmm, on further review I remain puzzled by the functional differences between the OpenCensus Agent and the OpenCensus Collector service if the Agent can still receive from multiple sources. I agree with you that Knative does not need to specify Agent vs. Collector for vendors.

What I really meant was should Knative allow sidecar as an option? It would be pretty difficult to vendors to bring their own agents as sidecars if Knative does not provide an easy path for queue-proxy sidecars. It also brings back the short-lived pod issue and concerns about sidecar perf like with the FluentD one.

mattmoor:
> What I really meant was should Knative allow sidecar as an option? 

Yes :)

> It would be pretty difficult to vendors to bring their own agents as sidecars if Knative does not provide an easy path for queue-proxy sidecars

The Binding pattern allows folks to do precisely this out-of-tree.

daemonset.yaml~
```diff
@@ -0,0 +1,176 @@
...
+    component: oc-collector-conf
+data:
+  oc-collector-config: |
+    receivers:
+      opencensus:
+# keepalive settings can help load balancing, see receiver/README.md for more info.
+        keepalive:
+          server-parameters:
+            max-connection-age: 120s
+            max-connection-age-grace: 30s
```
mattmoor: Remove trailing whitespace:
```suggestion
            max-connection-age-grace: 30s
```

k8s.yaml
```diff
@@ -0,0 +1,176 @@
...
+    component: oc-collector-conf
+data:
+  oc-collector-config: |
+    receivers:
+      opencensus:
+# keepalive settings can help load balancing, see receiver/README.md for more info.
+        keepalive:
+          server-parameters:
+            max-connection-age: 120s
+            max-connection-age-grace: 30s
```
mattmoor: Remove trailing whitespace:
```suggestion
            max-connection-age-grace: 30s
```

metrics/README.md
evankanderson: Ugh, so after a bit of experimentation, it appears that standardizing on `localhost` is exactly what we don't want to do, because each netns (i.e. each Pod) has its own `localhost` and there's no good way to inject a port-forwarding rule into each one.

I'll update this section to highlight that the default `localhost:55678` **won't** work for common cases, and call out this is something to work on with OTel.

metrics/README.md
evankanderson: I think once Google has a new export process that handles the Stackdriver things that are going on, we should start ripping out the non-`ocagent` backends.

daemonset.yaml~
evankanderson: Whoops, those shouldn't have been committed at all.

metrics/README.md
```diff
@@ -0,0 +1,103 @@
...
+present on the cluster. This will allow us to ship Knative components which
+attempt to reach out the Agent if present, and which simply retain local
+statistics for a short period of time if not.
+
+### Concerns
+
+- Unsure about the stability of the OpenCensus Agent (or successor). We're
+  currently investigating this, but the OpenCensus agent seems to have been
+  recommended by several others.
+- Running `fluentd` as a sidecar was very big (400MB) and had a large impact on
```
yanweiguo: Why do you mention `fluentd` here? I think you mean `OpenCensus agent`?

evankanderson: I mention `fluentd` here because at one point we ran `fluentd` as a sidecar, rather than as a DaemonSet. We switch to the DaemonSet due to the problems mentioned.

Running as a sidecar has the following benefits:

* Resource usage is properly accounted to the user's namespace
* Credentials are similarly associated with the user's namespace
* It's easier to configure `fluentd` to export for exactly container than to pick up all the containers on the node.

The drawbacks of running as a DaemonSet with our historical approach:

* Upgrades become trickier, as they need to roll out through the Deployment/Revision reconciliation which is "perpendicular" to the typical Knative rollout (i.e. it updates the running Pods of a Revision, rather than spawning a new revision).
* Requiring `fluentd` container to be present caused slow starts when the container hadn't already been loaded on the node.
* Running as a DaemonSet is closer to "logging as a service" model that seems long-term desirable, rather than "everyone figures out logging themselves". It also decouples the need for Knative controllers to know about fluent plugins and how to export logs, which simplifies `config-logging` for application workloads considerably.

mattmoor: What he means is that a topic that came up in our discussion was the potential for injecting fat sidecars materially affecting cold-start latency.  This was an issue with sidecar-based `fluentd`, and we'd like to understand the impact before encouraging the pattern for metrics.

My hot take is: `fluentd` is a beast, but the pattern would likely have been fine for `fluentbit`, so I'm skeptical this will matter if the metrics sidecar is a lean Go container.

yanweiguo: Thanks for explaining.

PR #954, opened 2019-12-20
For GKE clusters used in performance tests, always enable Workload Identity

chizhg:
When we run performance tests internally, the clusters must have Workload Identity enabled. If we enable Workload Identity on existing clusters, it takes ~ 5mins for each node, which is too long.

It does not hurt to enable it if it's not needed, so we always enable Workload Identity for perf test GKE clusters.


test/gke/request.go
```diff
@@ -83,12 +88,15 @@ func NewCreateClusterRequest(request *Request) (*container.CreateClusterRequest,
 	if request.NodeType == "" {
 		return nil, errors.New("node type cannot be empty")
 	}
+	if request.EnableWorkloadIdentity && request.Project == "" {
+		return nil, errors.New("project cannot be empty if you want to Workload Identity")
```
vagababov:
```suggestion
		return nil, errors.New("project cannot be empty if you want Workload Identity")
```

adrcunha:
> If the title is `always enable` then why do we need the variable? Just set `true` at the invocation site?

I think Chi means that Workload Identity is always enabled for performance tests, based on the code. If I'm correct, the description and title needs some tweaking to reflect that though.

testutils/clustermanager/perf-tests/pkg/cluster.go
```diff
@@ -230,11 +230,14 @@ func (gc *gkeClient) createClusterWithRetries(gcpProject, name string, config Cl
 		addons = strings.Split(config.Addons, ",")
 	}
 	req := &gke.Request{
+		Project:     gcpProject,
 		ClusterName: name,
 		MinNodes:    config.NodeCount,
 		MaxNodes:    config.NodeCount,
 		NodeType:    config.NodeType,
 		Addons:      addons,
+		// Always enable Workload Identity.
```
adrcunha: Can you use something like "Enable Workload Identity for perf tests because ...blabala..." so it's clear why we need this for perf clusters?

chizhg:
> > If the title is `always enable` then why do we need the variable? Just set `true` at the invocation site?
> 
> I think Chi means that Workload Identity is always enabled for performance tests, based on the code. If I'm correct, the description and title needs some tweaking to reflect that though.

Yes, thanks for helping me clarify it. Updated the title.

testutils/clustermanager/perf-tests/pkg/cluster.go
chizhg: Added some comments.
But since this is only used in internal tests, I didn't add much detail.

Issue #934, opened 2019-12-10
client_test.go should not require gcloud application default credentials


coryrc:
Cloned knative/pkg.
Ran `go test ./...`
All tests pass except `pkg/test/gke/client_test.go`
Ran `gcloud auth application-default login`
Now all tests pass.

I don't believe unit tests should require 3rd-party software to be configured.
Or if they do, this should be documented in this repo somehow.

coryrc:
cc @adrcunha as he appears to be creator of the file

adrcunha:
This was created by @chaodaiG . Seems like a simple fix; we never had any issues with that because credentials are already setup in a Prow job.

evankanderson:
From #957:

> I'm guessing this uses [`x.oauth2.google.FindDefaultCredentials`](https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials), which says:
> 
> It looks for credentials in the following places, preferring the first location found:
> 
> 1. A JSON file whose path is specified by the
>    `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
> 2. A JSON file in a location known to the gcloud command-line tool.
>    On Windows, this is `%APPDATA%/gcloud/application_default_credentials.json`.
>    On other systems, `$HOME/.config/gcloud/application_default_credentials.json`.
> 3. On Google App Engine standard first generation runtimes (<= Go 1.9) it uses
>    the `appengine.AccessToken` function.
> 4. On Google Compute Engine, Google App Engine standard second generation runtimes
>    (>= Go 1.11), and Google App Engine flexible environment, it fetches
>    credentials from the metadata server.
> 
> We should be able to set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to point to either an empty file, or a temp file with bogus credentials to make sure there are no GCP dependencies in our tests.

chizhg:
/assign
I'll work on the fix.

Issue #958, opened 2019-12-20
Google Application Creds required to run UNIT TESTS in pkg.


n3wscott:
Please fix:

```bash
--- FAIL: TestNewSDKClient (0.07s)
    client_test.go:46: Expected no error from request '<nil>', but got 'failed to create container service: 'google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.''
    client_test.go:49: Expected a valid client, but got nil
FAIL
FAIL	knative.dev/pkg/test/gke	1.268s
```

_Originally posted by @n3wscott in https://github.com/knative/pkg/pull/953#issuecomment-568045781_

evankanderson:
I think this is covered in #934 

/close

(duplicate)

PR #959, opened 2019-12-20
Adding a gke tag to let pkg tests run on a system without google creds installed

n3wscott:
Related to https://github.com/knative/pkg/issues/958

### DIFF
--- a/test/gke/addon.go
+++ b/test/gke/addon.go
@@ -1,3 +1,5 @@
+// +build gke
+
 /*
 Copyright 2019 The Knative Authors
 
--- a/test/gke/client.go
+++ b/test/gke/client.go
@@ -1,3 +1,5 @@
+// +build gke
+
 /*
 Copyright 2019 The Knative Authors
 
--- a/test/gke/client_test.go
+++ b/test/gke/client_test.go
@@ -1,3 +1,5 @@
+// +build gke
+
 /*
 Copyright 2019 The Knative Authors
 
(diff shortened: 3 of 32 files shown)


n3wscott:
fixes https://github.com/knative/pkg/issues/934

adrcunha:
This is the wrong fix. I agree that the unit tests should not be relying on real interactions with GKE. But tagging them is just silently excluding them from presubmit tests (which are currently running fine) and zeroing test coverage in that package (thus risking introducing bugs without authors noticing).

/hold

n3wscott:
> (which are currently running fine)

It is only running fine because it has gcloud and is authed to google.

I am not able to run `go test ./...` in knative/pkg locally. So I can't run the unit tests because the unit tests have external deps.

adrcunha:
> > (which are currently running fine)
> 
> It is only running fine because it has gcloud and is authed to google.

Yes, because gcloud is setup on the Prow jobs.

> I am not able to run `go test ./...` in knative/pkg locally. So I can't run the unit tests because the unit
> tests have external deps.

Right. That's why the right fix is to remove such dependency instead of knocking out the tests.

n3wscott:
@adrcunha anyone working on a fix? All this gke stuff breaks testing on anything else.

adrcunha:
Not that I am aware of.

chizhg:
> @adrcunha anyone working on a fix? All this gke stuff breaks testing on anything else.

I'll take care of it.

chizhg:
Fixed in https://github.com/knative/pkg/pull/963

/close

PR #961, opened 2019-12-23
Spoofing client to retry on all error

zsxking:
Add option to have spoofing client always retry regardless what error encountered.

Prepare for https://github.com/knative/serving/issues/6224

/cc vagababov

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -64,7 +64,7 @@ func (r *Response) String() string {
 // Interface defines the actions that can be performed by the spoofing client.
 type Interface interface {
 	Do(*http.Request) (*Response, error)
-	Poll(*http.Request, ResponseChecker) (*Response, error)
+	Poll(*http.Request, ResponseChecker, ...ErrorRetryChecker) (*Response, error)
 }
 
 // https://medium.com/stupid-gopher-tricks/ensuring-go-interface-satisfaction-at-compile-time-1ed158e8fa17
@@ -79,6 +79,10 @@ var (
 // See the apimachinery wait package:
 // https://github.com/kubernetes/apimachinery/blob/cf7ae2f57dabc02a3d215f15ca61ae1446f3be8f/pkg/util/wait/wait.go#L172
 type ResponseChecker func(resp *Response) (done bool, err error)
+
+// ErrorRetryChecker is used to determine if an error should be retried or not.
+// If an error should be retried, it should return true and the wrapped error to explain why to retry.
+type ErrorRetryChecker func(e error) (retry bool, err error)
 
 // SpoofingClient is a minimal HTTP client wrapper that spoofs the domain of requests
 // for non-resolvable domains.
@@ -199,7 +203,7 @@ func (sc *SpoofingClient) Do(req *http.Request) (*Response, error) {
 }
 
 // Poll executes an http request until it satisfies the inState condition or encounters an error.
-func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Response, error) {
+func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker, errorRetryCheckers ...ErrorRetryChecker) (*Response, error) {
 	var (
 		resp *Response
 		err  error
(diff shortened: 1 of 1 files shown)


test/spoof/spoof.go
```diff
@@ -134,10 +135,11 @@ func New(
...
 
 	sc := SpoofingClient{
-		Client:          &http.Client{Transport: roundTripper},
-		RequestInterval: requestInterval,
-		RequestTimeout:  RequestTimeout,
-		Logf:            logf,
+		Client:           &http.Client{Transport: roundTripper},
+		RequestInterval:  requestInterval,
+		RequestTimeout:   RequestTimeout,
+		RetryOnAllErrors: false,
```
vagababov: Where is it going to be set to `true`?

zsxking: It would be in the serving side. The prober for upgrade test. Somewhere around 
https://github.com/knative/serving/blob/master/test/prober.go#L177

vagababov: This needs to be properly guarded for thread access then I think.
Unless it's set in the constructor and immutable for the life of the client (then it should be set in the `New` and made private).

test/spoof/spoof.go
```diff
@@ -83,15 +83,28 @@ type ResponseChecker func(resp *Response) (done bool, err error)
...
+type SpoofingClientOptions struct {
+	RequestInterval  time.Duration
+	RequestTimeout   time.Duration
+	RetryOnAllErrors bool
 }
 
 // TransportOption allows callers to customize the http.Transport used by a SpoofingClient
 type TransportOption func(transport *http.Transport) *http.Transport
 
+func GetDefaultSpoofingClientOptions() SpoofingClientOptions {
```
vagababov: This dosen't have to be a func, but rather a package private variable.

test/clients.go
```diff
@@ -38,8 +38,15 @@ type KubeClient struct {
...
-func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool, opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
-	return spoof.New(client.Kube, logf, domain, resolvable, Flags.IngressEndpoint, opts...)
+func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool,
+	opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
+	return NewSpoofingClientWithOptions(client, logf, domain, resolvable, nil, opts...)
+}
+
+// NewSpoofingClientWithOptions returns a spoofing client to make requests with spoofing options
+func NewSpoofingClientWithOptions(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool,
+	clientOpts *spoof.SpoofingClientOptions, opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
```
vagababov:
```suggestion
	clientOpts spoof.SpoofingClientOptions, opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
```
no need to pass by pointer. Also saves you hassle later.

zsxking: It's easier to test for empty on pointer though, so that it can set default. But I guess it's OK to have this function to set default. Changing it to pass by value then.

test/spoof/spoof.go
vagababov: SG

test/clients.go
```diff
@@ -38,8 +38,15 @@ type KubeClient struct {
...
 // NewSpoofingClient returns a spoofing client to make requests
-func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool, opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
-	return spoof.New(client.Kube, logf, domain, resolvable, Flags.IngressEndpoint, opts...)
+func NewSpoofingClient(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool,
+	opts ...spoof.TransportOption) (*spoof.SpoofingClient, error) {
+	return NewSpoofingClientWithOptions(client, logf, domain, resolvable, spoof.GetDefaultSpoofingClientOptions(), opts...)
+}
+
+// NewSpoofingClientWithOptions returns a spoofing client to make requests with spoofing options
+func NewSpoofingClientWithOptions(client *KubeClient, logf logging.FormatLogger, domain string, resolvable bool,
```
evankanderson: It's kind of strange to have `spoof.SpoofingClientOptions, ...spoof.TransportOption`.

What about the following:

```golang
func NewSpoofingClient2(*KubeClient, logging.FormatLogger, string, bool, ...spoof.ClientOptions) (*spoof.SpoofingClient, error) {
  ...
`

And then:

```golang
type ClientOptions func(*SpoofingClient) *SpoofingClient

func HTTPOption(opt TransportOption) ClientOption {
  return func(s *SpoofingClient) *SpoofingClient {
    s.Client.Transport = opt(s.Client.Transport)
    return s
  }
}
```

You can then convert calls to `NewSpoofingClient` to `NewSpoofingClient2` at your leisure. Once all calls are converted, you can redefine `NewSpoofingClient` to the same call signature as the "2" version if desired.

test/spoof/spoof.go
```diff
@@ -228,6 +247,11 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker) (*Res
 			}
 			if isConnectionReset(err) {
 				sc.Logf("Retrying %s for connection reset: %v", req.URL, err)
+				return false, nil
+			}
+			if retryOnAllErrors {
```
mattmoor: I think we'd have significantly greater flexibility here if we offered a callback for evaluating whether to retry and the default value was `isConnectionReset` (which we'd want to make public to enable composition).

Not only would this let us plug in an option that allows retrying all errors, but also lets us customize the logging we do on such events, which feels like it matches your goal of "tracking downtime".

PR #965, opened 2019-12-26
Add helper methods in conditon_set

capri-xiyue:
Add helper methods in conditon_set.
We will need such helper methods in https://github.com/knative/eventing/issues/2187

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -153,28 +153,19 @@ func (r ConditionSet) Manage(status ConditionsAccessor) ConditionManager {
 // IsHappy looks at the happy condition and returns true if that condition is
 // set to true.
 func (r conditionsImpl) IsHappy() bool {
-	if c := r.GetCondition(r.happy); c == nil || !c.IsTrue() {
-		return false
-	}
-	return true
+	return r.GetHappyCondition().IsTrue()
 }
 
 // IsUnknown looks at the happy condition and returns true if that condition is
 // set to unknown.
 func (r conditionsImpl) IsUnknown() bool {
-	if c := r.GetCondition(r.happy); !c.IsUnknown() {
-		return false
-	}
-	return true
+	return r.GetHappyCondition().IsUnknown()
 }
 
 // IsFalse looks at the happy condition and returns true if that condition is
 // set to false.
 func (r conditionsImpl) IsFalse() bool {
-	if c := r.GetCondition(r.happy); c == nil || !c.IsFalse() {
-		return false
-	}
-	return true
+	return r.GetHappyCondition().IsFalse()
 }
 
 // GetHappyCondition finds and returns the Condition that matches the ConditionType happy.
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -52,6 +52,10 @@ type ConditionManager interface {
 	// set to true.
 	IsHappy() bool
 
+	// IsUnknown looks at the happy condition and returns true if that condition is
+	// set to unknown or that condition is nil.
+	IsUnknown() bool
+
 	// GetCondition finds and returns the Condition that matches the ConditionType
 	// previously set on Conditions.
 	GetCondition(t ConditionType) *Condition
@@ -143,6 +147,15 @@ func (r ConditionSet) Manage(status ConditionsAccessor) ConditionManager {
 // set to true.
 func (r conditionsImpl) IsHappy() bool {
 	if c := r.GetCondition(r.happy); c == nil || !c.IsTrue() {
+		return false
+	}
+	return true
+}
+
+// IsUnknown looks at the happy condition and returns true if that condition is
+// set to Unknown or that condition is nil.
+func (r conditionsImpl) IsUnknown() bool {
+	if c := r.GetCondition(r.happy); !c.IsUnknown() {
 		return false
 	}
 	return true
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -56,6 +56,10 @@ type ConditionManager interface {
 	// set to unknown or that condition is nil.
 	IsUnknown() bool
 
+	// IsFalse looks at the happy condition and returns true if that condition is
+	// set to false.
+	IsFalse() bool
+
 	// GetCondition finds and returns the Condition that matches the ConditionType
 	// previously set on Conditions.
 	GetCondition(t ConditionType) *Condition
@@ -156,6 +160,15 @@ func (r conditionsImpl) IsHappy() bool {
 // set to Unknown or that condition is nil.
 func (r conditionsImpl) IsUnknown() bool {
 	if c := r.GetCondition(r.happy); !c.IsUnknown() {
+		return false
+	}
+	return true
+}
+
+// IsFalse looks at the happy condition and returns true if that condition is
+// set to False.
+func (r conditionsImpl) IsFalse() bool {
+	if c := r.GetCondition(r.happy); c == nil || !c.IsFalse() {
 		return false
 	}
 	return true
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -52,7 +52,7 @@ type ConditionManager interface {
 	// set to true.
 	IsHappy() bool
 
-	// GetTopLevelCondition finds and returns the top level Condition((happy Condition).
+	// GetTopLevelCondition finds and returns the top level Condition (happy Condition).
 	GetTopLevelCondition() *Condition
 
 	// GetCondition finds and returns the Condition that matches the ConditionType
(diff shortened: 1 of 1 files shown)


mattmoor:
cc @dgerd @n3wscott

apis/condition_set.go
```diff
@@ -148,6 +159,29 @@ func (r conditionsImpl) IsHappy() bool {
 	return true
 }
 
+// IsUnknown looks at the happy condition and returns true if that condition is
+// set to Unknown or that condition is nil.
+func (r conditionsImpl) IsUnknown() bool {
```
vagababov: Implementation does not match the comment. If there's a `nil` it'll panic.
Also you can `return c == nil || c.IsUnknown`
Same below.

capri-xiyue: c.isUnknown() will check whether c is `nil`, if c is `nil`, it will return true. It won't panic.

capri-xiyue: Do you think the comment here is confusing since it touches the implementation details of c.isUnknown()?

vagababov: yes it presumes some things about internal implementation.,

capri-xiyue: I will modify the comment here to 
// IsUnknown looks at the happy condition and returns true if that condition is set to Unknown

vagababov: sure, what about `return r.GetCondition(r.happy).IsUnknown()` as your whole method?

apis/condition_set.go
```diff
@@ -148,6 +159,29 @@ func (r conditionsImpl) IsHappy() bool {
...
+	if c := r.GetCondition(r.happy); !c.IsUnknown() {
+		return false
+	}
+	return true
+}
+
+// IsFalse looks at the happy condition and returns true if that condition is
+// set to false.
+func (r conditionsImpl) IsFalse() bool {
+	if c := r.GetCondition(r.happy); c == nil || !c.IsFalse() {
```
vagababov: Same here.

apis/condition_set.go
```diff
@@ -52,6 +52,17 @@ type ConditionManager interface {
 	// set to true.
 	IsHappy() bool
 
+	// IsUnknown looks at the happy condition and returns true if that condition is
```
n3wscott: There is not a reason for the manager to use these methods.

capri-xiyue: I checked the comment of ConditionManager.
It said "ConditionManager allows a resource to operate on its Conditions using higher order operations".  I think the methods I add falls into this category. I think it makes sense to add `IsUnknown()` and `isFalse()` into `ConditionManager` if we want to differentiate between status `Unknown` and 'False` for resources.

capri-xiyue: In this https://github.com/knative/eventing/issues/2187, we talked about the standard of  True/False/Unknown status. This standard will be used in both serving and eventing. For this standard, we will need to differentiate between status `Unknown` and `False` for resources.

capri-xiyue: @n3wscott  Besides `type conditionsImpl`,  do we have other integrators who implements the ConditionManager?

apis/condition_set.go
```diff
@@ -52,6 +52,17 @@ type ConditionManager interface {
...
+	// set to unknown..
+	IsUnknown() bool
+
+	// IsFalse looks at the happy condition and returns true if that condition is
+	// set to false.
+	IsFalse() bool
```
yolocs: Can we just drop these two functions? Instead, a caller can just call `GetHappeyCondition().IsUnknown()`.

capri-xiyue: I thought the goal of ConditionManager is to make it easier for users to operate on resources' conditions and it prevents users from dealing with the business logic of underlying `Condition`.  
By following the logic you mentioned, should we also deprecate `IsHappy() bool` for ConditionManager? In addition, `GetHappeyCondition()` can also be dropped since we can use `GetCondition(t ConditionType) *Condition` to get the happy condition. I think whether we can add these two functions into ConditionManger should depend on what ConditionManager aims to provide.

n3wscott: The goal of the condition manager is to allow the manager to manage conditions on behalf of the controller.

`.happy` is used internally to understand which condition to propagate updates to. Externally the class already knows which is happy (which is a very silly name and should not be leaked more than it is) because it set from consts in the controlling class.

apis/condition_set.go
n3wscott: > Could you share me with the example that #965 will break the integrations for all other knative resources?

ah ok you are correct, it is just the internal impl that does the impl of this interface.

n3wscott: > For this standard, we will need to differentiate between status Unknown and False for resources.

We can already do this today. I think the helper methods make it a little unclear what is happening in the condition set. The Resource impl creates methods to help deal with the complexity, but I don't think it is the place for the condition manager to do this because what nil means and what true, false means for resources could differ. 

The bar for pkg code should be high and universally useful.

vaikas: Is there a reason why we couldn't introduce a new method called:
GetStatus() corev1.ConditionStatus

and in the Impl it just returns
r.GetCondition(r.happy).Status

Then, I _think_ (haven't thought this through) in our XXX_types.go (which means our Status duck) we should expose GetResourceStatus() or GetStatus() (which is wonky, so just spitballing here about the name) that would be more useful than the current IsReady() we have in our Status fields

Some random comments here:
Just to make sure I'm on the same page, when you say "We can already do this today.", do you mean that we can do this because we know if we've created either a Living or Batch condition set? Because they define which kind of a Condition you have (Ready, or Succeeded). If that's the reasoning (and I might have missed something), then I think having an explicit GetHappyCondition (Though, I'd probably rename that to be something like GetTopLevelCondition, or GetRootCondition, or something like that) would be much more clear. And, I think Living/Batch in the name does not make things much clearer either, they might as well be GetReadyConditionSet and GetSucceededConditionSet because those are the ConditionTypes they actually bake in (which we have to explain in the comments).
But, in any case, the ConditionSetManager is indeed meant only to make propagating things to the "happy" state easier, which it does but there are some assumptions built in that I personally think are little wonky. The second part of where this is used is the actual XXX_types where the resource Status fields expose the IsReady() by calling the IsHappy method in the ConditionSet, which does not give enough information to anything that's relying on propagating the status to the resource (because we need a tri-state, not boolean: True, False and Unknown).
For example, Sequence uses Subscriptions, and today it has to know which is the Happy condition in order to fetch the "happy" condition instead of asking for the top level. And while you can define something like a SubscriptionConditionReady = apis.ConditionReady
and pass that to the NewLivingConditionSet, it's only because the string values match (and we remove dupes) that they are the same thing (and if by mistake you create a NewBatchConditionSet and pass that to it, who knows what happens).
So, I think getting rid of the Happy in the name is a good thing and should probably be called GetTopLevelCondition or something like that for both ConditionManager as well as in our Status using that method to expose the actual ConditionStatus (or entire Top level condition) instead of boolean for various types, because I think you typically will just want to propagate (or check for ready) and having Is*** makes it harder.

capri-xiyue: Yep. I agree that "GetTopLevelCondition" is more clear. Since we have this `"GetTopLevelCondition()` method, should we deprecate `IsHappy() bool` in ConditionManager?

capri-xiyue:
@vaikas  @n3wscott Can someone help review this PR?

apis/condition_set.go
```diff
@@ -52,6 +52,9 @@ type ConditionManager interface {
 	// set to true.
 	IsHappy() bool
 
+	// GetTopLevelCondition finds and returns the top level Condition((happy Condition).
```
vaikas: nit:
Condition((happy Condition) => Condition (happy Condition)

apis/condition_set.go
```diff
@@ -142,10 +145,12 @@ func (r ConditionSet) Manage(status ConditionsAccessor) ConditionManager {
...
 // set to true.
 func (r conditionsImpl) IsHappy() bool {
-	if c := r.GetCondition(r.happy); c == nil || !c.IsTrue() {
-		return false
-	}
-	return true
+	return r.GetTopLevelCondition().IsTrue()
+}
+
+// GetTopLevelCondition finds and returns the top level Condition(happy Condition).
```
vaikas: nit Condition(happy Condition) => Condition (happy Condition).

apis/condition_set.go
```diff
@@ -52,6 +52,9 @@ type ConditionManager interface {
 	// set to true.
```
vaikas: Might consider updating this comment with something about that if you need additional details (for propagating the status in case of dependencies, to look at GetTopLevelCondition), and if you only need to know if something is indeed happy to use this method.

vaikas:
/lgtm
/hold

(I'd like nits fixed since they are comments, but since I'm in a different tz don't want to hold this up unnecessarily).

n3wscott:
/lgtm

Thanks for the update, this does look useful long term.

/hold cancel

PR #971, opened 2020-01-03
bump the default minimum version to 1.15

ZhiminXiang:


mattmoor:
/hold

Going to ping slack first.

mattmoor:
Let's give folks until Monday ~noon PST to speak up about conflicts.  In the meantime, if you want to stage the 1.16 upgrade, that'd be great :)

ZhiminXiang:
Yes. I will work on k8s 1.16 upgrade.

rgregg:
I wish there had been more visibility into this change. pkg isn't a repo I think most of us are tracking, so I didn't catch this PR earlier. I know that 1.15 is the oldest k8s supported version, but looking across the ecosystem most cloud providers are not staying caught up with k8s releases currently. 1.15 is pretty new for both Google and Azure, and still not supported on AWS. That seems like something we should be taking into consideration when making these decisions.

Did I miss a broader communication on this point in a TOC meeting?

mattmoor:
Given the release principles (which we've discussed multiple times at TOC and elsewhere) you should expect one of these every other Knative release.  This specific instance was announced and briefly discussed in slack, and all of the above *was* taken into consideration.  I also expect it to be called out in the release notes.

PR #973, opened 2020-01-04
Filter tracked sharedmain ConfigMaps based on optional label selector

antoineco:
### Proposed change

Ensure the ConfigMap watcher only tracks the ConfigMaps which contain the label defined by the optional `SYSTEM_RESOURCE_LABEL` env var if the latter is set and not empty.

### Rationale

Custom components (event sources, channels, ...) may be deployed to various namespaces which sometimes contain a large amount of ConfigMaps. For example, a typical [Kyma](https://kyma-project.io/) installation deploys >80 ConfigMaps inside the `kyma-system` namespace. This creates a surge of memory usage when watches expire and get re-instantiated.

image
_jumps from 9MiB to **25MiB**_

image
_stable-ish usage between 9MiB and **11MiB** with a label selector applied_

### DIFF
--- a/configmap/informed_watcher.go
+++ b/configmap/informed_watcher.go
@@ -18,12 +18,17 @@ package configmap
 
 import (
 	"errors"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	informers "k8s.io/client-go/informers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/client-go/informers"
 	corev1informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/informers/internalinterfaces"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 )
@@ -35,7 +40,7 @@ func NewDefaultWatcher(kc kubernetes.Interface, namespace string) *InformedWatch
 	return NewInformedWatcher(kc, namespace)
 }
 
-// NewInformedWatcherFromFactory watches a Kubernetes namespace for configmap changes.
+// NewInformedWatcherFromFactory watches a Kubernetes namespace for ConfigMap changes.
 func NewInformedWatcherFromFactory(sif informers.SharedInformerFactory, namespace string) *InformedWatcher {
 	return &InformedWatcher{
 		sif:      sif,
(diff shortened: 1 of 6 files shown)


system/env.go
```diff
@@ -22,11 +22,12 @@ import (
 )
 
 const (
-	NamespaceEnvKey = "SYSTEM_NAMESPACE"
+	NamespaceEnvKey   = "SYSTEM_NAMESPACE"
+	ConfigLabelEnvKey = "SYSTEM_CONFIG_LABEL"
```
mattmoor: The label that comes to mind here (which we used in the config validation webhook) is `foo.knative.dev/release`.  I think that we should probably call this the `SYSTEM_RESOURCE_LABEL` (even though both current uses are configmaps, we could have more in the future), and use it both places.

configmap/informed_watcher.go
```diff
@@ -47,17 +52,46 @@ func NewInformedWatcherFromFactory(sif informers.SharedInformerFactory, namespac
...
+		lo.LabelSelector = sel.Add(lr...).String()
+	}
+}
+
+// FilterConfigByLabelExists returns an "exists" label requirement for the
+// given label key.
+func FilterConfigByLabelExists(labelKey string) *labels.Requirement {
+	req, err := labels.NewRequirement(labelKey, selection.Exists, nil)
+	if err != nil {
+		panic(fmt.Sprintf("could not construct label requirement: %w", err))
```
mattmoor: Return the error and log.Fatal in sharedmain?

configmap/informed_watcher.go
```diff
@@ -47,17 +52,46 @@ func NewInformedWatcherFromFactory(sif informers.SharedInformerFactory, namespac
...
 		0,
 		informers.WithNamespace(namespace),
+		informers.WithTweakListOptions(addLabelRequirementsToListOptions(lr)),
 	), namespace)
 }
 
+// addLabelRequirementsToListOptions returns a function which injects label
+// requirements to existing metav1.ListOptions.
+func addLabelRequirementsToListOptions(lr []labels.Requirement) internalinterfaces.TweakListOptionsFunc {
+	if lr == nil {
```
mattmoor:
```suggestion
	if len(lr) == 0 {
```

system/env.go
antoineco: That's exactly the one I was thinking about. I can rename it, that makes sense.

configmap/informed_watcher.go
antoineco: Right, that's probably a better approach. on it.

injection/sharedmain/main.go
```diff
@@ -140,13 +143,23 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
...
+	if cmLabel := system.ResourceLabel(); cmLabel != "" {
+		req, err := configmap.FilterConfigByLabelExists(cmLabel)
+		if err != nil {
+			logger.With(zap.Error(err)).Fatalf("Failed to generate requirement for label %q")
+		}
+		cmLabelReqs = append(cmLabelReqs, *req)
+	}
 	// TODO(mattmoor): This should itself take a context and be injection-based.
-	cmw := configmap.NewInformedWatcher(kc, system.Namespace())
+	cmw := configmap.NewInformedWatcher(kc, system.Namespace(), cmLabelReqs...)
```
antoineco: `cmLabelReqs...` will simply be `nil` if the `SYSTEM_RESOURCE_LABEL` env var is not set, and no filter will be applied (current behavior).

configmap/informed_watcher.go
```diff
@@ -125,7 +158,7 @@ func (i *InformedWatcher) registerCallbackAndStartInformer(stopCh <-chan struct{
 	i.m.Lock()
 	defer i.m.Unlock()
 	if i.started {
-		return errors.New("watcher already started")
+		return fmt.Errorf("watcher already started")
```
n3wscott: this should be back to `errors.New`

configmap/informed_watcher.go
```diff
@@ -115,7 +148,7 @@ func (i *InformedWatcher) Start(stopCh <-chan struct{}) error {
 
 	// Wait until it has been synced (WITHOUT holing the mutex, so callbacks happen)
 	if ok := cache.WaitForCacheSync(stopCh, i.informer.Informer().HasSynced); !ok {
-		return errors.New("error waiting for ConfigMap informer to sync")
+		return fmt.Errorf("error waiting for ConfigMap informer to sync")
```
n3wscott: `errors.New`

configmap/informed_watcher_test.go
```diff
@@ -165,20 +168,108 @@ func TestInformedWatcher(t *testing.T) {
 	}
 }
 
+func TestFilterConfigByLabelExists(t *testing.T) {
+	testCases := map[string]struct {
+		input        string
+		expectOutStr string
+		expectErr    bool
+	}{
+		"Valid input": {
```
n3wscott: please also add a test with `input:""` ?

n3wscott:
neat! 

/lgtm

configmap/informed_watcher.go
antoineco: `fmt.Errorf` was created to avoid a circular import issue with `errors` in early Go versions. It's been fixed since then, but what's the point importing both ?

configmap/informed_watcher_test.go
antoineco: Good point

configmap/informed_watcher.go
vagababov: you're using a formatter function while having no format, which at least will scan the whole string once. No reason to do work that's totally unnecessary.

antoineco: OK, I didn't see the argument about performance. Thanks for clarifying.

configmap/informed_watcher_test.go
antoineco: [done](https://github.com/knative/pkg/compare/c975f8ea85cb8e268338e217d34c5c251496ffdf..0994948eed981540438e919532b1e697763abaac)

configmap/informed_watcher.go
antoineco: [done](https://github.com/knative/pkg/compare/c975f8ea85cb8e268338e217d34c5c251496ffdf..0994948eed981540438e919532b1e697763abaac)

configmap/informed_watcher.go
antoineco: [done](https://github.com/knative/pkg/compare/c975f8ea85cb8e268338e217d34c5c251496ffdf..0994948eed981540438e919532b1e697763abaac)

PR #978, opened 2020-01-07
Add support for OpenCensus Agent as a destination for Knative metrics.

evankanderson:
/assign @mattmoor @anniefu 

I tested this with exporting to stackdriver using the following config:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
  name: opencensus-agent-config
  namespace: knative-monitoring
data:
  opencensus-agent-config: |
    receivers:
      opencensus:
        address: ":55678"
    #   jaeger: {}
    #   zipkin: {}
    exporters:
      stackdriver:
        project: "kubecon-knative-2019"  # Optional, defaults to node project
        metric_prefix: "testing"
        enable_metrics: true
        enable_tracing: true
      prometheus:
        namespace: "namespace"
        address: ":8888"
---
apiVersion: v1
kind: Secret
metadata:
  name: stackdriver-key
  namespace: knative-monitoring
type: Opaque
data:
  credentials.json: <<<BASE64 JSON>>>
---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  labels:
  name: opencensus-agent
  namespace: knative-monitoring
spec:
  selector:
    matchLabels:
      component: opencensus-agent
  template:
    metadata:
      labels:
        component: opencensus-agent
    spec:
      containers:
      - command:
        - /ocagent_linux
        - --config=/conf/opencensus-agent-config.yaml
        env:
        - name: GOOGLE_APPLICATION_CREDENTIALS
          value: /secrets/credentials.json
        image: omnition/opencensus-agent:0.1.6
        name: opencensus-agent
        ports:
        - containerPort: 55678
          name: collect
          protocol: TCP
        - containerPort: 55679
          name: zpages
          protocol: TCP
        - containerPort: 8888
          name: prometheus
          protocol: TCP
        resources:
          limits:
            cpu: 500m
            memory: 500Mi
          requests:
            cpu: 100m
            memory: 100Mi
        volumeMounts:
        - mountPath: /conf
          name: opencensus-agent-config-vol
        - mountPath: /secrets
          name: stackdriver-key
      volumes:
      - configMap:
          items:
          - key: opencensus-agent-config
            path: opencensus-agent-config.yaml
          name: opencensus-agent-config
        name: opencensus-agent-config-vol
      - name: stackdriver-key
        secret:
          defaultMode: 420
          secretName: stackdriver-key
---
apiVersion: v1
kind: Service
metadata:
  labels:
  name: opencensus
  namespace: knative-monitoring
spec:
  ports:
  - name: opencensus
    port: 55678
    protocol: TCP
    targetPort: 55678
  selector:
    component: opencensus-agent
  type: ClusterIP
```

I had a little trouble tracking down how to find the exported metrics in Stackdriver, but I was able to trace the traffic off my machine... 😜


metrics/opencensus_exporter.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2019 The Knative Authors
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2020 The Knative Authors
```

metrics/config.go
```diff
@@ -80,6 +86,12 @@ type metricsConfig struct {
 	// writing the metrics to the stats.RecordWithOptions interface.
 	recorder func(context.Context, stats.Measurement, ...stats.Options) error
 
+	// ---- OpenCensus specific below ----
+	// collectorAddress is the address of the collector, if not `localhost:55678`
+	collectorAddress string
+	// Require mutual TLS. Defaults to "false" because
```
anniefu: incomplete sentence

metrics/opencensus_exporter.go
```diff
@@ -0,0 +1,38 @@
...
+	"go.opencensus.io/stats/view"
+	"go.uber.org/zap"
+)
+
+func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
+	opts := []ocagent.ExporterOption{ocagent.WithServiceName(config.component)}
+	if config.collectorAddress != "" {
+		opts = append(opts, ocagent.WithAddress(config.collectorAddress))
+	}
+	if !config.requireSecure {
```
anniefu: What happens when secure is required?

metrics/config.go
```diff
@@ -57,9 +57,15 @@ const (
 	Stackdriver metricsBackend = "stackdriver"
 	// Prometheus is used for Prometheus backend
 	Prometheus metricsBackend = "prometheus"
+	// OpenCensus is used to export to the OpenCensus Agent / Collector,
+	// which can send to many other services.
+	OpenCensus metricsBackend = "opencensus"
 
 	defaultBackendEnvName = "DEFAULT_METRICS_BACKEND"
 
+	collectorAddressKey = "metrics.collector"
```
anniefu: [nit]
"metrics.collector-address"
"metrics.collector-require-tls"

anniefu: I have found that exporting config-observability field names is useful once crossing repo boundaries. In particular for writing tests in knative/serving & passing queue-proxy EnvVars

metrics/config_test.go
```diff
@@ -93,13 +93,24 @@ var (
...
-				"metrics.backend-destination":      "prometheus",
+				"metrics.backend-destination":      "opencensus",
 				"metrics.reporting-period-seconds": "test",
 			},
 			Domain:    servingDomain,
 			Component: testComponent,
 		},
 		expectedErr: "invalid metrics.reporting-period-seconds value \"test\"",
+	}, {
+		name: "invalidOpenCensusSecuritySetting",
```
anniefu: validOpenCensusSecuritySetting test?

metrics/exporter.go
```diff
@@ -121,6 +121,12 @@ func isNewExporterRequired(newConfig *metricsConfig) bool {
 		return true
 	}
 
+	// If the OpenCensus address has changed, restart the collector.
+	// TODO(evankanderson): Should we just always restart the opencensus agent?
+	if newConfig.backendDestination == OpenCensus && newConfig.collectorAddress != cc.collectorAddress {
```
anniefu: && newConfig.requiresSecure != cc.requiresSecure?

metrics/exporter.go
```diff
@@ -121,6 +121,12 @@ func isNewExporterRequired(newConfig *metricsConfig) bool {
 		return true
 	}
 
+	// If the OpenCensus address has changed, restart the collector.
```
anniefu: restart the exporter? In bring-your-own-collector-service world, I don't think Knative should have the ability to or be responsible for restarting OpenCensus Service agents.

metrics/config.go
```diff
@@ -164,18 +176,28 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
...
-	case Stackdriver, Prometheus:
+	case Stackdriver, Prometheus, OpenCensus:
 		mc.backendDestination = lb
 	default:
 		return nil, fmt.Errorf("unsupported metrics backend value %q", backend)
 	}
 
+	if mc.backendDestination == OpenCensus {
+		mc.collectorAddress = ops.ConfigMap[collectorAddressKey]
+		if isSecure := ops.ConfigMap[collectorSecureKey]; isSecure != "" {
```
anniefu: Does strconv.ParseBool consider "" to be valid? If not, maybe just:
if isSecure, ok := ops.ConfigMap[collectorSecureKey]; ok {
// let empty string return invalid value error
}

evankanderson: It does not. I can add a test for that if you want.

https://golang.org/pkg/strconv/#ParseBool
> It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False. Any other value returns an error.

metrics/exporter.go
evankanderson: Yes, this was supposed to be exporter.

metrics/exporter.go
evankanderson: Broke this up, since it should be an `||` not an `&&`.

metrics/config_test.go
evankanderson: Good point, done.

(BTW, I find it strange that this table test doesn't simply have a *err where empty means you expect success. But that's for a different PR.)

metrics/config.go
evankanderson: Done.

metrics/config.go
evankanderson: Done.

metrics/opencensus_exporter.go
evankanderson: Secure is the default. Unfortunately, it's hard to get TLS set up for many users, so we default here to assuming that the cluster network is "secure enough".

metrics/config.go
evankanderson: Thanks!

metrics/config.go
anniefu: Oh, I don't think a test case is necessary to validate ParseBool behavior. I was just thinking to also error as invalid if someone uses empty string i.e. "metrics.collector-require-tls="
		if isSecure, ok := ops.ConfigMap[CollectorSecureKey]; ok {
			var err error
			if mc.requireSecure, err = strconv.ParseBool(isSecure); err != nil {
				return nil, fmt.Errorf("invalid %s value %q", CollectorSecureKey, isSecure)
			}
		}

metrics/opencensus_exporter.go
evankanderson: I'm assuming you mean the [`WithTLSCredentials`](https://godoc.org/contrib.go.opencensus.io/exporter/ocagent#WithTLSCredentials) option?

I can take a stab at a way to do this, but it's starting to go down a rabbit hole here:

> WithTLSCredentials allows the connection to use TLS credentials when talking to the server. It takes in grpc.TransportCredentials instead of say a Certificate file or a tls.Certificate, because the retrieving these credentials can be done in many ways e.g. plain file, in code tls.Config or by certificate rotation, so it is up to the caller to decide what to use.

I think a secret is probably the best choice, let me see if I can lift some from [`webbook`](https://github.com/knative/pkg/blob/master/webhook/webhook.go#L141).

metrics/config.go
```diff
@@ -57,9 +57,15 @@ const (
 	Stackdriver metricsBackend = "stackdriver"
 	// Prometheus is used for Prometheus backend
 	Prometheus metricsBackend = "prometheus"
+	// OpenCensus is used to export to the OpenCensus Agent / Collector,
+	// which can send to many other services.
+	OpenCensus metricsBackend = "opencensus"
 
 	defaultBackendEnvName = "DEFAULT_METRICS_BACKEND"
 
+	CollectorAddressKey = "metrics.collector-address"
```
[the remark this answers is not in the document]
evankanderson: OpenCensus can use the same port for both metrics and traces; I can call this `opencensus-address`, but I'd like to use the same address for both metrics and trace destination.

metrics/opencensus_exporter.go
```diff
@@ -0,0 +1,38 @@
...
+		opts = append(opts, ocagent.WithInsecure())
+	}
+	e, err := ocagent.NewExporter(opts...)
+	if err != nil {
+		logger.Errorw("Failed to create the OpenCensus exporter.", zap.Error(err))
+		return nil, err
+	}
+	logger.Infof("Created OpenCensus exporter with config: %+v.", *config)
+	// Start the server for Prometheus scraping
```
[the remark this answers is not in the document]
evankanderson: D'oh -- this got lost between my testing in serving and this file.

The comment above is also wrong. Thanks!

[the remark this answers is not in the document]
evankanderson: Yeah, but it was a good hint I'd missed something.

evankanderson:
I'm working on adding support for a TLS secret for metrics export. Any thoughts on the format for that?

metrics/config.go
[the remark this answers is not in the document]
anniefu: I'm good with "opencensus-address" :)

When we get around to using OpenCensus Service for tracing too, we should definitely look deprecating 'config-tracing'

evankanderson:
Adding TLS from a secret requires some larger surgery (the ability to supply a `SecretLister` to `pkg/metrics`.

I'd like to do that in a subsequent PR to keep this from snowballing too much.

I think we *do* need to support mutual TLS as an option, but I'll also point out that it is a feature creep over our current state, where either:

1. Everyone on the node / running in a namespace can access the current push secret (i.e. via GCE metadata server for GKE).
2. The data is sent unencrypted without any authentication anyway (via Prometheus).

Mind if I do a fast-follow with secrets for OpenCensus?

I don't expect any of this to be the default in the next release, but would like to give users the option to try this out if desired.

anniefu:
I do happen to have a PR out for adding a SecretLister/Watcher to pkg/metrics that could be repurposed :) https://github.com/knative/pkg/pull/925

Merging this PR and doing fast-follows sounds good to me.

I agree TLS is beyond what Knative currently has, but I imagine it will come in handy quickly as Knative adds more observability backends.

PR #980, opened 2020-01-08
Optimize locking behavior of the profiling handler.

markusthoemmes:
The profiling handler is in the path of our most performance critical components (especially the activator). Taking a write-lock on each request is probably a bad idea.

Replaced the mutex with an atomic flag. Lost reads are not critical in this code path and that should be the quickest solution in terms of avoiding contention.

/assign @vagababov


markusthoemmes:
/assign @mattmoor 

This seems to need superpowers.

PR #982, opened 2020-01-09
perf-tests cluster manager supports creating clusters in other environment

chizhg:
perf-tests cluster manager tool supports creating clusters in other environment, including `test`, `staging`, `prod` or with any other custom endpoint.

Note:
1. Part of the code is borrowed from `kubetest` in kubernetes test-infra - https://github.com/kubernetes/test-infra/blob/9d3c810352d830d3a29b8625df7f5dab25924eb7/kubetest/gke.go#L184, thanks @yt3liu for digging it out 
2. I believe the e2e-tests cluster manager tool will also need this, but I did not make the change in this PR as I do want to keep it smaller.

/cc @adrcunha 
/cc @yt3liu 
/cc @chaodaiG

### DIFF
--- a/test/gke/endpoint.go
+++ b/test/gke/endpoint.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gke
+
+import (
+    "regexp"
+)
+
+const (
+    testEnv = "test"
+    stagingEnv = "staging"
+    staging2Env = "staging2"
+    prodEnv = "prod"
+
+    testEndpoint = "https://test-container.sandbox.googleapis.com/"
+    stagingEndpoint = "https://staging-container.sandbox.googleapis.com/"
+    staging2Endpoint = "https://staging2-container.sandbox.googleapis.com/"
+    prodEndpoint = "https://container.googleapis.com/"
+)
+
+var urlRe = regexp.MustCompile(`https://.*/`)
+
+// ServiceEndpoint returns the container service endpoint for the given environment.
+func ServiceEndpoint(environment string) string {
+    var endpoint string
+    if environment != "" {
+        switch env := environment; {
+        case env == testEnv:
+            endpoint = testEndpoint
+        case env == stagingEnv:
+            endpoint = stagingEndpoint
+        case env == staging2Env:
+            endpoint = staging2Endpoint
+        case env == prodEnv:
+            endpoint = prodEndpoint
+        case urlRe.MatchString(env):
+            endpoint = env
+        }
+    }
+    return endpoint
+}
--- a/test/gke/endpoint.go
+++ b/test/gke/endpoint.go
@@ -17,6 +17,7 @@ limitations under the License.
 package gke
 
 import (
+    "fmt"
     "regexp"
 )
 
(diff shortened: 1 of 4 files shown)


test/gke/endpoint.go
```diff
@@ -0,0 +1,51 @@
...
+    "regexp"
+)
+
+const (
+    testEnv = "test"
+    stagingEnv = "staging"
+    prodEnv = "prod"
+
+    testEndpoint = "https://test-container.sandbox.googleapis.com/"
+    stagingEndpoint = "https://staging-container.sandbox.googleapis.com/"
```
yt3liu: Please add staging2 to list of endpoints.

testutils/clustermanager/perf-tests/main.go
```diff
@@ -44,7 +46,7 @@ func main() {
 		log.Fatal("Only one operation can be specified, either recreate or reconcile")
 	}
 
-	client, err := testPkg.NewClient()
+	client, err := testPkg.NewClient(gkeEnvironment)
```
adrcunha: You're defaulting --gke-environment to an unsupported environment (empty string). NewClient() should either fail or the default value should be a valid one or the flag should be mandatory.

testutils/clustermanager/perf-tests/pkg/cluster.go
```diff
@@ -44,8 +46,9 @@ type gkeClient struct {
 }
 
 // NewClient will create a new gkeClient.
-func NewClient() (*gkeClient, error) {
-	operations, err := gke.NewSDKClient()
+func NewClient(environment string) (*gkeClient, error) {
+	endpointOption := option.WithEndpoint(gke.ServiceEndpoint(environment))
```
adrcunha: You should validate the return of ServiceEndpoint and return a meaningful error.

testutils/clustermanager/perf-tests/main.go
chizhg: Done.
Default the flag to be "prod".

testutils/clustermanager/perf-tests/pkg/cluster.go
chizhg: I changed the logic a bit:
If the provided gke environment is empty or invalid, default the endpoint to be prod. So the return of ServiceEndpoint will be always valid, unless caller provides some custom endpoint like `https://invalid.endpoint`.

Does it make sense to you?

test/gke/endpoint.go
```diff
@@ -0,0 +1,56 @@
...
+        switch env := environment; {
+        case env == testEnv:
+            endpoint = testEndpoint
+        case env == stagingEnv:
+            endpoint = stagingEndpoint
+        case env == staging2Env:
+            endpoint = staging2Endpoint
+        case env == prodEnv:
+            endpoint = prodEndpoint
+        case urlRe.MatchString(env):
```
yt3liu: I'd believe you can set defaults by using the default case. 
https://yourbasic.org/golang/switch-statement/

testutils/clustermanager/perf-tests/pkg/cluster.go
adrcunha: I don't think an invalid endpoint (which includes the empty string) should be silently translated to prod, it's dangerous and counter-intuitive. Anyone calling NewClient() should be sure about the environment. If you want to keep backwards compatibility, rename this to NewClientForEnvironment() and make NewClient() call this method passing prod. But since this method is not widely used yet, I don't think you need to do such thing.

test/gke/endpoint.go
chizhg: Done.

testutils/clustermanager/perf-tests/pkg/cluster.go
chizhg: Make sense.
Added error handling.

test/gke/endpoint.go
```diff
@@ -0,0 +1,59 @@
...
+    prodEndpoint = "https://container.googleapis.com/"
+)
+
+var urlRe = regexp.MustCompile(`https://.*/`)
+
+// ServiceEndpoint returns the container service endpoint for the given environment.
+func ServiceEndpoint(environment string) (string, error) {
+    // Default to be prod.
+    var endpoint = prodEndpoint
+    if environment != "" {
```
adrcunha: You're still treating empty string as `prod` and I think it's dangerous, as discussed previously.

chizhg: Done.

PR #987, opened 2020-01-10
Provide an semantic equality func for apis.URL

dprotaso:
url.URL has an unexported type (UserInfo) which causes semantic
equality to panic unless we add a custom equality function


dprotaso:
let me move the func to the bottom of the file

PR #989, opened 2020-01-10
Refactor webhook to allow adding conversion support

dprotaso:
Testing changes with serving here: https://github.com/knative/serving/pull/6496

PR #993, opened 2020-01-14
Conversion Webhook Framework

dprotaso:
This builds on https://github.com/knative/pkg/pull/989


webhook/conversion_integration_test.go
```diff
@@ -0,0 +1,228 @@
+/*
+Copyright 2018 The Knative Authors
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2020 The Knative Authors
```

webhook/resourcesemantics/conversion/conversion.go
```diff
@@ -0,0 +1,200 @@
...
+	)
+
+	// TODO(dprotaso) - potentially error on unknown fields
+	if err = json.Unmarshal(inRaw.Raw, &in); err != nil {
+		err = fmt.Errorf("unable to unmarshal input: %s", err)
+		return
+	}
+
+	if inGVK.Version == conv.HubVersion {
+		hub = in
```
bbrowning: @pmorie tells me a useful pattern with conversion webhooks is to convert from one version to the same version in the case of newly added required fields and such. Should we allow that here - converting between two of the same versions? For things like when Serving did the v1alpha1 API changes without changing the version?

dprotaso: I'd consider the lemonade migration an exception at the time due to k8s limitations. Also the conversion webhook isn't invoked if the version being served is the same as the storage version.

dprotaso:
Rebased now that https://github.com/knative/pkg/pull/989 is merged

webhook/resourcesemantics/conversion/conversion.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	if outGVK.Version == conv.HubVersion {
+		out = hub
+	} else if err = hub.ConvertUp(ctx, out); err != nil {
+		err = fmt.Errorf("conversion failed to version %s for type %s -  %s", outGVK.Version, formatGVK(inGVK), err)
+		return
+	}
+
+	out.SetGroupVersionKind(outGVK)
```
dprotaso: We should be applying defaults since new fields may have been added which would need to be defaulted.

dprotaso:
/assign @mattmoor 

dropping hold once I rebase/update https://github.com/knative/serving/pull/6566

webhook/conversion.go
```diff
@@ -0,0 +1,71 @@
...
+		response := apixv1beta1.ConversionReview{
+			Response: c.Convert(ctx, review.Request),
+		}
+
+		if err := json.NewEncoder(w).Encode(response); err != nil {
+			http.Error(w, fmt.Sprintf("could encode response: %v", err), http.StatusInternalServerError)
+			return
+		}
+
+		// TODO(dprotaso) - figure out what metrics we want reported
```
dprotaso: Will make a separate issue for this

dprotaso:
Here's the integration into serving's webhook
https://github.com/knative/serving/pull/6566/files#diff-58de452513b7e8d8d3cfea23eb4ae6a8

dprotaso:
rebased on master

webhook/resourcesemantics/conversion/controller.go
```diff
@@ -0,0 +1,130 @@
...
+			Handler:    controller.HandleAll(c.Enqueue),
+		})
+
+		sentinel := c.EnqueueSentinel(types.NamespacedName{Name: name})
+
+		// Reconcile when the cert bundle changes.
+		secretInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+			FilterFunc: controller.FilterWithNameAndNamespace(system.Namespace(), options.SecretName),
+			Handler:    controller.HandleAll(sentinel),
+		})
```
mattmoor: This doesn't need to be in the loop

dprotaso: It does.

When the secret changes we need to reconcile all the CRDs we're tracking in order to update their webhook config. So the sentinel value actually changes on line 128

mattmoor: I missed that the Sentinel is changing, which is a bit weird.  Generally I'd expect `EnqueueSentinel` to be used with singleton controllers that aren't paying attention to the key at all (we probably shouldn't have even exposed a parameter, which in hindsight was probably a mistake).

Should this be using EnqueueKey instead of EnqueueSentinel?

dprotaso: > Should this be using EnqueueKey instead of EnqueueSentinel?

I think `EnqueueSentinel` still fits since we're reconciling a fixed set of objects (CRDs).

mattmoor: The spirit of `EnqueueSentinel` was to trigger a reconciliation where the key that's enqueued doesn't matter.  `EnqueueKey` should be used when what's enqueued matters, whether it's a discrete or unbounded set.

dprotaso: I'm confused - in the usage actioning on secret updates and enqueuing the fixed list of CRD keys to be reconciled. Passing `EnqueueKey` would enqueue the secret's key.

dprotaso: Unless you're suggesting write a custom informer handler where I invoke `EnqueueKey` N-times for the CRD resources we're watching.

dprotaso:
rebased

dprotaso:
/hold cancel

The failing serving tests I kept observing were marked as flakey in the continuous test-grid runs.

webhook/resourcesemantics/conversion/controller.go
mattmoor: I see now.  I guess the closest analog would be a global resync today, but because of the fixed keyspace you are able to do it with a fixed N informer watches.

I think I still have a mild preference for `EnqueueKey` with a fixed key over `EnqueueSentinel`, but this does end up looking a bit grosser due to the need for a thunk (if I'm not mistaken, I certainly need more ☕️ )

PR #998, opened 2020-01-21
Add support for client TLS and Resources to pkg/metrics

evankanderson:
Next steps here are to change all the clients to natively export resources and call the 3-argument form of `UpdateExporterFromConfigMap`, then update `UpdateExporterFromConfigMap` to take 3 arguments and remove the to/from Resource / tags code in `stackdriver_exporter.go`.

/assign @anniefu @yanweiguo

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -27,8 +27,12 @@ import (
 	"strings"
 	"time"
 
+	"go.opencensus.io/tag"
+
+	"go.opencensus.io/resource"
 	"go.opencensus.io/stats"
 	"go.uber.org/zap"
+	clientv1 "k8s.io/client-go/listers/core/v1"
 	"knative.dev/pkg/metrics/metricskey"
 )
 
@@ -86,11 +90,19 @@ type metricsConfig struct {
 	// writing the metrics to the stats.RecordWithOptions interface.
 	recorder func(context.Context, stats.Measurement, ...stats.Options) error
 
+	// secretsLister provides access for fetching Kubernetes Secrets from an
+	// informer cache.
+	secretsLister clientv1.SecretLister
+
 	// ---- OpenCensus specific below ----
 	// collectorAddress is the address of the collector, if not `localhost:55678`
 	collectorAddress string
 	// Require mutual TLS. Defaults to "false" because mutual TLS is hard to set up.
 	requireSecure bool
+	// resourceDescriptor is provides a function to extract an opencensus Resource
+	// (Resource describes an entity about which identifying information and
+	// metadata is exposed)
+	resourceDescriptor resource.Detector
 
 	// ---- Prometheus specific below ----
 	// prometheusPort is the port where metrics are exposed in Prometheus
@@ -147,14 +159,42 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 // Record applies the `ros` Options to `ms` and then records the resulting
 // measurements in the metricsConfig's designated backend.
 func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
+	var err error
 	if mc == nil || mc.recorder == nil {
+		ros, err = resourceToTags(ctx, ros...)
+		if err != nil {
+			return err
+		}
 		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(ms))...)
 	}
 	return mc.recorder(ctx, ms, ros...)
 }
 
+// resourceToTags is a backwards-compatibility shim to forward all the labels
+// on the Resource context to tags (used for Prometheus and Stackdriver during
+// the transition to using Resources).
+func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options, error) {
+	if r, ok := metricskey.FromContext(ctx); ok && r != nil {
+		tags := []tag.Mutator{}
+		for label, v := range r.Labels {
+			key, err := tag.NewKey(label)
+			if err != nil {
+				return nil, fmt.Errorf("Bad resource label %q can't be converted to a tag: %+v", label, err)
+			}
+			tags = append(tags, tag.Upsert(key, v))
+			fmt.Printf("Adding tag %q: %q\n", key, v)
+		}
+		return append(ros, stats.WithTags(tags...)), nil
+	}
+	return ros, nil
+}
+
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
 	var mc metricsConfig
+
+	// We don't check if this is `nil` right now, because this is a transition step.
+	// Eventually, this should be a startup check.
+	mc.secretsLister = ops.Secrets
 
 	if ops.Domain == "" {
 		return nil, errors.New("metrics domain cannot be empty")
@@ -239,6 +279,11 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
 				metricType := path.Join(mc.stackdriverMetricTypePrefix, ms.Measure().Name())
 
 				if servingOrEventing.Has(metricType) {
+					var err error
+					ros, err = resourceToTags(ctx, ros...)
+					if err != nil {
+						return err
+					}
 					return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(ms))...)
 				}
 				// Otherwise, skip (because it won't be accepted)
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -175,14 +175,13 @@ func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros .
 // the transition to using Resources).
 func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options, error) {
 	if r, ok := metricskey.FromContext(ctx); ok && r != nil {
-		tags := []tag.Mutator{}
+		tags := make([]tag.Mutator, 0, len(r.Labels))
 		for label, v := range r.Labels {
 			key, err := tag.NewKey(label)
 			if err != nil {
-				return nil, fmt.Errorf("Bad resource label %q can't be converted to a tag: %+v", label, err)
+				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %+v", label, err)
 			}
 			tags = append(tags, tag.Upsert(key, v))
-			fmt.Printf("Adding tag %q: %q\n", key, v)
 		}
 		return append(ros, stats.WithTags(tags...)), nil
 	}
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -159,9 +159,8 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 // Record applies the `ros` Options to `ms` and then records the resulting
 // measurements in the metricsConfig's designated backend.
 func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
-	var err error
 	if mc == nil || mc.recorder == nil {
-		ros, err = resourceToTags(ctx, ros...)
+		ros, err := resourceToTags(ctx, ros...)
 		if err != nil {
 			return err
 		}
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -179,7 +179,7 @@ func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options,
 		for label, v := range r.Labels {
 			key, err := tag.NewKey(label)
 			if err != nil {
-				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %+v", label, err)
+				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %w", label, err)
 			}
 			tags = append(tags, tag.Upsert(key, v))
 		}
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -74,8 +74,8 @@ type ExporterOptions struct {
 // when a config map is updated.
 // DEPRECATED: Callers should migrate to UpdateExporterFromConfigMap3 (which will be renamed
 // when all callers are migrated).
-func UpdateExporterFromConfigMap(component string, secrets clientv1.SecretLister, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
-	return UpdateExporterFromConfigMap3(component, secrets, logger)
+func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	return UpdateExporterFromConfigMap3(component, nil, logger)
 }
 
 // UpdateExporterFromConfigMap3 includes a corev1.SecretLister which is used to configure
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -20,6 +20,7 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	clientv1 "k8s.io/client-go/listers/core/v1"
 )
 
 var (
@@ -64,17 +65,31 @@ type ExporterOptions struct {
 	// See https://github.com/knative/serving/blob/master/config/config-observability.yaml
 	// for details.
 	ConfigMap map[string]string
+
+	// A lister for Secrets to allow dynamic configuration of outgoing TLS client cert.
+	Secrets clientv1.SecretLister `json:"-"`
 }
 
 // UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
 // when a config map is updated.
-func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+// DEPRECATED: Callers should migrate to UpdateExporterFromConfigMap3 (which will be renamed
+// when all callers are migrated).
+func UpdateExporterFromConfigMap(component string, secrets clientv1.SecretLister, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	return UpdateExporterFromConfigMap3(component, secrets, logger)
+}
+
+// UpdateExporterFromConfigMap3 includes a corev1.SecretLister which is used to configure
+// mTLS with the opencensus agent.
+// Callers should migrate to this function, which will be renamed to UpdateExporterFromConfigMap
+// once all callers in all repos are renamed.
+func UpdateExporterFromConfigMap3(component string, secrets clientv1.SecretLister, logger *zap.SugaredLogger) func(*corev1.ConfigMap) {
 	domain := Domain()
 	return func(configMap *corev1.ConfigMap) {
 		UpdateExporter(ExporterOptions{
 			Domain:    domain,
 			Component: component,
 			ConfigMap: configMap.Data,
+			Secrets:   secrets,
 		}, logger)
 	}
 }
--- a/metrics/metricskey/context.go
+++ b/metrics/metricskey/context.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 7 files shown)


metrics/metricskey/context.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2019 The Knative Authors
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2020 The Knative Authors
```

metrics/config.go
```diff
@@ -147,15 +159,43 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+// resourceToTags is a backwards-compatibility shim to forward all the labels
+// on the Resource context to tags (used for Prometheus and Stackdriver during
+// the transition to using Resources).
+func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options, error) {
+	if r, ok := metricskey.FromContext(ctx); ok && r != nil {
+		tags := []tag.Mutator{}
+		for label, v := range r.Labels {
+			key, err := tag.NewKey(label)
+			if err != nil {
+				return nil, fmt.Errorf("Bad resource label %q can't be converted to a tag: %+v", label, err)
```
vagababov:
```suggestion
				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %w", label, err)
```

metrics/config.go
```diff
@@ -147,15 +159,43 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
 	}
 	return mc.recorder(ctx, ms, ros...)
 }
 
+// resourceToTags is a backwards-compatibility shim to forward all the labels
+// on the Resource context to tags (used for Prometheus and Stackdriver during
+// the transition to using Resources).
+func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options, error) {
+	if r, ok := metricskey.FromContext(ctx); ok && r != nil {
+		tags := []tag.Mutator{}
```
vagababov:
```suggestion
		tags := make([]tag.Mutator,0,len(r.Labels))
```

metrics/config.go
```diff
@@ -147,15 +159,43 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+		for label, v := range r.Labels {
+			key, err := tag.NewKey(label)
+			if err != nil {
+				return nil, fmt.Errorf("Bad resource label %q can't be converted to a tag: %+v", label, err)
+			}
+			tags = append(tags, tag.Upsert(key, v))
+			fmt.Printf("Adding tag %q: %q\n", key, v)
```
vagababov: rm debug.

metrics/exporter.go
```diff
@@ -64,17 +65,31 @@ type ExporterOptions struct {
...
 // UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
 // when a config map is updated.
-func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+// DEPRECATED: Callers should migrate to UpdateExporterFromConfigMap3 (which will be renamed
+// when all callers are migrated).
+func UpdateExporterFromConfigMap(component string, secrets clientv1.SecretLister, logger *zap.SugaredLogger) func(configMap *corev1.ConfigMap) {
+	return UpdateExporterFromConfigMap3(component, secrets, logger)
+}
+
+// UpdateExporterFromConfigMap3 includes a corev1.SecretLister which is used to configure
```
vagababov: Why `3` and not `2` or `Ex` or `Temp`?

metrics/opencensus_exporter.go
```diff
@@ -36,3 +55,82 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+		return &resource.Resource{
+			Type:   metricskey.ResourceTypeKnativeBroker,
+			Labels: extractLabelsFromMap(m, metricskey.KnativeBrokerLabels.UnsortedList()...),
+		}, nil
+	}
+	return nil, nil
+}
+
+func extractLabelsFromMap(t *tag.Map, labels ...string) map[string]string {
+	ret := map[string]string{}
```
vagababov:
```suggestion
	ret := make(map[string]string, len(labels))
```

metrics/config.go
evankanderson: Done

metrics/config.go
evankanderson: Done

metrics/config.go
evankanderson: Whoops, thanks!

metrics/exporter.go
evankanderson: Because it has 3 arguments.

vagababov: But the old one also had 3 args, no? :-)

metrics/config.go
```diff
@@ -147,15 +159,42 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+// resourceToTags is a backwards-compatibility shim to forward all the labels
+// on the Resource context to tags (used for Prometheus and Stackdriver during
+// the transition to using Resources).
+func resourceToTags(ctx context.Context, ros ...stats.Options) ([]stats.Options, error) {
+	if r, ok := metricskey.FromContext(ctx); ok && r != nil {
+		tags := make([]tag.Mutator, 0, len(r.Labels))
+		for label, v := range r.Labels {
+			key, err := tag.NewKey(label)
+			if err != nil {
+				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %+v", label, err)
```
vagababov: there's also `%w` change here :)
```suggestion
				return nil, fmt.Errorf("bad resource label %q can't be converted to a tag: %w", label, err)
```

metrics/exporter.go
```diff
@@ -64,17 +65,31 @@ type ExporterOptions struct {
 	// See https://github.com/knative/serving/blob/master/config/config-observability.yaml
 	// for details.
 	ConfigMap map[string]string
+
+	// A lister for Secrets to allow dynamic configuration of outgoing TLS client cert.
+	Secrets clientv1.SecretLister `json:"-"`
```
vagababov: Do we need `json` tag at all?

metrics/metricskey/context.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+
+	"go.opencensus.io/resource"
+)
+
+// key is an unexported type for keys defined in this package.
+// This prevents collisions with keys defined in other packages.
+type key int
```
vagababov: Why not go with our favorite
```
type resourceKey struct{}
```
?

metrics/opencensus_exporter.go
```diff
@@ -36,3 +55,82 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+func defaultResourceDescriptor(ctx context.Context) (*resource.Resource, error) {
+	if r, ok := metricskey.FromContext(ctx); ok {
+		return r, nil
+	}
+	m := tag.FromContext(ctx)
+
+	if _, ok := m.Value(tag.MustNewKey(metricskey.LabelRevisionName)); ok {
+		return &resource.Resource{
+			Type:   metricskey.ResourceTypeKnativeRevision,
+			Labels: extractLabelsFromMap(m, metricskey.KnativeRevisionLabels.UnsortedList()...),
```
anniefu: So, unfortunately, the  "WithResourceDetector" option only runs the ResourceDetector once at time of exporter creation, so it can't capture dynamic label values like revision_name and configuration_name.

This isn't necessarily breaking problem since the correct values will still be passed along as normal metrics tags, but it will be very strange to have the resource labels differ from the metrics tags.

I'm not sure if the static nature of the Resource is by design for the ocagent or just a code limitation... I can find out from the Otel folks.

metrics/opencensus_exporter.go
```diff
@@ -36,3 +55,82 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+			if err != nil {
+				return nil, fmt.Errorf("Unable to fetch opencensus secret for %q, cannot use requireSecure=true: %+v", component, err)
+			}
+			return &cert, err
+		},
+	})
+}
+
+func certificateFetcher(secretName string, lister clientv1.SecretLister) (tls.Certificate, error) {
+	secret, err := lister.Secrets(system.Namespace()).Get(secretName)
```
anniefu: This has a problem similar to https://github.com/knative/pkg/issues/842 where the exporter should be re-created when the Secret changes.

metrics/opencensus_exporter.go
evankanderson: Thanks for pointing this out.

I'm guessing that the common case for OpenCensus is that each object reports statistics about itself, rather than for infrastructure which reports on behalf of other objects. Do you want to see about making an API change upstream for this case?

Looking at the code, it looks like `Exporter.resource` is used in the following places:

**Trace:**
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go#L232
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go#L507

**View:**
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go#L268
https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go#L540


Unfortunately, I think the assumption of a single resource may be somewhat baked in to the exporter, because the same resource is used for all the metrics exported from a View.

I'm wondering whether we should maintain multiple exporters in multi-tenant cases like the activator and controllers. We could do this without neeeding to change the interface by keeping a map from Resource -> Exporter in this OpenCensus module.

metrics/opencensus_exporter.go
```diff
@@ -36,3 +55,82 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+		logger.Errorf("No secret lister provided for component %q; cannot use requireSecure=true", component)
+		return nil
+	}
+	return credentials.NewTLS(&tls.Config{
+		GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			// We ignore the CertificateRequestInfo for now, and hand back a single fixed certificate.
+			// TODO(evankanderson): maybe do something SPIFFE-ier?
+			cert, err := certificateFetcher(component+"-opencensus", lister)
+			if errors.IsNotFound(err) {
+				cert, err = certificateFetcher("opencensus", lister)
```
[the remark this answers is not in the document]
evankanderson: The intent was actually to collide for simple cases where you want all the processes in the namespace to use the same secret.

metrics/opencensus_exporter.go
```diff
@@ -36,3 +55,82 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+			}
+			if err != nil {
+				return nil, fmt.Errorf("Unable to fetch opencensus secret for %q, cannot use requireSecure=true: %+v", component, err)
```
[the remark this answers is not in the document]
evankanderson: I'd rather do a search-and-replace when we do the migration, which will pick up the comments as well as uses like error strings.

metrics/record_test.go
```diff
@@ -98,6 +101,65 @@ func TestRecordEventing(t *testing.T) {
...
+			want: map[string]string{"foo": "bar", "hero": "aquaman"},
+		},
+	}
+
+	for _, backend := range []metricsBackend{Prometheus, Stackdriver, OpenCensus} {
+
+		for _, test := range testCases {
+			t.Run(test.name, func(t *testing.T) {
+				view.Register(v)
+				defer view.Unregister(v)
```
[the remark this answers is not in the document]
evankanderson: The `defer` is inside the `func(t *testing.T)` that is handed to `t.Run`, so it is actually called on each loop.

For proof that the defer is called at the end of the anonymous function, see:
https://play.golang.org/p/bhfj8VRtKrI

evankanderson:
Note that I'm still working on getting this up and running with TLS in a Knative cluster; my testing environment went kablooey (technical term) over the weekend.

metrics/config.go
```diff
@@ -147,15 +159,42 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 // Record applies the `ros` Options to `ms` and then records the resulting
 // measurements in the metricsConfig's designated backend.
 func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
+	var err error
```
vagababov: this can be folded into the `if`.

metrics/exporter.go
vagababov: Perhaps we need tags on the others as well, for uniformity?

metrics/opencensus_exporter.go
evankanderson: This fetches the Secret (from the Informer cache) every time that a new TLS connection is constructed, because the grpc `credentials.TransportCredentials` constructed from `NewTLS` will use the `tls.Config` to start the handshake: https://github.com/grpc/grpc-go/blob/master/credentials/tls.go#L70

In turn, the `tls.Config` is constructed with a [`GetClientCertificate` function, which says](https://godoc.org/crypto/tls#Config):

> GetClientCertificate, if not nil, is called when a server requests a
> certificate from a client. If set, the contents of Certificates will
> be ignored.
>
> If GetClientCertificate returns an error, the handshake will be
> aborted and that error will be returned. Otherwise
> GetClientCertificate must return a non-nil Certificate. If
> Certificate.Certificate is empty then no certificate will be sent to
> the server. If this is unacceptable to the server then it may abort
> the handshake.
>
> GetClientCertificate may be called multiple times for the same
> connection if renegotiation occurs or if TLS 1.3 is in use.

This is why we need to feed the SecretLister down into the OpenCensus agent, so that it can fetch the secret whenever the library attempts to `grpc.Dial()` the server.

metrics/exporter.go
evankanderson: I'm happy to use the default behavior for the others, though I can add tags if you think it's important that we hint that these are JSON-serialized in eventing-contrib (that seems to be the only place where to/from JSON is used at all).

metrics/config.go
evankanderson: Thanks, I'd previously had the `if` on one line with the assignment, but realized I needed to avoid aliasing `ros`.

metrics/opencensus_exporter.go
anniefu: Digging around the OpenCensus world, each OpenCensus Metric does actually support having a unique Resource field that can supercede the batch's Resource: https://github.com/census-instrumentation/opencensus-proto/blob/master/gen-go/metrics/v1/metrics.pb.go#L108

The ocagent does convert each row in a view individually, so we could modify it to re-run the ResourceDetector on each view.Data row: https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/transform_stats_to_metrics.go#L36 

I prefer your idea of doing an exporter per unique knative_revision Resource though because:
1. Even if we fix the ocagent exporter, OpenCensus collector service & other exporters rely on batching too, so they'd all need to respect Metric's Resource over the batched Resource. I know Stackdriver exporter will, but not sure about the others.
2. I'm not super familiar with contexts but it seems like the ocagent exporter's context isn't shared with the Knative controller's and won't have the required metrics tags anyway? https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/ocagent.go#L130

My only concern with one exporter per Resource is it might be a lot of exporters, but I think it would just be one exporter per revision? Would you agree on this set of Resource labels?
- revision_name
- configuration_name
- namespace_name
- service_name

Assuming the GCP stuff is going to be ripped out and shouldn't be part of the OpenCensus knative_revision Resource model (which can be added back on the collector service side).

metrics/opencensus_exporter.go
anniefu: Ah, I see, thanks for the detailed response! 

In terms of the keeping the SecretLister package local though, couldn't the metrics package maintain its own in-cluster kubeclient too?

metrics/opencensus_exporter.go
```diff
@@ -14,19 +14,38 @@ limitations under the License.
...
 	"contrib.go.opencensus.io/exporter/ocagent"
 	"go.opencensus.io/stats/view"
 	"go.uber.org/zap"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/apimachinery/pkg/api/errors"
+	clientv1 "k8s.io/client-go/listers/core/v1"
+	"knative.dev/pkg/system"
 )
 
 func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
```
anniefu: Separate from this PR, but it looks like WithServiceName does not have the same effects as component name does for Prometheus/Stackdriver backends.

For example, for Prometheus I'm seeing:
"namespace_actual_pods" instead of "autoscaler_actual_pods"
And Stackdriver I get:
"custom.googleapis.com/opencensus/actual_pods" instead of "custom.googleapis.com/opencensus/autoscaler/actual_pods"

For Stackdriver, it looks like setting the "view.Name" field for each view in stats_reporter to "autoscaler/actual_pods" would work. Haven't looked into Prometheus yet.

evankanderson: Thanks -- do you have quick instructions for setting up the environment you're using to validate this?

anniefu: I filled out the Stackdriver section of the config from your other PR, then deployed on GKE cluster, then the metrics show up in Stackdriver under 
"custom.googleapis.com/opencensus/actual_pods" with Monitored Resource Type "global".

The "testing" prefix only seems to show up in the display name, but not the metric name itself.

```
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
  name: opencensus-agent-config
  namespace: knative-monitoring
data:
  opencensus-agent-config: |
    receivers:
      opencensus:
        address: ":55678"
    #   jaeger: {}
    #   zipkin: {}
    exporters:
      stackdriver:
        # project: "kubecon-knative-2019"  # Optional, defaults to node project
        metric_prefix: "testing"
        enable_metrics: true
        enable_tracing: true
      prometheus:
        namespace: "namespace"
        address: ":8888"
# ---
# apiVersion: v1
# kind: Secret
# metadata:
#   name: stackdriver-key
#   namespace: knative-monitoring
# type: Opaque
# data:
#   credentials.json: <<<BASE64 JSON>>>
---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  labels:
  name: opencensus-agent
  namespace: knative-monitoring
spec:
  selector:
    matchLabels:
      component: opencensus-agent
  template:
    metadata:
      labels:
        component: opencensus-agent
    spec:
      containers:
      - command:
        - /ocagent_linux
        - --config=/conf/opencensus-agent-config.yaml
        # env:
        # - name: GOOGLE_APPLICATION_CREDENTIALS
        #   value: /secrets/credentials.json
        image: omnition/opencensus-agent:0.1.6
        name: opencensus-agent
        ports:
        - containerPort: 55678
          name: collect
          protocol: TCP
        - containerPort: 55679
          name: zpages
          protocol: TCP
        - containerPort: 8888
          name: prometheus
          protocol: TCP
        resources:
          limits:
            cpu: 500m
            memory: 500Mi
          requests:
            cpu: 100m
            memory: 100Mi
        volumeMounts:
        - mountPath: /conf
          name: opencensus-agent-config-vol
        # - mountPath: /secrets
        #   name: stackdriver-key
      volumes:
      - configMap:
          items:
          - key: opencensus-agent-config
            path: opencensus-agent-config.yaml
          name: opencensus-agent-config
        name: opencensus-agent-config-vol
      # - name: stackdriver-key
      #   secret:
      #     defaultMode: 420
      #     secretName: stackdriver-key
---
apiVersion: v1
kind: Service
metadata:
  labels:
  name: opencensus
  namespace: knative-monitoring
spec:
  ports:
  - name: opencensus
    port: 55678
    protocol: TCP
    targetPort: 55678
  selector:
    component: opencensus-agent
  type: ClusterIP
```

vagababov:
@evankanderson I presume we can close this one, due to all opencensus work?

PR #999, opened 2020-01-21
Adding ensure package with convenience helpers Error, NoError, ErrorWithMessage

cardil:
Writing a Go code makes a lot of repetition, regarding error checking. That's especially true for end user code like e2e tests when we expect that something will work. In other cases that's a bug and should code should panic then.

I'd like to add a `ensure` package that contains convenience helpers in those cases. Example usage can be seen in this short example:

https://github.com/cardil/eventing/blob/feature/upgrade-tests/test/prober/configuration.go#L50-L60

### DIFF
--- a/ensure/errors.go
+++ b/ensure/errors.go
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020 The Knative Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ensure
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"regexp"
+)
+
+// NoError will panic if given an error, as it was unexpected
+func NoError(err error) {
+	if err != nil {
+		panic(errors.WithMessage(err, "unexpected error"))
+	}
+}
+
+// Error will panic if given no error, as it expected one
+func Error(err error) {
+	if err == nil {
+		panic(errors.New("expecting error, but none given"))
+	}
+}
+
+// ErrorWithMessage will panic if given no error, or error message don't match provided regexp
+func ErrorWithMessage(err error, messageRegexp string) {
+	Error(err)
+	validErrorMessage := regexp.MustCompile(messageRegexp)
+	if !validErrorMessage.MatchString(err.Error()) {
+		panic(errors.WithMessage(
+			err,
+			fmt.Sprintf("given error doesn't match given regexp (%s)", messageRegexp),
+		))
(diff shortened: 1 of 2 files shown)


vagababov:
Can you explain why do we need this library?

cardil:
Having such helper tools greatly simplify the end use code, for example e2e tests.

When you write an e2e code you generally want specific thing, and you want to panic if something goes wrong.

Here is an example:

```go
func TestSomething(t *testing.T) {
  alpha, err := deployAlpha()
  if err != nil {
    panic(errors.WithMessage(err, "unexpected error"))
  }
  beta, err := deployBeta(alpha)
  if err != nil {
    panic(errors.WithMessage(err, "unexpected error"))
  }
  _, err := deployGamma(beta)
  if err != nil {
    panic(errors.WithMessage(err, "unexpected error"))
  }
}
```

with this PR I can write it like:

```go
func TestSomething(t *testing.T) {
  alpha, err := deployAlpha()
  ensure.NoError(err)
  beta, err := deployBeta(alpha)
  ensure.NoError(err)
  _, err := deployGamma(beta)
  ensure.NoError(err)
}
```

Above is much more readable and pleasant to see.

Here is a real world example: 

https://github.com/cardil/eventing/blob/feature/upgrade-tests/test/prober/configuration.go#L50-L60

vagababov:
Panicing in test is contituitive. You just do `t.Error` or `t.Fatal` depending on the context.

cardil:
`t.Fatal` panic in it's implementation

vagababov:
I don't think that's true. It calls `Goexit`.
Even if it did, it should be the only place where it happens.

cardil:
Sure. But, as I was saying it's useful in all situations in which you don't expect errors and don't want to handle each one of them.

It removes a bunch of conditionals that are repetitive and makes code easier to understand.

vagababov:
In general the Go philosophy has been against this kind of helpers.
Implementation with panics is basically bad. You don't want to panic on error. t.Error continuues on execution.

cardil:
@vagababov If you write libraries, then yes, you should return errors. If writing end code, you should just panic, assuming there shouldn't be errors, assuming you are walking the happy path.

Maybe below examples will be better to illustrate the idea. 

Instead of writing:

```go
func DeployAllComponents() error {
  alpha, err := deployAlpha()
  if err != nil {
    return errors.WithMessage(err, "unexpected error")
  }
  beta, err := deployBeta(alpha)
  if err != nil {
    return errors.WithMessage(err, "unexpected error")
  }
  _, err := deployGamma(beta)
  if err != nil {
    return errors.WithMessage(err, "unexpected error")
  }
  return nil
}

// execution isn't simple
err = DeployAllComponents()
if err != nil {
  panic(err)
}
```

with this PR I can write it like:

```go
func DeployAllComponents() {
  alpha, err := deployAlpha()
  ensure.NoError(err)
  beta, err := deployBeta(alpha)
  ensure.NoError(err)
  _, err := deployGamma(beta)
  ensure.NoError(err)
}

// execution is simple
DeployAllComponents()
```

Above is much more readable and easier to understand.

vagababov:
Your example in the original PR text is "e2e tests".
If you're using this for client, perhaps it should be in the client repo.

vagababov:
Is this still being pursued? Or we can close it?

cardil:
I've extracted that to [github.com/wavesoftware/go-ensure](https://github.com/wavesoftware/go-ensure) package. Closing.

PR #1003, opened 2020-01-22
status.ConvertTo can optionally accept predicates

dprotaso:
the intent is to allow the inclusion of addition ConditionTypes to be copied over if a
a predicate returns true

Motivation: https://github.com/knative/serving/pull/6542/files?file-filters%5B%5D=.go#r368663116

### DIFF
--- a/apis/duck/v1/status_types.go
+++ b/apis/duck/v1/status_types.go
@@ -99,17 +99,32 @@ func (s *Status) GetCondition(t apis.ConditionType) *apis.Condition {
 }
 
 // ConvertTo helps implement apis.Convertible for types embedding this Status.
-func (source *Status) ConvertTo(ctx context.Context, sink *Status) {
+//
+// By default apis.ConditionReady and apis.ConditionSucceeded will be copied over to the
+// sink. Other conditions types are tested against a list of predicates. If any of the predicates
+// return true the condition type will be copied to the sink
+func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ...func(apis.ConditionType) bool) {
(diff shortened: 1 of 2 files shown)


dprotaso:
From: https://github.com/knative/serving/pull/6542/files?file-filters%5B%5D=.go#r368663672

> Also, perhaps this is something where a conditionset can expose a compatible function?

ConditionSets are pretty rigid at this time and only contain terminal conditions & (ready|succeeded). I can imagine some non-terminal conditions being converted between different resource versions.

apis/duck/v1/status_types.go
```diff
@@ -99,17 +99,32 @@ func (s *Status) GetCondition(t apis.ConditionType) *apis.Condition {
 }
 
 // ConvertTo helps implement apis.Convertible for types embedding this Status.
-func (source *Status) ConvertTo(ctx context.Context, sink *Status) {
+//
+// By default apis.ConditionReady and apis.ConditionSucceeded will be copied over to the
+// sink. Other conditions types are tested against a list of predicates. If any of the predicates
+// return true the condition type will be copied to the sink
+func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ...func(apis.ConditionType) bool) {
```
mattmoor: I'd rather take a single predicate and leave conjunction/disjunction semantics up to the caller?

dprotaso: the `...` is there to ensure this isn't a breaking change

dprotaso: but it has that side effect of being a disjunction

mattmoor: 🤔 I like that even better, thanks.

PR #1004, opened 2020-01-22
Only intercept types that appear in Bindings.

mattmoor:
This removes the built-in types from being hard-coded, and we will now only intercept them if they are used in a Binding.

cc @bbrowning @vaikas


n3wscott:
Is there an eventing update to this PR too?

mattmoor:
Shouldn't need anything.  This just seeded stuff that would have been added dynamically anyways.

evankanderson:
It looks like the follow-on code doesn't use the CRDs, just the GVKs, so adding an extra set of GVKs doesn't really buy anything.

/lgtm

Issue #1005, opened 2020-01-22
Allow to specify PrometheusPort when creating metrics exporter 


yolocs:
## Problem

Currently https://github.com/knative/pkg/blob/master/metrics/exporter.go#L71 does allow specifying a `PrometheusPort` when creating the exporter. When a knative component uses metrics exporter along with serving queue-proxy (which also uses the metrics exporter), the same default port 9090 will be used and cause failure.

## Solution

Allow to specify PrometheusPort when creating metrics exporter so that a knative component could specify a different port than the default one.

## Reference

Eventing broker filter/ingress uses the metrics package. And when we try to create them as ksvc, they run into a port conflict as queue-proxy. https://github.com/knative/eventing/issues/2405

PR #1006, opened 2020-01-22
Allow to specify default exporter option when updating from ConfigMap

yolocs:
To fix: https://github.com/knative/pkg/issues/1005

### DIFF
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -14,6 +14,7 @@ limitations under the License.
 package metrics
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


yolocs:
@vaikas

metrics/exporter.go
```diff
@@ -79,6 +79,29 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
 	}
 }
 
+// UpdateExporterFromConfigMapWithDefaults returns a helper func that can be used to update the exporter
```
vaikas: Super nit but I find the name of this method a bit confusing. Are the Defaults here that then get overridden by the ConfigMap? I'd think Opts might be better, since you're providing the values to use to create the helper func?
UpdateExporterFromConfigMapWithOpts(opts ExporterOptions, logger...)
Also might wan to say that Component must not be specified but ConfigMap must not, which is also a bit confusing, since the ExporterOptions struct has this gem in it:
	// ConfigMap is the data from config map config-observability. Must be present.

yolocs: Make sense. I updated the comment to point out that ConfigMap must not be present because the value from the ConfigMap will be used instead.

PR #1008, opened 2020-01-23
add retry func for reconciler

nlopezgi:
https://github.com/knative/serving/pull/6162 added a retry mechanism for status updates that will be of use for:
- https://github.com/knative/eventing/issues/2296 &
- https://github.com/google/knative-gcp/issues/442
As suggested in https://github.com/google/knative-gcp/pull/508/files#r370294441 this PR upstreams the code to perform retries to this location.
After this PR I will send a PR to https://github.com/knative/serving removing their copy of retry.go, I will also update https://github.com/google/knative-gcp/pull/508 to use this, and likely take on https://github.com/knative/eventing/issues/2296.

fyi @nachocano


nlopezgi:
PR was failing on presubmit complaining I need to run ./hack/update-codegen.sh. Had to also run update-deps which added tons of files to vendor directory. Not too experienced with go, so not sure this should be expected of this PR.

vagababov:
Yes you need to run those, to prune the redundant files.
your chage should be 1-2 files at most.

nlopezgi:
> Yes you need to run those, to prune the redundant files.
> your chage should be 1-2 files at most.

Thanks, I think I got it to run properly this time. Please let me know if this looks good. Thanks!

reconciler/retry_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+package reconciler
+
+import (
+	"errors"
+	"testing"
+
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+
+	v1 "k8s.io/api/autoscaling/v1"
```
dgerd: nit: the package name is already v1 so the identifier is not necessary here.

```suggestion
	"k8s.io/api/autoscaling/v1"
```

dgerd:
/assign @mattmoor 
as I don't have approval powers here.

reconciler/retry.go
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright 2020 The Knative Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    https://www.apache.org/licenses/LICENSE-2.0
```
vagababov: Can you please use the standard license header.
We have this botched one in a few places, but we should not introduce new ones.
https://github.com/knative/pkg/blob/bec759b1d79ec5e1d2c00450f04477e85b7dc3fc/reconciler/events.go, e.g. -- see empty lines around the URL.

reconciler/retry_test.go
mattmoor: goimports adds these, leave it. (thank gomod 🙄 )

reconciler/retry.go
nlopezgi: fixed

nlopezgi:
@mattmoor this is ready to go; or let me know if there are any pending changes, thanks!

PR #1009, opened 2020-01-24
conditions were being duplicated since we broke vs. continuing

dprotaso:


apis/duck/v1/status_types.go
```diff
@@ -107,19 +107,21 @@ func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ..
...
 			conditions = append(conditions, c)
-			break
+			continue
 		}
 
 		for _, predicate := range predicates {
 			if predicate(c.Type) {
 				conditions = append(conditions, c)
-				break
+				continue condition_loop
```
vagababov: this I think was correct.

apis/duck/v1/status_types.go
```diff
@@ -107,19 +107,21 @@ func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ..
 	sink.ObservedGeneration = source.ObservedGeneration
 
 	conditions := make(apis.Conditions, 0, len(source.Conditions))
+
+condition_loop:
 	for _, c := range source.Conditions {
 		switch c.Type {
 		// Copy over the "happy" condition, which is the only condition that
```
vagababov: it might be better constructed as
```
if a == b || a == c {
  ... 
  continue
}
```

dprotaso: agreed

PR #1011, opened 2020-01-26
Escape test name before matching slack message history

chizhg:
1. When we use Slack API to retrieve message history, the message text is escaped. If the test name contains special characters, like in `IMC Channel Latency & Throughput`, the deduplication logic will be broken and we will keep sending alerts within short interval, see https://knative.slack.com/archives/CBDMABCTF/p1579647224000300. This PR fixes this issue by escaping the test name in matching message history (and also adds the test)

2. A minor improvement by adding `https://` prefix in the link, so that it can be actually shown as a link in the Slack message.

### DIFF
--- a/test/mako/alerter/alerter.go
+++ b/test/mako/alerter/alerter.go
@@ -61,7 +61,7 @@ func (alerter *Alerter) HandleBenchmarkResult(
 			summary := fmt.Sprintf("%s\n\nSee run chart at: %s\n\nSee aggregate chart at: %s",
 				output.GetSummaryOutput(),
 				output.GetRunChartLink(),
-				fmt.Sprintf("mako.dev/benchmark?benchmark_key=%s&tseconds=604800", benchmarkKey),
+				fmt.Sprintf("https://mako.dev/benchmark?benchmark_key=%s&tseconds=604800", benchmarkKey),
 			)
 			if alerter.githubIssueHandler != nil {
 				if err := alerter.githubIssueHandler.CreateIssueForTest(benchmarkName, summary); err != nil {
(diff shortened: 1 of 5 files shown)


test/mako/alerter/alerter.go
```diff
@@ -61,7 +61,7 @@ func (alerter *Alerter) HandleBenchmarkResult(
 			summary := fmt.Sprintf("%s\n\nSee run chart at: %s\n\nSee aggregate chart at: %s",
 				output.GetSummaryOutput(),
 				output.GetRunChartLink(),
-				fmt.Sprintf("mako.dev/benchmark?benchmark_key=%s&tseconds=604800", benchmarkKey),
+				fmt.Sprintf("https://mako.dev/benchmark?benchmark_key=%s&tseconds=604800", benchmarkKey),
```
vagababov:
```suggestion
				"https://mako.dev/benchmark?tseconds=604800&benchmark_key="+benchmarkKey,
```
If you want to keep sprintf, I'd highly recommend 60480 be defined as a constant `7*24*60*60`.

test/mako/alerter/slack/message.go
```diff
@@ -91,7 +92,8 @@ func (smh *MessageHandler) SendAlert(testName, summary string) error {
 			// do not send message again if alert for this test has been sent to
 			// the channel a short while ago
 			for _, message := range messageHistory {
-				if strings.Contains(message, decoratedTestName) {
+				// the message text queried from Slack will be escaped, so we also need to do escape here
+				if strings.Contains(message, html.EscapeString(decoratedTestName)) {
```
chaodaiG: I think the reason for text being escaped is caused by `slackutil`, and it should be `slackutil`'s responsibility to provide this function. WDYT?

test/mako/alerter/alerter.go
chizhg: Done.

test/mako/alerter/slack/message.go
chizhg: It makes sense. I moved the logic to `slackutil`. Thanks!

PR #1012, opened 2020-01-27
Use major/minor version of the k8s version to version check explicitly.

markusthoemmes:
Parsing the GitVersion is a little brittle as the parser doesn't seem to handle BUILD tags nicely. We can either do this or we can send a patch upstream to fix the parser. Or we use a different library 🤔 .

Alternative to #1013


markusthoemmes:
We went with another.

PR #1015, opened 2020-01-28
Give cloud-platform scope by default in creating GKE clusters

chizhg:
Background:
When we create a GKE cluster, we can specify the scopes, which means the set of Google API scopes to be made available on all of the node VMs under the "default" service account. If it's not specified, no scopes will be added.
Plus, communicating and authenticating with **gcr.io** requires `https://www.googleapis.com/auth/devstorage.read_only`. If it's not specified, the workloads on that cluster won't be able to pull images from private GCR.

This PR adds CloudPlatformScope in the base gke lib, so all clusters will be able to talk to all Google APIs. We can make it more fine granular in the future if needed.

/cc @chaodaiG 
/cc @yt3liu 
/cc @adrcunha

### DIFF
--- a/test/gke/request.go
+++ b/test/gke/request.go
@@ -114,6 +114,15 @@ func NewCreateClusterRequest(request *Request) (*container.CreateClusterRequest,
 					},
 					Config: &container.NodeConfig{
 						MachineType: request.NodeType,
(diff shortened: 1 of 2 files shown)


adrcunha:
Why this is an issue now?

PR #1024, opened 2020-01-30
Use e2 VMs instead of n1 for test clusters

adrcunha:
It's more efficient on Google Cloud.

Updated a couple of examples as well, to avoid copy&paste of code using n1 instances.


coryrc:
Have you benchmarked the difference? Reading the [page on machine types](https://cloud.google.com/compute/docs/machine-types) the main difference seems to be lack of SSD and I wouldn't expect that to matter for the test cluster (I think it could be bad for the Prow cluster, but that's not this PR).

PR #1025, opened 2020-01-30
Add RecordN method to permit recording of N measurements simultaneously.

vagababov:
In serving we have many places where we try to record several metrics at the same time.
Besides very repetetive code, this also brings inefficiences, since metrics code has to do _some_ work
for every call, which is completely avoidable.

This way we can record N measurements with one hit.

If we agree in principle, I'll extend the unit tests.

/assign @evankanderson @anniefu

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -84,7 +84,7 @@ type metricsConfig struct {
 
 	// recorder provides a hook for performing custom transformations before
 	// writing the metrics to the stats.RecordWithOptions interface.
-	recorder func(context.Context, stats.Measurement, ...stats.Options) error
+	recorder func(context.Context, []stats.Measurement, ...stats.Options) error
 
 	// ---- OpenCensus specific below ----
 	// collectorAddress is the address of the collector, if not `localhost:55678`
@@ -144,13 +144,19 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 	}
 }
 
+// RecordN applies the `ros` Options to each measurement in `mss` and then records the resulting
+// measurements in the metricsConfig's designated backend.
+func (mc *metricsConfig) RecordN(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
+	if mc == nil || mc.recorder == nil {
+		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+	}
+	return mc.recorder(ctx, mss, ros...)
+}
+
 // Record applies the `ros` Options to `ms` and then records the resulting
 // measurements in the metricsConfig's designated backend.
 func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
-	if mc == nil || mc.recorder == nil {
-		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(ms))...)
-	}
-	return mc.recorder(ctx, ms, ros...)
+	return mc.RecordN(ctx, []stats.Measurement{ms}, ros...)
 }
 
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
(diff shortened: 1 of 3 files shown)


metrics/config.go
```diff
@@ -144,13 +144,19 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 	}
 }
 
-// Record applies the `ros` Options to `ms` and then records the resulting
+// RecordN applies the `ros` Options to each measurement in `mss` and then records the resulting
```
anniefu: Could we call this RecordBatch? RecordN makes me think I am recording the same measurement N times instead of N different measurements.

metrics/config.go
```diff
@@ -235,14 +241,22 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
 		if !allowCustomMetrics {
 			servingOrEventing := metricskey.KnativeRevisionMetrics.Union(
 				metricskey.KnativeTriggerMetrics).Union(metricskey.KnativeBrokerMetrics)
-			mc.recorder = func(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
-				metricType := path.Join(mc.stackdriverMetricTypePrefix, ms.Measure().Name())
-
-				if servingOrEventing.Has(metricType) {
-					return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(ms))...)
+			mc.recorder = func(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
+				w := 0
```
anniefu: This is an efficient, but kind of hard to read block of code. 

I'm not going to say we shouldn't save the extra array allocation, but could we use more descriptive variables than "w" and "r"? I'm not sure what they're intended to stand for...

metrics/config.go
```diff
@@ -144,13 +144,19 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 	}
 }
 
-// Record applies the `ros` Options to `ms` and then records the resulting
+// RecordN applies the `ros` Options to each measurement in `mss` and then records the resulting
 // measurements in the metricsConfig's designated backend.
-func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
+func (mc *metricsConfig) RecordN(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
```
anniefu: Could you add a test for this function please?

metrics/config.go
vagababov: `writeIndex` and `readIndex` :-)
I can make them more readable, this was more of a PoC:)

metrics/config.go
vagababov: The function _is_ actually tested via Record. But yeah, as said in description — if we're pursuing this — I'll add tests :)

yanweiguo:
LGTM. Do you still have unit tests to extend as described in https://github.com/knative/pkg/pull/1025#issue-369322279?

vagababov:
As per discussion with @evankanderson I changed the external interface to accept list rather than slice and removed options, since they are practically unused.
This should make users even happier.

metrics/config.go
```diff
@@ -144,13 +144,19 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 	}
 }
 
-// Record applies the `ros` Options to `ms` and then records the resulting
+// RecordBatch applies the `ros` Options to each measurement in `mss` and then records the resulting
 // measurements in the metricsConfig's designated backend.
-func (mc *metricsConfig) Record(ctx context.Context, ms stats.Measurement, ros ...stats.Options) error {
+func (mc *metricsConfig) RecordBatch(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
```
evankanderson: What do you think about making this method non-exported? I don't think it's called anywhere outside this package.

evankanderson: Actually, even better, I don't think `*metricsConfig.Record` is called at all outside the package, so we should only need one of these. (This exists simply to support the `nil` mc and default `recorder` case.)

metrics/config.go
```diff
@@ -235,14 +241,24 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
...
+					metricType := path.Join(mc.stackdriverMetricTypePrefix, mss[rIdx].Measure().Name())
+					if servingOrEventing.Has(metricType) {
+						mss[wIdx] = mss[rIdx]
+						wIdx++
+					}
+					// Otherwise, skip the measurement (because it won't be accepted).
+				}
+				// Trim the list to the number of written objects.
+				mss = mss[:wIdx]
+				// Found no matched metrics.
```
evankanderson: Move this above line 256?

metrics/config.go
```diff
@@ -235,14 +241,24 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
...
+				for rIdx := 0; rIdx < len(mss); rIdx++ {
+					metricType := path.Join(mc.stackdriverMetricTypePrefix, mss[rIdx].Measure().Name())
+					if servingOrEventing.Has(metricType) {
+						mss[wIdx] = mss[rIdx]
+						wIdx++
+					}
+					// Otherwise, skip the measurement (because it won't be accepted).
+				}
+				// Trim the list to the number of written objects.
+				mss = mss[:wIdx]
```
evankanderson: Do you need to do this reassignment here, or just do the slice on line 261?

metrics/record_test.go
```diff
@@ -38,68 +38,87 @@ type cases struct {
...
+		metricsConfig: &metricsConfig{},
+		measurement:   measure.M(1),
+	}, {
+		name: "stackdriver backend with supported metric",
+		metricsConfig: &metricsConfig{
+			isStackdriverBackend:        true,
+			stackdriverMetricTypePrefix: "knative.dev/eventing/broker",
 		},
-	}
+		measurement: measure.M(2),
```
evankanderson: Why did this change from 5 to 2? (More generally, the reformatting on this file makes it hard to review.)

metrics/config.go
vagababov: Mostly for readability, I don't need reassignment

metrics/record_test.go
vagababov: comment above `// Increase the measurement value for each test` -- so I made comment reality :-)
I can split reformatting into a separate PR if you want, but this is superior formatting :-)

metrics/config.go
vagababov: `mess[:wIdx]...` is a bit hard to comprehend from the getgo.

PR #1028, opened 2020-01-31
Use the new batch record method

vagababov:
Subj


vagababov:
This one isn't better
/close

PR #1029, opened 2020-01-31
Use new RecordBatch method to join metric reporting

vagababov:
This is in 2 places in pkg that we can take obvious advantage.


controller/stats_reporter.go
```diff
@@ -244,7 +244,7 @@ func (r *reporter) ReportReconcile(duration time.Duration, key, success string)
 		return err
 	}
 
-	metrics.Record(ctx, reconcileCountStat.M(1))
-	metrics.Record(ctx, reconcileLatencyStat.M(int64(duration/time.Millisecond)))
+	metrics.RecordBatch(ctx, reconcileCountStat.M(1),
+		reconcileLatencyStat.M(int64(duration/time.Millisecond)))
```
evankanderson:
```suggestion
		reconcileLatencyStat.M(duration.Milliseconds())))
```

https://golang.org/pkg/time/#Duration.Milliseconds

?

webhook/stats_reporter.go
```diff
@@ -101,9 +101,9 @@ func (r *reporter) ReportRequest(req *admissionv1beta1.AdmissionRequest, resp *a
 		return err
 	}
 
-	metrics.Record(ctx, requestCountM.M(1))
-	// Convert time.Duration in nanoseconds to milliseconds
-	metrics.Record(ctx, responseTimeInMsecM.M(float64(d/time.Millisecond)))
+	metrics.RecordBatch(ctx, requestCountM.M(1),
+		// Convert time.Duration in nanoseconds to milliseconds
+		responseTimeInMsecM.M(float64(d/time.Millisecond)))
```
evankanderson: It's strange that we're using a "float" here but our units are 1/1000 of a "second".

Probably not a fix for this PR, but it makes me slightly sad.

vagababov: yeah. changed to d.Milliseconds() anyway.

controller/stats_reporter.go
vagababov: Done.

PR #1030, opened 2020-01-31
add KnativeReference

vaikas:
Introduce a KnativeReference resource to replace the existing uses of ObjectReference. Motivation for the change:
https://github.com/kubernetes/kubernetes/pull/87459

And partly motivated by this as well:
https://github.com/knative/eventing/issues/2470

I'll follow up with another PR to replace ObjectReference with this object. This should have no user visible changes, since the wire format is not changing.


apis/duck/v1/knative_reference_test.go
```diff
@@ -0,0 +1,103 @@
+/*
+Copyright 2019 The Knative Authors
```
mattmoor: Incorrect copyright year:
```suggestion
Copyright 2020 The Knative Authors
```

apis/duck/v1/knative_reference.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Name of the referent.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string `json:"name,omitempty"`
+
+	// API version of the referent.
+	APIVersion string `json:"apiVersion,omitempty"`
+}
+
+func (kr *KnativeReference) Validate(ctx context.Context) *apis.FieldError {
```
n3wscott: Can you also add a default and a method to inject the current object's namespace into the context? It can be a follow up

vaikas: I can add the method to inject it from an object, but I'm unclear what you think the default method should do?

PR #1033, opened 2020-02-01
Add extra flags for cluster creation in test

chizhg:
Support `--enable-workload-identity` and `--service-account` for cluster creation request.
I add them in the gke lib as all the tools built on it will automatically support these two flags.

/cc @chaodaiG

### DIFF
--- a/test/gke/request.go
+++ b/test/gke/request.go
@@ -18,18 +18,11 @@ package gke
 
 import (
 	"errors"
-	"flag"
 
 	container "google.golang.org/api/container/v1beta1"
 )
 
 const defaultGKEVersion = "latest"
-
-// Extra configurations we want to support for cluster creation request.
-var (
-	enableWorkloadIdentity = flag.Bool("enable-workload-identity", false, "whether to enable Workload Identity")
-	serviceAccount = flag.String("service-account", "", "service account that will be used on this cluster")
-)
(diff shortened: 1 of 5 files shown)


test/gke/request.go
```diff
@@ -18,12 +18,19 @@ package gke
...
+	"flag"
 
 	container "google.golang.org/api/container/v1beta1"
 )
 
 const defaultGKEVersion = "latest"
 
+// Extra configurations we want to support for cluster creation request.
+var (
+	enableWorkloadIdentity = flag.Bool("enable-workload-identity", false, "whether to enable Workload Identity")
```
chaodaiG: Is there any specific reason why the flags have to be parsed here instead of from the caller? Is this going to be consumed directly by something else?

chaodaiG: Might have asked the wrong question, just not sure if defining flags here is a good thing or not, since the indirect consumer might not have the knowledge of this unless it's documented in the README of consumers

chizhg: I think you are right. Moved the flags out of request.go, and define them in perf-tests cluster manager.
Note I'm not defining them in main.go as it would be verbose to pass them as parameters, and I think it makes sense to define them in cluster.go as the two flags are only used in cluster creation.

PR #1043, opened 2020-02-04
Add GetGroupVersionKind and Selector WithPod

savitaashture:
**Background:**
as per the comments [1](https://github.com/knative/serving-operator/pull/281#pullrequestreview-352138634) and [2](https://github.com/knative/serving-operator/pull/281#pullrequestreview-352373707) usage of `PodSpecable` duck type helps to reduce lot of duplicate code

**Issue:**
To use [duck](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L38) type for Deployment, DaemonSet, StatefulSet etc... `Selector` field is mandatory and which is missing in [WithPod](https://github.com/knative/pkg/blob/master/apis/duck/v1/podspec_types.go#L38) and also `GetGroupVersionKind` implementation


apis/duck/v1/podspec_types.go
```diff
@@ -44,7 +45,8 @@ type WithPod struct {
 
 // WithPodSpec is the shell around the PodSpecable within WithPod.
 type WithPodSpec struct {
-	Template PodSpecable `json:"template,omitempty"`
+	Selector *metav1.LabelSelector `json:"selector"`
```
markusthoemmes: This would break compatibility with our own PodSpecables, like Service and Configuration.

apis/duck/v1/podspec_types.go
```diff
@@ -82,6 +84,11 @@ func (*WithPod) GetListType() runtime.Object {
 	return &WithPodList{}
 }
 
+// GetGroupVersionKind returns a GroupVersionKind.
+func (*WithPod) GetGroupVersionKind() schema.GroupVersionKind {
```
markusthoemmes: Not sure if this is the right thing to do here as `WithPod` doesn't really exist. Should this read the actual GroupVersionKind from the typeMeta?

apis/duck/v1/podspec_types.go
savitaashture: If that's the case then we need to write new one because for controllers selector is mandatory

markusthoemmes: See my comment on the PR. I think we shouldn't actually reserialize from the PodSpecable as that's a lossy conversion (as you noticed).

Instead, we should only override the relevant part of the unstructured that's been updated and part of PodSpecable

savitaashture:
closing [PR](https://github.com/knative/serving-operator/pull/281) because it may break compatibility with `PodSpecables` by adding selector support

PR #1045, opened 2020-02-04
Add support for client TLS to pkg/metrics

evankanderson:
This separates out the TLS support (which is finished) in #998 from the Resource support (which is not).

I can try to write a fake OpenCensus gRPC endpoint to test the TLS handshake if you want; I didn't do this yet because I got ratholed on the Resource support.

/assign @anniefu 
/assign @vagababov


evankanderson:
Looks like I do need test coverage to be able to merge this.

Tests incoming in the next day or so.

metrics/opencensus_exporter.go
```diff
@@ -36,3 +45,36 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
+		},
+	})
+}
+
+func certificateFetcher(secretName string, lister clientv1.SecretLister) (tls.Certificate, error) {
+	secret, err := lister.Secrets(system.Namespace()).Get(secretName)
+	if err != nil {
+		return tls.Certificate{}, err
+	}
+	return tls.X509KeyPair(secret.Data["server-cert.pem"], secret.Data["server-key.pem"])
```
anniefu: [nit] consts?

metrics/exporter.go
```diff
@@ -65,17 +66,31 @@ type ExporterOptions struct {
...
 	ConfigMap map[string]string
+
+	// A lister for Secrets to allow dynamic configuration of outgoing TLS client cert.
+	Secrets clientv1.SecretLister `json:"-"`
 }
 
 // UpdateExporterFromConfigMap returns a helper func that can be used to update the exporter
 // when a config map is updated.
+// DEPRECATED: Callers should migrate to UpdateExporterFromConfigMap3 (which will be renamed
+// when all callers are migrated).
```
anniefu: Is the rename of UpdateExporterFromConfigMap3 necessary? Why not make it a slightly more descriptive name and then leave it as-is going forward?

metrics/opencensus_exporter.go
evankanderson: I'm not sure if there's a benefit to extracting these. I can do it if it seems important.

metrics/exporter.go
evankanderson: How about "ConfigMapWatcher"? I've updated this PR.

metrics/testdata/README.md
```diff
@@ -0,0 +1,14 @@
+# Test files for metrics
+
+The cert files were generated with:
+
+```shell
+openssl req -x509 -nodes -newkey dsa:<(openssl dsaparam 1024) -keyout client-key.pem -out client-cert.pem -days 10000
```
mattmoor: Remove trailing whitespace:
```suggestion
openssl req -x509 -nodes -newkey dsa:<(openssl dsaparam 1024) -keyout client-key.pem -out client-cert.pem -days 10000
```

metrics/testdata/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+```shell
+openssl req -x509 -nodes -newkey dsa:<(openssl dsaparam 1024) -keyout client-key.pem -out client-cert.pem -days 10000 
+```
+
+Note that there are some manual prompts later in the process. This seemed simpler than generating the certs in Go.
+
+To view the cert:
+```shell
+openssl x509 -noout -text -in client-cert.pem
```
mattmoor: Remove trailing whitespace:
```suggestion
openssl x509 -noout -text -in client-cert.pem
```

evankanderson:
I added two tests for OpenCensus, one without certs, and one with client and server certs (though it doesn't test actual metrics delivery to the other end, only setup and configuration).

metrics/opencensus_exporter.go
anniefu: I've found it's typically useful to export consts from knative/pkg for the other knative repo's references.

anniefu:
Small nit on exporting the TLS strings as consts for the sake of other Knative repos. But otherwise,
/lgtm

PR #1050, opened 2020-02-05
Add support for KUBECONFIG env var in the tests

duglin:
Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -64,11 +64,17 @@ func initializeFlags() *EnvironmentFlags {
 	flag.StringVar(&f.Cluster, "cluster", "",
 		"Provide the cluster to test against. Defaults to the current cluster in kubeconfig.")
 
-	var defaultKubeconfig string
-	if usr, err := user.Current(); err == nil {
-		defaultKubeconfig = path.Join(usr.HomeDir, ".kube/config")
+	// Use KUBECONFIG if available
+	defaultKubeconfig := os.Getenv("KUBECONFIG")
+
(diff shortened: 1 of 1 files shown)


lionelvillard:
I was waiting for this :-) !

/lgtm

test/e2e_flags.go
```diff
@@ -64,11 +64,17 @@ func initializeFlags() *EnvironmentFlags {
...
-		defaultKubeconfig = path.Join(usr.HomeDir, ".kube/config")
+	// Use KUBECONFIG if available
+	defaultKubeconfig := os.Getenv("KUBECONFIG")
+
+	// If KUBECONFIG env var isn't set then look for $HOME/.kube/config
+	if defaultKubeconfig == "" {
+		if usr, err := user.Current(); err == nil {
+			defaultKubeconfig = path.Join(usr.HomeDir, ".kube/config")
+		}
 	}
```
chaodaiG: nit, felt a log here would be helpful for debugging purpose

chaodaiG:
/approve
/hold

Feel free ignoring the nit and unhold if don't agree with it

test/e2e_flags.go
duglin: ok done

chaodaiG: Actually I didn't realize there was a flag override before seeing your change. I'm not quite convinced that the flag was parsed at the point of logging, so it might not even be doable here. Sorry for misleading

PR #1074, opened 2020-02-11
Modify TestLogger util function to support for Benchmark testing

savitaashture:
As part of [PR](https://github.com/knative/serving/pull/6807)

### DIFF
--- a/logging/testing/util.go
+++ b/logging/testing/util.go
@@ -36,8 +36,10 @@ func TestLogger(i interface{}) *zap.SugaredLogger {
 	switch v := i.(type) {
 	case *testing.B:
 		return zaptest.NewLogger(v, opts).Sugar()
+	case *testing.T:
+		return zaptest.NewLogger(v, opts).Sugar()
 	}
-	return zaptest.NewLogger(&testing.T{}, opts).Sugar()
+	return &zap.SugaredLogger{}
 }
 
 // ClearAll removes all the testing loggers.
--- a/logging/testing/util.go
+++ b/logging/testing/util.go
@@ -27,13 +27,17 @@ import (
 )
 
 // TestLogger gets a logger to use in unit and end to end tests
-func TestLogger(t *testing.T) *zap.SugaredLogger {
+func TestLogger(i interface{}) *zap.SugaredLogger {
 	opts := zaptest.WrapOptions(
 		zap.AddCaller(),
 		zap.Development(),
 	)
 
-	return zaptest.NewLogger(t, opts).Sugar()
+	switch v := i.(type) {
+	case *testing.B:
+		return zaptest.NewLogger(v, opts).Sugar()
+	}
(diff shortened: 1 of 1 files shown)


logging/testing/util.go
```diff
@@ -27,13 +27,17 @@ import (
...
 		zap.AddCaller(),
 		zap.Development(),
 	)
 
-	return zaptest.NewLogger(t, opts).Sugar()
+	switch v := i.(type) {
+	case *testing.B:
+		return zaptest.NewLogger(v, opts).Sugar()
+	}
+	return zaptest.NewLogger(&testing.T{}, opts).Sugar()
```
vagababov: this seems wrong, shouldn't it be `i.(*testing.T)`?

savitaashture: `i.(*testing.T)` and `&testing.T{}` both works

i dint use `i.(*testing.T)` because again need to add condition check in order to make sure its of type `testing.T` 
```
v, ok := i.(*testing.T)
```
and here as we know TestLogger can get type `testing.T` or `testing.B` so used directly

vagababov: but you're replacing the actual testing.T object with a new one. The old one had its state.

savitaashture: Ah okay modified thank you

logging/testing/util.go
```diff
@@ -27,13 +27,19 @@ import (
...
 		zap.AddCaller(),
 		zap.Development(),
 	)
 
-	return zaptest.NewLogger(t, opts).Sugar()
+	switch v := i.(type) {
+	case *testing.B:
+		return zaptest.NewLogger(v, opts).Sugar()
+	case *testing.T:
+		return zaptest.NewLogger(v, opts).Sugar()
```
markusthoemmes: Should we panic on default? That should never happen, right?

savitaashture: Currently  all e2e, UT, benchmark etc... pass `testing.B` or `testing.T` only
But still have written empty object at the end if both case fails

markusthoemmes: I'd let it panic 🤔. Nothing else is supposed to reach here and we have a bug in the tests if it does.

logging/testing/util.go
```diff
@@ -27,13 +27,19 @@ import (
 )
 
 // TestLogger gets a logger to use in unit and end to end tests
-func TestLogger(t *testing.T) *zap.SugaredLogger {
+func TestLogger(i interface{}) *zap.SugaredLogger {
```
markusthoemmes: You can actually just change this to `t zaptest.TestingT`. No need for type assertions then :).

markusthoemmes: Can we add this to `TestContextWithLogger` for completeness as well?

savitaashture: Sure 
Thank you :+1:

Issue #1060, opened 2020-02-06
duck.CreatePatch cannot handle "remove" properly in some cases


yolocs:
## Problem

https://github.com/knative/pkg/blob/master/apis/duck/patch.go uses `github.com/mattbaird/jsonpatch` which cannot handle json object removal properly and is no longer maintained upstream.

For example if I two podspec:

```
	p1 := &duckv1.WithPod{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "foo",
			Namespace: "foo-ns",
		},
		Spec: duckv1.WithPodSpec{
			Template: duckv1.PodSpecable{
				Spec: corev1.PodSpec{
					Containers: []corev1.Container{
						{
							Name:    "curl-1",
							Image:   "governmentpaas/curl-ssl",
							Command: []string{"/bin/sleep", "3650d"},
							Env: []corev1.EnvVar{
								{
									Name:  "KEY1",
									Value: "VAL1",
								},
							},
						},
						{
							Name:    "curl-2",
							Image:   "governmentpaas/curl-ssl",
							Command: []string{"/bin/sleep", "3650d"},
							Env: []corev1.EnvVar{
								{
									Name:  "KEY2",
									Value: "VAL2",
								},
							},
						},
					},
				},
			},
		},
	}

	p2 := &duckv1.WithPod{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "foo",
			Namespace: "foo-ns",
		},
		Spec: duckv1.WithPodSpec{
			Template: duckv1.PodSpecable{
				Spec: corev1.PodSpec{
					Containers: []corev1.Container{
						{
							Name:    "curl-1",
							Image:   "governmentpaas/curl-ssl",
							Command: []string{"/bin/sleep", "3650d"},
							Env: []corev1.EnvVar{
								{
									Name:  "KEY3",
									Value: "VAL3",
								},
							},
						},
					},
				},
			},
		},
	}
```

## Expected Behavior

I expect patch generated is something like

```
[
  {
    "op": "remove",
    "path": "/spec/template/spec/containers/1"
  },
  {
    "op": "replace",
    "path": "/spec/template/spec/containers/0/env",
    "value": [
      {
        "name": "KEY3",
        "value": "VAL3"
      }
    ]
  }
]
```

## Actual Behavior

Actual patch

```
[
  {
    "op": "remove",
    "path": "/spec/template/spec/containers/0"
  },
  {
    "op": "remove",
[...]

PR #1078, opened 2020-02-11
Update jsonpatch lib that correctly handle object removal

yolocs:
To fix: https://github.com/knative/pkg/issues/1060
Interestingly, the new lib is a fork based on the original jsonpatch lib. So there is much code change introduced!


yolocs:
@mattmoor

apis/duck/patch.go
```diff
@@ -20,7 +20,7 @@ import (
 	"encoding/json"
 
 	jsonmergepatch "github.com/evanphx/json-patch"
-	"github.com/mattbaird/jsonpatch"
+	jsonpatch "gomodules.xyz/jsonpatch/v2"
```
mattmoor: For posterity, this is a fork of what we were using: https://github.com/gomodules/jsonpatch

yolocs: Wonder what made you drop it?

PR #1080, opened 2020-02-12
allow filtering on schema.GroupKind

dprotaso:
In addition deprecated usage of `Filter` with the introduction of `FilterGroupVersionKind`

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -86,13 +86,44 @@ func HandleAll(h func(interface{})) cache.ResourceEventHandler {
 // Filter makes it simple to create FilterFunc's for use with
 // cache.FilteringResourceEventHandler that filter based on the
 // schema.GroupVersionKind of the controlling resources.
+//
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -98,6 +107,28 @@ func Filter(gvk schema.GroupVersionKind) func(obj interface{}) bool {
...
+// schema.GroupKind of the controlling resources.
+func FilterGroupKind(gk schema.GroupKind) func(obj interface{}) bool {
+	return func(obj interface{}) bool {
+		if object, ok := obj.(metav1.Object); ok {
+			owner := metav1.GetControllerOf(object)
+			if owner == nil {
+				return false
+			}
+
+			ownerGV, err := schema.ParseGroupVersion(owner.APIVersion)
```
vagababov: since you're ignoring error, you can just `return err==nil && ownerGV.Group...`

PR #1081, opened 2020-02-12
Auto set observed generation

vaikas:
I can't think of a single case where we wouldn't want to do this for our resources and can think of many cases where it's been a source of bugs.

The one reason why we wouldn't want to do this is that if this is reconciling objects that do not expose the Status.ObservedGeneration field however. Since have that in stub however as well, seemed like the intent is that the resources will have these fields, but I can see the stub breaking is less of an issue than the generated one.

If I had enough template fu, maybe we could look and if the type implements this will we do it.

But in any case, just wanted to throw it out there as a thought.

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -236,6 +236,10 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
 		}
 	}
 
+        // Since the Reconciler observed this object, update the Status.ObservedGeneration
+        // to match the objects ObservedGeneration.
+        resource.Status.ObservedGeneration = original.Generation
+
 	// Synchronize the status.
 	if equality.Semantic.DeepEqual(original.Status, resource.Status) {
 		// If we didn't change anything then don't call updateStatus.
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler_stub.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler_stub.go
@@ -115,7 +115,6 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, o *{{.type|raw}}) {{.rec
 
 	// TODO: add custom reconciliation logic here.
 
-	o.Status.ObservedGeneration = o.Generation
 	return newReconciledNormal(o.Namespace, o.Name)
 }
 


dprotaso:
cc @dgerd related: https://github.com/knative/serving/issues/4937

n3wscott:
The reason I did not do this is there is nothing that changes the resource for you in the generated reconcilers so that means you can use this as a read only observer type like my graph program or as a first class reconciler like w use in Knative

I don't think we should do with without a way to turn it off

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -236,6 +236,10 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
 		}
 	}
 
+        // Since the Reconciler observed this object, update the Status.ObservedGeneration
+        // to match the objects ObservedGeneration.
+        resource.Status.ObservedGeneration = original.Generation
```
mattmoor: I don't think that this alone is enough.  If `ReconcileKind` returned an error before updating `Status` then we might inadvertently convey readiness before we mean to by auto-bumping ObsGen before the conditions are (successfully) updated.  I would be supportive of doing this whenever the return code is not an error 🤷‍♂

mattmoor: @n3wscott has a good point too, we [will] use this for a couple resources in Serving that don't touch status, and shouldn't update this.  I'd be ok with having a way of turning all the status stuff off.

mattmoor: e.g. https://github.com/knative/serving/pull/6813

vaikas:
@n3wscott I think for that case, having a:
ObserveKind method might be more appropriate, but shrug :)

@mattmoor I don't quite understand in which case you should not still bump the observed generation even in the case of error? Each of the uses does the same two lines at the beginning of their reconcilekind:
initializeconditions
set observedgeneration

I'm totes fine for making it optional and not super wedded to it, but given the long history of bugs that have been done here. It's my understanding that ObservedGeneration is meant to convey to the user that a reconciler has taken a look and I don't see how that would convey or be any different from how it's being used in all the reconcilers that are being migrated to the new generic reconciler.

Again, not super wedded just something that thought might be useful.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
dgerd: > we might inadvertently convey readiness before we mean to by auto-bumping ObsGen before the conditions are (successfully) updated. 

I am not sure I follow this. Both this and the status of conditions are updated in `updateStatus`. No? This does however mean that reconcilers have to follow the contract of always setting Conditions before returning.

>  I'd be ok with having a way of turning all the status stuff off.

Yes i agree with this.

dgerd:
> The reason I did not do this is there is nothing that changes the resource for you in the generated reconcilers so that means you can use this as a read only observer type like my graph program or as a first class reconciler like w use in Knative

There is some talk in various K8s issues around multiple observedGenerations for situations when you have multiple controllers working on the same object. Knowing the observed generation is valuable for read-only reconcilers or autoscaling-like reconcilers, but I agree that clobbering the 'primary' controllers observedGeneration is not the right approach. I do not think K8s offers guidance here yet.

Issue #1007, opened 2020-01-23
[Feature Track] Active/passive HA


pmorie:
This issue is for the active/passive HA feature track, document [here](https://docs.google.com/document/d/1zHtVjVmzivO2PEhZkpqNo-M9cPnmfvBWMYzP8ehVXtM/edit#).

/kind feature

pmorie:
Haha, we're not done. Next steps here:

- Determine how to horizontally scale webhooks
- Determine whether to separate reconcilers and webhooks for the same resource
- #1048

vagababov:
Do I miss something about Webhooks just being naturally Horizontally scalable by virtue of being stateless (well modulo certs)

pmorie:
@vagababov The webhook parts are naturally horizontally scalable. We have adopted an idiomatic architecture of running the reconcilers and webhooks for resources in the same process. We decided to introduce the additional main methods to insulate webhook binaries from the leader election changes for now. Personally, I would rather migrate the reconcilers into the main controllers and disentangle them from the webhooks; we skipped doing it initially because we didn't easily align in a few beats of discussion.

vagababov:
Ah, I didn't know you run them combined :)

PR #1084, opened 2020-02-12
Introduce sharedmain.WebhookMain*

pmorie:
WebhookMain* now implements the old main flow - while Main* no longer serves webhooks. This allows us to retain the existing behavior for webhooks - which require additional investigation / decision making on how to deal with HA concerns.

Related to #1007

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -93,120 +93,113 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
 	return logging.NewConfigFromConfigMap(loggingConfigMap)
 }
 
+// Main runs the generic main flow for non-webhook controllers with a new
+// context. Use WebhookMainWith* if you need to serve webhooks.
 func Main(component string, ctors ...injection.ControllerConstructor) {
 	// Set up signals so we handle the first shutdown signal gracefully.
 	MainWithContext(signals.NewContext(), component, ctors...)
 }
 
+// MainWithContext runs the generic main flow for non-webhook controllers. Use
+// WebhookMainWithContext if you need to serve webhooks.
 func MainWithContext(ctx context.Context, component string, ctors ...injection.ControllerConstructor) {
-	var (
-		masterURL = flag.String("master", "",
-			"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
-		kubeconfig = flag.String("kubeconfig", "",
-			"Path to a kubeconfig. Only required if out-of-cluster.")
-	)
-	flag.Parse()
-
-	cfg, err := GetConfig(*masterURL, *kubeconfig)
-	if err != nil {
-		log.Fatalf("Error building kubeconfig: %v", err)
-	}
-	MainWithConfig(ctx, component, cfg, ctors...)
+	MainWithConfig(ctx, component, ParseAndGetConfigOrDie(), ctors...)
 }
 
+// MainWithConfig runs the generic main flow for non-webhook controllers. Use
+// WebhookMainWithConfig if you need to serve webhooks.
 func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, ctors ...injection.ControllerConstructor) {
 	log.Printf("Registering %d clients", len(injection.Default.GetClients()))
 	log.Printf("Registering %d informer factories", len(injection.Default.GetInformerFactories()))
 	log.Printf("Registering %d informers", len(injection.Default.GetInformers()))
 	log.Printf("Registering %d controllers", len(ctors))
 
-	// Report stats on Go memory usage every 30 seconds.
-	msp := metrics.NewMemStatsAll()
-	msp.Start(ctx, 30*time.Second)
+	MemStatsOrDie(ctx)
 
-	if err := view.Register(msp.DefaultViews()...); err != nil {
-		log.Fatalf("Error exporting go memstats view: %v", err)
-	}
-
-	// Adjust our client's rate limits based on the number of controller's we are running.
+	// Adjust our client's rate limits based on the number of controllers we are running.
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -93,74 +93,128 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
...
+	// Based on the reconcilers we have linked, build up the set of controllers to run.
+	controllers := make([]*controller.Impl, 0, len(ctors))
+	for _, cf := range ctors {
+		ctrl := cf(ctx, cmw)
+		controllers = append(controllers, ctrl)
 	}
 
+	profilingHandler := profiling.NewHandler(logger, false)
+
+	// Watch the logging config map and dynamically update logging levels.
```
vaikas: maybe move this  and observability die above setting up the controllers so it's grouped by where the pieces necessary to construct it are set up? My reasoning being that if all the dies are as early as possible, just makes it easier to reason about the code?
Again, not a big deal :)

injection/sharedmain/main.go
```diff
@@ -245,3 +287,89 @@ func flush(logger *zap.SugaredLogger) {
...
+
+// SetupLoggerOrDie sets up the logger using the config from the given context
+// and returns a logger and atomic level, or dies by calling log.Fatalf.
+func SetupLoggerOrDie(ctx context.Context, component string) (*zap.SugaredLogger, zap.AtomicLevel) {
+	loggingConfig, err := GetLoggingConfig(ctx)
+	if err != nil {
+		log.Fatalf("Error reading/parsing logging configuration: %v", err)
+	}
+	logger, atomicLevel := logging.NewLoggerFromConfig(loggingConfig, component)
+	ctx = logging.WithLogger(ctx, logger)
```
vaikas: You might want to document this in the function comments. I think the intent is that this will also modify the incoming context and attach a logger to it. I think it would be good to document that because it's not (to me) immediately clear that's what's happening. Alternatively, if the intent is for the caller to then create / decorate their context with the logger is to return above with:
return logging.NewLoggerFromConfig(loggingConfig, component).

injection/sharedmain/main.go
```diff
@@ -245,3 +287,89 @@ func flush(logger *zap.SugaredLogger) {
...
+	return cmw
+}
+
+// WatchLoggingConfigOrDie establishes a watch of the logging config or dies by
+// calling log.Fatalf.
+func WatchLoggingConfigOrDie(ctx context.Context, cmw *configmap.InformedWatcher, logger *zap.SugaredLogger, atomicLevel zap.AtomicLevel, component string) {
+	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(),
+		metav1.GetOptions{}); err == nil {
+		cmw.Watch(logging.ConfigMapName(), logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else if !apierrors.IsNotFound(err) {
```
vaikas: What happens if it is not found? Seems like it should also fail? Or put another way, is there a reason not to die on any error?

injection/sharedmain/main.go
```diff
@@ -245,3 +287,89 @@ func flush(logger *zap.SugaredLogger) {
...
+
+// WatchObservabilityConfigOrDie establishes a watch of the logging config or
+// dies by calling log.Fatalf.
+func WatchObservabilityConfigOrDie(ctx context.Context, cmw *configmap.InformedWatcher, profilingHandler *profiling.Handler, logger *zap.SugaredLogger, component string) {
+	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(metrics.ConfigMapName(),
+		metav1.GetOptions{}); err == nil {
+		cmw.Watch(metrics.ConfigMapName(),
+			metrics.UpdateExporterFromConfigMap(component, logger),
+			profilingHandler.UpdateFromConfigMap)
+	} else if !apierrors.IsNotFound(err) {
```
vaikas: ditto here, what happens if it fails for another reason?

injection/sharedmain/main.go
markusthoemmes: This is a copy of the code as it lives in the Main function today. This has been changed not to cause the component to die a while ago.

injection/sharedmain/main.go
markusthoemmes: Same here.

injection/sharedmain/main.go
markusthoemmes: I think the code as it stands is even wrong. `WithLogger` will create a new context that you have no access to in this case.

I agree, we should probably just do the `WithLogger` call right after calling `SetupLoggerOrDie` at the calling side.

injection/sharedmain/main.go
vaikas: Then I think at the very least we should clearly document what's going on. There are cases where this will NOT establish a watch nor die? Seems like if there's any reason (temporary can't reach), this will fail? Or maybe I'm just misunderstanding something :)

injection/sharedmain/main.go
pmorie: Whoops, good catch. I will fix

injection/sharedmain/main.go
markusthoemmes: See https://github.com/knative/pkg/pull/872 for the broader reasoning of that behavior. I agree we should add into the comment that this will not watch the logging config if it doesn't exist (but die on any other error)

injection/sharedmain/main.go
pmorie: Ack, I tried to preserve the existing flow as much as possible for lower review overhead, but reading it now it's a bit apart. I'll rejigger

vaikas:
/lgtm
Thanks!!

PR #1086, opened 2020-02-13
[GenRec] Keep resource in sync when setting and removing finalizers.

n3wscott:
Update the resource if it is patched for finalizer mods to let status update work the first try.

/assign @mattmoor


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -248,7 +248,7 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 
 		// Set and update the finalizer on resource if r.reconciler
 		// implements Finalizer.
-		if err := r.setFinalizerIfFinalizer(ctx, resource); err != nil {
+		if resource, err := r.setFinalizerIfFinalizer(ctx, resource); err != nil {
```
mattmoor:
```suggestion
		var err error
		if resource, err = r.setFinalizerIfFinalizer(ctx, resource); err != nil {
```

PR #1097, opened 2020-02-14
[Genreconciler] Fix group names when client groups differ than their directory name.

n3wscott:
Integrating with Tekton, we found an issue where the generated reconciler called out to the wrong groupVersion from the client because that reconciler was from a directory named differently than the target api group (we do not do this in Knative in general).

This will correct that.

Also adding two sample apis in `apis/test` that are generated into `client/test` to be able to compile the output of the generators from within `pkg`.


hack/update-test-codegen.sh
```diff
@@ -0,0 +1,52 @@
...
+KNATIVE_CODEGEN_PKG=${KNATIVE_CODEGEN_PKG:-$(cd ${REPO_ROOT_DIR}; ls -d -1 $(dirname $0)/../vendor/knative.dev/pkg 2>/dev/null || echo ../pkg)}
+
+(
+  cd ${REPO_ROOT_DIR}
+  rm -rf ${REPO_ROOT_DIR}/knative.dev/pkg/apis/test
+)
+
+${CODEGEN_PKG}/generate-groups.sh "deepcopy,client,informer,lister" \
+  knative.dev/pkg/client/test knative.dev/pkg/apis/test \
+  "example:v1alpha1" \
```
mattmoor: You can combine these invocations `"example:v1alpha1 pub:v1alpha1"`

mattmoor: In fact, I think not combining them is a bug (unless you output to different test directories)

On that topic, checking the output in will result in our vendoring it into all downstream repos.  I appreciate you adding examples that we can use as tests, but could I bother you a little more to move this so that it's only generates by presubmit (with a `go test` sniff test)?

n3wscott: You can't. It puts them in the same client interface and the generators explode from k8s

n3wscott: IMG-20200214-WA0000

mattmoor: Yeah, but run in sequence I'd bet that `samplev1alpha1.SampelV1alpha1Client` only contains one of them.  Can we change the groupName of the examples to not so overtly collide?  I care more about the `go test` comment and not vendoring this TBH.

n3wscott: I think if you target the same dir twice it will not have the first client in the clientset. I think this is a bug in k8s code generators in general, or working as intended and you should combine the client group into a single dir, or do it like this and target two dirs

n3wscott: You can checkout this fork and see what I am talking about. I wanted to do what you are asking but the generators for client did not like it at all

mattmoor: I missed that you were targeting separate dirs (sips ☕️ ) can we avoid committing the result but still get test coverage?

n3wscott: Another option is to name the second client group like SamplePub with ` // +groupGoName=SamplePub,,`

n3wscott: That lets you gen them in the same dir and at the same time without conflict

n3wscott: The point of checking in the gen code is that if needs to compile. I bet we could add a gitignore for client/test, that feels a little better...

n3wscott:
/hold

I will make it so the generated code is put into a temp, compiled, and then deleted as part of a test

.gitignore
```diff
@@ -12,3 +12,6 @@ bazel-*
 
 # Emacs files
 *~
+
+# Generated Test Code
+test/genclient
```
mattmoor: Add trailing newline:
```suggestion
test/genclient

```

n3wscott:
Flake? 

```
workqueue_test.go:114: Wrong value metric adds value 3 want 2
```

/retest

PR #1101, opened 2020-02-18
Renews a webhook cert before it expires

mpetason:
This will remove the webhook secret(webhook-certs) so that a new secret is created if the NotAfter/Expiration date is less than a week.

### DIFF
--- a/webhook/certificates/certificates.go
+++ b/webhook/certificates/certificates.go
@@ -31,6 +31,11 @@ import (
 	certresources "knative.dev/pkg/webhook/certificates/resources"
 )
 
+const (
+	// Time used for updating a certificate before it expires. Default 1 week.
+	oneWeek = 168 * time.Hour
+)
+
 type reconciler struct {
 	client       kubernetes.Interface
 	secretlister corelisters.SecretLister
@@ -70,18 +75,18 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
 	} else if _, haskey := secret.Data[certresources.CACert]; !haskey {
 		logger.Infof("Certificate secret %q is missing key %q", r.secretName, certresources.CACert)
 	} else {
-		// It has all of the keys, it's good.
+		// Check the expiration date of the certificate to see if it needs to be updated
 		cert, err := tls.X509KeyPair(secret.Data[certresources.ServerCert], secret.Data[certresources.ServerKey])
 		if err != nil {
-			logger.Infof("error creating pem from certificate and key")
-		}
-		certData, err := x509.ParseCertificate(cert.Certificate[0])
-		if err != nil {
-			logger.Infof("error parsing certificate: ", err.Error)
-		}
-		oneWeek := 168 * time.Hour
-		if certData.NotAfter.Sub(time.Now()) > oneWeek {
-			return nil
+			logger.Errorf("error creating pem from certificate and key: %v", err)
+		} else {
+			certData, err := x509.ParseCertificate(cert.Certificate[0])
+			if err != nil {
+				logger.Errorf("error parsing certificate: %v", err)
+			}
+			if certData.NotAfter.Sub(time.Now()) > oneWeek {
+				return nil
+			}
 		}
 	}
 	// Don't modify the informer copy.
--- a/webhook/certificates/certificates.go
+++ b/webhook/certificates/certificates.go
@@ -33,7 +33,7 @@ import (
 
 const (
 	// Time used for updating a certificate before it expires. Default 1 week.
-	oneWeek = 168 * time.Hour
+	oneWeek = 7 * 24 * time.Hour
 )
 
 type reconciler struct {
@@ -78,13 +78,12 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
 		// Check the expiration date of the certificate to see if it needs to be updated
 		cert, err := tls.X509KeyPair(secret.Data[certresources.ServerCert], secret.Data[certresources.ServerKey])
 		if err != nil {
-			logger.Errorf("error creating pem from certificate and key: %v", err)
+			logger.Warnf("Error creating pem from certificate and key: %v", err)
 		} else {
 			certData, err := x509.ParseCertificate(cert.Certificate[0])
 			if err != nil {
-				logger.Errorf("error parsing certificate: %v", err)
-			}
-			if certData.NotAfter.Sub(time.Now()) > oneWeek {
+				logger.Errorf("Error parsing certificate: %v", err)
+			} else if time.Now().Add(oneWeek).Before(certData.NotAfter) {
 				return nil
 			}
 		}
--- a/webhook/certificates/certificates.go
+++ b/webhook/certificates/certificates.go
@@ -18,6 +18,9 @@ package certificates
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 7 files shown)


mpetason:
Definitely looking for feedback on this as it is my first Knative PR. 

Issue this addresses - https://github.com/knative/serving/issues/6537

webhook/webhook.go
```diff
@@ -164,6 +167,15 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
 				if err != nil {
 					return nil, err
 				}
+				certData, err := x509.ParseCertificate(cert.Certificate[0])
+				if err != nil {
+					logger.Info("failed to parse certificate: " + err.Error())
+				}
+				oneWeek := 168 * time.Hour
+				if certData.NotAfter.Sub(time.Now()) < oneWeek {
+					wh.Client.CoreV1().Secrets(system.Namespace()).Delete(secret.Name, &metav1.DeleteOptions{})
```
mattmoor: The secret isn't created by the reconciler, so this will not work as intended.

I'd change here: https://github.com/knative/pkg/blob/1cc7b7152937735a29c9d6fdaf3b244684aa850e/webhook/certificates/certificates.go#L70

mpetason: That makes sense. I was actually able to delete the secret so many times that it just didn't come back and couldn't figure out how to trigger it to be created again. I'll look into this and make updates.

mattmoor: I just wrote a surprisingly similar check for auto-TLS here: https://github.com/mattmoor/http01-solver/blob/4f7f1732d30512a74cbeb5d88341dfeb26c546ba/pkg/reconciler/certificate/certificate.go#L70

dereferencing that here: https://github.com/mattmoor/http01-solver/blob/4f7f1732d30512a74cbeb5d88341dfeb26c546ba/pkg/reconciler/certificate/resources/secret.go#L62-L66

The logic should be set up in a way so that if you just add this check things should automatically refresh the cert, we just only do it for malformed certs today.

webhook/certificates/certificates.go
```diff
@@ -68,7 +71,18 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
 		logger.Infof("Certificate secret %q is missing key %q", r.secretName, certresources.CACert)
 	} else {
 		// It has all of the keys, it's good.
-		return nil
+		cert, err := tls.X509KeyPair(secret.Data[certresources.ServerCert], secret.Data[certresources.ServerKey])
+		if err != nil {
+			logger.Infof("error creating pem from certificate and key")
```
mattmoor: We should include the error in what we log.

webhook/certificates/certificates.go
```diff
@@ -68,7 +71,18 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
...
+		}
```
mattmoor: everything below this "if" should be wrapped in an else block to avoid errors due to the failed condition.

webhook/certificates/certificates.go
```diff
@@ -68,7 +71,18 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
...
+		certData, err := x509.ParseCertificate(cert.Certificate[0])
+		if err != nil {
+			logger.Infof("error parsing certificate: ", err.Error)
```
mattmoor:
```suggestion
			logger.Infof("error parsing certificate: %v", err)
```

webhook/certificates/certificates.go
```diff
@@ -68,7 +71,18 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
...
+		}
+		oneWeek := 168 * time.Hour
```
mattmoor: I believe you can make this a const

webhook/certificates/certificates.go
mpetason: I'll update this, thank you for the suggestion.

webhook/certificates/certificates.go
mpetason: Will do. I'm open to suggestions to a shorter period of time as well, I just went with 1 week instead of 48 hours.

mpetason:
Making sure that the updates are correct before I write a test. The test might take me a little bit longer since I haven't done a lot with table tests.

webhook/certificates/certificates.go
```diff
@@ -67,8 +75,19 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
...
+		// Check the expiration date of the certificate to see if it needs to be updated
+		cert, err := tls.X509KeyPair(secret.Data[certresources.ServerCert], secret.Data[certresources.ServerKey])
+		if err != nil {
+			logger.Errorf("error creating pem from certificate and key: %v", err)
+		} else {
+			certData, err := x509.ParseCertificate(cert.Certificate[0])
+			if err != nil {
+				logger.Errorf("error parsing certificate: %v", err)
+			}
+			if certData.NotAfter.Sub(time.Now()) > oneWeek {
```
mattmoor: I think this should be an "else if" as well.  Otherwise this LGTM.  thanks!

webhook/certificates/certificates.go
```diff
@@ -28,6 +31,11 @@ import (
 	certresources "knative.dev/pkg/webhook/certificates/resources"
 )
 
+const (
+	// Time used for updating a certificate before it expires. Default 1 week.
+	oneWeek = 168 * time.Hour
```
vagababov: `default one week` is kind of strange here, since it's a constant.
I'd also write `7*24*time.hour`, which is more expressive.

webhook/certificates/certificates.go
```diff
@@ -67,8 +75,19 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
 	} else if _, haskey := secret.Data[certresources.CACert]; !haskey {
 		logger.Infof("Certificate secret %q is missing key %q", r.secretName, certresources.CACert)
 	} else {
-		// It has all of the keys, it's good.
-		return nil
+		// Check the expiration date of the certificate to see if it needs to be updated
+		cert, err := tls.X509KeyPair(secret.Data[certresources.ServerCert], secret.Data[certresources.ServerKey])
+		if err != nil {
+			logger.Errorf("error creating pem from certificate and key: %v", err)
```
vagababov:
```suggestion
			logger.Errorf("Error creating PEM from certificate and key: %v", err)
```
Logger errors are complete sentences.
Since we're ignoring them, should they rather be `Warnf`? Same below?

webhook/certificates/certificates.go
```diff
@@ -67,8 +75,19 @@ func (r *reconciler) reconcileCertificate(ctx context.Context) error {
...
+		} else {
+			certData, err := x509.ParseCertificate(cert.Certificate[0])
+			if err != nil {
+				logger.Errorf("error parsing certificate: %v", err)
```
vagababov:
```suggestion
				logger.Errorf("Error parsing certificate: %v", err)
```

webhook/certificates/certificates.go
vagababov: What about `time.Since(certData.NotAfter)`? Might need a `-oneWeek`.

mattmoor: I think that makes things less readable.  If I were going to rewrite this I'd probably put:
```
if time.Now().Add(oneWeek) < certData.NotAfter
```

vagababov:
Well. `Before(x)` :-) times in go somehow not comparable 😁

webhook/certificates/certificates.go
mpetason: Makes sense, so remove the comment about default and change from 168 to 7*24?

webhook/certificates/certificates.go
mpetason: Will do, wasn't sure which we should use. Both were suggested so I went with the same error type that was being used further up. I'll change over to Warnf.

webhook/certificates/certificates.go
mattmoor: I think leaving the comment is fine, but now it's more clearly accurate :D

webhook/certificates/certificates.go
vagababov: Error will print stacktrace etc. 
Given you ignore the error, I'd go with Infof/Warnf. I have no preference which.

webhook/certificates/certificates.go
mpetason: Should I change this, or leave it how it is?

vagababov: I'd go with Matt's `NotAfter` is always hard to understand.

mattmoor:
This LGTM, let's get a test for the new paths and get it in.  Thanks!

mpetason:
Fixes knative/serving/issues/6537

Issue #1098, opened 2020-02-15
First-class support for class-based reconcilers.


mattmoor:
/area API
/kind bug
/kind feature

There is a bug today with `// +genreconciler`-based reconcilers that using finalizers and are class-based.  Essentially we have no way to keep the wrong class' reconciler from adding finalizers to resources they shouldn't handle or (worse!) removing them from resources they shouldn't handle.

We know in general which resources we define are going to be class-based, so my recommendation here would be to adopt the following solution.

```
// This is a normal type
// +genreconciler
...

// This is a class-based type
// +genreconciler:class=networking.knative.dev/ingress.class
...
```

The idea here is that we would tell the codegen infrastructure the name of the class annotation.  This would then influence the code-generator to add an additional argument to the appropriate constructors:

```
// Normal +genreconciler
fooreconciler.NewImpl(ctx, c)

// One with a class.
barreconciler.NewImpl(ctx, c, "my.class.value")
```

n3wscott:
Can we add that to the optional stuct?

n3wscott:
Or!! How about allow a options filter function callback to be passed back!!

mattmoor:
I think what I suggest above is harder to hold wrong because it requires a different signature iff it is class-based.  I think adding stuff to the options struct that is only used for class-based stuff gets confusing (what if I specify it when I'm not using a class-based type?!?).

n3wscott:
ok I like the args idea. I will give it a try

/assign

PR #1103, opened 2020-02-19
Implement strongly typed class reconcilers

n3wscott:
Fixes: https://github.com/knative/pkg/issues/1098

Adding a new annotation config to allow us to configure strongly classed reconcilers, such as Eventing's Broker, or Serving's Ingress.

This will be a breaking API change for types that opt into this behavior. The main changes are the addition of ` classValue string` to `NewImpl` and `NewReconciler` (for tests)

```go
NewImpl(ctx context.Context, r Interface, classValue string, optionsFns ...controller.OptionsFn) *controller.Impl
```

```go
NewReconciler(ctx context.Context, logger *zap.SugaredLogger, client versioned.Interface, lister pubv1alpha1.BarLister, recorder record.EventRecorder, r Interface, classValue string, options ...controller.Options) controller.Reconciler
```

This feature is enabled with a golang annotation argument:

```go
// +genreconciler:class=example.com/filter.class
```

### DIFF
--- a/codegen/cmd/injection-gen/generators/packages.go
+++ b/codegen/cmd/injection-gen/generators/packages.go
@@ -144,7 +144,6 @@ type Tags struct {
 
 	GenerateDuck       bool
 	GenerateReconciler bool
-	ClassFilter        *string
 }
 
 func (t Tags) NeedsInformerInjection() bool {
--- a/codegen/cmd/injection-gen/generators/packages.go
+++ b/codegen/cmd/injection-gen/generators/packages.go
@@ -144,6 +144,7 @@ type Tags struct {
 
 	GenerateDuck       bool
 	GenerateReconciler bool
+	ClassFilter        *string
 }
 
 func (t Tags) NeedsInformerInjection() bool {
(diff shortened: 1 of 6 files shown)


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -191,6 +196,11 @@ type reconcilerImpl struct {
 
 	// reconciler is the implementation of the business logic of the resource.
 	reconciler Interface
+
+	{{if .hasClass}}
+	// classValue is the resource annotation[{{ .class }}] instance value this reconciler instance filters on.
```
mattmoor: Remove trailing whitespace:
```suggestion
	// classValue is the resource annotation[{{ .class }}] instance value this reconciler instance filters on.
```

mattmoor: Remove trailing whitespace:
```suggestion
	// classValue is the resource annotation[{{ .class }}] instance value this reconciler instance filters on.
```

codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -165,13 +165,26 @@ func MustParseClientGenTags(lines []string) Tags {
 	}
 
 	values := types.ExtractCommentTags("+", lines)
-	// log.Printf("GOT values %v", values)
+
 	_, ret.GenerateDuck = values["genduck"]
 	_, ret.GenerateReconciler = values["genreconciler"]
+	_, ret.GenerateReconciler = values["genreconciler:class"]
```
mattmoor: This will overwrite what's above 🤦‍♂

n3wscott: yeah that is fine and intended.

I did not want to require `// +genreconciler ` AND `// genreconciler:class=foo.bar`

n3wscott: note it is not the value, but the existence of that tag.

n3wscott: oooh wait wait... I need an or


fml

PR #1106, opened 2020-02-19
Revert #1045, causes sharedmain to need global secrets list

evankanderson:
This reverts commit 945b5567088aa7272584324665e428e3eb9b45da.

/assign @pmorie


pmorie:
Just a note, the revert is necessary because the existing change requires cluster-level permissions on secrets, which is a very high degree of privilege and incompatible with how we grant permissions to brokers currently.

evankanderson:
Thanks for noticing this, Paul!

Rolling back to make a clean fix without time pressure.

PR #1108, opened 2020-02-20
Clear unhappy propagation status of the top level condition if cleared.

n3wscott:
If you have a condition set: [Ready, Foo, Bar] the following happens:

Foo is marked unknown
Bar is marked unknown

Ready will be marked unknown with the Reason/Message of Bar.

Later, if Bar is marked true, the code would not clear the propagated old Bar status.

Now Foo's status will be propagated and Bar's status cleared off Ready.

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -242,23 +242,38 @@ func (r conditionsImpl) ClearCondition(t ConditionType) error {
 // MarkTrue sets the status of t to true, and then marks the happy condition to
 // true if all other dependents are also true.
 func (r conditionsImpl) MarkTrue(t ConditionType) {
-	// set the specified condition
+	// Save the original status of t.
+	org := r.GetCondition(t).DeepCopy()
+	orgTL := r.GetTopLevelCondition().DeepCopy()
+
+	// Set the specified condition.
 	r.SetCondition(Condition{
 		Type:     t,
 		Status:   corev1.ConditionTrue,
 		Severity: r.severity(t),
 	})
 
-	// check the dependents.
+	// Check the dependents.
 	for _, cond := range r.dependents {
 		c := r.GetCondition(cond)
-		// Failed or Unknown conditions trump true conditions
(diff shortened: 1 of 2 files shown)


apis/condition_set.go
```diff
@@ -242,23 +242,38 @@ func (r conditionsImpl) ClearCondition(t ConditionType) error {
...
 	})
 
-	// check the dependents.
+	// Check the dependents.
 	for _, cond := range r.dependents {
 		c := r.GetCondition(cond)
-		// Failed or Unknown conditions trump true conditions
+		// Failed or Unknown conditions trump true conditions.
 		if !c.IsTrue() {
+			// Update the happy condition if the current ready condition is
```
evankanderson: Why do this comparison, rather than always updating this?

apis/condition_set.go
```diff
@@ -242,23 +242,38 @@ func (r conditionsImpl) ClearCondition(t ConditionType) error {
...
+			// marked not ready because of this condition.
+			if org.Reason == orgTL.Reason && org.Message == orgTL.Message {
+				r.SetCondition(Condition{
+					Type:     r.happy,
+					Status:   c.Status,
+					Reason:   c.Reason,
+					Message:  c.Message,
+					Severity: r.severity(r.happy),
+				})
+			}
```
evankanderson: I'm not sure if this is right... I think we should be marking the top-level condition based on the min() of the conditions in the list, but this might mark the condition `Unknown` even if one of the dependents is `False`.

condition A = True
condition B = Unknown
condition C = False

I think this will early-exit on B and not mark the top-level as False.

apis/condition_set.go
n3wscott: because you want the last status to change bad to be reflected, not the first you find.

apis/condition_set.go
n3wscott: yeah I just thought of this... looking into it

PR #1109, opened 2020-02-20
Roll forward #1045

evankanderson:
I switched from calling `Get()` from a global secrets lister to a KubeClient Get in a namespace.

This fixes RBAC problems with the eventing InMemory Channel controller that manifest as:

    E0220 04:09:57.632572       1 reflector.go:123] runtime/asm_amd64.s:1357: Failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:knative-eventing:imc-controller" cannot list resource "secrets" in API group "" at the cluster scope

(The other option is to grant every Knative controller derived from SharedMain read/list access to all cluster Secrets, which seems excessive.

/assign @anniefu


evankanderson:
Merged with Paul's sharedmain change, needs a new LGTM

injection/sharedmain/main.go
```diff
@@ -324,8 +325,17 @@ func WatchLoggingConfigOrDie(ctx context.Context, cmw *configmap.InformedWatcher
 func WatchObservabilityConfigOrDie(ctx context.Context, cmw *configmap.InformedWatcher, profilingHandler *profiling.Handler, logger *zap.SugaredLogger, component string) {
 	if _, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(metrics.ConfigMapName(),
 		metav1.GetOptions{}); err == nil {
+		// NOTE: Do not use secrets.Get(ctx) here to get a SecretLister, as it will register
```
anniefu: Could we pass the kubeclient to the ConfigMapWatcher func instead of the secretFetcher and have the metrics pkg enforce this logic? 

It seems easy to make this mistake for any non sharedmain controllers. 

(Also, whoops, should have made this comment in the first pass...)

evankanderson: I'd like to be able to switch from kubeclient to something which actually uses a well-scoped informer, but I don't think that the current amount of apiserver traffic makes this blocking.

Also, it's much easier for tests and other clients to mock out a single fetch function than kubeclient.

anniefu: Hmm, could we put the logic and NOTE in a static helper function on the metric pkg that non sharedmain controllers could also use then? That way swap-ability and testing ease remains, but we can help avoid other controllers using secrets.Get(ctx)

evankanderson: You mean a helper for `WatchObservabilityConfigOrDie`? Sure.

evankanderson:
Updated.

injection/sharedmain/main.go
anniefu: I was thinking just "GetDefaultSecretFetcher(name string, kubeclient kubeclient) func(name string) (*corev1.Secret, error)"

Either one's good with me though!

anniefu:
Thanks!
/lgtm

PR #1112, opened 2020-02-20
Nil check conditions

n3wscott:
This change:

- finding the next best dependent condition that should be set on ready when ready changes true.
- Remove conditions from ducks/v1alpha1 (not used)
- Remove KResource from ducks/v1alpha1 (not used)


n3wscott:
/hold
writing unit tests for condition

n3wscott:
/unhold

Unit tests added.

n3wscott:
/hold

integrating downstream.

apis/condition_set.go
```diff
@@ -260,8 +260,8 @@ func (r conditionsImpl) MarkTrue(t ConditionType) {
 		if !c.IsTrue() {
 			// Update the happy condition if the current ready condition is
 			// marked not ready because of this condition.
-			if org.Reason == orgTL.Reason && org.Message == orgTL.Message {
-				r.propagateFailure(org, orgTL)
+			if org.GetReason() == orgTL.GetReason() && org.GetMessage() == orgTL.GetMessage() {
```
mattmoor: `org` and `orgTL` only exist for this predicate.  Instead of copying on 246/247, check this condition there.  Is would also be significantly clearer to nil check the condition there vs. adding nil-sensitive getters.

I'd also structure this differently.  Rather than adding `propagateFailure`, I would create a method for finding the new `*apis.Condition` that should be reflected in `Ready` (extracted from `propagateFailure`).  Why a pointer?  `nil` means do the SetCondition (below) and non-nil means SetCondition to the returned Condition.

n3wscott: but how do you know when the ready reason should be cleared if you don't capture what they were before setting the t value?

grantr:
The KResource tracker is used in the broker reconciler: https://github.com/knative/eventing/blob/1e30195e75055266e676a6fa97fc23e8ee12371e/pkg/reconciler/broker/trigger.go#L211

Does this PR break that?

mattmoor:
/hold

For @grantr's question.

n3wscott:
> The KResource tracker is used in the broker reconciler: https://github.com/knative/eventing/blob/1e30195e75055266e676a6fa97fc23e8ee12371e/pkg/reconciler/broker/trigger.go#L211
> 
> Does this PR break that?

v1 is used: `duckv1.KResource`

I am deleting v1alpha1, which is not used.

Eventing uses a mix of v1 and v1beta1 KResource.

mattmoor:
@grantr does that use the v1beta1 or v1\ KResource?

n3wscott:
> @grantr does that use the v1beta1 or v1\ KResource?

the link you pointed to does not use KResource: 

https://github.com/knative/eventing/blob/1e30195e75055266e676a6fa97fc23e8ee12371e/pkg/duck/listable.go#L114

grantr:
:+1:
/hold cancel

n3wscott:
/unhold

Eventing builds with this pkg with a minor test update for bad assumptions.

n3wscott:
https://github.com/knative/eventing/pull/2615

apis/duck/v1alpha1/conditions_types.go
```diff
@@ -1,203 +0,0 @@
-/*
-Copyright 2018 The Knative Authors
```
nachocano: isn't this being used in eventing? See https://github.com/knative/eventing/blob/master/pkg/reconciler/broker/controller.go#L98

nachocano: oops, missed the previous comments... discard

PR #1118, opened 2020-02-23
Pass verbosity level on to klog

coryrc:
This was left out of #1022

### DIFF
--- a/test/logging/logging.go
+++ b/test/logging/logging.go
@@ -163,7 +163,7 @@ func InitializeLogger() {
 
 		// Set klog/glog verbosities (works with and without proxy libraries)
 		klogLevel := klog.Level(0)
-		klogLevel.Set(fmt.Sprint(Verbosity))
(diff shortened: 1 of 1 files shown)


test/logging/logging.go
```diff
@@ -159,6 +161,10 @@ func InitializeLogger() {
 		logger = zap.New(zapCore)
 		zap.ReplaceGlobals(logger) // Gets used by klog/glog proxy libraries
 
+		// Set klog/glog verbosities (works with and without proxy libraries)
+		klogLevel := klog.Level(0)
+		klogLevel.Set(fmt.Sprint(Verbosity))
```
vagababov: Verbosity.String()?

coryrc: I accidentally copied an old version, it shouldn't be capitalized.

It's an int and int doesn't have a .String() AFAICT.

vagababov: If it's an int, then just `strconv.Itoa`.

PR #1119, opened 2020-02-24
Auto-update dependencies

mattmoor:
Produced via:
  `dep ensure -update knative.dev/test-infra`
/assign n3wscott
/cc n3wscott


n3wscott:
I think this is related to my change adding the genreconciler test client in test. @pmorie was going to add a delete call after the presumbits happen. I will follow up with Paul today 👍

PR #1120, opened 2020-02-24
Use interfaces when testing conditions so we don't need versioned methods

dprotaso:
### Downstream PRs

- [x] knative/serving - https://github.com/knative/serving/pull/6985
- [x] knative/eventing - noop (no usage)
- [x] knative/eventing-contrib - noop (no usage)
- [x] knative/sample-controller - noop (no usage)
- [x] knative/sample-source - noop (no usage)


dprotaso:
hmm

apis/testing/conditions.go
```diff
@@ -20,14 +20,12 @@ import (
...
 	"knative.dev/pkg/apis"
-	duckv1b1 "knative.dev/pkg/apis/duck/v1beta1"
 	"knative.dev/pkg/test"
 )
 
 // CheckCondition checks if condition `c` on `cc` has value `cs`.
-// DEPRECATED: Use versioned test helper
-func CheckCondition(s *duckv1b1.Status, c apis.ConditionType, cs corev1.ConditionStatus) error {
-	cond := s.GetCondition(c)
+func CheckCondition(a apis.ConditionAccessor, c apis.ConditionType, cs corev1.ConditionStatus) error {
```
n3wscott: Why not use the condtionManager interface?

dprotaso: That interface is huge for what these helpers need

dprotaso: secondly we're passing `duck.Status` types here

n3wscott: You might also want gettoplevel in the new interface

dprotaso:
`knative/eventing` doesn't use these helpers so that's done

dprotaso:
`knative/eventing-contrib` doesn't use these helpers so that's done

dprotaso:
`knative/sample-controller` doesn't use these helpers so that's done

dprotaso:
`knative/sample-source` doesn't use these helpers so that's done

apis/testing/conditions.go
dprotaso: `duck.Status` doesn't have that method so don't want to add it

dprotaso:
@n3wscott PTAL again

PR #1121, opened 2020-02-24
Delete ./test/testgenclient after presubmit is finished.

n3wscott:
An oversight left the gen test client around and it is causing issues with other automation that did not gen the test client because of deps changes.

/assign @mattmoor 
/cc @pmorie


test/presubmit-tests.sh
```diff
@@ -24,12 +24,16 @@ export DISABLE_MD_LINTING=1
...
-main $@
+function pre_build_tests() {
+  # Test the custom code generators. This makes sure we can compile the output
+  # of the injection generators.
+  $(dirname $0)/test-reconciler-codegen.sh
+  return 0
+}
+
+
+main $@
```
mattmoor: Add trailing newline:
```suggestion
main $@

```

PR #1126, opened 2020-02-25
Move map functions to PKG

mattmoor:
These are from `knative.dev/serving/pkg/resources/meta[_test].go`

### DIFF
--- a/kmeta/map.go
+++ b/kmeta/map.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


kmeta/map.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	for k, v := range b {
+		out[k] = v
+	}
+	return out
+}
+
+// FilterMap creates a copy of the provided map, filtering out the elements
+// that match `filter`.
+// nil `filter` is accepted.
```
dprotaso: hmm.. but why

vagababov: Why not? :) Probably at some invocation sites it was required.

PR #1133, opened 2020-02-27
Add filtering to class based controller stubs.

markusthoemmes:
Breaking this out from the broader discussion I started on Slack around class-based reconcilers.
Regardless of what the outcome is, I think we should encourage users to add the class filters to their informers to reduce churn.

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_controller.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller.go
@@ -154,7 +154,10 @@ const (
 	defaultControllerAgentName = "{{.type|lowercaseSingular}}-controller"
 	defaultFinalizerName       = "{{.type|allLowercasePlural}}.{{.group}}"
 	defaultQueueName           = "{{.type|allLowercasePlural}}"
-	{{if .hasClass}}classAnnotationKey = "{{ .class }}"{{end}}
+	{{if .hasClass}}
+	// ClassAnnotationKey points to the annotation for the class of this resource.
+	ClassAnnotationKey = "{{ .class }}"
+	{{end}}
 )
 
 // NewImpl returns a {{.controllerImpl|raw}} that handles queuing and feeding work from
--- a/codegen/cmd/injection-gen/generators/reconciler_controller_stub.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller_stub.go
@@ -87,6 +87,18 @@ func (g *reconcilerControllerStubGenerator) GenerateType(c *generator.Context, t
 			Package: "knative.dev/pkg/configmap",
 			Name:    "Watcher",
 		}),
+		"classAnnotationKey": c.Universe.Variable(types.Name{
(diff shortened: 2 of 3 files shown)


codegen/cmd/injection-gen/generators/reconciler_controller_stub.go
```diff
@@ -106,15 +118,27 @@ func NewController(
...
 	r := &Reconciler{}
-	impl := {{.reconcilerNewImpl|raw}}(ctx, r{{if .hasClass}}, "default"{{end}})
+	impl := {{.reconcilerNewImpl|raw}}(ctx, r{{if .hasClass}}, classValue{{end}})
 
 	logger.Info("Setting up event handlers.")
 
+	{{if .hasClass}}
+	{{.type|lowercaseSingular}}Informer.Informer().AddEventHandler({{.filterHandler|raw}}{
+		FilterFunc: classFilter,
+		Handler:    controller.HandleAll(impl.Enqueue),
```
mattmoor: I like this, we should definitely do this regardless.  thanks!

markusthoemmes:
@mattmoor I think this piece should be non-contentious.

codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -154,7 +154,10 @@ const (
 	defaultControllerAgentName = "{{.type|lowercaseSingular}}-controller"
 	defaultFinalizerName       = "{{.type|allLowercasePlural}}.{{.group}}"
 	defaultQueueName           = "{{.type|allLowercasePlural}}"
-	{{if .hasClass}}classAnnotationKey = "{{ .class }}"{{end}}
+	{{if .hasClass}}
+	// ClassAnnotationKey points to the annotation for the class of this resource.
+	ClassAnnotationKey = "{{ .class }}"
```
n3wscott: you want to export it?

codegen/cmd/injection-gen/generators/reconciler_controller_stub.go
n3wscott: yea nice.

codegen/cmd/injection-gen/generators/reconciler_controller.go
n3wscott: maybe what you are after is an accessor? like `GetClass` and `SetClass` implemented in the reconciler as exported functions? Then you never need to know what the class annotation key (or the details of how it is set). This would save us from rewriting a bunch of things if we move to an annotation or a spec field for some reason

mattmoor: IIUC this is to make the comment the source of truth, which is something I was struggling with us splitting.  So 👍 from me 😉

markusthoemmes: Sounds good, I can explore that direction!

PR #1136, opened 2020-03-02
Add an `--upgrade` flag to `./hack/update-deps.sh`

mattmoor:
This flag upgrades the list of dependencies that we choose to track at HEAD.

cc @n3wscott @coryrc @chaodaiG @vdemeester

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -23,8 +23,26 @@ set -o pipefail
 
 cd ${ROOT_DIR}
 
+# The list of dependencies that we track at HEAD and periodically
+# float forward in this repository.
+FLOATING_DEPS=(
+  "knative.dev/test-infra"
+)
+
+# Parse flags to determine any we should pass to dep.
+DEP_FLAGS=()
+while [[ $# -ne 0 ]]; do
+  parameter=$1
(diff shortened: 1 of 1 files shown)


hack/update-deps.sh
```diff
@@ -23,8 +23,26 @@ set -o pipefail
...
+FLOATING_DEPS=(
+  "knative.dev/test-infra"
+)
+
+# Parse flags to determine any we should pass to dep.
+DEP_FLAGS=""
+while [[ $# -ne 0 ]]; do
+  parameter=$1
+  case ${parameter} in
+    --upgrade) DEP_FLAGS="-update ${FLOATING_DEPS[@]}" ;;
```
n3wscott: can you try this script with two or more floaters? I thought there should only be one `-update` flag?

n3wscott: OH nevermind... I forgot we are passing in --upgrade, not the deps.

IGNORE ME

also...

:fire: bash

mattmoor: This only yields one flag, and I did try with two.  I have a tweak to this incoming to fix the error where `""` is passed to `dep` without `--upgrade`, testing locally now.

PR #1142, opened 2020-03-05
Fix crash looping when leader election config doesn't exist

pmorie:
This fixes an issue where a controller will crashloop if a leader election config doesn't exist.


houshengbo:
/lgtm
Tested with operators.

PR #1147, opened 2020-03-07
Add a function to parse all env config set by Prow

chizhg:
Add a function to parse all env config set by Prow

### DIFF
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -18,7 +18,11 @@ limitations under the License.
 
 package prow
 
-import "github.com/kelseyhightower/envconfig"
+import (
+	"fmt"
+
+	"github.com/kelseyhightower/envconfig"
+)
 
 type EnvConfig struct {
 	CI          bool   `required:"true"`
@@ -33,12 +37,14 @@ type EnvConfig struct {
 	PullBaseRef string `split_words:"true"`
 	PullBaseSha string `split_words:"true"`
 	PullRefs    string `split_words:"true"`
-	PullNumber  string `split_words:"true"`
+	PullNumber  uint   `split_words:"true"`
 	PullPullSha string `split_words:"true"`
 }
 
 func GetEnvConfig() (*EnvConfig, error) {
 	var ec EnvConfig
-	err := envconfig.Process("", &ec)
-	return &ec, err
+	if err := envconfig.Process("", &ec); err != nil {
+		return nil, fmt.Errorf("failed getting environment variables for Prow: %w", err)
+	}
+	return &ec, nil
 }
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -24,6 +24,9 @@ import (
 	"github.com/kelseyhightower/envconfig"
 )
 
+// EnvConfig consists of all the environment variables that can be set in a Prow job,
+// check https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
+// for more information.
 type EnvConfig struct {
 	CI          bool   `required:"true"`
 	Artifacts   string `required:"true"`
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// env.go provides a central point to read all environment variables defined by Prow.
+
+package prow
+
+import "github.com/kelseyhightower/envconfig"
+
+type EnvConfig struct {
+	CI          bool   `required:"true"`
+	Artifacts   string `required:"true"`
+	JobName     string `required:"true" split_words:"true"`
+	JobType     string `required:"true" split_words:"true"`
+	JobSpec     string `required:"true" split_words:"true"`
+	BuildID     string `required:"true" envconfig:"BUILD_ID"`
+	ProwJobID   string `required:"true" envconfig:"PROW_JOB_ID"`
+	RepoOwner   string `split_words:"true"`
+	RepoName    string `split_words:"true"`
+	PullBaseRef string `split_words:"true"`
+	PullBaseSha string `split_words:"true"`
+	PullRefs    string `split_words:"true"`
+	PullNumber  string `split_words:"true"`
+	PullPullSha string `split_words:"true"`
+}
+
+func GetEnvConfig() (*EnvConfig, error) {
+	var ec EnvConfig
+	err := envconfig.Process("", &ec)
+	return &ec, err
+}
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -19,6 +19,7 @@ limitations under the License.
 package prow
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/kelseyhightower/envconfig"
@@ -44,14 +45,15 @@ type EnvConfig struct {
 	PullPullSha string `split_words:"true"`
 }
 
+// GetEnvConfig returns values of all the environment variables that can be possibly set in a Prow job.
 func GetEnvConfig() (*EnvConfig, error) {
-	if !IsCI() {
-		return nil, fmt.Errorf("this function is not expected to be called from a non-CI environment")
-	}
-
 	var ec EnvConfig
 	if err := envconfig.Process("", &ec); err != nil {
 		return nil, fmt.Errorf("failed getting environment variables for Prow: %w", err)
+	}
+
+	if !ec.CI {
+		return nil, errors.New("this function is not expected to be called from a non-CI environment")
 	}
 	return &ec, nil
 }
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -28,13 +28,13 @@ import (
 // check https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
 // for more information.
 type EnvConfig struct {
-	CI          bool   `required:"true"`
-	Artifacts   string `required:"true"`
-	JobName     string `required:"true" split_words:"true"`
-	JobType     string `required:"true" split_words:"true"`
-	JobSpec     string `required:"true" split_words:"true"`
-	BuildID     string `required:"true" envconfig:"BUILD_ID"`
-	ProwJobID   string `required:"true" envconfig:"PROW_JOB_ID"`
+	CI          bool
+	Artifacts   string
+	JobName     string `split_words:"true"`
+	JobType     string `split_words:"true"`
+	JobSpec     string `split_words:"true"`
+	BuildID     string `envconfig:"BUILD_ID"`
+	ProwJobID   string `envconfig:"PROW_JOB_ID"`
 	RepoOwner   string `split_words:"true"`
 	RepoName    string `split_words:"true"`
 	PullBaseRef string `split_words:"true"`
--- a/test/prow/env.go
+++ b/test/prow/env.go
@@ -45,6 +45,10 @@ type EnvConfig struct {
 }
 
 func GetEnvConfig() (*EnvConfig, error) {
+	if !IsCI() {
+		return nil, fmt.Errorf("this function is not expected to be called from a non-CI environment")
+	}
+
 	var ec EnvConfig
 	if err := envconfig.Process("", &ec); err != nil {
 		return nil, fmt.Errorf("failed getting environment variables for Prow: %w", err)
--- a/test/prow/env_test.go
+++ b/test/prow/env_test.go
@@ -16,9 +16,17 @@ limitations under the License.
 
 package prow
 
-import "testing"
+import (
+    "os"
+    "testing"
+)
 
 func TestGetEnvConfig(t *testing.T) {
+    isCI := os.Getenv("CI")
+    // Set it to the original value
+    defer os.Setenv("CI", isCI)
+
+    os.Setenv("CI", "true")
     if IsCI() {
         ec, err := GetEnvConfig()
         t.Logf("EnvConfig is: %v", ec)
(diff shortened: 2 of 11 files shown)


test/prow/env.go
```diff
@@ -0,0 +1,50 @@
...
+
+package prow
+
+import (
+	"fmt"
+
+	"github.com/kelseyhightower/envconfig"
+)
+
+type EnvConfig struct {
```
chaodaiG: Is there a struct defined in Prow that we can use instead of maintaining this struct by ourselves?

test/prow/env.go
```diff
@@ -0,0 +1,50 @@
...
+	PullBaseSha string `split_words:"true"`
+	PullRefs    string `split_words:"true"`
+	PullNumber  uint   `split_words:"true"`
+	PullPullSha string `split_words:"true"`
+}
+
+func GetEnvConfig() (*EnvConfig, error) {
+	var ec EnvConfig
+	if err := envconfig.Process("", &ec); err != nil {
+		return nil, fmt.Errorf("failed getting environment variables for Prow: %w", err)
```
chaodaiG: Is this s a runtime error?

test/prow/env_test.go
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prow
+
+import "testing"
+
+func TestGetEnvConfig(t *testing.T) {
+    if IsCI() {
```
chaodaiG: I think it would be better to mock this function for unit test

test/prow/prow_test.go
```diff
@@ -64,8 +64,22 @@ func TestInvalidJobPath(t *testing.T) {
...
 }
 
+func TestIsCI(t *testing.T) {
+	isCI := os.Getenv("CI")
+	// Set it to the original value
+	defer os.Setenv("CI", isCI)
+
+	os.Setenv("CI", "true")
+	ic := IsCI()
+	if !ic {
```
chaodaiG:
```suggestion
	if ic := IsCI(); !ic {
```

test/prow/prow_test.go
```diff
@@ -64,8 +64,22 @@ func TestInvalidJobPath(t *testing.T) {
...
+		t.Fatalf("Expected: %t, actual: %t", true, false)
```
chaodaiG:
```suggestion
		t.Fatal("Expected: true, actual: false")
```

test/prow/env.go
chizhg: Good point.
I did some search but k8s test-infra only have the env var names defined as private variables, see https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/downwardapi/jobspec.go#L80. They also have a function `EnvForType` that can return the environment variables as a map based on the job type, which I think cannot satisfy our needs either.

test/prow/env.go
chizhg: I think an error makes more sense as people can also call this function from a non-CI environment. In that case, it's expected to have an error and I think an error is better than a `panic`.

test/prow/env_test.go
chizhg: Done.

test/prow/prow_test.go
chizhg: Done.

test/prow/prow_test.go
chizhg: Done

test/prow/env.go
chaodaiG: So what will happen in CI? Is it going to fail all CI jobs if this ever failed? What could make it throw?

test/prow/env.go
chaodaiG: ack. that's sad

test/prow/env.go
chizhg: Not sure if I understand your question correctly. 
There are two cases:

1. If this function is called from a non-CI environment, it will always return an error, I have changed the function to return an error for such scenario in a different way.

2. If the function is called from a CI environment and it returns an error, it means Prow has some breaking change in the required fields, see https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables, which should be rare, but in this case, we'll need to make the corresponding update here.

chaodaiG: Sorry for not making it clear. I assume this function will be called in all Prow jobs in the future, and don't think it's generally the best idea to take Prow down because of possibly irrelevant API breakage. MY personally preference would be a unit test to capture this rather than failing at runtime. Can we make this function to do best effort or have a fall back strategy so it doesn't fail that hard?

chizhg: I got what you meant and I agree with it.
The easiest change is to remove all `required:"true"` from the struct tags so there won't be an error simply because of Prow spec changes. Please let me know if you have any further thoughts.

chaodaiG: sgtm. Please make a unit test so that it can capture broken API

chizhg: Hmm, what do you expect the unit test to do? Checking all required env vars are not empty? IMO such unit test is not useful in preventing breaking changes, and will block other PRs if one day Prow indeed makes one of the required env vars as optional

chaodaiG: What I was thinking about is your previous implementation, if possible we could use that one as a unit test. But I agree with you that we should not block pkg PRs because of this reason

chizhg: Hmm, so what's your current suggestion now that I changed the previous implementation?

chaodaiG: I think it's good for now

test/prow/env_test.go
```diff
@@ -0,0 +1,44 @@
...
+        t.Fatalf("Error getting envconfig for Prow: %v", err)
+    }
+    if !ec.CI {
+        t.Fatal("Expected CI to be true but is false")
+    }
+
+    os.Setenv("CI", "false")
+    _, err = GetEnvConfig()
+    if err == nil {
+        t.Fatalf("Expected an error if called from a non-CI environment but got nil")
```
vagababov:
```suggestion
        t.Fatal("Expected an error if called from a non-CI environment but got nil")
```

test/prow/prow.go
```diff
@@ -108,6 +108,15 @@ type Finished struct {
 // Metadata contains metadata in finished.json
 type Metadata map[string]interface{}
 
+// IsCI returns whether the current environment is a CI environment.
+func IsCI() bool {
+	isCI := os.Getenv("CI")
+	if isCI == "true" {
+		return true
+	}
+	return false
```
vagababov:
```suggestion
      return strings.EqualFold(os.Getenv("CI"), "true")
```

test/prow/env.go
```diff
@@ -0,0 +1,57 @@
...
+	PullNumber  uint   `split_words:"true"`
+	PullPullSha string `split_words:"true"`
+}
+
+func GetEnvConfig() (*EnvConfig, error) {
+	if !IsCI() {
+		return nil, fmt.Errorf("this function is not expected to be called from a non-CI environment")
+	}
+
+	var ec EnvConfig
```
vagababov: Why not process the envconfig anyway and then just check `ec.CI` field?

test/prow/env_test.go
```diff
@@ -0,0 +1,44 @@
...
+    t.Logf("EnvConfig is: %v", ec)
+    if err != nil {
+        t.Fatalf("Error getting envconfig for Prow: %v", err)
+    }
+    if !ec.CI {
+        t.Fatal("Expected CI to be true but is false")
+    }
+
+    os.Setenv("CI", "false")
+    _, err = GetEnvConfig()
```
vagababov: fold the error in the `if`.

chizhg: Done.

test/prow/env.go
chizhg: I initially set `CI` as `"required: true"` but as I removed it later then, so it now makes sense to change to what you suggested. 
Done.

test/prow/prow.go
chizhg: Thanks for the trick! Done.

test/prow/prow.go
```diff
@@ -108,6 +108,11 @@ type Finished struct {
 // Metadata contains metadata in finished.json
 type Metadata map[string]interface{}
 
+// IsCI returns whether the current environment is a CI environment.
```
vagababov: Now, do we need this function anymore?

test/prow/env.go
```diff
@@ -0,0 +1,58 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// env.go provides a central point to read all environment variables defined by Prow.
+
+package prow
+
+import (
```
mattmoor: Format Go code:
```suggestion
import (
	"errors"
```

test/prow/env_test.go
```diff
@@ -0,0 +1,43 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prow
+
+import (
+    "os"
+    "testing"
```
mattmoor: Format Go code:
```suggestion
	"os"
	"testing"
```

test/prow/env_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func TestGetEnvConfig(t *testing.T) {
+    isCI := os.Getenv("CI")
+    // Set it to the original value
+    defer os.Setenv("CI", isCI)
```
mattmoor: Format Go code:
```suggestion
	isCI := os.Getenv("CI")
	// Set it to the original value
	defer os.Setenv("CI", isCI)
```

test/prow/env_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+    os.Setenv("CI", "true")
+    ec, err := GetEnvConfig()
+    t.Logf("EnvConfig is: %v", ec)
+    if err != nil {
+        t.Fatalf("Error getting envconfig for Prow: %v", err)
+    }
+    if !ec.CI {
+        t.Fatal("Expected CI to be true but is false")
+    }
```
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "true")
	ec, err := GetEnvConfig()
	t.Logf("EnvConfig is: %v", ec)
	if err != nil {
		t.Fatalf("Error getting envconfig for Prow: %v", err)
	}
	if !ec.CI {
		t.Fatal("Expected CI to be true but is false")
	}
```

test/prow/env_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+    os.Setenv("CI", "false")
+    if _, err = GetEnvConfig(); err == nil {
+        t.Fatal("Expected an error if called from a non-CI environment but got nil")
+    }
```
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "false")
	if _, err = GetEnvConfig(); err == nil {
		t.Fatal("Expected an error if called from a non-CI environment but got nil")
	}
```

test/prow/env.go
```diff
@@ -0,0 +1,59 @@
...
```
mattmoor: Format Go code:
```suggestion
import (
	"errors"
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	"os"
	"testing"
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	isCI := os.Getenv("CI")
	// Set it to the original value
	defer os.Setenv("CI", isCI)
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "true")
	ec, err := GetEnvConfig()
	t.Logf("EnvConfig is: %v", ec)
	if err != nil {
		t.Fatalf("Error getting envconfig for Prow: %v", err)
	}
	if !ec.CI {
		t.Fatal("Expected CI to be true but is false")
	}
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "false")
	if _, err = GetEnvConfig(); err == nil {
		t.Fatal("Expected an error if called from a non-CI environment but got nil")
	}
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	"os"
	"testing"
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	isCI := os.Getenv("CI")
	// Set it to the original value
	defer os.Setenv("CI", isCI)
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "true")
	ec, err := GetEnvConfig()
	t.Logf("EnvConfig is: %v", ec)
	if err != nil {
		t.Fatalf("Error getting envconfig for Prow: %v", err)
	}
	if !ec.CI {
		t.Fatal("Expected CI to be true but is false")
	}
```

test/prow/env_test.go
mattmoor: Format Go code:
```suggestion
	os.Setenv("CI", "false")
	if _, err = GetEnvConfig(); err == nil {
		t.Fatal("Expected an error if called from a non-CI environment but got nil")
	}
```

test/prow/prow.go
chizhg: We still need it. As if we have a test flow that can be run both locally and in the CI environment, we need to use this function to decide whether we should call `GetEnvConfig` to get the env vars set by Prow.

chizhg:
/test 
pull-knative-pkg-unit-tests

vagababov:
Please don't force push... it makes review experience horrible :(

test/prow/prow.go
vagababov: So it'll be used in future code?

chizhg: Yes it will. More specifically, in the current shell script we have an environment variable called `IS_PROW`, later on when we rewrite the shell scripts into Go, we will replace that env var with this function.

chizhg:
> Please don't force push... it makes review experience horrible :(

Sorry I didn't realized that... Will avoid doing force push in the future.

PR #1148, opened 2020-03-07
Add MarkTrueWithReason function to ConditionManager

nak3:
Currently we cannot set a condition to true with the reason.

When using `MarkTrue()` function, it does not allow us to add the reason and message.
Also, if we use `SetCondition()` instead, it does not update the happy condition.

So, this PR adds new `MarkTrueWithReason(reason, message)` function to solve it.

For example in the Serving project, we want to manage KCert's condition, like below:

#### After using this path for https://github.com/knative/serving/pull/7163:

```
$ kubectl get rt hello-example -o yaml
    ...
    conditions:
    - lastTransitionTime: "2020-03-07T08:30:11Z"
      message: autoTLS is not enabled
      reason: AutoTLSNotEnabled
      status: "True"
      type: CertificateProvisioned

$ kubectl get rt
NAME            URL                                        READY   REASON
hello-example   http://hello-example.default.example.com   True
```

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -75,6 +75,10 @@ type ConditionManager interface {
 	// true if all dependents are true.
 	MarkTrue(t ConditionType)
 
+	// MarkTrueWithReason sets the status of t to true with the reason, and then marks the happy
+	// condition to true if all dependents are true.
+	MarkTrueWithReason(t ConditionType, reason, messageFormat string, messageA ...interface{})
+
 	// MarkUnknown sets the status of t to Unknown and also sets the happy condition
 	// to Unknown if no other dependent condition is in an error state.
 	MarkUnknown(t ConditionType, reason, messageFormat string, messageA ...interface{})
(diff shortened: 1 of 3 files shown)


apis/condition_set_test.go
```diff
@@ -111,6 +111,12 @@ func TestNonTerminalCondition(t *testing.T) {
 		t.Errorf("MarkUnknown(Foo) = %v, wanted %v", got, want)
 	}
 
+	// Setting the other "terminal" condition makes Ready true.
+	manager.MarkTrueWithReason("Foo", "", "")
```
mattmoor: We could test this a little better :)

nak3: Yes, updated the tests.

nak3:
Could you please take a look @mattmoor @vaikas ?
https://github.com/knative/serving/pull/7163 needs to wait this library function if it will be available or not.

apis/condition_set.go
```diff
@@ -253,7 +257,25 @@ func (r conditionsImpl) MarkTrue(t ConditionType) {
...
+		Status:   corev1.ConditionTrue,
+		Reason:   reason,
+		Message:  fmt.Sprintf(messageFormat, messageA...),
+		Severity: r.severity(t),
+	})
+	r.markHappy(t)
+}
 
+// markHappy marks the happy condition to true if all other dependents are also true.
+func (r conditionsImpl) markHappy(t ConditionType) {
```
markusthoemmes: Maybe something like `recomputeHappiness` or similar is more fitting? We're not marking things happy unconditionally, we check recheck if we can and do if we can.

n3wscott:
nice, we had never needed a reason for true, but there is nothing that says you can't have one...

/lgtm
/approve

apis/condition_set.go
n3wscott: I love that method name.

PR #1154, opened 2020-03-11
Drop the logic to create missing secrets.

mattmoor:
This is a carry-over from when this was originally written and we had webhooks and secrets being GC'd due to a bad owner reference.

### DIFF
--- a/webhook/certificates/certificates_test.go
+++ b/webhook/certificates/certificates_test.go
@@ -64,17 +64,8 @@ func TestReconcile(t *testing.T) {
 		Key:     key,
 		Objects: []runtime.Object{secret},
 	}, {
-		Name:        "secret does not exist",
-		Key:         key,
-		WantCreates: []runtime.Object{secret},
-	}, {
-		Name:    "secret does not exist, create error",
-		Key:     key,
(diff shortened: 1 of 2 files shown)


webhook/certificates/certificates_test.go
```diff
@@ -176,9 +167,20 @@ func TestReconcileMakeSecretFailure(t *testing.T) {
...
+		Name:    "malformed secret",
 		Key:     key,
 		WantErr: true,
+		Objects: []runtime.Object{&corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: system.Namespace(),
+			},
+			Data: map[string][]byte{
+				// certresources.ServerKey:  []byte("missing"),
```
vagababov: rm ?

mattmoor: When I initially wrote these I had intentionally left them in to make clear which keys were excluded where.  See also the content `"present"` vs. `"missing"`.

mattmoor: We can convey this other ways, but we should fix the whole file. 😇

PR #1155, opened 2020-03-11
Use metrics.backend-destination for request metrics backend as well

yanweiguo:
We allow empty value when we first introduced `metrics.request-metrics-backend-destination` flag to disable request metrics from queue proxy.

We set `Prometheus` as default value in serving 0.12 release, which means we no longer allow disabling request metrics from queue proxy.

It doesn't make sense that we have two flags, i.e. `metrics.backend-destination` and `metrics.request-metrics-backend-destination`, to control what backend to send metrics to. The goal is to keep one, `metrics.backend-destination`.

So switch to use `metrics.backend-destination` instead of `metrics.request-metrics-backend-destination`. Next step is to remove this flag in serving.

### DIFF
--- a/metrics/config_observability_test.go
+++ b/metrics/config_observability_test.go
@@ -49,12 +49,12 @@ func TestObservabilityConfiguration(t *testing.T) {
 				Name:      ConfigMapName(),
 			},
 			Data: map[string]string{
-				"logging.enable-var-log-collection":           "true",
-				"logging.revision-url-template":               "https://logging.io",
-				"logging.enable-probe-request-log":            "true",
-				"logging.write-request-logs":                  "true",
-				"logging.request-log-template":                `{"requestMethod": "{{.Request.Method}}"}`,
-				"metrics.request-metrics-backend-destination": "stackdriver",
+				"logging.enable-var-log-collection": "true",
+				"logging.revision-url-template":     "https://logging.io",
+				"logging.enable-probe-request-log":  "true",
+				"logging.write-request-logs":        "true",
+				"logging.request-log-template":      `{"requestMethod": "{{.Request.Method}}"}`,
+				"metrics.backend-destination":       "stackdriver",
 			},
 		},
 	}, {
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -86,7 +86,7 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		oc.EnableProbeRequestLog = strings.EqualFold(eprl, "true")
 	}
 
-	if mb, ok := configMap.Data["metrics.request-metrics-backend-destination"]; ok {
+	if mb, ok := configMap.Data["metrics.backend-destination"]; ok {
 		oc.RequestMetricsBackend = mb
 	} else {
 		oc.RequestMetricsBackend = DefaultRequestMetricsBackend


mattmoor:
@evankanderson given the proximity to what you've been looking at, can you TAL?

mattmoor:
@yanweiguo I'd like Evan to look at this, since he's taken on the ocagent stuff.

I don't have (sufficient) context there, and I'd like to make sure we're aligned with the direction this stuff is heading.

/hold

mdemirhan:
I have three concerns/questions with this one:
1. Do we know if there are customers using one different than the other? If not, then this is good.

2. We still have examples pointing to the old name in config-observability. We should remove `metrics.request-metrics-backend-destination` from there as well.

3. When we disable this and for people that still have the extra value in their config map, this should not cause an issue, right? I don't remember, but do we fail if we find extra values in config maps?

anniefu:
I share @mdemirhan's concerns. Though for 3, extra map values in the configmaps are just ignored. I think we should still respect metrics.request-backend-destination first and then check metrics.backend-destination. We can add a warning that metrics.request-backend-destination will be deprecated if anyone is using and eventually completely remove it.

From my point of view, this change is orthogonal to the opencensus agent work and will still be relevant after we move forward with opencensus.

mattmoor:
@anniefu you have more context than me, so I'm not gonna block things.  Hopefully Evan agrees, but he's had >3 weeks 😅 

Free feel to `/unhold`

yanweiguo:
> 1. Do we know if there are customers using one different than the other? If not, then this is good.

We don't have such data. It doesn't make sense to me they users use one backend for request metrics and but another one for other metrics although it could happen.

> 2. We still have examples pointing to the old name in config-observability. We should remove `metrics.request-metrics-backend-destination` from there as well.

Will do it in serving repository.

> 3. When we disable this and for people that still have the extra value in their config map, this should not cause an issue, right? I don't remember, but do we fail if we find extra values in config maps?

As @anniefu mentioned, extra map values in the configmaps are just ignored.

> I think we should still respect metrics.request-backend-destination first and then check metrics.backend-destination. We can add a warning that metrics.request-backend-destination will be deprecated if anyone is using and eventually completely remove it.

Use metrics.backend-destination for metrics.request-metrics-backend-destination if it is not set is also a breaking change since it defaults to Prometheus. We can show warning logs and mark metrics.request-metrics-backend-destination deprecated for at least a release before we eventually completely remove, although it won't prevent breaking change. If this is what you prefer, I will do that.

metrics/config_observability_test.go
```diff
@@ -49,12 +49,12 @@ func TestObservabilityConfiguration(t *testing.T) {
...
-				"logging.enable-probe-request-log":            "true",
-				"logging.write-request-logs":                  "true",
-				"logging.request-log-template":                `{"requestMethod": "{{.Request.Method}}"}`,
-				"metrics.request-metrics-backend-destination": "stackdriver",
+				"logging.enable-var-log-collection": "true",
+				"logging.revision-url-template":     "https://logging.io",
+				"logging.enable-probe-request-log":  "true",
+				"logging.write-request-logs":        "true",
+				"logging.request-log-template":      `{"requestMethod": "{{.Request.Method}}"}`,
+				"metrics.backend-destination":       "stackdriver",
```
anniefu, line 57: Shouldn't line 67 be "stackdriver" now?

anniefu:
I see your point about this being a breaking change either way... and I assume there's no way for us to really know if people are using `metrics.request-backend-destination` in a way that would get broken...

To add some details about how this will interact with the future opencensus world, after we switch to opencensus agent, we'll want to deprecate the Stackdriver/Prometheus exporters (replaced by OpenCensus Collector Service pods that handle exporting to Stackdriver, Prometheus, etc) and opencensus will be the only option for `metrics.backend-destination`. 

Once we're confident in opencensus, we can remove `metrics.backend-destination` and `metrics.request-backend-destination` altogether, along with many other options in config-observability.


Given that, maybe it's not worth making this change? Though admittedly, I think any deprecation is still a long ways off.

yanweiguo:
> Once we're confident in opencensus, we can remove `metrics.backend-destination` and `metrics.request-backend-destination` altogether, along with many other options in config-observability.
> 
> Given that, maybe it's not worth making this change? Though admittedly, I think any deprecation is still a long ways off.

SGTM. Closing this.

Issue #1152, opened 2020-03-10
[webhook] reduce memory usage of the webhook


dprotaso:
## Expected Behavior
Webhook memory usage should remain somewhat linear with the resources that are being **_watched_**

## Actual Behavior
@sravankumar777 reported high memory usage here: https://github.com/knative/serving/issues/7195

I think it's because our informers are watching _all_ secrets here: https://github.com/knative/pkg/blob/master/webhook/certificates/controller.go#L59

## Steps to Reproduce the Problem

1. See https://github.com/knative/serving/issues/7195

## Additional Info
We should ensure we're scoping other informers appropriately as well

mattmoor:
There is an injection context function that can be used to scope the injected informers to a single namespace.  I think we'll also want the capacity to do this per-type as there are contexts where I could see us doing this for some types (Secrets, ConfigMaps) and not for others (Bindings).

mattmoor:
It would be good to get a heap profile of what's going on in these situations too rather than just guessing 👼

bbrowning:
Here's a screenshot of a heap profile for a case I recently ran into. If I'm reading it right, about 83% of the in_use heap space is from reading secrets from all over the cluster?
![Screenshot from 2020-03-11 14-29-49](https://user-images.githubusercontent.com/56071/76451007-21708f80-63a5-11ea-9c4a-87d4843fd62a.png)

mattmoor:
@bbrowning can you confirm that this is fixed at HEAD?

mattmoor:
cc @markusthoemmes

dprotaso:
PR that has the fix: https://github.com/knative/pkg/issues/1152 which is part of 0.14 serving release

sravankumar777:
@dprotaso & @mattmoor, looks like the issue for webhook pod memory usage is not resolved with knative-serving v0.14.0 release.

Below are the memory usage dashboard for two different environment, on which knative-serving is deployed. 

<img width="1340" alt="webhook-memory-usage-1" src="https://user-images.githubusercontent.com/18139717/84020949-ac6e1f80-a9be-11ea-88b5-95690e3992e3.png">
<img width="1340" alt="webhook-memory-usage-2" src="https://user-images.githubusercontent.com/18139717/84020952-ae37e300-a9be-11ea-9375-cf62451773ef.png">

```
[...]

PR #1156, opened 2020-03-12
Limit listers to the system names for secrets.

vagababov:
This is to restrict somewhat the memory consumption
of the webhook, which is now linear with number of namespaces
in the system.

For #1152 
/assign @dprotaso mattmoor

### DIFF
--- a/webhook/certificates/controller.go
+++ b/webhook/certificates/controller.go
@@ -22,6 +22,7 @@ import (
 	// Injection stuff
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	secretinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/secret"
+	"knative.dev/pkg/injection"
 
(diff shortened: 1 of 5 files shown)


webhook/certificates/controller.go
```diff
@@ -41,6 +42,7 @@ func NewController(
 ) *controller.Impl {
 
 	client := kubeclient.Get(ctx)
+	ctx = injection.WithNamespaceScope(ctx, system.Namespace())
```
mattmoor: This won't do what you think

vagababov: Well, that's how I parse the comment here. https://github.com/knative/pkg/blob/master/injection/context.go#L30
What will it do then?

vagababov: Well, apparently it'll do what I think it will do, but I have to put the line elsewhere :)

PR #1157, opened 2020-03-12
Add JSONTracePred

ian-mi:
Allows the use of an arbitrary predicate when polling a zipkin trace
ID. Enables the creation of tests in which the number of trace spans
may vary.


vagababov:
Can we add the test as well, then? :)

ian-mi:
Sorry the test is in eventing: https://github.com/knative/eventing/pull/2752, but if these helpers would be useful for serving I can move them here.

vagababov:
This general library, so it is better to test upstream, methinks.

ian-mi:
Do you mean a unit test for this helper? There are not currently unit tests for any of the zipkin helpers so that would increase the scope of this PR somewhat.

vagababov:
Hm, I thought there were tests here... but ok :|
/lgtm

ian-mi:
Thanks! I still need owner approval to merge, though.

chizhg:
Ah sorry missed that.
/approve

PR #1161, opened 2020-03-16
Add support for nonNamespaced generated reconcilers

nachocano:
- As with genClient, added `genreconciler:nonNamespaced` to provide cluster-scope resources generated reconcilers.

/assign @n3wscott

### DIFF
--- a/codegen/cmd/injection-gen/generators/packages.go
+++ b/codegen/cmd/injection-gen/generators/packages.go
@@ -169,9 +169,11 @@ func MustParseClientGenTags(lines []string) Tags {
 	_, ret.GenerateDuck = values["genduck"]
 
 	_, genRec := values["genreconciler"]
+	_, genRecNonNamespaced := values["genreconciler:nonNamespaced"]
 	_, genRecClass := values["genreconciler:class"]
-	// Generate Reconciler code if genreconciler OR genreconciler:class exist.
-	if genRec || genRecClass {
+
+	// Generate Reconciler code if genreconciler OR genreconciler:nonNamespaced OR genreconciler:class exist.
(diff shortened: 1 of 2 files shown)


mattmoor:
Superficially this LGTM.  @n3wscott any thoughts since you've done the most here?

codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -169,9 +169,11 @@ func MustParseClientGenTags(lines []string) Tags {
 	_, ret.GenerateDuck = values["genduck"]
 
 	_, genRec := values["genreconciler"]
+	_, genRecNonNamespaced := values["genreconciler:nonNamespaced"]
```
n3wscott: Use the real client key, no need to add a new key. 
```
// +genclient:nonNamespaced
```

n3wscott:
Rather than duplicating all the places we make the `Get` and `Update` calls, perhaps it would be better to do something like 

```
{{if .nonNamespaced}}
client := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}()
{{else}}
client := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}(existing.Namespace)
{{end}}

_, err = client.UpdateStatus(existing)
```

What do you think?


Also you will need to add a new resource that is generated at cluster scope to the test client to confirm this all compiles.

nachocano:
@n3wscott all comments make sense. Thanks for reviewing, will try to get back to this later today and will let you know when is ready for another look

n3wscott:
closed via https://github.com/knative/pkg/pull/1175

PR #1162, opened 2020-03-16
Add helper function to read webhook port number from env var

jihuin:
Users can use WEBHOOK_PORT env var to configure the port that webhook uses.
To read the env var, replace `Port: 8443` with `Port: webhook.Port(8443)` in [main.go](https://github.com/knative/serving/blob/0f1c5af53e0cb5172135d0e9d2d4b7e59e276187/cmd/webhook/main.go#L220).

### DIFF
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -17,17 +17,23 @@ limitations under the License.
 package webhook
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 )
 
 const portEnvKey = "WEBHOOK_PORT"
 
-// Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
+// Port returns the webhook port set by portEnvKey, or default port if env var is not set.
 func Port(defaultPort int) int {
-	port, err := strconv.Atoi(os.Getenv(portEnvKey))
-	if err != nil || port == 0 {
+	if os.Getenv(portEnvKey) == "" {
 		return defaultPort
+	}
+	port, err := strconv.Atoi(os.Getenv(portEnvKey))
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert the environment variable %q : %v", portEnvKey, err))
+	} else if port == 0 {
+		panic(fmt.Sprintf("the environment variable %q can't be zero", portEnvKey))
 	}
 	return port
 }
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -26,7 +26,7 @@ const portEnvKey = "WEBHOOK_PORT"
 // Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
 func Port(defaultPort int) int {
 	port, err := strconv.Atoi(os.Getenv(portEnvKey))
-	if err != nil {
+	if err != nil || port == 0 {
 		return defaultPort
 	}
 	return port
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -24,8 +24,8 @@ import (
 
 const portEnvKey = "WEBHOOK_PORT"
 
-// Port returns the webhook port set by portEnvKey, or default port if env var is not set.
-func Port(defaultPort int) int {
+// PortFromEnv returns the webhook port set by portEnvKey, or default port if env var is not set.
+func PortFromEnv(defaultPort int) int {
 	if os.Getenv(portEnvKey) == "" {
 		return defaultPort
 	}
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"os"
+	"strconv"
+)
+
+const portEnvKey = "WEBHOOK_PORT"
+
+// Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
+func Port(defaultPort int) int {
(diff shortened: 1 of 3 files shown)


webhook/env.go
```diff
@@ -0,0 +1,33 @@
...
+	"os"
+	"strconv"
+)
+
+const portEnvKey = "WEBHOOK_PORT"
+
+// Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
+func Port(defaultPort int) int {
+	port, err := strconv.Atoi(os.Getenv(portEnvKey))
+	if err != nil {
```
vagababov: I'd also check if `port ==0`

jihuin: Done. Thanks.

webhook/env.go
```diff
@@ -0,0 +1,33 @@
...
+
+const portEnvKey = "WEBHOOK_PORT"
+
+// Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
+func Port(defaultPort int) int {
+	port, err := strconv.Atoi(os.Getenv(portEnvKey))
+	if err != nil || port == 0 {
+		return defaultPort
+	}
+	return port
```
mattmoor: Could we get unit test coverage of the various cases?

jihuin: Working on adding the unit test. Thanks.

jihuin: Done. This helper function doesn't keep user from using invalid default port, so the invalid default port is not tested.

webhook/env_test.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// portEnvKey is unset when testing missing input.
+			if tc.name != testMissingInputName {
+				os.Setenv(portEnvKey, tc.in)
+			}
+			defer os.Unsetenv(portEnvKey)
+
+			if got := Port(testDefaultPort); got != tc.want {
+				t.Errorf("got %v, want %v", got, tc.want)
```
savitaashture:
```suggestion
				t.Errorf("got %d, want %d", got, tc.want)
```

webhook/env_test.go
```diff
@@ -0,0 +1,70 @@
...
+package webhook
+
+import (
+	"os"
+	"testing"
+)
+
+const (
+	testDefaultPort      = 8888
+	testMissingInputName = "MissingInput" // portEnvKey is not found.
```
vagababov: I don't think we need this constant.

jihuin: It's used to determine whether or not to set the portEnvKey env var. The "MissingInput" test case corresponds to current yaml that doesn't contain portEnvKey.

vagababov: I know, but you can just use "Missing" constant inline.
Up to you though.

webhook/env.go
```diff
@@ -0,0 +1,33 @@
...
+	"strconv"
+)
+
+const portEnvKey = "WEBHOOK_PORT"
+
+// Port returns the webhook port set by portEnvKey, or default port if env var is invalid.
+func Port(defaultPort int) int {
+	port, err := strconv.Atoi(os.Getenv(portEnvKey))
+	if err != nil || port == 0 {
+		return defaultPort
```
dprotaso: if `WEBHOOK_PORT` is set but not a valid port # we shouldn't fallback to the default port and instead return an error

ie. "9090 " (extra whitespace) will cause this to fallback to the default which is non-obvious

jihuin: Done. Thanks.

webhook/env.go
n3wscott: port == 0 is ok, the http lib will pick one for you

markusthoemmes: Do we want to allow 0 then?

webhook/env.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+	"os"
+	"strconv"
+)
+
+const portEnvKey = "WEBHOOK_PORT"
+
+// Port returns the webhook port set by portEnvKey, or default port if env var is not set.
+func Port(defaultPort int) int {
```
dprotaso, line 28: I think `Port` isn't clear enough that it tries to read the environment variable and then fallsback on the default.

How about  this mouthful `PortFromEnv`

jihuin: Done. Thanks.

webhook/env.go
vagababov: I don't think so, since bind will work, but we don't really query where we're bound so we cannot register the endpoint then, I think.

webhook/env_test.go
```diff
@@ -0,0 +1,83 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			// portEnvKey is unset when testing missing input.
+			if tc.name != testMissingInputName {
+				os.Setenv(portEnvKey, tc.in)
+			}
+			defer os.Unsetenv(portEnvKey)
+
+			defer func() {
+				if r := recover(); r == nil && tc.wantPanic {
+					t.Errorf("did not panic")
```
vagababov, line 72:
```suggestion
					t.Error("Did not panic")
```

webhook/env_test.go
```diff
@@ -0,0 +1,83 @@
...
+				} else if r != nil && !tc.wantPanic {
+					t.Errorf("got unexpected panic")
```
vagababov, line 74:
```suggestion
					t.Error("Got unexpected panic")
```

webhook/env_test.go
```diff
@@ -0,0 +1,83 @@
...
+				}
+			}()
+
+			if got := PortFromEnv(testDefaultPort); got != tc.want {
+				t.Errorf("got %d, want %d", got, tc.want)
```
vagababov, line 79:
```suggestion
				t.Errorf("PortFromEnv = %d, want: %d", got, tc.want)
```

webhook/env_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// portEnvKey is unset when testing missing input.
+			if tc.name != testMissingInputName {
+				os.Setenv(portEnvKey, tc.in)
+			}
+			defer os.Unsetenv(portEnvKey)
+
+			defer func() {
```
vagababov, line 70: we cam probably join the defers.

jihuin: Done.

PR #1163, opened 2020-03-17
Allow the ability to match unset class for genreconciler.

liu-cong:
The effects of this change is shown in https://github.com/knative/eventing/pull/2761/

/assign @vaikas

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_controller.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller.go
@@ -164,7 +164,7 @@ const (
 // the queue through an implementation of {{.controllerReconciler|raw}}, delegating to
 // the provided Interface and optional Finalizer methods. OptionsFn is used to return
 // {{.controllerOptions|raw}} to be used but the internal reconciler.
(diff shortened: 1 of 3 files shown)


codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -164,7 +164,7 @@ const (
 // the queue through an implementation of {{.controllerReconciler|raw}}, delegating to
 // the provided Interface and optional Finalizer methods. OptionsFn is used to return
 // {{.controllerOptions|raw}} to be used but the internal reconciler.
-func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValue string{{end}}, optionsFns ...{{.controllerOptionsFn|raw}}) *{{.controllerImpl|raw}} {
```
n3wscott: We had looked at this before and it is too easy for multiple controllers to do this wrong and both race to reconcile a resource.

/hold

liu-cong:
#2910

PR #1165, opened 2020-03-17
Updating duckv1.SourceStatus to include CloudEventAttributes

nachocano:
- Needed for adding EventType registry information. See https://github.com/knative/eventing/pull/2770 and https://github.com/knative/eventing-contrib/pull/1036. 
- Is worth taking a look at the context from the eventing PR first.

/assign @lionelvillard @n3wscott @vaikas @mattmoor


apis/duck/v1/source_types.go
```diff
@@ -81,6 +81,22 @@ type SourceStatus struct {
...
+
+// CloudEventAttributes specifies the attributes that a Source
+// uses as part of its CloudEvents.
+type CloudEventAttributes struct {
+
+	// Type refers to the CloudEvent type attribute.
+	Type string `json:"type,omitempty"`
+
+	// Source is the CloudEvents source attribute.
+	Source string `json:"source,omitempty"`
```
mikehelmick: Not sure we can reliably populate the `ce-source` attribute on a control plane component like this.

Source can be highly variable for a given event producer for a given `ce-type`.

nachocano: no, we can't. We already have a lot of variability in our current sources in eventing and eventing-contrib. That's why it's optional. Or it shouldn't even be here. If it's not, then we cannot do the refactor while maintaining the current EventType functionality... And I kind of wanted to have the refactor first, and then address EventType (I'm more inclined towards removing broker and source from it, although not entirely sure about the latter.)

n3wscott: ok, just wanted to get Mikes input here, thanks Mike!

mikehelmick: talked w/ Nacho - we'll leave this in for now, but still have an issue open to revisit this in a larger conversation on discovery.

nachocano:
Synced offline with @mikehelmick , he is fine with letting this in so that we can maintain current functionality and unblock the refactor PRs:
https://github.com/knative/eventing/pull/2770
https://github.com/knative/eventing-contrib/pull/1036

We will revisit EventType in https://github.com/knative/eventing/issues/2750

apis/duck/v1/source_types.go
nachocano: sounds good, thanks for the input guys!

PR #1167, opened 2020-03-20
update go-cmpt to work with 1.14.

vagababov:
Mostly to fix https://github.com/google/go-cmp/issues/167, but we also pinned at some
random commit, rather than at a release version.

As is upgrade to  go1.14 would cause `go test -race` fail in a few places.

/assign mattmoor

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -220,7 +220,7 @@
   version = "v1.3.2"
 
 [[projects]]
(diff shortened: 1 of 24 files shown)


vagababov:
meh, fixed.

vagababov:
(ノಠ益ಠ)ノ彡┻━┻

vagababov:
@googlebot I signed it!

Issue #1140, opened 2020-03-04
[webhook] support distribution authors who want to add their custom defaulting and validation


dprotaso:
## Expected Behavior
As a distribution author it should be easy to include additional defaulting & validation logic without modifying the core API types

## Actual Behavior

It's actually non-trivial to add a vendor specific default & validation. We want to avoid [leaking things in the upstream API types](https://github.com/knative/serving/pull/7038#issuecomment-592171078).

A potential workaround is to define your [own stub type](https://github.com/knative/serving/pull/7081/files#diff-aa0a40159a3087a142da90ae81d95862R30-R38) but this is error prone because

1) When defaulting - if your stub doesn't simply passthrough _all_ fields of the type you'll actually _remove_ fields
2) Keeping stubs up to date with the upstream will be tedious - ie. when new fields are added

Thus we need something better for this situation

dprotaso:
cc @whaught @trshafer

dprotaso:
## Background

To elaborate on the issue our defaulting and validating webhooks are tightly integrated with the `resourcesemantics.GenericCRD` interface

https://github.com/knative/pkg/blob/1893541a0fac5885032c8ec170fd81234de13f4a/webhook/resourcesemantics/interface.go#L26-L30

https://github.com/knative/pkg/blob/1893541a0fac5885032c8ec170fd81234de13f4a/apis/interfaces.go#L27-L35


This works fine for our types which are versioned and strongly typed but there's been a need to apply similar logic across different types and/or versions. 

## Use Cases

### I) Custom defaulting on a type across numerous versions

In https://github.com/knative/serving/pull/7038 we wanted to add some labels to Revisions. 

### Approaches to the problem

#### 1. Add the defaulting logic to our types in `pkg/apis`

The initial approach wasn't ideal because it was adding networking specific labels to our Revision defaulting.

#### 2. Webhook + Stub Solution

Proposed in https://github.com/knative/serving/pull/7081 was to use a separate webhook controller to apply the defaults. This required the use of a stub as follows

```go
type RevisionStub struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	// TODO - add a general solution that captures
	// top level properties in addition to spec & status
	Spec   runtime.RawExtension `json:"spec"`
	Status runtime.RawExtension `json:"status"`
}
func (*RevisionStub) SetDefault(context.Context){...}
[...]

PR #1172, opened 2020-03-24
Allow Unstructured callback from Validating Webhook

whaught:
Issue #1140

**Proposed Changes**
Allow Validating webhook to register generic Unstructured callbacks so that implementers can create generic, stateful validation outside of the APIs package.

This will later be used in Serving to implement DryRun validation of PodSpec for issue #3425


webhook/resourcesemantics/validation/validation.go
```diff
@@ -238,21 +245,27 @@ func (ac *reconciler) validate(ctx context.Context, req *admissionv1beta1.Admiss
...
 		// discretion over (our portion of) the message that the user sees.
 		return err
 	}
 
-	return nil
-}
+	// Generically callback if any are provided for the resource.
+	if callback, ok := ac.callbacks[gvk]; ok {
+		var unstruct unstructured.Unstructured
+		unstruct.NewEmptyInstance()
```
vagababov: This is just a copy ctor, methinks? I'd just remove this line

webhook/resourcesemantics/validation/validation.go
```diff
@@ -238,21 +245,27 @@ func (ac *reconciler) validate(ctx context.Context, req *admissionv1beta1.Admiss
...
 		// Return the error message as-is to give the validation callback
 		// discretion over (our portion of) the message that the user sees.
 		return err
 	}
 
-	return nil
-}
+	// Generically callback if any are provided for the resource.
+	if callback, ok := ac.callbacks[gvk]; ok {
+		var unstruct unstructured.Unstructured
```
vagababov:
```suggestion
		unstruct := &unstructured.Unstructured{}
```

webhook/resourcesemantics/validation/validation.go
```diff
@@ -46,11 +48,15 @@ import (
...
+
 // reconciler implements the AdmissionController for resources
 type reconciler struct {
-	name     string
-	path     string
-	handlers map[schema.GroupVersionKind]resourcesemantics.GenericCRD
+	name      string
+	path      string
+	handlers  map[schema.GroupVersionKind]resourcesemantics.GenericCRD
+	callbacks map[schema.GroupVersionKind]Callback
```
vagababov: can there be more than one callback per type?

webhook/resourcesemantics/validation/validation.go
whaught: Ooops sorry that was outdated. It's removed.

webhook/resourcesemantics/validation/validation.go
whaught: I considered making it a list of callbacks, but I suspect the pattern will be to create a callback in Main and convert to the concrete type. Then use one callback to shim out to one or many pieces of logic.

webhook/resourcesemantics/validation/validation.go
whaught: Done

webhook/resourcesemantics/validation/validation.go
```diff
@@ -46,11 +48,15 @@ import (
 
 var errMissingNewObject = errors.New("the new object may not be nil")
 
+// Callback is a generic function to be called by a consumer of validation
+type Callback func(context.Context, unstructured.Unstructured) error
```
vagababov: Should this take ptr to Unstructured (the contents is a map, which a referenced object, so 🤷‍♂), since you can't do much with actual object (all its methods require pointer receiver).

whaught: Sure. I'm not sure if it's a big difference, but let's go with the pointer.

webhook/resourcesemantics/validation/validation.go
```diff
@@ -238,21 +245,25 @@ func (ac *reconciler) validate(ctx context.Context, req *admissionv1beta1.Admiss
...
-	return nil
-}
+	// Generically callback if any are provided for the resource.
+	if callback, ok := ac.callbacks[gvk]; ok {
+		uns, err := runtime.DefaultUnstructuredConverter.ToUnstructured(newObj)
+		if err != nil {
+			return fmt.Errorf("cannot convert new object to unstructured: %v", err)
+		}
+		unstruct := &unstructured.Unstructured{}
+		unstruct.SetUnstructuredContent(uns)
```
mattmoor: Given that we're not dealing with mutations, I bet it is more efficient to simply decode the raw object from `newBytes` into an unstructured, but otherwise this seems reasonable to me.

whaught: https://github.com/knative/pkg/pull/1177

Sure I've written a benchmark to test those two options. You're right - straight decoding is faster.

Output of 10,000 trials:
Converting took: 315.162774ms
Decoding took: 263.158939ms

webhook/resourcesemantics/validation/validation.go
```diff
@@ -238,21 +244,24 @@ func (ac *reconciler) validate(ctx context.Context, req *admissionv1beta1.Admiss
...
 	}
 
-	return nil
-}
+	// Generically callback if any are provided for the resource.
+	if callback, ok := ac.callbacks[gvk]; ok {
+		unstruct := &unstructured.Unstructured{}
+		newDecoder := json.NewDecoder(bytes.NewBuffer(newBytes))
+		if err := newDecoder.Decode(&unstruct); err != nil {
+			return fmt.Errorf("cannot decode incoming new object: %v", err)
```
vagababov:
```suggestion
			return fmt.Errorf("cannot decode incoming new object: %w", err)
```

webhook/resourcesemantics/validation/validation_context.go
```diff
@@ -0,0 +1,57 @@
...
+	// DryRun indicates that this request will not be persisted.
+	// Any operations with persistent side-effects should be skipped.
+	DryRun bool
+
+	// KubeClient is a client for interacting with k8s APIs.
+	KubeClient kubernetes.Interface
+}
+
+// NewValidationContext copies relevant objects from the request to a single structure to be read during validation.
+func NewValidationContext(ctx context.Context, ac *reconciler, req *admissionv1beta1.AdmissionRequest) (vc *Context) {
```
dprotaso: golang's context is essentially a bag of key => values and deadline/cancellation support.

You don't need to wrap the request context in a new struct but instead decorate the existing one - see: https://golang.org/pkg/context/#example_WithValue & https://github.com/knative/pkg/blob/master/logging/logger.go

The controller also provides a `withContext` so we may only need to decorate this instance with only the request specific things - ie. operation, dry run.

you may not need to need to pass the kubeclient on this context since it should be available on the context that's passed to the webhook controllers [here](https://github.com/knative/serving/blob/ee26e64f2158112933834945620477af540efc77/cmd/webhook/main.go#L106).

webhook/resourcesemantics/validation/controller.go
```diff
@@ -38,6 +38,7 @@ func NewAdmissionController(
 	ctx context.Context,
 	name, path string,
 	handlers map[schema.GroupVersionKind]resourcesemantics.GenericCRD,
+	callbacks map[schema.GroupVersionKind]Callback,
```
dprotaso: This is ugly but can we make this the last parameter and make it variadic & thus optional? We'll need the obvious validation.

Otherwise this becomes a breaking change.

webhook/resourcesemantics/validation/validation_context.go
dprotaso: From: https://github.com/knative/pkg/issues/1140#issuecomment-598877765

>It might be useful to decorate the context with whether we're in a DryRun or not. e.g. webhook.IsInDryRun(ctx), so that things can know and act accordingly.

whaught: I've removed the wrapper in favor of more parameters for clarity.

webhook/resourcesemantics/validation/controller.go
whaught: Done!

webhook/resourcesemantics/validation/validation.go
```diff
@@ -238,21 +244,26 @@ func (ac *reconciler) validate(ctx context.Context, req *admissionv1beta1.Admiss
...
 		logger.Errorw("Failed the resource specific validation", zap.Error(err))
 		// Return the error message as-is to give the validation callback
 		// discretion over (our portion of) the message that the user sees.
 		return err
 	}
 
-	return nil
-}
+	// Generically callback if any are provided for the resource.
+	if callback, ok := ac.callbacks[gvk]; ok {
```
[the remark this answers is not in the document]
whaught: Moved some things around to flow better.

webhook/resourcesemantics/validation/controller.go
```diff
@@ -40,17 +40,29 @@ func NewAdmissionController(
...
 
 	client := kubeclient.Get(ctx)
 	vwhInformer := vwhinformer.Get(ctx)
 	secretInformer := secretinformer.Get(ctx)
 	options := webhook.GetOptions(ctx)
 
+	// This not ideal, we are using a variadic argument to effectively make callbacks optional
+	// This allows this addition to be non-breaking to consumers of /pkg
+	// TODO: once all sub-repos have adoped this, we might move this back to a traditional param.
+	unwrappedCallbacks := map[schema.GroupVersionKind]Callback{}
```
dprotaso: I'd keep it simple and panic if len(callbacks) > 1 

I'm suggesting we opt for the panic because fetching the clients above panic as well if the context isn't decorated.

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,190 @@
...
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/webhook"
+	"knative.dev/pkg/webhook/resourcesemantics"
+)
+
+var errMissingNewObject = errors.New("the new object may not be nil")
+
+// Callback is a generic function to be called by a consumer of validation
+type Callback func(ctx context.Context, unstructured *unstructured.Unstructured, dryRun bool, opVerb admissionv1beta1.Operation) error
```
dprotaso: We already have behaviour to decorate the context with the request op etc. so I'd like to be consistent and decorate the context.

Thus making the callback
```go
type Callback func(ctx context.Context, unstructured *unstructured.Unstructured) error
```

dprotaso: helpers such as - `webhook.IsCreate/Update/DryRun` also would allow us to migrate to admission v1 without needing to change much code downstream

vagababov: While we do have that, that's an antipattern (API obscurity and each lookup is an O(N) operation).
Perhaps (maybe not here, but as a followup), we should reconsider and switch to explicit
```
type WHContext struct {
  context.Context
  inUpdate,dryRun bool
  ...
}
```
So it'll work both as a context and explicitly denote the params.
/cc @mattmoor

whaught: My fear is that API implementers will forget about the dryrun scenario if it's buried in the context which could be dangerous, or ignore that the Delete verb (currently unsupported by Validatable) is super important.

Using https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39#9e77 as a guidance:
"Just because you’ve taken that ugliness and obscured it inside Context.Value doesn’t make your API or design better. Obscurity is the opposite of good API design."

Putting important-for-control-flow values in as arguments or at least a strongly type struct helps point future implementers toward self-documenting functions whose contracts are clear.

Introducing a brand-new interface into /pkg might be a really good opportunity to reconsider this pattern.

whaught: If you look back at commit "Merge branch 'vcontext' into unstructured" https://github.com/knative/pkg/pull/1172/commits/e1bacc8e358f95070a6f4696e98c9d792796e15f

that was my first attempt at what @vagababov is suggesting

dprotaso: >Perhaps (maybe not here, but as a followup), we should reconsider and switch to explicit

Long term I agree. I would want this to be part of bigger webhook v2 rewrite that better layers these concerns.

vagababov: Should we file issues for that? So it's not forgotten until it floats up next time?

webhook/resourcesemantics/validation/controller.go
whaught: Good point. This will be super annoying down the line if anyone ever has > 1

webhook/resourcesemantics/validation/validation_admit.go
whaught: Everything moved to the context.

apis/contexts.go
```diff
@@ -198,3 +199,32 @@ func AllowDifferentNamespace(ctx context.Context) context.Context {
...
+// IsDryRun indicates that this request is in DryRun mode.
+func IsDryRun(ctx context.Context) bool {
+	return ctx.Value(isDryRun{}) != nil
+}
+
+// Key to attach an operation verb to the context.
+type opVerb struct{}
+
+// WithOpVerb is used to attach an operation verb to the context.
+func WithOpVerb(ctx context.Context, op admissionv1beta1.Operation) context.Context {
```
dprotaso: This package already has helpers for `IsInCreate`  & `IsInUpdate` - if we need additional operation checks we can add them in the same style.

webhook/resourcesemantics/validation/controller.go
```diff
@@ -40,17 +40,32 @@ func NewAdmissionController(
...
+	if len(callbacks) > 1 {
+		panic("NewAdmissionController may not be called with multiple callback maps")
+	}
+	unwrappedCallbacks := map[schema.GroupVersionKind]Callback{}
+	for _, m := range callbacks {
+		for k, v := range m {
+			unwrappedCallbacks[k] = v
+		}
+	}
+
```
dprotaso: we can drop the loops and just check for `len(callbacks) == 1`

apis/contexts.go
whaught: The problem with those interfaces is that in theory they can both be set! Plus you'd need to make a whole new type for every verb as they are supported/added.

For Callback to be a good generic interface let's try to simply pass the verb along for use. If we ever decide to move Validatable logic over to callback, we can switch these as well.

webhook/resourcesemantics/validation/controller.go
whaught: Done!

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -76,6 +79,29 @@ var (
 		}: &InnerDefaultResource{},
 	}
 
+	callbacks = map[schema.GroupVersionKind]Callback{
+		{
+			Group:   "pkg.knative.dev",
+			Version: "v1alpha1",
+			Kind:    "Resource",
+		}: Callback{
```
mattmoor, line 87: Format Go code:
```suggestion
		}: {
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -76,6 +79,29 @@ var (
...
+			supportedVerbs: map[admissionv1beta1.Operation]bool{
+				admissionv1beta1.Create: true,
+				admissionv1beta1.Update: true,
+			},
+			callback: resourceCallback},
+		{
+			Group:   "pkg.knative.dev",
+			Version: "v1beta1",
+			Kind:    "Resource",
+		}: Callback{
```
mattmoor, line 97: Format Go code:
```suggestion
		}: {
```

apis/contexts.go
whaught: Okay new idea. We can use IsDelete from context, but have a list of supported verbs for the Callback. Thoughts?

dprotaso: sure

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+	"knative.dev/pkg/webhook/resourcesemantics"
+)
+
+var errMissingNewObject = errors.New("the new object may not be nil")
+
+// Callback is a generic function to be called by a consumer of validation
+type Callback struct {
+	Callback func(ctx context.Context, unstructured *unstructured.Unstructured) error
+
+	SupportedVerbs map[admissionv1beta1.Operation]bool
```
dprotaso, line 43: maybe just make this a list?

dprotaso: Additionally - do you mind just aliasing the Operation types? Thus the use of v1beta1 is an implementation detail and we won't need downstream changes when we swap out the API for v1 in the future (ie. next release)

ie. the top level webhook package

```
type Operation = v1beta1.Operation

const (
    Create  Operation = v1beta1.Create
    Update  Operation = v1beta1.Update
    ...
```

whaught: Aliased into validation.

 I think keeping the map is desirable (really I want a Set) so you can't repeat verbs and makes it easier to not have to search the slice.

whaught: Sorry moved the alias into the root webhook package. per your suggestion we probably want this everywhere

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if err := ac.callback(ctx, request, gvk); err != nil {
+		return webhook.MakeErrorStatus("validation callback failed: %v", err)
+	}
+
+	return &admissionv1beta1.AdmissionResponse{Allowed: true}
+}
+
+// decodeRequest deserializes the old and new GenericCrds from the incoming request and sets up the context.
```
vagababov, line 78:
```suggestion
// decodeRequestAndPrepareContext deserializes the old and new GenericCrds from the incoming request and sets up the context.
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/webhook"
+	"knative.dev/pkg/webhook/resourcesemantics"
+)
+
+var errMissingNewObject = errors.New("the new object may not be nil")
+
+// Callback is a generic function to be called by a consumer of validation
+type Callback struct {
+	Callback func(ctx context.Context, unstructured *unstructured.Unstructured) error
```
vagababov, line 41: I am not sure I said this before, but it's a bit tautological here. May be just `C` of `F`. :)

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+		Version: kind.Version,
+		Kind:    kind.Kind,
+	}
+
+	ctx, resource, err := ac.decodeRequestAndPrepareContext(ctx, request, gvk)
+	if err != nil {
+		return webhook.MakeErrorStatus("decoding request failed: %v", err)
+	}
+
+	err = validate(ctx, resource, request)
```
vagababov, line 66: I'd fold it in, just like below.
```
if err := v(); err != nil {
}
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+	// Decode json to a GenericCRD
+	var newObj resourcesemantics.GenericCRD
+	if len(newBytes) != 0 {
+		newObj = handler.DeepCopyObject().(resourcesemantics.GenericCRD)
+		newDecoder := json.NewDecoder(bytes.NewBuffer(newBytes))
+		if ac.disallowUnknownFields {
+			newDecoder.DisallowUnknownFields()
+		}
+		if err := newDecoder.Decode(&newObj); err != nil {
+			return ctx, nil, fmt.Errorf("cannot decode incoming new object: %v", err)
```
vagababov, line 104:
```suggestion
			return ctx, nil, fmt.Errorf("cannot decode incoming new object: %w", err)
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+
+	var oldObj resourcesemantics.GenericCRD
+	if len(oldBytes) != 0 {
+		oldObj = handler.DeepCopyObject().(resourcesemantics.GenericCRD)
+		oldDecoder := json.NewDecoder(bytes.NewBuffer(oldBytes))
+		if ac.disallowUnknownFields {
+			oldDecoder.DisallowUnknownFields()
+		}
+		if err := oldDecoder.Decode(&oldObj); err != nil {
+			return ctx, nil, fmt.Errorf("cannot decode incoming old object: %v", err)
```
vagababov, line 116:
```suggestion
			return ctx, nil, fmt.Errorf("cannot decode incoming old object: %w", err)
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+		logger.Infof("Unhandled webhook validation operation, letting it through %v", req.Operation)
+		return nil
+	}
+
+	// None of the validators will accept a nil value for newObj.
+	if resource == nil {
+		return errMissingNewObject
+	}
+
+	if err := resource.Validate(ctx); err != nil {
```
vagababov, line 157: I'd move logging to the callsite (or just ignore, since it's being wrapped and passed up there as well) and just unify the paths into
`return resource.Validate(ctx)`.

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	if toDecode == nil {
+		logger := logging.FromContext(ctx)
+		logger.Errorf("No incoming object found: %v for verb %v", gvk, req.Operation)
+		return nil
+	}
+
+	// Generically callback if any are provided for the resource.
+	if c, ok := ac.callbacks[gvk]; ok {
+		if b, supported := c.SupportedVerbs[req.Operation]; supported && b {
```
vagababov, line 183:
```suggestion
		if c.SupportedVerbs[req.Operation] {
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,197 @@
...
+			unstruct := &unstructured.Unstructured{}
+			newDecoder := json.NewDecoder(bytes.NewBuffer(toDecode))
+			if err := newDecoder.Decode(&unstruct); err != nil {
+				return fmt.Errorf("cannot decode incoming new object: %w", err)
+			}
+
+			if err := c.Callback(ctx, unstruct); err != nil {
+				return err
+			}
```
vagababov, line 192:
```suggestion
			return c.Callback(ctx, unstruct)
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -76,6 +79,29 @@ var (
...
+			Version: "v1beta1",
+			Kind:    "Resource",
+		}: {
+			SupportedVerbs: map[webhook.Operation]bool{
+				webhook.Create: true,
+				webhook.Update: true,
+			},
+			Callback: resourceCallback},
+	}
+
```
vagababov, line 104: We're doing different layout usually
```suggestion
	callbacks = map[schema.GroupVersionKind]Callback{
		{
			Group:   "pkg.knative.dev",
			Version: "v1alpha1",
			Kind:    "Resource",
		}: {
			SupportedVerbs: map[webhook.Operation]bool{
				webhook.Create: true,
				webhook.Update: true,
			},
			Callback: resourceCallback
		},{
			Group:   "pkg.knative.dev",
			Version: "v1beta1",
			Kind:    "Resource",
		}: {
			SupportedVerbs: map[webhook.Operation]bool{
				webhook.Create: true,
				webhook.Update: true,
			},
			Callback: resourceCallback
		}}
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -241,6 +277,84 @@ func TestAdmitCreates(t *testing.T) {
 	}
 }
 
+func resourceCallback(ctx context.Context, uns *unstructured.Unstructured) error {
+
```
vagababov, line 281:
```suggestion
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -241,6 +277,84 @@ func TestAdmitCreates(t *testing.T) {
...
+	var resource Resource
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(uns.UnstructuredContent(), &resource); err != nil {
+		return err
+	}
+
+	if apis.IsDryRun(ctx) {
+		return fmt.Errorf("dryRun fail")
```
vagababov, line 288:
```suggestion
		return errors.New("dryRun fail")
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -241,6 +277,84 @@ func TestAdmitCreates(t *testing.T) {
...
+	}
+
+	if resource.Spec.FieldForCallbackValidation != "" &&
+		resource.Spec.FieldForCallbackValidation != "magic value" {
+		return fmt.Errorf(resource.Spec.FieldForCallbackValidation)
```
vagababov, line 293:
```suggestion
		return errors.New(resource.Spec.FieldForCallbackValidation)
```

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -19,6 +19,7 @@ package validation
 import (
 	"context"
 	"encoding/json"
```
mattmoor, line 21: Format Go code:
```suggestion
	"encoding/json"
	"errors"
```

mattmoor, line 21: Format Go code:
```suggestion
	"encoding/json"
	"errors"
```

mattmoor, line 21: Format Go code:
```suggestion
	"encoding/json"
	"errors"
```

webhook/resourcesemantics/validation/validation_admit.go
whaught: Renamed to Func. Thanks!

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,194 @@
...
+	"knative.dev/pkg/webhook/resourcesemantics"
+)
+
+var errMissingNewObject = errors.New("the new object may not be nil")
+
+// Callback is a generic function to be called by a consumer of validation
+type Callback struct {
+	Func func(ctx context.Context, unstructured *unstructured.Unstructured) error
+
+	SupportedVerbs map[webhook.Operation]bool
```
dprotaso, line 43: > I think keeping the map is desirable (really I want a Set) so you can't repeat verbs and makes it easier to not have to search the slice.

I still think the map is odd for folks constructing these callbacks - since you can set the value to false.

One approach I can think of:

Searching the slice can be avoided by processing and converting it to an internal map during construction of the webhook/controller. Validation can be done at that time as well (if need).


```go
// composite key that let's you flatten callbacks
type entry struct{
  gvk schema.GroupVersionKind
  op admissionv1beta1.Operation
}

type CallbackFunc func(ctx context.Context, *unstructured.Unstrucuted)error
...
// internal map
callbacks := map[entry]CallbackFunc
...
// invocation - minus error checks
cb := callbacks[entry{gvk, op}]
cb(ctx, unstructured)
```

thoughts?

vagababov: I'd just go with ctor
`NewCallback(f func(...)..., v ...webhook.Operation) *Callback`
And actually probably kept both private and and added `Do(unstructured)` method on the type.

whaught: I went with the second suggestion which seems simplest, but perhaps that's only because it's closer to what I already had (and allows the user to pass a list and we turn it into the Set internally only).

Let me know what you think

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,210 @@
...
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/webhook"
+	"knative.dev/pkg/webhook/resourcesemantics"
+)
+
+var errMissingNewObject = errors.New("the new object may not be nil")
+
+// Callback is a generic function to be called by a consumer of validation
+type Callback struct {
+	// funcion is the callback to be invoked
```
dprotaso, line 41:
```suggestion
	// function is the callback to be invoked
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -0,0 +1,210 @@
...
+	function func(ctx context.Context, unstructured *unstructured.Unstructured) error
+
+	// supportedVerbs are the verbs supported for the callback.
+	// The function will only be called on these acitons.
+	supportedVerbs map[webhook.Operation]bool
+}
+
+// NewCallback creats a new callback function to be invoked on supported vebs.
+func NewCallback(function func(context.Context, *unstructured.Unstructured) error, supportedVerbs []webhook.Operation) (c *Callback) {
```
dprotaso, line 50:
```suggestion
// NewCallback creates a new callback function to be invoked on supported vebs.
func NewCallback(function func(context.Context, *unstructured.Unstructured) error, supportedVerbs []webhook.Operation) (c Callback) {
```

You can drop the pointer to simplify things

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -38,9 +38,25 @@ var errMissingNewObject = errors.New("the new object may not be nil")
...
+	f func(ctx context.Context, unstructured *unstructured.Unstructured) error
 
-	SupportedVerbs map[webhook.Operation]bool
+	// supported are the verbs supported for the callback.
+	// The function will only be called on these acitons.
+	supported map[webhook.Operation]bool
+}
+
+// NewCallback creats a new callback function to be invoked on supported vebs.
+func NewCallback(function func(context.Context, *unstructured.Unstructured) error, supportedVerbs []webhook.Operation) (c *Callback) {
```
vagababov:
```suggestion
func NewCallback(function func(context.Context, *unstructured.Unstructured) error, supportedVerbs ...webhook.Operation) (c *Callback) {
```
I'd go this way.

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -38,9 +38,25 @@ var errMissingNewObject = errors.New("the new object may not be nil")
...
+	m := make(map[webhook.Operation]bool)
+	for _, op := range supportedVerbs {
+		if _, has := m[op]; has {
+			panic("duplicates verbs not allowed")
+		}
+		m[op] = true
+	}
+	c = &Callback{f: function, supported: m}
+	return c
```
vagababov:
```suggestion
	return &Callback{f: function, supported: m}
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -38,9 +38,25 @@ var errMissingNewObject = errors.New("the new object may not be nil")
...
 
-	SupportedVerbs map[webhook.Operation]bool
+	// supported are the verbs supported for the callback.
+	// The function will only be called on these acitons.
+	supported map[webhook.Operation]bool
+}
+
+// NewCallback creats a new callback function to be invoked on supported vebs.
+func NewCallback(function func(context.Context, *unstructured.Unstructured) error, supportedVerbs []webhook.Operation) (c *Callback) {
+	m := make(map[webhook.Operation]bool)
```
vagababov, line 51:
```suggestion
	m := make(map[webhook.Operation]struct{})
```
Doesn't really matter here but in general that's how go does sets.

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -38,9 +38,25 @@ var errMissingNewObject = errors.New("the new object may not be nil")
...
+	for _, op := range supportedVerbs {
+		if _, has := m[op]; has {
+			panic("duplicates verbs not allowed")
```
vagababov, line 54:
```suggestion
			panic("duplicate verbs not allowed")
```

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -179,14 +195,14 @@ func (ac *reconciler) callback(ctx context.Context, req *admissionv1beta1.Admiss
 
 	// Generically callback if any are provided for the resource.
 	if c, ok := ac.callbacks[gvk]; ok {
-		if b, supported := c.SupportedVerbs[req.Operation]; supported && b {
+		if b, supported := c.supported[req.Operation]; supported && b {
```
vagababov: you only put `true` in the map, so just presence, i..e `supported == true` is enough.

webhook/resourcesemantics/validation/validation_admit.go
whaught: sure. Same thing to me - I think I saw both bool and struct referenced on golang forums

webhook/resourcesemantics/validation/validation_admit.go
whaught: sure. perhaps easier to read

webhook/resourcesemantics/validation/validation_admit.go
whaught: I like it.

PR #1173, opened 2020-03-24
Introduce key/value store interface + cm backed version of it.

vaikas:
Especially in sources you might need to checkpoint work, or save some other state. Introduce a key/value store that's backed by ConfigMap to make it easier to work with that state.


kvstore/kvstore.go
```diff
@@ -0,0 +1,28 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kvstore
+
+type KVStore interface {
```
n3wscott: I would call this `Interface` the context comes from the kvstore package.

kvstore/kvstore.go
```diff
@@ -0,0 +1,28 @@
...
+	// Load loads the configstore from the backing store
+	Load() error
```
n3wscott: would you ever want to pass down a context.Context here or other methods? like, for logging... it is easy to add it early.

kvstore/kvstore_cm.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"knative.dev/pkg/logging"
+)
+
+type ConfigMapKVStore struct {
```
n3wscott: to you want to export this? None of the internals are exported.

kvstore/kvstore_cm.go
```diff
@@ -0,0 +1,123 @@
...
+	_ KVStore = (*ConfigMapKVStore)(nil)
+)
+
+func NewConfigMapKVStore(name string, namespace string, clientset v1.CoreV1Interface) *ConfigMapKVStore {
+
+	return &ConfigMapKVStore{name: name, namespace: namespace, cmClient: clientset.ConfigMaps(namespace)}
+}
+
+// Init initializes ConfigMapKVStore either by loading or creating an empty one.
+func (cs *ConfigMapKVStore) Init(ctx context.Context) error {
```
n3wscott: It would be nice to a) have Init in another interface next to `KVStore` or b) add Init to `KVStore`

kvstore/kvstore_cm.go
```diff
@@ -0,0 +1,123 @@
...
+	name      string
+	namespace string
+	data      map[string]string
+}
+
+var (
+	_ KVStore = (*ConfigMapKVStore)(nil)
+)
+
+func NewConfigMapKVStore(name string, namespace string, clientset v1.CoreV1Interface) *ConfigMapKVStore {
```
n3wscott: You might want to just return `KVStore` here.

vaikas: I was thinking that I didn't want to expose Init, but yeah, it's better. done.

n3wscott:
Out of the gate with 91% code coverage.

/LGTM

PR #1176, opened 2020-03-26
Implement WaitForEndpointSteadyState helper

mgencur:
* this helper waits until a condition is met for the given number of
times in a row
* this can be useful if requests go through multiple different
components (e.g. routers) and we want to wait until all the components
are ready and requests going through any of the components will succeed

PR for knative/serving will be sent shortly: TBD


mgencur:
Closing for now until I find a better solution.

PR #1177, opened 2020-03-27
[DO NOT SUBMIT] Benchmarking decoding options

whaught:
Benchmark Converting incoming requests from GenericCRD to unstructured.Unstructured vs simply decoding raw bytes twice.


webhook/resourcesemantics/validation/validation_test.go
```diff
@@ -185,6 +215,32 @@ func TestUnknownFieldFails(t *testing.T) {
 		`validation failed: cannot decode incoming new object: json: unknown field "foo"`)
 }
 
+func TestBenchDecode(t *testing.T) {
```
markusthoemmes: Why not use golang's benchmark tooling? https://golang.org/pkg/testing/#hdr-Benchmarks

whaught: This wasn't intended for public consumption so much as me testing a theory for https://github.com/knative/pkg/pull/1172

but I'm also a Go newbie and wasn't aware of the Benchmark package. Thanks - Today I Learned! I'll definitely check this out in the future.

whaught:
This was just intended to test alternatives. Not to be reviewed.

/close

PR #1179, opened 2020-03-29
Switch to Go modules

chizhg:
Switch pkg from Go dep to Go modules.

/hold


hack/tools.go
```diff
@@ -0,0 +1,25 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tools
+
+// This package imports things required by this repository, to force `go mod` to see them as dependencies
+import (
+    _ "knative.dev/test-infra/scripts"
+    _ "k8s.io/code-generator"
```
mattmoor: Format Go code:
```suggestion
	_ "k8s.io/code-generator"
	_ "knative.dev/test-infra/scripts"
```

n3wscott:
I would leave the toml and lock files until we can migrate all the other things too

go.mod
```diff
@@ -0,0 +1,78 @@
+module knative.dev/pkg
```
matzew, line 1: will this than be also done for all the oter knative repos ?

chizhg: The finally goal will be switching all Knative projects to use Go modules. In case you have concerns or questions, please join #productivity for discussions :)

chizhg:
> I would leave the toml and lock files until we can migrate all the other things too

Good point. My latest commit does not delete the two files.

hack/update-deps.sh
```diff
@@ -30,19 +30,24 @@ FLOATING_DEPS=(
 )
 
 # Parse flags to determine any we should pass to dep.
-DEP_FLAGS=()
+GO_GET=0
```
coryrc, line 33: `declare -i GO_GET=0` would be better

hack/update-deps.sh
```diff
@@ -30,19 +30,24 @@ FLOATING_DEPS=(
...
   esac
   shift
 done
-readonly DEP_FLAGS
+readonly GO_GET
 
-# Ensure we have everything we need under vendor/
-dep ensure ${DEP_FLAGS[@]}
+if (( GO_GET )); then
+  go get -d ${FLOATING_DEPS[@]}
```
coryrc, line 45: Can you run `shellcheck` on this file? I'm guessing the arg should be double-quoted

vaikas:
/lgtm
/approve

Let's get this in so we can unblock the dependent changes and reduce the breakage time

PR #1182, opened 2020-03-31
Fix the defaulting for the leader election config map

vagababov:
- Add defaulting for all but the component, which are system dependent
- Improve test coverage
- Clean up formatting
- Other test changes

/assign mattmoor @pmorie 

@pmorie , Paul do you have a bug for this cleanup?


vagababov:
Rebased :-) needs re-LGTM :)

PR #1187, opened 2020-04-02
Do not set to unknown status when all dependents passed happy check

nak3:
`accessor.GetConditions()` in `findUnhappyDependent()` gets status
filed from runtime. So, if an object has following status:

```
  status:
    address:
      url: http://hello-example.default.svc.cluster.local
    conditions:
    - lastTransitionTime: "2020-04-02T09:49:03Z"
      status: "True"
      type: AllTrafficAssigned
    - lastTransitionTime: "2020-04-02T09:49:03Z"
      message: autoTLS is not enabled
      reason: AutoTLSNotEnabled
      severity: Info
      status: "True"
      type: CertificateProvisioned
    - lastTransitionTime: "2020-04-02T09:49:05Z"
      status: "True"
      type: IngressReady
    - lastTransitionTime: "2020-04-02T09:49:05Z"
      status: "True"
      type: Ready
```

`accessor.GetConditions()` returns `AllTrafficAssigned`,
`CertificateProvisioned`, `IngressReady` (and `Ready`).

Now, current code (`len(r.dependents) != len(conditions)`) sets to
`Unknown` if all conditions are not unhappy. However, it does not work
if we added a new dependent and downgraded the cluster because
the number of dependents is less than current status conditions.

This patch changes to the condition check because if all dependents
are verified, it should not set to unknown status.

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -334,7 +334,7 @@ func (r conditionsImpl) findUnhappyDependent() *Condition {
 	}
 
 	// If something was not initialized.
-	if len(r.dependents) != len(conditions) {
+	if len(r.dependents) > len(conditions) {
 		return &Condition{
 			Status: corev1.ConditionUnknown,
 		}
--- a/apis/condition_set_impl_test.go
+++ b/apis/condition_set_impl_test.go
@@ -768,6 +768,45 @@ func TestMarkTrue(t *testing.T) {
 			Type:   ConditionReady,
 			Status: corev1.ConditionTrue,
 		},
+	}, {
+		name: "all happy but not cover all dependents",
+		conditions: Conditions{{
+			Type:    ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  "LongStory",
+			Message: "Set manually",
+		}, {
+			Type:   "Foo",
+			Status: corev1.ConditionTrue,
+		}},
+		mark:           "Foo",
+		conditionTypes: []ConditionType{"Foo", "Bar"}, // dependents is more than conditions.
+		happy:          false,
+		happyWant: &Condition{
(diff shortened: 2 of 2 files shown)


nak3:
Without this fix, we need to do staging release for a new dependents. which I am trying to do in https://github.com/knative/serving/pull/7163.

/assign @mattmoor

apis/condition_set.go
```diff
@@ -334,7 +334,7 @@ func (r conditionsImpl) findUnhappyDependent() *Condition {
 	}
 
 	// If something was not initialized.
-	if len(r.dependents) != len(conditions) {
+	if len(r.dependents) > len(conditions) {
```
mattmoor, line 337: I don't understand the purpose of this change, perhaps a unit test to illustrate the problem case?

apis/condition_set.go
```diff
@@ -334,7 +334,7 @@ func (r conditionsImpl) findUnhappyDependent() *Condition {
 	}
 
 	// If something was not initialized.
-	if len(r.dependents) != len(conditions) {
```
n3wscott, line 337: I think you need to write a few unit tests to confirm what you are seeing and make the correct logic change in `init`

If I read your description right, this wants to check if there are extra conditions that are not dependents. The change looks correct but I would like to see it tested.

nak3:
Thank you @mattmoor @n3wscott and sorry for missing unit test. Updated.

And yes, as @n3wscott mentioned, this wants to check if there are extra conditions that are not dependents. Without this fix, this test https://github.com/knative/pkg/pull/1187/files#diff-12b91f6470199861728218f25bc23aaeR790 fails.

chizhg:
@googlebot rescan

PR #1189, opened 2020-04-03
Retry to get logging configmap when failed

nak3:
When deploying pods with Istio sidecar, it takes a few seconds until
network is configured. In fact, Serving's system pods always fail when
Istio sidecar is enbled.

To fix it, this patch changes to retry to get logging configmap for 5
seconds.

Fixes https://github.com/knative/serving/issues/7486


injection/sharedmain/main.go
```diff
@@ -89,15 +90,27 @@ func GetConfig(masterURL, kubeconfig string) (*rest.Config, error) {
...
+		loggingConfigMap, err = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			} else {
+				log.Printf("Failed to get logging configmap: %v", err)
+				return false, nil
+			}
 		}
+		return true, nil
```
vagababov, line 107: `return err == nil || apierrors.IsNotFound(err), nil`

nak3:
@vagababov Sorry, my bad. Updated.

PR #1190, opened 2020-04-03
Allow a 'None' metrics backend, to basically do nothing.

markusthoemmes:
This is useful if you want to disable any surfacing of metrics whatsoever.


metrics/exporter.go
```diff
@@ -195,6 +195,8 @@ func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.
 		e, err = newStackdriverExporter(config, logger)
 	case Prometheus:
 		e, err = newPrometheusExporter(config, logger)
+	case None:
+		e, err = nil, nil
```
vagababov, line 199:
```suggestion
```

vagababov: They're already nil.

PR #1191, opened 2020-04-03
Fix a few simple linter warnings: unused code, govet warnings etc.

markusthoemmes:
Because deadcode is bad code!


chizhg:
/cc @coryrc 
For the test logging change.

PR #1192, opened 2020-04-03
Add augmented logger to the context to pass it down.

markusthoemmes:
Otherwise it's just unused (the reference is never picked up anywhere).

/assign @dprotaso

### DIFF
--- a/webhook/resourcesemantics/conversion/conversion.go
+++ b/webhook/resourcesemantics/conversion/conversion.go
@@ -109,11 +109,11 @@ func (r *reconciler) convert(
 	out := outZygote.DeepCopyObject().(ConvertibleObject)
 
 	hubGVK := inGVK.GroupKind().WithVersion(conv.HubVersion)
-	logger = logger.With(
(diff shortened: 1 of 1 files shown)


webhook/resourcesemantics/conversion/conversion.go
```diff
@@ -109,11 +109,11 @@ func (r *reconciler) convert(
 	out := outZygote.DeepCopyObject().(ConvertibleObject)
 
 	hubGVK := inGVK.GroupKind().WithVersion(conv.HubVersion)
-	logger = logger.With(
+	ctx = logging.WithLogger(ctx, logger.With(
```
vagababov, line 112: Given this logger was not used at all, may be we just delete it?

markusthoemmes: @dprotaso will know which way is best I'd hope.

mattmoor: Thanks, I had vaguely recalled there being one of these.  In general, I think this is probably the right change.

PR #1208, opened 2020-04-09
Sinkbinding Webhook should label namespace for inclusion

lberk:
By labelling a target namespace `bindings.knative.dev/inclusion: true` we can then set the psbinding webhook to exlusion and not have the webhook hit every single request to the apiserver (but still allow the psbinding requests we *want* to reconcile).

### DIFF
--- a/webhook/psbinding/reconciler.go
+++ b/webhook/psbinding/reconciler.go
@@ -43,6 +43,12 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
+var jsonLabelPatch = map[string]interface{}{
+	"metadata": map[string]interface{}{
+		"labels": map[string]string{"bindings.knative.dev/include": "true"},
+	},
+}
+
 // BaseReconciler helps implement controller.Reconciler for Binding resources.
 type BaseReconciler struct {
 	// The GVR of the "primary key" resource for this reconciler.
@@ -232,29 +238,30 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
 func (r *BaseReconciler) labelNamespace(ctx context.Context, subject tracker.Reference) error {
 
 	// Determine the GroupVersionResource of the subject reference
-	gv, err := schema.ParseGroupVersion("v1")
-	if err != nil {
-		logging.FromContext(ctx).Infof("Error parsing GroupVersion %v: %v", "v1", err)
-		return err
+	gvr := schema.GroupVersionResource{
+		Group:    "",
+		Version:  "v1",
+		Resource: "namespaces",
 	}
-	gvr := apis.KindToResource(gv.WithKind("Namespace"))
 
 	// Don't remove any existing labels or add ours more than once
-	unstructuredLabels, err := r.DynamicClient.Resource(gvr).Get(subject.Namespace, metav1.GetOptions{})
+	_, namespaceLister, err := r.Factory.Get(gvr)
 	if err != nil {
-		logging.FromContext(ctx).Infof("Error getting labels from namespace: %s: %v", subject.Namespace, err)
+		logging.FromContext(ctx).Infof("Error getting lister for namespaces: '%v': %v", gvr, err)
+		return err
 	}
-	if unstructuredLabels != nil {
-		labels := unstructuredLabels.GetLabels()
-		if labels["bindings.knative.dev/exclude"] != "" || labels["bindings.knative.dev/include"] != "" {
-			return nil
-		}
+	namespaceObject, err := namespaceLister.Get(subject.Namespace)
+	if apierrs.IsNotFound(err) {
+		logging.FromContext(ctx).Infof("")
+		return err
+	} else if err != nil {
+		return err
 	}
-	patch, err := json.Marshal(map[string]interface{}{
-		"metadata": map[string]interface{}{
-			"labels": map[string]string{"bindings.knative.dev/include": "true"},
-		},
-	})
+	labels := (namespaceObject.(*duckv1.WithPod)).GetLabels()
+	if labels["bindings.knative.dev/exclude"] != "" || labels["bindings.knative.dev/include"] != "" {
+		return nil
+	}
+	patch, err := json.Marshal(jsonLabelPatch)
 	if err != nil {
 		logging.FromContext(ctx).Infof("Error generating json patch: %v, to namespace: %s", err, subject.Namespace)
 		return nil
@@ -262,6 +269,7 @@ func (r *BaseReconciler) labelNamespace(ctx context.Context, subject tracker.Ref
 	_, err = r.DynamicClient.Resource(gvr).Patch(subject.Namespace, types.MergePatchType, patch, metav1.PatchOptions{})
 	if err != nil {
 		logging.FromContext(ctx).Infof("Error applying patch to namespace: %s: %v", subject.Namespace, err)
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


matzew:
@mattmoor  any thoughts on this ?

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
 
+func (r *BaseReconciler) labelNamespace(ctx context.Context, subject tracker.Reference) error {
+
+	// Determine the GroupVersionResource of the subject reference
+	gv, err := schema.ParseGroupVersion("v1")
+	if err != nil {
+		logging.FromContext(ctx).Infof("Error parsing GroupVersion %v: %v", "v1", err)
+		return err
+	}
+	gvr := apis.KindToResource(gv.WithKind("Namespace"))
```
mattmoor, line 240:
```suggestion
	gvr := schema.GroupVersionResource{
		Group: "",
		Version: "v1",
		Resource: "namespaces",
	}
```

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+
+	// Don't remove any existing labels or add ours more than once
+	unstructuredLabels, err := r.DynamicClient.Resource(gvr).Get(subject.Namespace, metav1.GetOptions{})
```
mattmoor, line 243: This should probably be using a lister from a namespace informer.

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+	if err != nil {
+		logging.FromContext(ctx).Infof("Error getting labels from namespace: %s: %v", subject.Namespace, err)
```
mattmoor, line 245: return here?

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+	}
+	patch, err := json.Marshal(map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"labels": map[string]string{"bindings.knative.dev/include": "true"},
+		},
+	})
+	if err != nil {
+		logging.FromContext(ctx).Infof("Error generating json patch: %v, to namespace: %s", err, subject.Namespace)
+		return nil
+	}
```
mattmoor, line 261: This is effectively a constant

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+	_, err = r.DynamicClient.Resource(gvr).Patch(subject.Namespace, types.MergePatchType, patch, metav1.PatchOptions{})
+	if err != nil {
+		logging.FromContext(ctx).Infof("Error applying patch to namespace: %s: %v", subject.Namespace, err)
```
mattmoor, line 264: return err?

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +229,44 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+		return err
+	}
+	gvr := apis.KindToResource(gv.WithKind("Namespace"))
+
+	// Don't remove any existing labels or add ours more than once
+	unstructuredLabels, err := r.DynamicClient.Resource(gvr).Get(subject.Namespace, metav1.GetOptions{})
+	if err != nil {
+		logging.FromContext(ctx).Infof("Error getting labels from namespace: %s: %v", subject.Namespace, err)
+	}
+	if unstructuredLabels != nil {
```
mattmoor, line 247: We shouldn't need to deal with this case 🤔

webhook/psbinding/table_test.go
```diff
@@ -1926,6 +1993,31 @@ func mustTU(t *testing.T, ro duck.OneOfOurs) *unstructured.Unstructured {
...
+	}
+	action.ActionImpl = actionImpl
+	action.Name = namespace
+	action.PatchType = types.MergePatchType
+
+	patch, _ := json.Marshal(map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"labels": map[string]string{"bindings.knative.dev/include": "true"},
+		},
+	})
```
mattmoor, line 2017: This could share the ~constant I pointed our above.

webhook/psbinding/reconciler.go
```diff
@@ -232,36 +238,38 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+		return err
+	} else if err != nil {
+		return err
 	}
-	patch, err := json.Marshal(map[string]interface{}{
-		"metadata": map[string]interface{}{
-			"labels": map[string]string{"bindings.knative.dev/include": "true"},
-		},
-	})
+	labels := (namespaceObject.(*duckv1.WithPod)).GetLabels()
```
mattmoor, line 260: This is sort of abusing the duck factory because it's not a WithPod.  I was thinking we'd setup a namespace informer in the way our other reconcilers do this and store the lister on the reconciler.

webhook/psbinding/reconciler.go
```diff
@@ -232,36 +238,38 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+		return err
 	}
-	if unstructuredLabels != nil {
-		labels := unstructuredLabels.GetLabels()
-		if labels["bindings.knative.dev/exclude"] != "" || labels["bindings.knative.dev/include"] != "" {
-			return nil
-		}
+	namespaceObject, err := namespaceLister.Get(subject.Namespace)
+	if apierrs.IsNotFound(err) {
+		logging.FromContext(ctx).Infof("")
```
mattmoor, line 255: Empty log statement?

webhook/psbinding/reconciler.go
```diff
@@ -43,6 +43,12 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
+var jsonLabelPatch = map[string]interface{}{
+	"metadata": map[string]interface{}{
+		"labels": map[string]string{"bindings.knative.dev/include": "true"},
```
matzew, line 48: Want to use https://github.com/knative/pkg/blob/master/apis/duck/const.go#L32  constant, instead of this string?

webhook/psbinding/reconciler.go
```diff
@@ -229,6 +235,46 @@ func (r *BaseReconciler) RemoveFinalizer(ctx context.Context, fb kmeta.Accessor)
...
+	}
+	namespaceObject, err := namespaceLister.Get(subject.Namespace)
+	if apierrs.IsNotFound(err) {
+		logging.FromContext(ctx).Infof("")
+		return err
+	} else if err != nil {
+		return err
+	}
+	labels := (namespaceObject.(*duckv1.WithPod)).GetLabels()
+	if labels["bindings.knative.dev/exclude"] != "" || labels["bindings.knative.dev/include"] != "" {
```
matzew, line 261: wanna use the constants here too:

* https://github.com/knative/pkg/blob/master/apis/duck/const.go#L26
* https://github.com/knative/pkg/blob/master/apis/duck/const.go#L32

webhook/psbinding/table_test.go
```diff
@@ -1229,6 +1236,12 @@ func TestBaseReconcile(t *testing.T) {
 					},
 				},
 			},
+			&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "foo",
+					Labels: map[string]string{"bindings.knative.dev/include": "true"},
```
matzew, line 1242: constanst?

lberk:
@mattmoor I think that seems reasonable to me -- would 2. be something we could perhaps address in a follow up PR?

I recognize the limitation of leaving the label, would you be open to me creating an issue on that as well to ensure it doesn't get forgotten? Maybe when removing the finalizer we can check if there are any other `psobj`'s in the `subject.Namespace` and remove the label then (assuming none are left)?

mattmoor:
/lgtm
/approve

Yeah, a follow-up and issue SGTM.  We can discuss options in the issue?

PR #1218, opened 2020-04-10
Add callback operations to ValidatingWebhookConfiguration rules

### LINKED ISSUE #1207: Validating Webhook Configuration rules should include callback operations
## Expected Behavior
The `Rules` field in `ValidatingWebhookConfiguration` should include additional verbs supported in callbacks (such as the `Delete` operation), along with the default verbs `Create` and `Update`.

## Actual Behavior
The `ValidatingWebhookConfiguration` only contains rules for `Create` and `Update`, so requests with a `Delete` operation are never validated even if they have a callback.

1. Create a validating admission controller that includes a callback to validate a CRD on deletion
2. Send a `DELETE` request to that CRD
3. No validation occurs

## Additional Info
I believe this can be fixed by passing in the supported verbs in each `Callback` to the `reconcileValidatingWebhook` method: https://github.com/knative/pkg/blob/master/webhook/resourcesemantics/validation/reconcile_config.go#L89
That way, if a callback supports a Delete operation, the webhook configuration can include the Delete operation in its `Rules` list. Otherwise, Delete requests would be filtered out before reaching the callback.

I got the impression that delete validations are currently intended to be supported, based on the handling of the Delete operation in the callback logic: https://github.com/knative/pkg/blob/master/webhook/resourcesemantics/validation/validation_admit.go#L184

juliaguo:
Fixes #1207

Add the supported operations in reconciler callbacks to the webhook configuration. Right now, DELETE requests are intended to be supported by callbacks, but never make it to the `Admit()` method of the admission controller, since the webhook configuration only includes `CREATE` and `UPDATE` operations in its rules.

### DIFF
--- a/webhook/resourcesemantics/validation/reconcile_config.go
+++ b/webhook/resourcesemantics/validation/reconcile_config.go
@@ -93,7 +93,7 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 	for gvk := range ac.handlers {
 		plural := strings.ToLower(inflect.Pluralize(gvk.Kind))
 
-		rules = append(rules, admissionregistrationv1beta1.RuleWithOperations{
+		newRule := admissionregistrationv1beta1.RuleWithOperations{
 			Operations: []admissionregistrationv1beta1.OperationType{
 				admissionregistrationv1beta1.Create,
 				admissionregistrationv1beta1.Update,
@@ -103,7 +103,20 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 				APIVersions: []string{gvk.Version},
 				Resources:   []string{plural + "/*"},
 			},
-		})
+		}
+
+		// Add supported verbs from callbacks to the rule operations
+		callback, ok := ac.callbacks[gvk]
+		if ok {
+			for verb := range callback.supportedVerbs {
+				operationType := admissionregistrationv1beta1.OperationType(verb)
+				if operationType != admissionregistrationv1beta1.Create && operationType != admissionregistrationv1beta1.Update {
+					newRule.Operations = append(newRule.Operations, operationType)
+				}
+			}
+		}
+
+		rules = append(rules, newRule)
 	}
 
 	// Sort the rules by Group, Version, Kind so that things are deterministically ordered.
--- a/webhook/resourcesemantics/validation/reconcile_config.go
+++ b/webhook/resourcesemantics/validation/reconcile_config.go
@@ -106,8 +106,7 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 		}
 
 		// Add supported verbs from callbacks to the rule operations
-		callback, ok := ac.callbacks[gvk]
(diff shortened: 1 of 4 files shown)


whaught:
Thanks Julia! It was my intention to look at other verbs in the future and now we should be able to configure the config properly. This lgtm and thanks for correcting my bad comments!

/assign @dprotaso

webhook/resourcesemantics/validation/reconcile_config.go
```diff
@@ -103,7 +103,20 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 				APIVersions: []string{gvk.Version},
 				Resources:   []string{plural + "/*"},
 			},
-		})
+		}
+
+		// Add supported verbs from callbacks to the rule operations
+		callback, ok := ac.callbacks[gvk]
+		if ok {
```
whaught, line 110: if callback, ok := ac.callbacks[gvk]; ok {

juliaguo: done

webhook/resourcesemantics/validation/reconcile_config.go
```diff
@@ -103,7 +103,19 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 				APIVersions: []string{gvk.Version},
 				Resources:   []string{plural + "/*"},
 			},
-		})
+		}
+
+		// Add supported verbs from callbacks to the rule operations
+		if callback, ok := ac.callbacks[gvk]; ok {
```
mattmoor, line 109: As I said in chat, I'd like to see us address the whole problem here, and not just have a band-aid.  We should expand when `Validate` is called to include Delete, improve the context tagging, and ensure `oldObj` is the receiver to Validate calls.

whaught: I think this is a step in the right direction.

Here is where we decode oldObj for the delete action:
https://github.com/knative/pkg/blob/4e57475bc87c1aba3e7fcc0207b593dc7186eb83/webhook/resourcesemantics/validation/validation_admit.go#L185

whaught: https://github.com/knative/pkg/pull/1219

I think this is the rest of what we want here

juliaguo: Thank you @whaught! @mattmoor would #1219 + this PR be sufficient?

mattmoor: looks like I hadn't hit submit on my comments on the other PR 🤦 

I think the other PR makes a majority of this PR not necessary.  I think the only bit it's missing is updating the webhook to include Delete, which can be simpler than what's in this PR because this is dynamic and with the other PR it can still be a fixed set.

juliaguo: That sounds good to me, the other PR can just add the Delete operation statically

whaught:
/close

This is obsolete in favor of #1219

PR #1219, opened 2020-04-11
Allow Delete verb on validation callbacks

whaught:
Issue #1207
Issue  #1140

**Proposed Changes**

- Allow Delete verb on validation callbacks
- Include IsInDelete on the context
    - Note: I'm not a huge fan of the pattern of IfInVerb on context  rather than just storing the verb. It robs us of the ability to Switch on verb and  allows for the possibility of more than one set. That said, its how it is.
- Test callback on a delete verb

### DIFF
--- a/apis/contexts_test.go
+++ b/apis/contexts_test.go
@@ -49,6 +49,21 @@ func TestContexts(t *testing.T) {
 		check: IsInCreate,
 		want:  false,
 	}, {
+		name:  "is in delete",
+		ctx:   WithinDelete(ctx),
+		check: IsInDelete,
+		want:  true,
+	}, {
+		name:  "not in delete (bare)",
+		ctx:   ctx,
+		check: IsInDelete,
+		want:  false,
+	}, {
+		name:  "not in delete (create)",
+		ctx:   WithinCreate(ctx),
+		check: IsInDelete,
+		want:  false,
+	}, {
 		name:  "is in update",
 		ctx:   WithinUpdate(ctx, struct{}{}),
 		check: IsInUpdate,
@@ -144,7 +159,7 @@ func TestContexts(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			got := tc.check(tc.ctx)
 			if tc.want != got {
-				t.Errorf("check() = %v, wanted %v", tc.want, got)
+				t.Errorf("check() = %v, wanted %v", got, tc.want)
 			}
 		})
 	}
--- a/webhook/resourcesemantics/validation/validation_admit.go
+++ b/webhook/resourcesemantics/validation/validation_admit.go
@@ -161,6 +161,8 @@ func validate(ctx context.Context, resource resourcesemantics.GenericCRD, req *a
 	switch req.Operation {
 	case admissionv1beta1.Create, admissionv1beta1.Update:
 		// Supported verbs
+	case admissionv1beta1.Delete:
+		return nil // Validation handled by optional Callback, but not validatable.
 	default:
 		logger.Infof("Unhandled webhook validation operation, letting it through %v", req.Operation)
 		return nil
--- a/webhook/resourcesemantics/validation/validation_admit.go
+++ b/webhook/resourcesemantics/validation/validation_admit.go
@@ -131,6 +131,16 @@ func (ac *reconciler) decodeRequestAndPrepareContext(
 		}
 	}
 
+	ctx = apis.WithUserInfo(ctx, &req.UserInfo)
+	ctx = context.WithValue(ctx, kubeclient.Key{}, ac.client)
+	if req.DryRun != nil && *req.DryRun {
+		ctx = apis.WithDryRun(ctx)
+	}
+
+	if newObj != nil && oldObj != nil && req.SubResource == "" {
+		ctx = apis.WithinSubResourceUpdate(ctx, oldObj, req.SubResource)
+	}
+
 	switch req.Operation {
 	case admissionv1beta1.Update:
 		ctx = apis.WithinUpdate(ctx, oldObj)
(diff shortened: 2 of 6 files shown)


apis/contexts_test.go
```diff
@@ -144,7 +159,7 @@ func TestContexts(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			got := tc.check(tc.ctx)
 			if tc.want != got {
-				t.Errorf("check() = %v, wanted %v", tc.want, got)
+				t.Errorf("check() = %v, wanted %v", got, tc.want)
```
mattmoor, line 162: whoops, good catch 😂

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -131,19 +131,21 @@ func (ac *reconciler) decodeRequestAndPrepareContext(
...
+	if newObj != nil && oldObj != nil && req.SubResource == "" {
+		ctx = apis.WithinSubResourceUpdate(ctx, oldObj, req.SubResource)
+	}
+
 	ctx = apis.WithUserInfo(ctx, &req.UserInfo)
 	ctx = context.WithValue(ctx, kubeclient.Key{}, ac.client)
-
 	if req.DryRun != nil && *req.DryRun {
 		ctx = apis.WithDryRun(ctx)
 	}
```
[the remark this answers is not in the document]
whaught: Sure we can do that!

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -131,19 +131,21 @@ func (ac *reconciler) decodeRequestAndPrepareContext(
...
+	switch req.Operation {
+	case admissionv1beta1.Update:
+		ctx = apis.WithinUpdate(ctx, oldObj)
+	case admissionv1beta1.Create:
 		ctx = apis.WithinCreate(ctx)
+	case admissionv1beta1.Delete:
+		ctx = apis.WithinDelete(ctx)
 	}
+
+	if newObj != nil && oldObj != nil && req.SubResource == "" {
```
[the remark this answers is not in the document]
whaught: I don't think so, although the presence of both old and new implies update. The unit test doesn't assume that.

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -131,19 +131,21 @@ func (ac *reconciler) decodeRequestAndPrepareContext(
...
-		} else {
-			ctx = apis.WithinSubResourceUpdate(ctx, oldObj, req.SubResource)
-		}
-	} else {
+	switch req.Operation {
+	case admissionv1beta1.Update:
+		ctx = apis.WithinUpdate(ctx, oldObj)
+	case admissionv1beta1.Create:
 		ctx = apis.WithinCreate(ctx)
+	case admissionv1beta1.Delete:
```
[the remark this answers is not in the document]
whaught: Sure we can add that statically. It's easier to read. The other PR is clever and dynamically adds it only if there is a hook which supports delete - maybe that's fancier than we really need anyway.

dprotaso:
This log will get printed when deletes happen - can we just add a new case and return?

https://github.com/knative/pkg/blob/478afaf36495e3bafd447d7a61b19c073848dd3c/webhook/resourcesemantics/validation/validation_admit.go#L161-L167

whaught:
> This log will get printed when deletes happen - can we just add a new case and return?
> 
> https://github.com/knative/pkg/blob/478afaf36495e3bafd447d7a61b19c073848dd3c/webhook/resourcesemantics/validation/validation_admit.go#L161-L167

Sure we can add a return and skip the log line, It's sort of an odd log since only Create/Update are in the config for the webhook (and now delete). Presumably it currently never gets hit.

dprotaso:
cc @mattmoor for approval

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -158,6 +161,8 @@ func validate(ctx context.Context, resource resourcesemantics.GenericCRD, req *a
 	switch req.Operation {
 	case admissionv1beta1.Create, admissionv1beta1.Update:
 		// Supported verbs
+	case admissionv1beta1.Delete:
+		return nil // Validation handled by optional Callback, but not validatable.
```
mattmoor, line 165: 🤔 I don't understand the goal of excluding this.  We return `oldObj` above, so that we will get the moral equivalent of:
```go
oldObj.Validate(apis.WithinDelete(ctx))
```

This lets us add code to `Validate()` that can say:
```go
if apis.IsInDelete(ctx) {
    // Check for reasons to reject Delete
}
```

whaught: the various validate() calls are not known to be compatible with Delete yet and might break. I would prefer to knock that out in a different PR and assess that consumers of /pkg work with that potentially unsafe change.

whaught: This specific case was added in response to @dprotaso who worried the log line is misleading since Callbacks may be registered against delete.

mattmoor: I'm fine with following up, but the point of returning `oldObj` was to make it so that `.Validate()` is a nop for implementations that haven't seen this stuff before.

It is possible that if they assume it is either create or update that the introduction of delete could still be problematic, but that seems more niche than downstream things needing to handle a nil receiver on Delete calls, which would blow up *everywhere* 😅

whaught: that's fair... the oldObj should still be valid against, say, create.

Let's get this in and I'll patch /serving with this line knocked out (and run the tests). If everything tests fine I'll drop that line in another PR

PR #1229, opened 2020-04-21
Release leader election lock as soon as we exit.

markusthoemmes:
This releases the lease as soon as the provided context is cancelled.

While we're technically not quite done exiting once we see the cancel signal, the time overlap should be good enough as we'll stop the controller as soon as we see the cancel signal and the newly acquired leaders need to launch their informers etc. first. That should be close enough to being already shut down.

In the worst worst case, we have a very brief time of two controllers stamping each other. I think that's seldom enough to not warrant more orchestration logic.

Note: There's currently a bug in releasing the lock successfully, being fixed here: https://github.com/kubernetes/kubernetes/pull/88192

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -474,6 +474,7 @@ func RunLeaderElected(ctx context.Context, logger *zap.SugaredLogger, run func(c
 				logger.Fatal("leaderelection lost")
 			},
 		},
+		ReleaseOnCancel: true,
 		// TODO: use health check watchdog, knative/pkg#1048
 		Name: component,
 	})


mattmoor:
@markusthoemmes FYI there is also a data race that is fixed in the 1.18 client: https://github.com/kubernetes/kubernetes/pull/87899

Basically the background `renew` races with the `release` on cancel because `ctx` isn't threaded into the `tryAcquireOrRenew`.

The data race detection is (intermittently) freaking out about this in some of the unit testing I added for per-reconciler leader election: https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_pkg/1303/pull-knative-pkg-unit-tests/1257806940013072384/build-log.txt

PR #1230, opened 2020-04-21
Add a controller option to specify a custom finalizer name.

### LINKED ISSUE #1227: genreconciler should be able to override the finalizer name
## Expected Behavior

The MT GitHub source sets 2 finalizers, one by the control-plane component to delete the external webhook and one by the data-plane component to make sure the internal cache is properly synchronized.

## Actual Behavior

Only one finalizer can be set.

1.
2.
3.

markusthoemmes:
Adds an option to the controller to optionally override the finalizer name to use in the generated reconciler.

Fixes #1227.

**Note:** This is a quick hackup without actually bikeshedding it with peeps. Feel free to bin if this is entirely the wrong direction.

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/controller.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/controller.go
@@ -74,10 +74,11 @@ func NewImpl(ctx context.Context, r Interface, optionsFns ...controller.OptionsF
 	}
 
 	rec := &reconcilerImpl{
-		Client:     apiextensionsclient.Get(ctx),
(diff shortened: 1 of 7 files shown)


n3wscott:
/lgtm
/approve

This is exactly why I made the options like that 😄

grantr:
I think this is causing unit test failures on https://github.com/knative/eventing/pull/3039. Seems like some resources are now getting finalizers when they were not before.

mattmoor:
Good catch, Grant!

grantr:
Well thanks, but it was actually unit tests that caught it :grin:

mattmoor:
You tracked it back here! 😉

PR #1234, opened 2020-04-22
Drop 'no longer exist' line to DEBUG.

markusthoemmes:
These lines are printed fairly heavily, especially when we delete resources without issues. They definitely are not error worthy at this point and given their misleading nature, DEBUG might be best.


markusthoemmes:

Unrelated.

PR #1235, opened 2020-04-22
Add specific callback for the tracker to delete items eagerly.

markusthoemmes:
Currently, eviction of keys is timing based which results in unexpected enqueues of items that no longer exist, especially if you create, delete, recreate objects quickly.

This adds a specific `OnDelete` function to the tracker to eagerly delete items and adds helpers to create a respective event handler.

### DIFF
--- a/reconciler/testing/tracker.go
+++ b/reconciler/testing/tracker.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"knative.dev/pkg/kmeta"
 	"knative.dev/pkg/tracker"
 )
 
@@ -31,7 +33,7 @@ type NullTracker = FakeTracker
 // FakeTracker implements Tracker.
 type FakeTracker struct {
 	sync.Mutex
-	references []tracker.Reference
+	references map[tracker.Reference]map[types.NamespacedName]struct{}
 }
 
 var _ tracker.Interface = (*FakeTracker)(nil)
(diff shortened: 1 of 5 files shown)


markusthoemmes:
I think as implemented, this is garbage.

I think I need to inverse the relationship and not watch of the references being deleted but rather the parents being deleted.

From myself in Slack:
> So I'd need an OnDeletedObject which runs through all the matcher arrays and removes the actual objects from being enlisted. That becomes a little hairy because I need to register that with the other informers, mainly the "main" ones, i.e. the "route" in the route reconciler while the tracker is driven by the config and revision reconcilers

Gonna look at this with fresh eyes tomorrow.

markusthoemmes:
Reworked to have a callback that's to be called by the "parent" informer. Example usage in https://github.com/knative/serving/pull/7679

tracker/interface.go
```diff
@@ -70,6 +70,11 @@ type Interface interface {
 	// OnChanged is a callback to register with the InformerFactory
 	// so that we are notified for appropriate object changes.
 	OnChanged(obj interface{})
+
+	// OnDeletedParent is a callback to register with the InformerFactory
+	// so that we are notified for deletions of a watching parent to
+	// remove the respective tracking.
+	OnDeletedParent(obj interface{})
```
mattmoor, line 77: Going to paint the shed!

To me "Parent" evokes a controller relationship (e.g. this resource produced the resources being watched), which may be the case, but isn't always (e.g. Route -> Config/Rev, or ObjectRefs).

I think the name I'd use is probably `Observer` to avoid the implied hierarchical relationship.  WDYT?

markusthoemmes: SG!

reconciler/testing/tracker.go
```diff
@@ -39,6 +39,9 @@ var _ tracker.Interface = (*FakeTracker)(nil)
 // OnChanged implements OnChanged.
 func (*FakeTracker) OnChanged(interface{}) {}
 
+// OnDeletedParent implements OnDeletedParent.
+func (*FakeTracker) OnDeletedParent(interface{}) {}
```
mattmoor: should this delete from `n.references`?

markusthoemmes:
Yeah that was when I thought these would be on one informer. With the necessary split, there seems to be little gain in helpers.

```golang
routeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
	DeleteFunc: c.tracker.OnDeletedParent,
})
```

Is what the boilerplate looks like.

reconciler/testing/tracker.go
markusthoemmes: Good point, it likely should. I'll rearrange things.

Issue #1137, opened 2020-03-03
genreconciler should be able to override the controller agent name


dprotaso:
## Expected Behavior
The [Serving's GC reconciler](https://github.com/knative/serving/blob/b79fef014aaf1dd7bc06574cd2bfdc08a2aaad0f/pkg/reconciler/gc/controller.go#L36) has a controller agent name of [configuration-controller](https://github.com/knative/serving/blob/master/pkg/client/injection/reconciler/serving/v1/configuration/controller.go#L38)

It's used in the event broadcaster [here](https://github.com/knative/serving/blob/master/pkg/client/injection/reconciler/serving/v1/configuration/controller.go#L67) which makes the source of the events confusing.

## Actual Behavior
We should be able to override the the controller agent name so side-effects (logs & events) are have a clear origin

andrew-su:
/reopen

The change is only in pkg right now, still need to be updated in serving

dprotaso:
this was done in https://github.com/knative/serving/pull/7116

PR #1240, opened 2020-04-23
Add ability to override controller agent name

shashwathi:
fixes https://github.com/knative/pkg/issues/1137

cc @andrew-su

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_controller.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller.go
@@ -138,6 +138,10 @@ func (g *reconcilerControllerGenerator) GenerateType(c *generator.Context, t *ty
 			Package: "knative.dev/pkg/controller",
 			Name:    "OptionsFn",
 		}),
+		"controllerGetControllerAgentName": c.Universe.Function(types.Name{
+			Package: "knative.dev/pkg/controller",
+			Name:    "GetControllerAgentName",
+		}),
 		"contextContext": c.Universe.Type(types.Name{
 			Package: "context",
 			Name:    "Context",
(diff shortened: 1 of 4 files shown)


andrew-su:
@googlebot I consent.

andrew-su:
`optionsFns` is executed at the end of the controller implementation. The name change is to affect the `EventRecorder` which is used by the `reconcilerImpl`. Our option would be to iterate over the option functions once at the beginning and again at the end. Thoughts?

markusthoemmes:
@andrew-su alternatively, you can also delay the recorder creation until after the options are applied at the end. Just remove it from the struct initialization and set it at the end via `r.recorder = ...`. I'd avoid two places to do stuff with options honestly.

andrew-su:
@markusthoemmes Thanks for the suggestion. I definitely want to avoid doing options twice. Was putting it out there in hopes of hearing a better suggestion.

shashwathi:
Serving PR to show the changes: https://github.com/knative/serving/pull/7697
Testing grid link : https://prow.knative.dev/pr-history/?org=knative&repo=serving&pr=7697
Latest commit on the serving PR  has passed all the tests.

andrew-su:
@n3wscott @markusthoemmes Could you please take another look?

n3wscott:
Looks great! thanks

/lgtm
/approve

codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -212,11 +192,43 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 		if opts.FinalizerName != "" {
 			rec.finalizerName = opts.FinalizerName
 		}
+		if opts.AgentName != "" {
+			rec.Recorder = createRecorder(ctx, opts.AgentName)
```
mattmoor, line 196: nit: I'd declare a string `agentName := defaultControllerAgentName` above the loop, set that here, and remove the if below.

shashwathi: We still need if condition below because that condition checks if recorder is `nil`.

PR #1247, opened 2020-04-25
Change StartAll to take context.

mattmoor:
This has bugged me since we started using `ctx`, which containers a `stopCh` of sorts as `Done()`.  This is somewhat for consistency, but by using `ctx` explicitly we enable ourselves to take advantage of more contextual information.

I did a quick scan of call sites under knative.dev and github.com/tektoncd on my local GOPATH (I have many repos cloned) and the good news is that the `sharedmain` change should be the place through which the vast majority of calls occur, however, the one outlier I found is the KPA which calls this manually.  I will stage a PR to manually import pkg into serving to fix this once this lands.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -330,9 +330,10 @@ func (c *Impl) EnqueueKeyAfter(key types.NamespacedName, delay time.Duration) {
 }
 
 // Run starts the controller's worker threads, the number of which is threadiness.
-// It then blocks until stopCh is closed, at which point it shuts down its internal
-// work queue and waits for workers to finish processing their current work items.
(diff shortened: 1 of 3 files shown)


mattmoor:
/hold

I was focused on the `StartAll` breaking change and didn't think through the `Run` breaking change 🤦

mattmoor:
/hold

My PR pulling this (and other stuff) into serving is crash looping things.  I'm gonna take a deeper look.

mattmoor:
/unhold

Ok the downstream problem isn't this, it's the logging change, which I'm reverting in https://github.com/knative/pkg/pull/1248

mattmoor:
I rebased this and I'm gonna update the downstream PR to make sure things are good now.

PR #1239, opened 2020-04-23
Use loggers when emitting count messages

mattysweeps:
Fixes #1225

https://github.com/knative/pkg/blob/08b31df9e66a7ed7dbbfaf65de817f09d66cb79d/injection/sharedmain/main.go#L139
https://github.com/knative/pkg/blob/08b31df9e66a7ed7dbbfaf65de817f09d66cb79d/injection/sharedmain/main.go#L216

These lines:
```
log.Printf("Registering %d clients", len(injection.Default.GetClients()))
log.Printf("Registering %d informer factories", len(injection.Default.GetInformerFactories()))
log.Printf("Registering %d informers", len(injection.Default.GetInformers()))
log.Printf("Registering %d controllers", len(ctors))
```

Should use the zap logger, instead of printing directly to stderr

PR #1248, opened 2020-04-25
Revert "Use loggers when emitting count messages"

mattmoor:
Reverts knative/pkg#1239

Line 151 is what sets up the clients on context, so the way this is set up fetching the kubeclient fails and we `OrDie` all the time. 🤦


mattmoor:
cc @mattysweeps

PR #1251, opened 2020-04-27
adding tools to master in prep for go mod.

n3wscott:


hack/tools.go
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tools
+
+// This package imports things required by this repository, to force `go mod` to see them as dependencies
+import (
+	_ "k8s.io/code-generator"
```
mattmoor, line 23: I thought you needed to list individual code-generators, do you not?

Should this list the main package for our code-generator as well?

chizhg: This is how k8s/sample-controller does - https://github.com/kubernetes/sample-controller/blob/master/hack/tools.go#L22

n3wscott: I don't think this is working, I do think you have to include the individual code-generators

n3wscott: it turns out you need both.

vaikas:
shouldn't the Gopkg.* files be totally gone?

n3wscott:
Not migrating to go mod here, just prep for downstream to include the hack dir in downstream vendor dirs.

hack/tools.go
chizhg: Hmm, they are already imported in https://github.com/kubernetes/code-generator/blob/master/tools.go#L24-L34. I think it should be enough for Go modules?

PR #1253, opened 2020-04-28
Inject config into controller context.

markusthoemmes:
This makes the config used to start all the clients available on the context starting the controllers. That enables us to use that config to create secondary clients in the controllers with the same config.

/assign @mattmoor @n3wscott

### DIFF
--- a/injection/context.go
+++ b/injection/context.go
@@ -18,6 +18,8 @@ package injection
 
 import (
 	"context"
+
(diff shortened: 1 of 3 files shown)


n3wscott:
I have done so many hacks to do this same thing.

/lgtm

n3wscott:
I am not sure I can approve but we will try

/approve

PR #1256, opened 2020-04-28
Skip doing any work if `Also` receives no errors

vagababov:
1. if no errors are passed skip doing stuff (checks, allocations, etc)
2. if there's one error passed and it's empty (the same)
  - this is a quite popular pattern where in API validation we do:
    `x.Also(validateY())` and if that succeeded then it'll be an empty
    error. Which is in reality most of the time.
3. Improve tests by better printing
4. better test coverage shows commutativity of Also(a, b) and
   Also(a).Also(b)

/assign mattmoor

### DIFF
--- a/apis/field_error_test.go
+++ b/apis/field_error_test.go
@@ -595,10 +595,10 @@ not without this: bar.C`,
 
 			if test.want != "" {
 				if got, want := fe.Error(), test.want; got != want {
-					t.Errorf("%s: Error() = %v, wanted %v", test.name, got, want)
+					t.Errorf("\nGot:  %v,\nwant: %v diff(-want,+got)\n%s", got, want, cmp.Diff(want, got))
 				}
 			} else if fe != nil {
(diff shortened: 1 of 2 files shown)


apis/field_error_test.go
```diff
@@ -743,6 +745,33 @@ func TestAlsoStaysNil(t *testing.T) {
 	}
 }
 
+func TestAlsoMultiple(t *testing.T) {
+	var err *FieldError
+
+	errs := []*FieldError{
+		&FieldError{
```
mattmoor, line 752: Format Go code:
```suggestion
		{
```

apis/field_error_test.go
```diff
@@ -743,6 +745,33 @@ func TestAlsoStaysNil(t *testing.T) {
...
+			Message: "1",
+			Paths:   []string{"bar"},
+		},
+		&FieldError{
```
mattmoor, line 756: Format Go code:
```suggestion
		{
```

vagababov:
Fixed the nits.

PR #1257, opened 2020-04-29
Explicitly name controller filters.

markusthoemmes:
Everytime I read the generic "Filter" or "FilterGroupVersionKind" my brain needs to do an extra roundtrip to realize that this actually filters on the **controller** having that GVK/GK. This adds new functions that explictly state that to avoid that roundtrip.

Old functions are just deprecated so this can be rolled out without a downstream break.

/assign @dprotaso

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -98,13 +98,13 @@ func Filter(gvk schema.GroupVersionKind) func(obj interface{}) bool {
 //
 // Deprecated: Use FilterControlledByGVK instead.
 func FilterGroupVersionKind(gvk schema.GroupVersionKind) func(obj interface{}) bool {
-	return FilterControlledByGVK(gvk)
+	return FilterControllerGVK(gvk)
 }
 
-// FilterControlledByGVK makes it simple to create FilterFunc's for use with
+// FilterControllerGVK makes it simple to create FilterFunc's for use with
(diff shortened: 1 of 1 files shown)


controller/controller.go
```diff
@@ -95,7 +95,16 @@ func Filter(gvk schema.GroupVersionKind) func(obj interface{}) bool {
...
+//
+// Deprecated: Use FilterControlledByGVK instead.
 func FilterGroupVersionKind(gvk schema.GroupVersionKind) func(obj interface{}) bool {
+	return FilterControlledByGVK(gvk)
+}
+
+// FilterControlledByGVK makes it simple to create FilterFunc's for use with
+// cache.FilteringResourceEventHandler that filter based on the
+// schema.GroupVersionKind of the controlling resources.
+func FilterControlledByGVK(gvk schema.GroupVersionKind) func(obj interface{}) bool {
```
dprotaso, line 107:
```suggestion
func FilterControllerGVK(gvk schema.GroupVersionKind) func(obj interface{}) bool {
```

Thoughts on this slight change?

markusthoemmes: wfm!

markusthoemmes: Pushed.

PR #1258, opened 2020-04-29
Fix malformed license headers.

markusthoemmes:
These are just hideous to look at and license files should be consistent throughout the repo. So this fixes that.

/assign @mattmoor

### DIFF
--- a/metrics/opencensus_exporter.go
+++ b/metrics/opencensus_exporter.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2020 The Knative Authors
+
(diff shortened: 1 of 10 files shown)


metrics/opencensus_exporter.go
```diff
@@ -1,9 +1,12 @@
 /*
-Copyright 2020 The Knative Authors
+Copyright 2018 The Knative Authors
```
mattmoor, line 2: Incorrect copyright year:
```suggestion
Copyright 2020 The Knative Authors
```

Issue #1226, opened 2020-04-17
ObservedGeneration should only be set if Status is fully updated


grantr:
## Expected Behavior
_Followup from [this thread in Slack #serving-api](https://knative.slack.com/archives/CA4DNJ9A4/p1586958789227800)_

If ObservedGeneration == Generation, I expect that the object's Status reflects the actual status of the resource.

## Actual Behavior
If the reconcile is written to return errors early, then the Status may not be fully accurate.

Example from @mattmoor:
```go
foo.Status.ObsGen = foo.Gen
if bar, err := reconcileFoo(); err != nil {
  // What happens here?
  return err
} if bar.Something() {
  bar.Status.MarkSomething()
}
foo.Status.MarkReady()
return nil
```

If the reconciler returns on line 4, then the statuses of the `Something` and `Ready` conditions actually reflect the previous generation's status, not the current generation.

## Additional Info
One solution mentioned in the thread is to re-initialize status to fully unknown at the beginning of every reconcile so that no information is carried over from the previous generation. This works for non-timestamped status fields like `address`, but is suboptimal for conditions as it would make `lastTransitionTime` inaccurate.

Instead of replacing each condition, we could mark existing conditions before reconcile and unmark them when updated during reconcile. After the reconcile, any conditions that are still marked are set to unknown. This preserves the functionality of `lastTransitionTime`.

@dprotaso @markusthoemmes @mattmoor @vaikas

mattmoor:
> One solution mentioned in the thread is to re-initialize status to fully unknown at the beginning of every reconcile so that no information is carried over from the previous generation. This works for non-timestamped status fields like address, but is suboptimal for conditions as it would make lastTransitionTime inaccurate.

This isn't entirely accurate.  The proposal in the thread was scoped to `conditions`, and `lastTransitionTime`'s use of `apis.VolatileTime` combined with how we guard status updates is specifically designed to avoid timestamp churn.

A while back I actually tried to do this and uncovered a few places in Serving that have a pretty hard dependency on the imperative use of Status as state, so I gave up, but I like the suggestion @vaikas made that we not have a bad default behavior because of a few bad apples.

To avoid back-compat issues, I'd propose that we introduce `ResetConditions()` to force all of a resources conditions to Unknown, and shift all of our controllers (that can) over to this model.

whaught:
I tried to take a stab at this issue with PR #1261

[...]

PR #1261, opened 2020-04-29
Set observed gen in gen-reconciler

whaught:
Issue #1226

Automatically bump ObservedGeneration in the generated reconciler and set status to unknown for unhandled errors.

looks for `// +genducklogic` to run this part of the code generation. This can serve as a jumping-off point to share other arbitrary logic among duck-type shaped resources.

### DIFF
--- a/apis/test/example/v1alpha1/foo_types.go
+++ b/apis/test/example/v1alpha1/foo_types.go
@@ -28,6 +28,7 @@ import (
 
 // +genclient
 // +genreconciler
+// +genducklogic
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // Foo is for testing.
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -40,6 +40,7 @@ type reconcilerReconcilerGenerator struct {
 	reconcilerClass    string
 	hasReconcilerClass bool
 	nonNamespaced      bool
+	genDuckLogic       bool
 
 	groupGoName  string
 	groupVersion clientgentypes.GroupVersion
@@ -75,6 +76,7 @@ func (g *reconcilerReconcilerGenerator) GenerateType(c *generator.Context, t *ty
 		"class":         g.reconcilerClass,
 		"hasClass":      g.hasReconcilerClass,
 		"nonNamespaced": g.nonNamespaced,
+		"genDuckLogic":  g.genDuckLogic,
 		"controllerImpl": c.Universe.Type(types.Name{
 			Package: "knative.dev/pkg/controller",
 			Name:    "Impl",
@@ -212,8 +214,10 @@ type reconcilerImpl struct {
 // Check that our Reconciler implements controller.Reconciler
 var _ controller.Reconciler = (*reconcilerImpl)(nil)
 
-// Creates a {{.conditionSet|raw}} to manipulate resource status conditions
+{{if .genDuckLogic}}
+// Creates the {{.conditionSet|raw}} to manipulate resource status conditions
 var condSet = apis.NewLivingConditionSet()
+{{end}}
 
 `
 
@@ -324,6 +328,7 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		}
 	}
 
+	{{if .genDuckLogic}}
 	// Bump observed generation to denote that we have processed this
 	// generation regardless of success or failure.
 	resource.Status.ObservedGeneration = resource.Generation
@@ -339,6 +344,7 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 				apis.ConditionReady, "", "unsucessfully observed a new generation")
 		}
 	}
+	{{end}}
 
 	// Synchronize the status.
 	if equality.Semantic.DeepEqual(original.Status, resource.Status) {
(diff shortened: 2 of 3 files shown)


antoineco:
I like the idea but have 2 suggestions:
* Make that behavior part of every reconciler instead of being opt-in. We don't require devs to decide whether they want the reconciler to call UpdateStatus on their behalf, so why would they decide if they those extra initializations? I think people who want the flexibility won't use reconciler-gen anyway.
* If we insist on making it opt-in, please make your change in the existing tag (`+genreconciler:in it status=true`, false if missing). That would avoid clutter and I think it's generally a better UX.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -320,6 +328,24 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		}
 	}
 
+	{{if .genDuckLogic}}
```
n3wscott, line 331: the code will get here for deleted resources as well and it is not allowed to update generation for a deleted object

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -320,6 +328,24 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
+	// Bump observed generation to denote that we have processed this
+	// generation regardless of success or failure.
+	resource.Status.ObservedGeneration = resource.Generation
+
+	if resource.Status.ObservedGeneration != original.Status.ObservedGeneration && reconcileEvent != nil {
+		originalRc := original.Status.GetCondition(apis.ConditionReady)
+		rc := resource.Status.GetCondition(apis.ConditionReady)
+		// if a new generation is observed and reconciliation reported an error event
+		// the reconciler should change the ready state. By default we will set unknown.
+		if equality.Semantic.DeepEqual(originalRc, rc) {
```
n3wscott, line 341: but this could be that it went from ready true to ready true if the update was something minor

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -320,6 +328,24 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
+			logger.Warnw("A reconconiler observed a new generation without updating the resource status")
+			condSet.Manage(&resource.Status).MarkUnknown(
```
n3wscott, line 343: I don't think you should change the ready condition directly. It needs a dependent condition that cases the unreadiness.

whaught:
> I like the idea but have 2 suggestions:
> 
> * Make that behavior part of every reconciler instead of being opt-in. We don't require devs to decide whether they want the reconciler to call UpdateStatus on their behalf, so why would they decide to disable just those extra initializations? I think people who want the flexibility won't use reconciler-gen anyway.
> * If we insist on making it opt-in, please make your change in the existing tag (`+genreconciler:initstatus=true`, false if missing). That would avoid clutter and I think it's generally a better UX.

The reason this isn't part of every reconciler is because not ever reconciler follows the duck-type code shape. Specifically some eventing resources don't have observed generation and this code fails for those.

The reason this is opt-in is partly to allow to allow generating for non-duck resource but also for safety as we switch to it. I want to migrate resources to it one-by one by adding this tag and removing their reconciler's handling of observed gen. This will let me test it safely as I go through that process. Unfortunately +genreconciler:classname is already a thing to allow reconcilers the option of specifying a type to reconciler. Adding more options to that tag might get pretty cluttered (we could comma separate them).

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
whaught: This only runs if reconcileEvent != nil should we should be in a (potential) error case.

apis/test/example/v1alpha1/foo_types.go
```diff
@@ -28,6 +28,7 @@ import (
 
 // +genclient
 // +genreconciler
+// +genducklogic
```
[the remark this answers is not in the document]
whaught: I won't pretend to be good at naming. I'm not sure I see the interface problem. The idea was "generate duck-type shaped common logic".

Can you suggest an alternative? @antoineco also suggested making it another option on +genreconciler if you have any thoughts on that (although that already has a :classname option)

antoineco:
@whaught that's a fair point, I didn't consider not every type implements these.

If the tag's elements are `key=value` separated by colons, I think we should be safe adding new parameters to it.

whaught:
> @whaught that's a fair point, I didn't consider not every type implements these.
> 
> If the tag's elements are `key=value` separated by colons, I think we should be safe adding new parameters to it.

The tags are not key=value. You can read the parsing here: https://github.com/knative/pkg/blob/83edd9f30de46d085fc980fbb0561c5a7d8e76fc/codegen/cmd/injection-gen/generators/packages.go#L191

it currently expects "genreconciler:class". We could change this and then go update existing usages of it, make it key value, and comma separate multiple tags. I'm not sure if this is really a worthy migration in the scope of what this PR is trying to accomplish for the cost of just adding a new tag.

antoineco:
@whaught the current expected format is ` +genreconciler:class=<class>`, which looks to me like a format that could _very easily_ accept new parameters separated by colon or commas.

Besides, there are at most 4 types that currently set this class parameter in the code base.

I simply can't see a case for a new tag here, but I'm also happy to be proven wrong.

whaught:
@antoineco you're right about the existing format. Although a parser would need to be written to allow for other options. The argument is merely that a new tag doesn't cost anything and allows for options of its own. This isn't intended to be the only common generated logic, but a starting point.

In the future I hope to be able to include other common factors like initializing defaults/conversions and common metrics on conditions.
Eventually I was forseeing something like +genducklogic:observegen=true,commonmetrics=true,initdefaults=true

Just thought it might get cluttered, and in line with what @n3wscott said, these aren't properties of a reconciler (which makes no resource shape assumptions) but are properties of the API duck resource.

I'm happy to go the other way on this if folks are passionate.

whaught:
/close

pulled this logic into #1284 with an interface that makes it more easily testable

PR #1262, opened 2020-04-29
bump go.mod to go 1.14

dprotaso:
## Proposed Changes

From https://github.com/golang/go/wiki/Modules

> When the main module contains a top-level vendor directory and its go.mod file specifies go 1.14 or higher, the go command now defaults to -mod=vendor for operations that accept that flag.

This preserves our prior behaviour of using packages from our vendor dir

**Release Note**

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module knative.dev/pkg
 
-go 1.13
+go 1.14
 
 require (
 	cloud.google.com/go v0.55.0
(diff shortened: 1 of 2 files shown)


vagababov:
will this build with go1.14?

dprotaso:
i would think the build tests would use go1.14 but it seems like it doesn't

> go version go1.13 linux/amd64

dprotaso:
From [#test](https://knative.slack.com/archives/CA1DTGZ2N/p1588349446122100)

>chizhg  12:10 PM
Hi all, we have switched all the Prow jobs running on master branch to use Go 1.14, please feel free to start using new features in Go 1.14, and let us know if there are any issues!

PR #1263, opened 2020-04-29
Add a helper func to set a default metric config for unit tests

yanweiguo:
There exists a race condition that recording the data points before the metric config is initialized, at which point whether to record is unknown. This causes problem when the backend is Stackdriver and unsupported metrics are actually reported.

As Prometheus is the default backend, the global metric config should always be non-empty.

This PR only add a helper func to set a default metric config for unit tests so that it can be used for downstream repos before the actual breaking change.

/assign @anniefu @vagababov


vagababov:
/hold
It seems the UT failures are reasonable

anniefu:
/lgtm

There may be a separate issue here that until the first ConfigMap created signal is picked up for config-observability, metrics aren't recorded. It depends on whether opencensus somehow buffers Record calls until a valid exporter is created or just drops them.

yanweiguo:
/hold cancel

Tests fixed

/assign @mattmoor for testing files changes in other directories.

yanweiguo:
> Is this a breaking change for the downstream repos? Are we going to need to link a new library from a bunch of test code?

Yeah. How about:

1. Adding the test helper func first in pkg repo.
1. Updating pkg repo in downstream repos (no breaking change).
1. Changing the downstream repos to use the new test lib for test code (set a default test metric config).
1. Changing the actual behavior in pkg repo.

yanweiguo:
Reverted the actual behavior change. This PR only add the helper func to be used for downstream repos. @mattmoor

yanweiguo:
/hold cancel

nil check is still needed after #1271 as it doesn't fix the race condition.

I'll move forward with the steps mentioned in https://github.com/knative/pkg/pull/1263#issuecomment-621603510

@mattmoor could you take a look?

mattmoor:
/lgtm
/approve
/hold

Just putting a hold in case any manual staging is needed, but if not feel free to release for the auto-update.

yanweiguo:
/hold cancel

This PR won't break downstream repos.

Issue #807, opened 2019-10-25
Potential deadlock in knative.dev/pkg/metrics when using k8s.io/client-go/kubernetes


anniefu:
When using knative.dev/metrics/pkg, the kubeclient has an unexpected dependency on the knative.dev/pkg/metrics pkg and its internal lock (metricsMux) that prevents the kubeclient from being used freely within the metrics pkg.

Here are the two uses of the metricsMux that don't play well.

knative.dev/pkg/metrics/exporter.go
- exporter.go protects its internal state, the current metrics config and current metrics exporter, with a Mutex (metricsMux) https://github.com/knative/pkg/blob/master/metrics/exporter.go#L28 
- Whenever exporter.go updates the exporter, it needs to obtain the lock to read internal state
- Updating the exporter also requires using the kubeclient to read a K8s Secret specified in the metrics config https://github.com/knative/pkg/blob/master/metrics/stackdriver_exporter.go#L177 
- Potentially other kubeclient things could be done in the future

k8s.io/client-go/
- The k8s.io/client-go/tools/metrics pkg provides a hook for users to record the latency of the kubernetes rest clients. https://github.com/kubernetes/client-go/blob/master/tools/metrics/metrics.go#L40-43
- knative.dev/metrics/pkg sets this hook 
 https://github.com/knative/pkg/blob/master/metrics/metrics.go#L115-126
- On every kubeclient request, the hook calls metrics.Record() https://github.com/knative/pkg/blob/master/metrics/record.go#L41 
- The call to metrics.Record() needs to obtains the knative.dev/pkg/metrics pkg's internal lock to read the metrics config

The effect of this is every call to kubeclient uses the metrics pkg to Record a metric, and needs to lock metricsMux. Every update to the metrics exporter needs to call the kubeclient, and needs to lock metricsMux.

This prevents fully locking the metrics exporter update section to prevent race conditions.

Ideally we find a way to make the kubeclient usable anywhere with the metrics package, or we make it very clear to developers that they cannot use it within a lock through a test that will fail or dev-oriented error.

## Expected Behavior
You can use the client-go kubeclient while locked with the metricsMux.

## Actual Behavior
Deadlock

## Steps to Reproduce the Problem
Within the knative.dev/pkg/metrics pkg
1. Add a call to metricsMux.Lock()
2. Within the lock scope, attempt to use the client-go kubeclient
3. Deadlock
```
import (
  corev1 "k8s.io/api/core/v1"
[...]

PR #1271, opened 2020-05-01
Make updating metrics config and metrics exporter atomic

anniefu:
Avoid using an out-of-date metrics config with a newly created exporter when config-observability is updated. See #1263 for description of potential race condition issues.

This change also works around the potential deadlock with using k8s.io/client-go/kubernetes library within the metrics package described in #807.

The tl;dr of the deadlock cycle is the kubeclient will try and record a latency metric using knative/pkg metrics.record while knative/pkg UpdateExporter will try and use kubeclient to get a Secret for new exporters.

### DIFF
--- a/metrics/config_test.go
+++ b/metrics/config_test.go
@@ -271,8 +271,8 @@ var (
 						Name: "opencensus",
 					},
 					Data: map[string][]byte{
-						"client-cert.pem": []byte{},
-						"client-key.pem":  []byte{},
+						"client-cert.pem": {},
+						"client-key.pem":  {},
 					},
 				}).Get,
 			},
@@ -287,8 +287,8 @@ var (
 						Name: "opencensus",
 					},
 					Data: map[string][]byte{
-						"client-cert.pem": []byte{},
-						"client-key.pem":  []byte{},
+						"client-cert.pem": {},
+						"client-key.pem":  {},
 					},
 				},
 			},
--- a/metrics/config_test.go
+++ b/metrics/config_test.go
@@ -32,6 +32,7 @@ import (
 	"knative.dev/pkg/metrics/metricstest"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // TODO UTs should move to eventing and serving, as appropriate.
@@ -201,6 +202,12 @@ var (
 				},
 				Domain:    servingDomain,
 				Component: testComponent,
+				Secrets: fakeSecretList(corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      StackdriverSecretNameDefault,
+						Namespace: StackdriverSecretNamespaceDefault,
+					},
+				}).Get,
 			},
 			expectedConfig: metricsConfig{
 				domain:                            servingDomain,
@@ -215,6 +222,12 @@ var (
 					GCPLocation: "us-west1",
 					ClusterName: "cluster",
 					UseSecret:   true,
+				},
+				secret: &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      StackdriverSecretNameDefault,
+						Namespace: StackdriverSecretNamespaceDefault,
+					},
 				},
 			},
 			expectedNewExporter: true,
@@ -253,6 +266,15 @@ var (
 				},
 				Domain:    servingDomain,
 				Component: testComponent,
+				Secrets: fakeSecretList(corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "opencensus",
+					},
+					Data: map[string][]byte{
+						"client-cert.pem": []byte{},
+						"client-key.pem":  []byte{},
+					},
+				}).Get,
 			},
 			expectedConfig: metricsConfig{
 				domain:             servingDomain,
@@ -260,6 +282,15 @@ var (
 				backendDestination: OpenCensus,
 				collectorAddress:   "external-svc:55678",
 				requireSecure:      true,
+				secret: &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "opencensus",
+					},
+					Data: map[string][]byte{
+						"client-cert.pem": []byte{},
+						"client-key.pem":  []byte{},
+					},
+				},
 			},
 			expectedNewExporter: true,
 		}, {
@@ -514,6 +545,10 @@ var (
 		}}
 )
 
+func successTestsInit() {
+	SetStackdriverSecretLocation(StackdriverSecretNameDefault, StackdriverSecretNamespaceDefault)
+}
+
 func TestGetMetricsConfig(t *testing.T) {
 	for _, test := range errorTests {
 		t.Run(test.name, func(t *testing.T) {
@@ -525,6 +560,7 @@ func TestGetMetricsConfig(t *testing.T) {
 		})
 	}
 
+	successTestsInit()
 	for _, test := range successTests {
 		t.Run(test.name, func(t *testing.T) {
 			defer ClearAll()
@@ -558,6 +594,7 @@ func TestGetMetricsConfig_fromEnv(t *testing.T) {
 
 func TestIsNewExporterRequiredFromNilConfig(t *testing.T) {
 	setCurMetricsConfig(nil)
+	successTestsInit()
 	for _, test := range successTests {
 		t.Run(test.name, func(t *testing.T) {
 			defer ClearAll()
@@ -685,6 +722,7 @@ func TestIsNewExporterRequired(t *testing.T) {
 func TestUpdateExporter(t *testing.T) {
 	setCurMetricsConfig(nil)
 	oldConfig := getCurMetricsConfig()
+	successTestsInit()
 	for _, test := range successTests[1:] {
 		t.Run(test.name, func(t *testing.T) {
 			defer ClearAll()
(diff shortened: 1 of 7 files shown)


metrics/config_test.go
```diff
@@ -253,13 +266,31 @@ var (
 				},
 				Domain:    servingDomain,
 				Component: testComponent,
+				Secrets: fakeSecretList(corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "opencensus",
+					},
+					Data: map[string][]byte{
+						"client-cert.pem": []byte{},
+						"client-key.pem":  []byte{},
```
mattmoor, line 275: Format Go code:
```suggestion
						"client-cert.pem": {},
						"client-key.pem":  {},
```

metrics/config_test.go
```diff
@@ -253,13 +266,31 @@ var (
...
 				backendDestination: OpenCensus,
 				collectorAddress:   "external-svc:55678",
 				requireSecure:      true,
+				secret: &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "opencensus",
+					},
+					Data: map[string][]byte{
+						"client-cert.pem": []byte{},
+						"client-key.pem":  []byte{},
```
mattmoor, line 291: Format Go code:
```suggestion
						"client-cert.pem": {},
						"client-key.pem":  {},
```

metrics/exporter.go
```diff
@@ -130,6 +131,12 @@ func UpdateExporterFromConfigMapWithOpts(opts ExporterOptions, logger *zap.Sugar
 // to prevent a race condition between reading the current configuration
 // and updating the current exporter.
 func UpdateExporter(ops ExporterOptions, logger *zap.SugaredLogger) error {
+	// We don't check if ops.secrets is `nil` right now, because this is a transition step.
```
yanweiguo, line 134: Any open issue to track this? Are we going to do it before we migrate to OpenTelemetry? If not, I'd prefer not add the comments for the part after `Eventually`.

Also createMetricsConfig seems the better place for this comment.

metrics/exporter.go
```diff
@@ -177,9 +189,9 @@ func isNewExporterRequired(newConfig *metricsConfig) bool {
 }
 
 // newMetricsExporter gets a metrics exporter based on the config.
-// This function is not implicitly thread-safe.
+// This function must be called with the metricsMux reader (or writer) locked.
 func newMetricsExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, error) {
-	ce := getCurMetricsExporter()
+	ce := curMetricsExporter
```
yanweiguo, line 194: `ce` is only used once in this func. We can just use `curMetricsExporter` directly.

metrics/exporter.go
```diff
@@ -244,6 +260,10 @@ func setCurMetricsConfig(c *metricsConfig) {
 // Return value indicates whether the exporter is flushable or not.
 func FlushExporter() bool {
 	e := getCurMetricsExporter()
+	return flushExporterUnlocked(e)
+}
+
+func flushExporterUnlocked(e view.Exporter) bool {
```
yanweiguo, line 266: Maybe `flushGivenExporter`. Since you pass the exporter as parameter, we don't need to care about whether it is locked or not.

metrics/stackdriver_exporter.go
```diff
@@ -140,21 +140,20 @@ func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (v
...
 // On error, an empty array of client options is returned.
-func getStackdriverExporterClientOptions(sdconfig *StackdriverClientConfig) ([]option.ClientOption, error) {
+func getStackdriverExporterClientOptions(config *metricsConfig) ([]option.ClientOption, error) {
 	var co []option.ClientOption
-	if sdconfig.UseSecret && useStackdriverSecretEnabled {
-		secret, err := getStackdriverSecret(sdconfig)
-		if err != nil {
-			return co, err
+
+	if config.stackdriverClientConfig.UseSecret {
```
yanweiguo, line 146: If `useStackdriverSecretEnabled` is false, config.secret will be nil then an error will be thrown. This is different compared to original code. Is it intended change?

metrics/opencensus_exporter.go
```diff
@@ -47,35 +48,39 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
 	if lister == nil {
-		logger.Errorf("No secret lister provided for component %q; cannot use requireSecure=true", component)
+		return nil, fmt.Errorf("No secret lister provided for component %q; cannot use requireSecure=true", component)
+	}
+	secret, err := lister(component + "-opencensus")
+	if errors.IsNotFound(err) {
+		secret, err = lister("opencensus")
+	}
+	if err != nil {
+		return nil, fmt.Errorf("Unable to fetch opencensus secret for %q, cannot use requireSecure=true: %+v", component, err)
```
vagababov, line 64:
```suggestion
		return nil, fmt.Errorf("unable to fetch opencensus secret for %q, cannot use requireSecure=true: %+v", component, err)
```

metrics/opencensus_exporter.go
```diff
@@ -47,35 +48,39 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
...
-// credentialFetcher attempts to locate a secret containing TLS credentials
+// getOpenCensusSecret attempts to locate a secret containing TLS credentials
 // for communicating with the OpenCensus Agent. To do this, it first looks
 // for a secret named "<component>-opencensus", then for a generic
 // "opencensus" secret.
-func credentialFetcher(component string, lister SecretFetcher, logger *zap.SugaredLogger) credentials.TransportCredentials {
+func getOpenCensusSecret(component string, lister SecretFetcher) (*corev1.Secret, error) {
 	if lister == nil {
-		logger.Errorf("No secret lister provided for component %q; cannot use requireSecure=true", component)
+		return nil, fmt.Errorf("No secret lister provided for component %q; cannot use requireSecure=true", component)
```
vagababov, line 57:
```suggestion
		return nil, fmt.Errorf("no secret lister provided for component %q; cannot use requireSecure=true", component)
```

metrics/stackdriver_exporter.go
```diff
@@ -140,21 +140,20 @@ func newStackdriverExporter(config *metricsConfig, logger *zap.SugaredLogger) (v
...
+		if config.secret == nil {
+			return co, fmt.Errorf("No secret provided for component %q; cannot use stackdriver-use-secret=true", config.component)
```
vagababov, line 148:
```suggestion
			return co, fmt.Errorf("no secret provided for component %q; cannot use stackdriver-use-secret=true", config.component)
```

metrics/stackdriver_exporter_test.go
```diff
@@ -471,48 +473,45 @@ func assertStringsEqual(t *testing.T, description string, expected string, actua
...
 		secretNamespace = StackdriverSecretNamespaceDefault
+		useStackdriverSecretEnabled = false
 	}()
 
-	sdConfig := &StackdriverClientConfig{
-		ProjectID:   "project",
-		GCPLocation: "us-west2",
-		ClusterName: "cluster",
-		UseSecret:   false,
+	testName, testNamespace := "test-name", "test-namespace"
```
vagababov, line 485: `const`?

metrics/stackdriver_exporter_test.go
```diff
@@ -471,48 +473,45 @@ func assertStringsEqual(t *testing.T, description string, expected string, actua
...
-	if _, e1 := getStackdriverExporterClientOptions(sdConfig); e1 == nil {
-		t.Errorf("Expected a known error when getting exporter options with Secrets enabled (cannot create valid kubeclient in tests). Did the function run as expected?")
+	// Once SetStackdriverSecretLocation has been called, attempts to get the secret should complete.
+	SetStackdriverSecretLocation(testName, testNamespace)
+	sec, err = getStackdriverSecret(secretFetcher)
+	if err != nil {
+		t.Errorf("Got unexpected error when getting secret: %v", err)
+	}
+	if sec == nil {
+		t.Errorf("expected secret to be non-nil if there is no error and SetStackdriverSecretLocation has been called")
```
vagababov, line 513: no formatting in this call
```suggestion
		t.Error("expected secret to be non-nil if there is no error and SetStackdriverSecretLocation has been called")
```

metrics/exporter.go
anniefu: I filed #1273 and linked it here. I know we currently do all the ExporterOptions checks in createMetricsConfig, but I think they are more appropriate in exporter.go where ExporterOptions is actually defined. Might as well catch the error as early as possible in the call stack.

The new UpdateExporterFromConfigMapWithOpts does this.

metrics/stackdriver_exporter.go
anniefu: Tricky to see, but this error is actually returned and logged as just a warning in newStackdriverExporter. 

This message is really the only sign to metrics package consumers that they need to call SetStackdriverSecretLocation to enable Stackdriver secret option. I added a comment that SetStackdriverSecretLocation must be called to make it clearer to Knative developers.

The SetStackdriverSecretLocation requirement was to prevent accidental attempts to read and use the default Secret namespace/name ("default/stackdriver-service-account-key") if a user turned the flag on in config-observability, but knative/serving really wanted the Secret to be "knative-serving/stackdriver-service-account-key").

yanweiguo:
@googlebot rescan

PR #1274, opened 2020-05-01
Add elapsed wait time in timeout message.

tcnghia:
This would give us some ideas about the total wait time when we look at logs, to make it easier to decide if we want to adjust the timeout setting.

### DIFF
--- a/network/transports.go
+++ b/network/transports.go
@@ -19,6 +19,7 @@ package network
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"time"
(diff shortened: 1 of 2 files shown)


network/transports.go
```diff
@@ -89,7 +91,8 @@ func dialBackOffHelper(ctx context.Context, network, address string, bo wait.Bac
 		}
 		return c, nil
 	}
-	return nil, errDialTimeout
+	elapsed := time.Now().Sub(start)
+	return nil, fmt.Errorf("timed out dialing(%.2fs)", elapsed.Seconds())
```
vagababov, line 95:
```suggestion
	return nil, fmt.Errorf("timed out dialing after %.2fs", elapsed.Seconds())
```

network/transports_test.go
```diff
@@ -90,8 +90,9 @@ func TestDialWithBackoff(t *testing.T) {
 		c.Close()
 		t.Error("Unexpected success dialing")
 	}
-	if err != errDialTimeout {
-		t.Errorf("Error = %v, want: %v", err, errDialTimeout)
+	expectedErrPrefix := "timed out dialing"
```
vagababov, line 93:
```suggestion
	const expectedErrPrefix = "timed out dialing"
```

PR #1277, opened 2020-05-02
Create our own comment parser for codegen

whaught:
Issue #1226

Per comments in PR #1226 there was some dissatisfaction with our comment parsing.
This parser allows for codegen tags of the form` +foo:key=value,key2=value2`

We currently cheat by using the k8s parser and search for the prefix `+foo:key` to achieve such a result, but that doesn't allow for multiple comma-separated values per line.

Although this is a break from the current parser: https://github.com/kubernetes/gengo/blob/master/types/comments.go
We're already kind of abusing it with our "+genreconciler:class" prefix search.

This also reduces the number of times we fully scan the comments.

### DIFF
--- a/codegen/cmd/injection-gen/generators/comment_parser.go
+++ b/codegen/cmd/injection-gen/generators/comment_parser.go
@@ -53,7 +53,7 @@ func ExtractCommentTags(marker string, lines []string) map[string]map[string]str
 
 			opts := out[options[0]]
 			if opts == nil {
-				opts = map[string]string{}
+				opts = make(map[string]string, len(vals))
 			}
 
 			for _, pair := range vals {
--- a/codegen/cmd/injection-gen/generators/comment_parser.go
+++ b/codegen/cmd/injection-gen/generators/comment_parser.go
@@ -40,10 +40,7 @@ func ExtractCommentTags(marker string, lines []string) map[string]map[string]str
 	out := map[string]map[string]string{}
 	for _, line := range lines {
 		line = strings.TrimSpace(line)
-		if len(line) == 0 {
-			continue
-		}
-		if !strings.HasPrefix(line, marker) {
+		if len(line) == 0 || !strings.HasPrefix(line, marker) {
 			continue
 		}
 
--- a/codegen/cmd/injection-gen/generators/comment_parser.go
+++ b/codegen/cmd/injection-gen/generators/comment_parser.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2020 The Knative Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package generators
+
+import "strings"
+
+// Adapted from the k8s.io comment parser https://github.com/kubernetes/gengo/blob/master/types/comments.go
+
+// ExtractCommentTags parses comments for lines of the form:
+//
+//   'marker' + ':' "key=value,key2=value2".
+//
+// Values are optional; empty map is the default.  A tag can be specified more than
+// one time and all values are returned.  If the resulting map has an entry for
+// a key, the value (a slice) is guaranteed to have at least 1 element.
+//
+// Example: if you pass "+" for 'marker', and the following lines are in
+// the comments:
+//   +foo:key=value1,key2=value2
+//   +bar
+//
+// Then this function will return:
+//   map[string]map[string]string{"foo":{"key":value1","key2":"value2"}, "bar": {}}
+//
+// Users are not expected to repeat values.
+func ExtractCommentTags(marker string, lines []string) map[string]map[string]string {
+	out := map[string]map[string]string{}
+	for _, line := range lines {
+		line = strings.Trim(line, " ")
+		if len(line) == 0 {
+			continue
+		}
+		if !strings.HasPrefix(line, marker) {
+			continue
+		}
+
+		options := strings.SplitN(line[len(marker):], ":", 2)
+		if len(options) == 2 {
+			vals := strings.Split(options[1], ",")
+
+			opts := out[options[0]]
+			if opts == nil {
+				opts = map[string]string{}
+			}
+
+			for _, pair := range vals {
+				kv := strings.SplitN(pair, "=", 2)
+				if len(kv) == 2 {
+					opts[kv[0]] = kv[1]
+				} else if len(kv) == 1 {
+					opts[kv[0]] = ""
+				}
+			}
+			out[options[0]] = opts
+
+		} else if len(options) == 1 {
+			if _, has := out[options[0]]; !has {
+				out[options[0]] = nil
+			}
+		}
+	}
+	return out
+}
--- a/codegen/cmd/injection-gen/generators/comment_parser.go
+++ b/codegen/cmd/injection-gen/generators/comment_parser.go
@@ -65,7 +65,7 @@ func ExtractCommentTags(marker string, lines []string) map[string]map[string]str
 			} else {
 				out[options[0]] = opts
 			}
-		} else if len(options) == 1 {
+		} else if len(options) == 1 && options[0] != "" {
 			if _, has := out[options[0]]; !has {
 				out[options[0]] = nil
 			}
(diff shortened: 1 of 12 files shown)


codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,76 @@
...
+			continue
+		}
+
+		options := strings.SplitN(line[len(marker):], ":", 2)
+		if len(options) == 2 {
+			vals := strings.Split(options[1], ",")
+
+			opts := out[options[0]]
+			if opts == nil {
+				opts = map[string]string{}
```
vagababov, line 56:
```suggestion
				opts = make(map[string]string, len(vals))
```

codegen/cmd/injection-gen/generators/comment_parser_test.go
```diff
@@ -0,0 +1,81 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package generators
+
+import "testing"
+
+func TestParseComments(t *testing.T) {
+
```
vagababov, line 21:
```suggestion
```

codegen/cmd/injection-gen/generators/comment_parser_test.go
```diff
@@ -0,0 +1,81 @@
...
+		t.Errorf("Failed to extract single key got=%t,%v want=true,{\"key\":\"value\"}", ok, val)
+	}
+
+	if val, ok := extracted["manypairs"]; !ok || val["key1"] != "value1" || val["key2"] != "value2" {
+		t.Errorf("Failed to extract single key got=%t,%v want=true,{\"key\":\"value\"}", ok, val)
+	}
+}
+
+func TestMergeDuplicates(t *testing.T) {
+
```
vagababov, line 56:
```suggestion
```

codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -188,22 +188,27 @@ func MustParseClientGenTags(lines []string) Tags {
...
-	comments := append(append([]string{}, t.SecondClosestCommentLines...), t.CommentLines...)
-	_, nonNamespaced := types.ExtractCommentTags("+", comments)["genclient:nonNamespaced"]
-	return nonNamespaced
+func isNonNamespaced(tags map[string]map[string]string) bool {
+	vals, has := tags["genclient"]
+	if !has {
+		return false
+	}
+	_, has = vals["nonNamespaced"]
+	return has
```
vagababov, line 211:
```suggestion
	return has && vals["nonNamespaced"]
```

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			for _, pair := range vals {
+				kv := strings.SplitN(pair, "=", 2)
+				if len(kv) == 2 {
```
vagababov, line 61:
```suggestion
				if kv := strings.SplitN(pair, "=", 2); len(kv) == 2 {
```

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,76 @@
...
+					opts[kv[0]] = kv[1]
+				} else if len(kv) == 1 {
+					opts[kv[0]] = ""
+				}
+			}
+			out[options[0]] = opts
+
```
vagababov, line 68:
```suggestion
```

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,74 @@
...
+func ExtractCommentTags(marker string, lines []string) map[string]map[string]string {
+	out := map[string]map[string]string{}
+	for _, line := range lines {
+		line = strings.Trim(line, " ")
+		if len(line) == 0 {
+			continue
+		}
+		if !strings.HasPrefix(line, marker) {
+			continue
+		}
```
vagababov, line 48:
```suggestion
		if len(line) == 0 || !strings.HasPrefix(line, marker){
			continue
		}
```

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,74 @@
...
+//   +bar
+//
+// Then this function will return:
+//   map[string]map[string]string{"foo":{"key":value1","key2":"value2"}, "bar": {}}
+//
+// Users are not expected to repeat values.
+func ExtractCommentTags(marker string, lines []string) map[string]map[string]string {
+	out := map[string]map[string]string{}
+	for _, line := range lines {
+		line = strings.Trim(line, " ")
```
vagababov, line 42: https://golang.org/pkg/strings/#TrimSpace, perhaps. What if it's a tab?

codegen/cmd/injection-gen/generators/packages.go
vagababov: Looks the same.

antoineco:
LGTM :+1:

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,71 @@
...
+		if len(line) == 0 || !strings.HasPrefix(line, marker){
```
mattmoor, line 43: Format Go code:
```suggestion
		if len(line) == 0 || !strings.HasPrefix(line, marker) {
```

codegen/cmd/injection-gen/generators/comment_parser.go
whaught: This was all just copying the existing k8s parser behavior. Sure let's trimspace

codegen/cmd/injection-gen/generators/packages.go
whaught: This suggestion does not compile. vals["nonNamespaced"] is not boolean.

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -0,0 +1,71 @@
...
+			if opts == nil {
+				opts = make(map[string]string, len(vals))
+			}
+
+			for _, pair := range vals {
+				if kv := strings.SplitN(pair, "=", 2); len(kv) == 2 {
+					opts[kv[0]] = kv[1]
+				} else if len(kv) == 1 {
+					opts[kv[0]] = ""
+				}
```
vagababov, line 61: Perhaps
```
if pair == "" {
  continue
}
```
then you don't need to check for `len(kv)==1` (and avoid calling split at all).

codegen/cmd/injection-gen/generators/comment_parser_test.go
```diff
@@ -0,0 +1,79 @@
...
+func TestParseComments(t *testing.T) {
+	comment := []string{
+		"This is an example comment to parse",
+		"",
+		" +foo",
+		"+bar",
+		"+with:option",
+		"+pair:key=value",
+		"+manypairs:key1=value1,key2=value2",
+	}
```
vagababov, line 29: so you can add a test, where: "a=b,,,,,c=d"

codegen/cmd/injection-gen/generators/packages.go
vagababov: hm, for reason I thought it's a map to `bool`...

codegen/cmd/injection-gen/generators/comment_parser_test.go
whaught: Done

codegen/cmd/injection-gen/generators/comment_parser.go
whaught: I think the split is reasonable to detect the '=' char. We might be over-optimizing a bit a this point. This really just needs to work for happy cases - we can assume developers with malformed comments won't get generated code correctly.

antoineco:
Prow says
> Code coverage is below threshold (50%), fail presubmit workflow intentionally

Probably due to the 0% coverage in `packages.go`, which is surprising because the file already existed in the repo before this PR, so coverage should be tackled separately.

codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -188,22 +188,27 @@ func MustParseClientGenTags(lines []string) Tags {
...
+	return ExtractCommentTags("+", comments)
+}
+
+func extractReconcilerClassTag(tags map[string]map[string]string) (classname string, has bool) {
+	vals, has := tags["genreconciler"]
+	if !has {
+		return
 	}
-	return "", false
+	classname, _ = vals["class"]
```
mattmoor, line 201: This doesn't properly set `has` anymore, and breaks serving.

mattmoor: I'm tweaking this, and will attempt to roll things forward

PR #1282, opened 2020-05-04
Add support for reconcilers to update annotations and labels

duglin:
We have cases where our reconciler will update annotations or labels and we'd like those to be copied back into etcd.

/cc @n3wscott 

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -367,18 +367,41 @@ func (r *reconcilerImpl) updateStatus(existing *{{.type|raw}}, desired *{{.type|
 			}
 		}
 
-		// If there's nothing to update, just return.
-		if reflect.DeepEqual(existing.Status, desired.Status) {
-			return nil
+		fullUpdate := false
+		statusUpdate := false
+
+		if !reflect.DeepEqual(existing.GetAnnotations(),desired.GetAnnotations()) {
+			existing.ObjectMeta.Annotations = desired.ObjectMeta.Annotations
+			fullUpdate = true
 		}
 
-		existing.Status = desired.Status
+		if !reflect.DeepEqual(existing.GetLabels(),desired.GetLabels()) {
+			existing.ObjectMeta.Labels = desired.ObjectMeta.Labels
+			fullUpdate = true
+		}
+
+		if !reflect.DeepEqual(existing.Status, desired.Status) {
+			existing.Status = desired.Status
+			statusUpdate = true
+		}
+
+		// If there's nothing to update, just return.
+		if !fullUpdate && !statusUpdate {
+			return nil
+		}
 
 		{{if .nonNamespaced}}
 		updater := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}()
 		{{else}}
 		updater := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}(existing.Namespace)
 		{{end}}
+
+		if fullUpdate {
+			_, err = updater.Update(existing)
+		} else {
+			_, err = updater.UpdateStatus(existing)
+		}
+
 		_, err = updater.UpdateStatus(existing)
 		return err
 	})


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -367,18 +367,41 @@ func (r *reconcilerImpl) updateStatus(existing *{{.type|raw}}, desired *{{.type|
 			}
 		}
 
+		fullUpdate := false
+		statusUpdate := false
+
+		if !reflect.DeepEqual(existing.GetAnnotations(),desired.GetAnnotations()) {
```
n3wscott, line 373: this are not imported correctly, see other examples

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -367,18 +367,41 @@ func (r *reconcilerImpl) updateStatus(existing *{{.type|raw}}, desired *{{.type|
...
-		existing.Status = desired.Status
-
 		{{if .nonNamespaced}}
 		updater := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}()
 		{{else}}
 		updater := r.Client.{{.group}}{{.version}}().{{.type|apiGroup}}(existing.Namespace)
 		{{end}}
+
+		if fullUpdate {
+			_, err = updater.Update(existing)
```
n3wscott, line 400: this makes me nervous to be honest. It is not clear that it is ok for a controller to add labels and annotations. You should do that in a mutating webhook.

n3wscott: @mattmoor wdyt?

lionelvillard: I wonder if instead of `Update` we should use `Patch`. That would make me less nervous.

mattmoor: A few things:
1. With the `/status` sub-resource, I don't believe `Update` will actually update the status, so as written this is probably breaking.
2. I agree with Scotty's sentiment above, that in general this is an anti-pattern and we've been trying to use `// +genreconciler` to codify best practices.  In the past we (Serving) have only used this as an escape hatch to "touch" the resource, but expanded generally I can see a number of problems with this.

For example, some parent resources are particular about the labels and annotations on child resources, so by doing this, we make it possible (maybe even likely) that a child controller will fight the parent controller.  Combined with the bug from `1.` this could result in no status ever really getting published because the `UpdateStatus` calls are always elided in favor of `Update`s 😬 

If you had to do this, I'd recommend using `Patch` from the `ReconcileKind` as @lionelvillard suggests, but I'd be very careful about spinning resource mutations (these can be incredibly hard to detect, especially when the mutations are outside of `spec:` where `metadata.generation` soaring is a tell-tale sign).

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
duglin: Not sure what you mean by "not imported correctly".

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -367,18 +367,41 @@ func (r *reconcilerImpl) updateStatus(existing *{{.type|raw}}, desired *{{.type|
...
 
+		fullUpdate := false
+		statusUpdate := false
+
+		if !reflect.DeepEqual(existing.GetAnnotations(),desired.GetAnnotations()) {
+			existing.ObjectMeta.Annotations = desired.ObjectMeta.Annotations
+			fullUpdate = true
+		}
+
+		if !reflect.DeepEqual(existing.GetLabels(),desired.GetLabels()) {
```
markusthoemmes, line 378: Is `DeepEqual` the right operation? I think users should be allowed to label/annotate downstream resources at their will and not get that reconciled away? If anything, we should ensure that "our" annotations and labels are properly set? 🤔

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
antoineco: "formatted" maybe? Like this missing space after the comma for example, which would cause `go fmt` to fail on generated files.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
antoineco: We need to compare `currrent` to `desired` using `relect.DeepDerivative(desired, current)` from `k8s.io/apimachinery/third_party/forked/golang/reflect`. It returns `true` if there are additional values in `current` but none of the values from `desired` have changed.

markusthoemmes: We'll then also need to use `UnionMaps` instead of straight up overriding things.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
antoineco: I'm also not convinced the injected reconciler should do this on the user's behalf. I think it's the role of the user to decide which extraneous values are OK to have inside `current`, and which are not.
This is a slippery slope.

duglin:
Now that we have Annotations in Status, this isn't needed any more.
thanks @vaikas

PR #1283, opened 2020-05-04
add source metrics to report general errors and timeout

### LINKED ISSUE #1281: Source metrics does not support reporting client error
## Expected Behavior
The source stats reporter should have tags for representing client errors (eg. when `http.Client.Do()` returns an error, or other client, like Kafka...).

## Actual Behavior
No such tags.

lionelvillard:
Fixes #1281

### DIFF
--- a/metrics/metricskey/constants.go
+++ b/metrics/metricskey/constants.go
@@ -29,6 +29,12 @@ const (
 	// LabelNamespaceName is the label for immutable name of the namespace that the service is deployed
 	LabelNamespaceName = "namespace_name"
 
+	// ContainerName is the container for which the metric is reported.
+	ContainerName = "container_name"
+
+	// PodName is the name of the pod for which the metric is reported.
+	PodName = "pod_name"
+
 	// LabelResponseCode is the label for the HTTP response status code.
 	LabelResponseCode = "response_code"
 
--- a/metrics/metricskey/constants.go
+++ b/metrics/metricskey/constants.go
@@ -29,17 +29,17 @@ const (
 	// LabelNamespaceName is the label for immutable name of the namespace that the service is deployed
 	LabelNamespaceName = "namespace_name"
 
-	// ContainerName is the container for which the metric is reported.
-	ContainerName = "container_name"
-
-	// PodName is the name of the pod for which the metric is reported.
-	PodName = "pod_name"
-
 	// LabelResponseCode is the label for the HTTP response status code.
 	LabelResponseCode = "response_code"
 
 	// LabelResponseCodeClass is the label for the HTTP response status code class. For example, "2xx", "3xx", etc.
 	LabelResponseCodeClass = "response_code_class"
+
+	// LabelResponseError is the label for client error. For HTTP, A non-2xx status code doesn't cause an error.
+	LabelResponseError = "response_error"
+
+	// LabelResponseTimeout is the label timeout.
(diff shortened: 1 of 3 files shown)


metrics/utils.go
```diff
@@ -16,11 +16,39 @@ limitations under the License.
...
 
 // ResponseCodeClass converts an HTTP response code to a string representing its response code class.
 // E.g., The response code class is "5xx" for response code 503.
 func ResponseCodeClass(responseCode int) string {
 	// Get the hundred digit of the response code and concatenate "xx".
 	return strconv.Itoa(responseCode/100) + "xx"
 }
+
+// MaybeInsertInt conditionally insert the tag when cond is true.
+func MaybeInsertInt(key tag.Key, value int, cond bool) tag.Mutator {
```
antoineco, line 33: `ConditionallyDoXYZ` feels less uncertain than `MaybeDoXYZ` to me.

antoineco, line 33: Do you mean `InsertIntTag`? I think it's important to clarify, since the package name doesn't indicate the context here (`metrics` vs. `tag`)

lionelvillard: That's a fairly common terminology in functional programming (see the Maybe monad).. But I digress. `Maybe` is shorter :-)

lionelvillard: good point! Fixing...

metrics/metricskey/constants.go
```diff
@@ -29,18 +29,18 @@ const (
 	// LabelNamespaceName is the label for immutable name of the namespace that the service is deployed
 	LabelNamespaceName = "namespace_name"
 
-	// ContainerName is the container for which the metric is reported.
-	ContainerName = "container_name"
```
yanweiguo, line 33: Why are these two vars removed? They are used in downstream repos, for example: https://github.com/knative/serving/blob/master/pkg/metrics/key.go#L35

metrics/utils.go
antoineco: If shorter means better I can't really argue.

metrics/utils.go
```diff
@@ -16,11 +16,39 @@ limitations under the License.
...
 // E.g., The response code class is "5xx" for response code 503.
 func ResponseCodeClass(responseCode int) string {
 	// Get the hundred digit of the response code and concatenate "xx".
 	return strconv.Itoa(responseCode/100) + "xx"
 }
+
+// MaybeInsertIntTag conditionally insert the tag when cond is true.
+func MaybeInsertIntTag(key tag.Key, value int, cond bool) tag.Mutator {
+	if cond {
+		return tag.Insert(key, strconv.Itoa(value))
```
yanweiguo, line 35: You may want to use `tag.Upsert` instead of `tag.Insert` as `insert` won't update existing key value pair. See https://godoc.org/go.opencensus.io/tag#Insert

yanweiguo:
/hold

Some comments haven't be solved.

metrics/utils.go
lionelvillard: I think we want `Insert` here. This is just an utility function wrapping `tag.Insert`.

yanweiguo: I mean using `tag.Upsert` instead of `tag.Insert` in `source/source_stats_reporter.go`. Then this func will be `MaybeUpsertIntTag` and warpping `tag.Upsert`. Anyway this is something we can improve in separate PR.

PR #1284, opened 2020-05-04
Create a common method for processing reconciliation on generic KRShaped

whaught:
Issue #1226

- Creates a post-process function for reasoning generically about KRShaped resources
    - Automatically bumps ObservedGeneration
    - Sets Ready=Unknown if an error is encountered and the reconciler did not update conditions.

This is not yet invoked, the next PR will integrate this with genreconciler.


reconciler/reconcile_common.go
```diff
@@ -0,0 +1,49 @@
...
+	// generation regardless of success or failure.
+	newStatus.ObservedGeneration = new.GetObjectMeta().Generation
+
+	if newStatus.ObservedGeneration != old.GetStatus().ObservedGeneration && reconcileEvent != nil {
+		oldRc := old.GetStatus().GetCondition(apis.ConditionReady)
+		rc := newStatus.GetCondition(apis.ConditionReady)
+		// if a new generation is observed and reconciliation reported an error event
+		// the reconciler should change the ready state. By default we will set unknown.
+		if equality.Semantic.DeepEqual(oldRc, rc) {
+			logger.Warnw("A reconconiler observed a new generation without updating the resource status")
```
vagababov, line 45:
```suggestion
			logger.Warn("A reconconiler observed a new generation without updating the resource status")
```

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,128 @@
...
+			Conditions:         v1.Conditions{*foo, *bar},
+		},
+	}
+	return old, new
+}
+
+func TestPostProccessReconcileBumpsGeneration(t *testing.T) {
+	old, new := makeResources()
+
+	oldShape := interface{}(old).(v1.KRShaped)
```
vagababov, line 69: Do you need this `interface{}()` cast? Feel redundant?

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,49 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	"knative.dev/pkg/apis"
+	v1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/logging"
+)
+
+var condSet = apis.NewLivingConditionSet()
+
+// PostProccessReconcile contains logic to apply after reconciliation of a resource.
+func PostProccessReconcile(ctx context.Context, old v1.KRShaped, new v1.KRShaped, reconcileEvent Event) {
```
grantr, line 31: spelling nit: `PostProcess`

reconciler/reconcile_common_test.go
whaught: I believe so because old is a struct type and type assertions are only allowed between interfaces. Please do correct me if that's wrong (I'm a golang amateur)

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+	}
+	return old, new
+}
+
+func TestPostProcessReconcileBumpsGeneration(t *testing.T) {
+	old, new := makeResources()
+
+	oldShape := v1.KRShaped(old)
+	newShape := interface{}(new).(v1.KRShaped)
```
vagababov, line 70: here too?

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"knative.dev/pkg/apis"
+	v1 "knative.dev/pkg/apis/duck/v1"
+)
+
+func makeResources() (*v1.KResource, *v1.KResource) {
+	foo := &apis.Condition{
```
antoineco, line 31: The pointer is not necessary since your de-reference right below.

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"knative.dev/pkg/apis"
+	v1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/logging"
+)
+
+var condSet = apis.NewLivingConditionSet()
```
antoineco, line 28: Does it need to be a global var?

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if newShape.GetStatus().ObservedGeneration != newShape.GetObjectMeta().Generation {
+		t.Errorf("Expected observed generation bump got=%d want=%d", new.Status.ObservedGeneration, new.Generation)
+	}
+
+	// The old/new conditions were changed, expect that ready remains unchanged
+	if rc := new.Status.GetCondition(apis.ConditionReady); rc.Status != originalReadyStatus {
+		t.Errorf("Expected unchanged ready status got=%s want=%s", rc.Status, originalReadyStatus)
+	}
+}
```
antoineco, line 128: I think a table test would be a good fit here:
```go
testCases := []struct {
    name           string
    reconcileEvent *Event
    newCondition   *apis.Condition
}{
    {
        name:           "sets Unknown condition if returned event and no condition change",
        reconcileEvent: NewEvent(corev1.EventTypeWarning, exampleStatusFailed, ""),
    },
    {
        name:           "leaves condition unchanged if returned event and condition changed",
        reconcileEvent: NewEvent(corev1.EventTypeWarning, exampleStatusFailed, ""),
        newCondition:   &apis.Condition{Type: apis.ConditionReady, Status: corev1.ConditionFalse},
    },
}
```

antoineco:
A general question about this PR: why would we want to force the Ready condition to Unknown instead of letting the user decide what to do with it in case of error?

reconciler/reconcile_common.go
whaught: nah

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,49 @@
...
+	newStatus.ObservedGeneration = new.GetObjectMeta().Generation
+
+	if newStatus.ObservedGeneration != old.GetStatus().ObservedGeneration && reconcileEvent != nil {
+		oldRc := old.GetStatus().GetCondition(apis.ConditionReady)
+		rc := newStatus.GetCondition(apis.ConditionReady)
+		// if a new generation is observed and reconciliation reported an error event
+		// the reconciler should change the ready state. By default we will set unknown.
+		if equality.Semantic.DeepEqual(oldRc, rc) {
+			logger.Warn("A reconconiler observed a new generation without updating the resource status")
+			condSet.Manage(newStatus).MarkUnknown(apis.ConditionReady, "", "unsucessfully observed a new generation")
```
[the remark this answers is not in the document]
whaught: Sure we can include one. This is intended to be catching unhandled failures of the reconciler (Ideally implementers of reconciler always set a condition).

whaught:
> A general question about this PR: why would we want to force the Ready condition to Unknown instead of letting the user decide what to do with it in case of error?

We always want the implementer of the reconciler to be setting a condition when there is a failure. However now we will be incrementing observed generation automatically.

The client is typically waiting for obsGen == gen && ready to be terminal. If we were ready=true and then bumped generation on a failure without resetting the condition, clients would erroneously stop listening. The solution here is to catch failures and if the ready condition is unchanged, move it to unknown.

In a correctly implemented reconciler, the hope is that this status never gets hit. One alternative to this might be to set this status on generation bump _before_ the reconciler code runs and not bother with the old/new comparison?

antoineco:
What about a transient failure? E.g. some external service the reconciler depends on is unreachable, but the reconciled objects can be assumed to be _just fine_, unless we observe that failure for a certain amount of time, or a certain amount of retries? In that case I want some level of control, Knative shouldn't do some magic and unconditionally update the `Ready` condition on my behalf, other components may depend on it.

whaught:
> I agree that this might be OK for some/most implementations, but what about a transient failure? E.g. some external service the reconciler depends on is unreachable, but the reconciled objects can be assumed to be _just fine_, unless we observe that failure for a certain amount of time, or a certain amount of retries? In that case I want some level of control, Knative shouldn't do some magic and unconditionally update the `Ready` condition on my behalf, other components may depend on it.

I think in that case it would be reasonable for the reconciler to set conditions to _something_ (perhaps ready=true) as it processes the new generation rather than returning a reconciler.Event

antoineco:
That would be fine, but your code forces the Ready condition to `Unknown` as long as `Semantic.DeepEqual` returns `true`, so if the Ready condition is already `True`, it will flip to `Unknown`.

whaught:
> That would be fine, but your code forces the Ready condition to `Unknown` as long as `Semantic.DeepEqual` returns `true`, so if the Ready condition is already `True`, it will flip to `Unknown`.

Sorry I've changed the strategy with the above gist. I think preprocessing will be much simpler than trying to compare to the old resource.

reconciler/reconcile_common_test.go
whaught: I've changed this a bunch - should be much simpler now.

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,54 @@
...
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/logging"
+)
+
+const failedGenerationBump = "NewObservedGenFailure"
+
+// PreProcessReconcile contains logic to apply before reconciliation of a resource.
+func PreProcessReconcile(ctx context.Context, new duckv1.KRShaped) {
+	var condSet = apis.NewLivingConditionSet()
```
n3wscott, line 31: we might have to add extra logic here to make sure that Batch (Ready vs Succeeded) style condition sets also work, tekton uses this.

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/logging"
+)
+
+const failedGenerationBump = "NewObservedGenFailure"
+
+// PreProcessReconcile contains logic to apply before reconciliation of a resource.
+func PreProcessReconcile(ctx context.Context, new duckv1.KRShaped) {
```
n3wscott, line 30: using `new` is sometimes not so nice to go because `new` is also a keyword. We do it sometimes but it makes a warning

reconciler/reconcile_common.go
whaught: Done!

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,70 @@
...
+// PostProcessReconcile contains logic to apply after reconciliation of a resource.
+func PostProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
+	logger := logging.FromContext(ctx)
+	newStatus := resource.GetStatus()
+
+	// Bump observed generation to denote that we have processed this
+	// generation regardless of success or failure.
+	newStatus.ObservedGeneration = resource.GetObjectMeta().GetGeneration()
+
+	rc := newStatus.GetCondition(apis.ConditionReady)
```
n3wscott, line 55: looks like this needs the same treatment as above with the getDefaultConditionSet

whaught: ah missed that one. should be corrected now

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func getDefaultConditionSet(resource duckv1.KRShaped) (apis.ConditionSet, error) {
+	switch resource.GetTopLevelConditionType() {
+	case apis.ConditionReady:
+		return apis.NewLivingConditionSet(), nil
+	case apis.ConditionSucceeded:
+		return apis.NewBatchConditionSet(), nil
+	default:
+		return apis.ConditionSet{}, errors.Errorf("No ConditionSet found for type %s", resource.GetTypeMeta().Kind)
```
antoineco, line 68: I don't think we want to introduce a new dependency just for this, did you mean `fmt.Errorf`?

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/markbates/inflect v1.0.4
 	github.com/openzipkin/zipkin-go v0.2.2
+	github.com/pkg/errors v0.9.1
```
antoineco, line 28: See comment about `fmt.Errorf`.

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,70 @@
...
+	newStatus.ObservedGeneration = resource.GetObjectMeta().GetGeneration()
+
+	rc := newStatus.GetCondition(resource.GetTopLevelConditionType())
+	if rc.Reason == failedGenerationBump {
+		logger.Warn("A reconconiler observed a new generation without updating the resource status")
+	}
+}
+
+func getDefaultConditionSet(resource duckv1.KRShaped) (apis.ConditionSet, error) {
+	switch resource.GetTopLevelConditionType() {
```
antoineco, line 62: Why didn't we use the term `Happy` instead of `TopLevel` here? Not saying it's good or bad, but it's just the term condition sets use right now.

reconciler/reconcile_common.go
antoineco, line 68: > Error strings should not be capitalized
> -- https://github.com/golang/go/wiki/CodeReviewComments#error-strings

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"context"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+	v1 "knative.dev/pkg/apis/duck/v1"
```
antoineco, line 28: Double import, `v1` and `duckv1` are the same.

vendor/modules.txt
```diff
@@ -224,6 +224,7 @@ github.com/openzipkin/zipkin-go/reporter
 github.com/openzipkin/zipkin-go/reporter/http
 github.com/openzipkin/zipkin-go/reporter/recorder
 # github.com/pkg/errors v0.9.1
+## explicit
```
antoineco, line 227: See comment about `fmt.Errorf`

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,70 @@
...
+		return
+	}
+
+	newStatus := resource.GetStatus()
+	if newStatus.ObservedGeneration != resource.GetObjectMeta().GetGeneration() {
+		// Reset ready to unknown. The reconciler is expected to overwrite this.
+		manager := condSet.Manage(newStatus)
+		manager.MarkUnknown(
+			manager.GetTopLevelCondition().Type, failedGenerationBump, "unsucessfully observed a new generation")
+	}
```
antoineco, line 43: Don't we want to also call `InitializeConditions` here?

reconciler/reconcile_common.go
```diff
@@ -0,0 +1,70 @@
...
+	logger := logging.FromContext(ctx)
+	newStatus := resource.GetStatus()
+
+	// Bump observed generation to denote that we have processed this
+	// generation regardless of success or failure.
+	newStatus.ObservedGeneration = resource.GetObjectMeta().GetGeneration()
+
+	rc := newStatus.GetCondition(resource.GetTopLevelConditionType())
+	if rc.Reason == failedGenerationBump {
+		logger.Warn("A reconconiler observed a new generation without updating the resource status")
```
antoineco, line 57: Typo "reconconiler" -> "reconciler"

reconciler/reconcile_common.go
whaught: @n3wscott on https://github.com/knative/pkg/pull/1298 :

> I would call this GetTopLevelConditionType
> 
> happy was a tongue and cheek joke for the internal code.

I myself am a bit indifferent. The first iteration had "GetHappyConditionType()"

reconciler/reconcile_common.go
whaught: My fear was that I don't necessarily want to clobber _all_ the conditions and I believe they should have been initialized in the admission webhook at this point?

antoineco: Controllers [call that method usually](https://github.com/knative/eventing/blob/e51b8a490e58762ce1cbda695785f277d6595294/pkg/client/injection/reconciler/sources/v1alpha2/containersource/stub/reconciler.go#L51-L52), webhooks only initialize the `spec` if I'm not mistaken (not every type has an admission webhook, for some third-party types the OpenAPI schema present in the CRD is sufficient).

I'm OK leaving that as it is, I was just curious what you thought about it.

whaught: This is also not a complete condition set (we don't make any assumptions about the dependant sup-conditions here)

reconciler/reconcile_common.go
antoineco: Alright, glad that Scott could confirm 👍

reconciler/reconcile_common.go
whaught: hmm maybe we will want the full conditionset so that we can do this and at some point have the preprocessing do all the init stuff. I may reconsider this as another PR?

n3wscott:
This is great, nice work. I think this will blend well with our controllers and let us drop some easily forgotten copy/paste logic from the current reconcilers. 

👏 

/lgtm
/approve

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	readyCond := apis.Condition{
+		Type:    apis.ConditionType(topLevelCond),
+		Status:  corev1.ConditionTrue,
+		Message: "Something something bar",
+	}
+
+	return &duckv1.KResource{
+		ObjectMeta: metav1.ObjectMeta{
+			Generation: 1,
```
vagababov, line 44: For fun of it:
```suggestion
			Generation: 42,
```
? Or any other number. To make sure some defaulting somewhere does not just set 1.

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,99 @@
...
+			name:                      "top level Ready",
+			initTopLevelCond:          "Ready",
+			expectedTopLevelCondition: apis.ConditionReady,
+		},
+		{
+			name:                      "top level Succeeded",
+			initTopLevelCond:          "Succeeded",
+			expectedTopLevelCondition: apis.ConditionSucceeded,
+		},
+	}
```
vagababov, line 70:
```suggestion
	}{{
			name:                      "top level Ready",
			initTopLevelCond:          "Ready",
			expectedTopLevelCondition: apis.ConditionReady,
		},{
			name:                      "top level Succeeded",
			initTopLevelCond:          "Succeeded",
			expectedTopLevelCondition: apis.ConditionSucceeded,
		}}
```

reconciler/reconcile_common_test.go
```diff
@@ -0,0 +1,96 @@
...
+		expectedTopLevelCondition apis.ConditionType
+	}{{
+			name:                      "top level Ready",
+			initTopLevelCond:          "Ready",
+			expectedTopLevelCondition: apis.ConditionReady,
+		},{
+			name:                      "top level Succeeded",
+			initTopLevelCond:          "Succeeded",
+			expectedTopLevelCondition: apis.ConditionSucceeded,
+		}}
```
mattmoor, line 67: Format Go code:
```suggestion
		name:                      "top level Ready",
		initTopLevelCond:          "Ready",
		expectedTopLevelCondition: apis.ConditionReady,
	}, {
		name:                      "top level Succeeded",
		initTopLevelCond:          "Succeeded",
		expectedTopLevelCondition: apis.ConditionSucceeded,
	}}
```

mattmoor, line 67: Format Go code:
```suggestion
		name:                      "top level Ready",
		initTopLevelCond:          "Ready",
		expectedTopLevelCondition: apis.ConditionReady,
	}, {
		name:                      "top level Succeeded",
		initTopLevelCond:          "Succeeded",
		expectedTopLevelCondition: apis.ConditionSucceeded,
	}}
```

PR #1289, opened 2020-05-05
Fail reconciliation when a ReconcileEvent is wrapped inside an error

### LINKED ISSUE #1266: Implement retryable API events in injected Reconcilers
/area API
/kind feature

I like the way generated Reconcilers can extract an event from the returned error and record that event via the injected EventRecorder. 👍 

Unfortunately, there is no way to return an "error" event _and_ retry the reconciliation. What I want is to be able to return an actual error while still being able to provide a better classification than "InternalError".

I suggest we support wrapping events inside a specific error type, similarly to how `controller.NewPermanentError` creates a special type of error, and take that as a hint that the reconciliation should be retried after the event is recorded.

I'm gonna look into this over the weekend unless someone objects.

/assign

antoineco:
Allows errors to be re-enqueued event if they wrap a `ReconcilerEvent`.

The event is recorded whether it's wrapped or not, but a non-wrapped `ReconcilerEvent` is not considered as a `Reconcile()` error (current behaviour).

Too bad Kubernetes doesn't have a `corev1.EventTypeError`, _sighs_.

Closes #1266

---

examples:

```golang
// causes a re-enqueue
myevent := reconciler.NewEvent(corev1.EventTypeWarning, oopsReason, "oops")
return fmt.Errorf("this is bad: %w", myEvent)
```
```golang
// does NOT cause a re-enqueue
return reconciler.NewEvent(corev1.EventTypeWarning, oopsReason, "oops")
```
```golang
// reminder: permanentError can be used to prevent a re-enqueue regardless
// of the contents of the error
myevent := reconciler.NewEvent(corev1.EventTypeWarning, oopsReason, "oops")
return controller.NewPermanentError(fmt.Errorf("this is bad: %w", myEvent))
```

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -339,13 +339,19 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		if reconciler.EventAs(reconcileEvent, &event) {
 			logger.Infow("Returned an event", zap.Any("event", reconcileEvent))
 			r.Recorder.Eventf(resource, event.EventType, event.Reason, event.Format, event.Args...)
+
+			// the event was wrapped inside an error, consider the reconciliation as failed
+			if _, isEvent := reconcileEvent.(*reconciler.ReconcilerEvent); !isEvent {
+				return reconcileEvent
+			}
 			return nil
-		} else {
-			logger.Errorw("Returned an error", zap.Error(reconcileEvent))
-			r.Recorder.Event(resource, {{.corev1EventTypeWarning|raw}}, "InternalError", reconcileEvent.Error())
-			return reconcileEvent
 		}
+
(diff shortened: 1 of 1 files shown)


antoineco:
@dprotaso @n3wscott friendly ping.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -339,13 +339,19 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		if reconciler.EventAs(reconcileEvent, &event) {
```
n3wscott, line 339: Let's follow up by moving the reconciler event logic out of the generated code and into the reconciler base that @whaught is working on.

n3wscott: it would be nice to write some tests on this area of the codebase.

antoineco: Good idea, I'll see how I can help.

PR #1292, opened 2020-05-05
Hide metadata as a private

vagababov:
I am pretty sure it is unused outside of the package.

/assign @chaodaiG @chizhg

### DIFF
--- a/testutils/metahelper/client/client_test.go
+++ b/testutils/metahelper/client/client_test.go
@@ -99,8 +99,8 @@ func TestSync(t *testing.T) {
 		if (err == nil && data.expErr) || (err != nil && !data.expErr) {
 			log.Fatalf("%s\ngot: '%v', want: '%v'", errMsg, err, data.expErr)
 		}
-		if !reflect.DeepEqual(c.MetaData, data.expMetadata) {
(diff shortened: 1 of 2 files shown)


chaodaiG:
Agree, the real APIs are Get and Set

/lgtm
/approve

testutils/metahelper/client/client_test.go
```diff
@@ -99,8 +98,8 @@ func TestSync(t *testing.T) {
 		if (err == nil && data.expErr) || (err != nil && !data.expErr) {
 			log.Fatalf("%s\ngot: '%v', want: '%v'", errMsg, err, data.expErr)
 		}
-		if !reflect.DeepEqual(c.MetaData, data.expMetadata) {
-			log.Fatalf("%s\ngot: '%v', want: '%v'", errMsg, c.MetaData, data.expMetadata)
+		if !cmp.Equal(c.metadata, data.expMetadata) {
+			log.Fatalf("%s\ngot: '%v', want: '%v'", errMsg, c.metadata, data.expMetadata)
```
antoineco, line 102: Not directly related to your changes, but while you're at it, how about printing the result of `cmd.Diff` instead of dumping those maps in the logs?

PR #1293, opened 2020-05-05
Use go run to run codegen

### LINKED ISSUE #1294: update-codegen.sh produces inconsistent output
## Expected Behavior
Checking out master and running ./hack/update-codegen.sh should produce no changes.

## Actual Behavior
Running ./hack/update-codegen.sh produces changes to the generated controllers.

1. Checkout the master branch.
2. Run ./hack/update-codegen.sh

## Additional Info
Using go 1.14.2

ian-mi:
Fixes #1294
cc @grantr @chizhg

### DIFF
--- a/hack/generate-knative.sh
+++ b/hack/generate-knative.sh
@@ -46,12 +46,8 @@ APIS_PKG="$3"
 GROUPS_WITH_VERSIONS="$4"
 shift 4
 
-(
-  # To support running this script from anywhere, we have to first cd into this directory
-  # so we can install the tools.
-  cd $(dirname "${0}")
-  go install ../codegen/cmd/injection-gen
-)
+# To support running this script from anywhere, we have to first cd into this directory.
+cd $(dirname "${0}")/..
 
 function codegen::join() { local IFS="$1"; shift; echo "$*"; }
 
@@ -89,7 +85,7 @@ if grep -qw "injection" <<<"${GENS}"; then
   # Clear old injection
   rm -rf ${OUTPUT_PKG}
 
-  ${GOPATH}/bin/injection-gen \
+  go run knative.dev/pkg/codegen/cmd/injection-gen \
     --input-dirs $(codegen::join , "${FQ_APIS[@]}") \
     --versioned-clientset-package ${VERSIONED_CLIENTSET_PKG} \
     --external-versions-informers-package ${EXTERNAL_INFORMER_PKG} \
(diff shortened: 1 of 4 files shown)


coryrc:
I'd rather keep them as executables so it's easier to transition to having all our tools in the [prow-tests](https://github.com/knative/test-infra/tree/master/images/prow-tests) image.

Can you instead update the `go install` or `go get` to get whichever copy you would otherwise be running?

ian-mi:
I'm not sure I understand the requirement. I think we should always be building and running the vendored version of the generators rather than including them in the prow image.

chizhg:
> I'd rather keep them as executables so it's easier to transition to having all our tools in the [prow-tests](https://github.com/knative/test-infra/tree/master/images/prow-tests) image.
> 
> Can you instead update the `go install` or `go get` to get whichever copy you would otherwise be running?

@coryrc this is not a tool maintained by EngProd and the version has to be consistent with the vendored k8s versions, so it does not make much sense to move it to `prow-tests` image.

/lgtm
/assign @mattmoor

hack/update-codegen.sh
```diff
@@ -64,7 +64,7 @@ ${CODEGEN_PKG}/generate-groups.sh "deepcopy" \
   --go-header-file ${REPO_ROOT_DIR}/hack/boilerplate/boilerplate.go.txt
 
 # Depends on generate-groups.sh to install bin/deepcopy-gen
```
coryrc, line 66: Ah, now that I understand the requirement, you aren't using the vendored version here AFAICT, but whatever happens to be in `$GOPATH/src/k8s.io`, at least as far as I can understand from reading `go help packages`

It think should be `go run $(dirname $0)/../vendor/k8s.io/code-generator/cmd/deepcopy-gen`

The other instance happens to be okay because this is knative.dev/pkg... but you may want to change it to be consistent

coryrc:
/hold
I don't think it will have the effect you desire (see comment I added)
But if you are sure, do /hold cancel

hack/generate-knative.sh
```diff
@@ -89,7 +85,7 @@ if grep -qw "injection" <<<"${GENS}"; then
   # Clear old injection
   rm -rf ${OUTPUT_PKG}
 
-  ${GOPATH}/bin/injection-gen \
+  go run knative.dev/pkg/codegen/cmd/injection-gen \
```
mattmoor, line 88: I thought you had to pass the list of Go files to go run...

coryrc: directory is fine

hack/update-codegen.sh
ian-mi: I'm not sure this is true. I tried removing code-generator from my GOPATH and the script succeeded without pulling code-generator back into GOPATH. Since the repo is now using go modules so GOPATH should not be used, and since the go.mod version is >= 1.14 go will default to running with -mod=vendor. Therefore go run should run by default using vendored modules. I also tested this by vendoring a change to deepcopy gen and checking for the changed output.

coryrc: okay, I forgot go modules affects a lot of functionality of go itself. Glad it worked!

n3wscott: vendor acts as GOPATH in go mod projects

PR #1295, opened 2020-05-05
ran ./update-codegen.sh

dprotaso:
Our `knative.dev/pkg`'s codegen script wasn't invoked 

cc @shashwathi @andrew-su


dprotaso:
I dunno why `test-infra` didn't catch this @coryrc - since it's using the default build runner

andrew-su:
/lgtm

Thanks for catching that. It usually complains when update-codegen.sh isn't ran as part of `pull-knative-pkg-build-tests`.

andrew-su:
For context, https://github.com/knative/pkg/pull/1240 was the change that update-codegen.sh should have been invoked for

ian-mi:
Verify codegen was run on that PR, I believe the issue is #1294. codegen seems to be inconsistent since the switch to go modules.

mattmoor:
I think I covered this in my fix in https://github.com/knative/pkg/pull/1296?

dprotaso:
yup looks like it

PR #1296, opened 2020-05-05
Properly set hasClass

mattmoor:
This was changed in #1277

/hold

I wanna verify in serving first

### DIFF
--- a/codegen/cmd/injection-gen/generators/packages.go
+++ b/codegen/cmd/injection-gen/generators/packages.go
@@ -193,13 +193,13 @@ func extractCommentTags(t *types.Type) map[string]map[string]string {
 	return ExtractCommentTags("+", comments)
 }
 
-func extractReconcilerClassTag(tags map[string]map[string]string) (classname string, has bool) {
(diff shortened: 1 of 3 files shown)


codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -193,13 +193,13 @@ func extractCommentTags(t *types.Type) map[string]map[string]string {
...
-	if !has {
-		return
+func extractReconcilerClassTag(tags map[string]map[string]string) (string, bool) {
+	vals, ok := tags["genreconciler"]
+	if !ok {
+		return "", false
 	}
-	classname, _ = vals["class"]
-	return
+	classname, has := vals["class"]
```
whaught, line 201: is `return vals["class"]` the same here?

mattmoor: No, I tried that first :(

Go special cases subscript on maps to have one or two values, but it isn't smart enough to use the latter here.

mattmoor:
/hold cancel

This works for me

PR #1298, opened 2020-05-05
Include GetTopLevelConditionType for KRShaped

whaught:
Issue #1226

- Each type will report its expected happy condition
- KRShaped will detect its happy type by looking at its own conditions.
- Added a unit test

At the moment each implementer either makes a LivingConditionSet manager or a BatchConditionSet manager within each resource. However this allows for polymorphic logic since each type will report its own happy condition.

Longer term we could simplify by just having GetConditionSet(resource) and the manager can figure out what the condition ought to be. This will make it easier to share common logic.

### DIFF
--- a/apis/duck/v1/kresource_type.go
+++ b/apis/duck/v1/kresource_type.go
@@ -34,7 +34,7 @@ type KRShaped interface {
 
 	GetStatus() *Status
 
-	GetHappyConditionType() apis.ConditionType
+	GetTopLevelConditionType() apis.ConditionType
 }
 
 // Asserts KResource conformance with KRShaped
@@ -91,8 +91,8 @@ func (t *KResource) GetStatus() *Status {
 	return &t.Status
 }
 
-// GetHappyConditionType retrieves the happy condition of this resource. Implements the KRShaped interface.
-func (t *KResource) GetHappyConditionType() apis.ConditionType {
+// GetTopLevelConditionType retrieves the happy condition of this resource. Implements the KRShaped interface.
+func (t *KResource) GetTopLevelConditionType() apis.ConditionType {
 	// Note: KResources are unmarshalled from existing resources. This will only work properly for resources that
 	// have already been initialized to their type.
 	if cond := t.Status.GetCondition(apis.ConditionSucceeded); cond != nil {
--- a/apis/duck/v1/kresource_type.go
+++ b/apis/duck/v1/kresource_type.go
@@ -33,6 +33,8 @@ type KRShaped interface {
 	GetTypeMeta() *metav1.TypeMeta
 
 	GetStatus() *Status
+
+	GetHappyConditionType() apis.ConditionType
 }
 
 // Asserts KResource conformance with KRShaped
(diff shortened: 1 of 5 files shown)


apis/duck/v1/kresource_type.go
```diff
@@ -33,6 +33,8 @@ type KRShaped interface {
 	GetTypeMeta() *metav1.TypeMeta
 
 	GetStatus() *Status
+
+	GetHappyConditionType() apis.ConditionType
```
n3wscott, line 37: I would call this GetTopLevelConditionType

happy was a tongue  and cheek joke for the internal code.

apis/duck/v1/kresource_type.go
```diff
@@ -88,3 +90,13 @@ func (t *KResource) GetTypeMeta() *metav1.TypeMeta {
 func (t *KResource) GetStatus() *Status {
 	return &t.Status
 }
+
+// GetHappyConditionType retrieves the happy condition of this resource. Implements the KRShaped interface.
+func (t *KResource) GetHappyConditionType() apis.ConditionType {
+	// Note: KResources are unmarshalled from existing resources. This will only work properly for resources that
+	// have already been initialized to their type.
+	if cond := t.Status.GetCondition(apis.ConditionSucceeded); cond != nil {
```
n3wscott, line 98: this will not work for un-initialized resources, the condition set will be an empty list.

apis/duck/v1/kresource_type_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"knative.dev/pkg/apis"
+)
+
+func TestGetHappyCondition(t *testing.T) {
+	resource := KResource{}
+
+	condSet := apis.NewLivingConditionSet("Foo")
+	mgr := condSet.Manage(resource.GetStatus())
+	mgr.InitializeConditions()
```
n3wscott, line 30: I think for resources that have not yet called InitializeConditions() the api will be confused

apis/test/example/v1alpha1/fiz_types.go
```diff
@@ -103,3 +103,8 @@ func (f *ClusterFiz) GetTypeMeta() *metav1.TypeMeta {
 func (f *ClusterFiz) GetStatus() *duckv1.Status {
 	return &f.Status.Status
 }
+
+// GetHappyConditionType retrieves the happy condition of this resource. Implements the KRShaped interface.
+func (f *ClusterFiz) GetHappyConditionType() apis.ConditionType {
+	return apis.ConditionReady
```
n3wscott, line 109: ooh, it is set in the api. ok I got it a bit better

apis/duck/v1/kresource_type.go
n3wscott: sorry I did not see the final impl, this will work but it wil also require all downstreams to update

whaught: I have that included as a comment. The idea is that each API shape reports its type and gets initialized. KResource is a generic struct we unmarshal to, but not something that directly manages init.

apis/duck/v1/kresource_type_test.go
whaught: Agreed.

apis/test/example/v1alpha1/fiz_types.go
whaught: Yea downstream will set like this. After this, I'll keep incorporating here: https://github.com/knative/serving/pull/7847

Annoyingly I'll have to ensure I have the right type per every resource (which as a newbie I don't know off the top of my head)

apis/duck/v1/kresource_type.go
```diff
@@ -33,6 +33,8 @@ type KRShaped interface {
 	GetTypeMeta() *metav1.TypeMeta
 
 	GetStatus() *Status
+
+	GetTopLevelConditionType() apis.ConditionType
```
antoineco, line 37: Too late, but I'm curious why we didn't stick to the current naming with here (`Happy` instead of `TopLevel`).

Issue #1181, opened 2020-03-31
RFC: Per-reconciler leader election (and leader-aware reconcilers)


mattmoor:
/area API
/kind cleanup

I've been thinking a bit about leader election, and the way we run things today we basically allocate N backup pods that wait to become the leader and effectively do nothing until they become the leader, they are essentially there as scheduling reservations.  

How we achieve this is by wrapping the *entire* controller bootstrap process in a giant resource lock, which kicks things off when it acquires the lock and terminates the process if it should lose it.  This means that our Reconcilers get no benefit from horizontal scaling because no load is spread, and in fact they pay a penalty of having to also run the leader election logic as well.

Given 10 reconcilers living in a single controller, by leader electing at the pod level, leader failovers result in a standby pod spinning up 10 reconcilers worth of state before they can become functional again.  Given that we build none of that state in advance, it doesn't seem like a particularly hot failover.

---

The `tl;dr` of what I've been considering is this:
1. Have per `controller.Impl` / `Reconciler` leaders
2. Run all of the informers and controllers, eliding* reconciliation when not the leader.
3. When we become the leader, global resync.

_* - We may want to run leaderelection-aware reconcilers even when not the leader, e.g. if the resources program routing information for an admission control webhook._

The cons of this approach are:
 - Higher resource utilization for non-leader pods because they have the full informer cache.
 - A multiple on the number of leader locks we would need, and a corresponding increase in API Server load.

The pros of this approach are:
 - Leaders can be spread across replicas, so instead of having 10 on the single leader, if I run 3 replicas, I might only run 3-4 reconcilers per Pod.  So our controllers start to see benefits from horizontal scaling.
 - Hot failovers, which best case cost O(nop global resync).  This is amplified by the above because generally fewer things should be failing over each time a pod fails.

---

I think it'd be interesting to try this with our webhooks, which don't currently perform leader election, and in some cases (e.g. Bindings), need the informer cache to program the webhook.  If we decide to move ahead with this approach, then I think we should probably integrate with the `// +genreconciler` work, which I believe should have all of the information to do all(?) of the heavy lifting for (now) the vast majority of our reconcilers.

I'd love to try this out, but my schedule is chaotic and I don't want to be a bottleneck if others have time and are interested.

cc @pmorie @markusthoemmes @dprotaso @vaikas @vagababov for thoughts (LMK if I am missing any pros/cons)

lionelvillard:
Here some random thoughts:
[...]

PR #1301, opened 2020-05-05
Implement the first wave of per-reconciler leaderelection.

mattmoor:
Detailed design: https://docs.google.com/document/d/1i_QHjQO2T3SNv49xjZLWlivcc0UvZN1Tbw2NKxThkyM/edit

Issue: https://github.com/knative/pkg/issues/1181

WIP as we discuss the doc, but functionally complete and comments are welcome.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -805,6 +805,7 @@ func TestStartAndShutdown(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	go func() {
@@ -872,6 +873,7 @@ func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	go func() {
@@ -884,7 +886,9 @@ func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
 		// We expect to be promoted immediately, since there is no
 		// ElectorBuilder attached to the context.
 	case <-doneCh:
-		t.Error("StartAll finished early.")
+		t.Fatal("StartAll finished early.")
+	case <-time.After(10 * time.Second):
+		t.Error("Timed out waiting for StartAll.")
 	}
 
 	cancel()
@@ -897,7 +901,7 @@ func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
 	}
 
 	if got, want := r.count, 0; got != want {
-		t.Errorf("count = %v, wanted %v", got, want)
+		t.Errorf("reconcile count = %v, wanted %v", got, want)
 	}
 }
 
@@ -938,6 +942,7 @@ func TestStartAndShutdownWithLeaderAwareWithLostElection(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	ctx = leaderelection.WithStandardLeaderElectorBuilder(ctx, kc, cc)
 	doneCh := make(chan struct{})
 
@@ -965,7 +970,7 @@ func TestStartAndShutdownWithLeaderAwareWithLostElection(t *testing.T) {
 	}
 
 	if got, want := r.count, 0; got != want {
-		t.Errorf("count = %v, wanted %v", got, want)
+		t.Errorf("reconcile count = %v, wanted %v", got, want)
 	}
 }
 
@@ -976,6 +981,7 @@ func TestStartAndShutdownWithWork(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", reporter)
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	impl.EnqueueKey(types.NamespacedName{Namespace: "foo", Name: "bar"})
@@ -1001,10 +1007,10 @@ func TestStartAndShutdownWithWork(t *testing.T) {
 	}
 
 	if got, want := r.count, 1; got != want {
-		t.Errorf("Count = %v, wanted %v", got, want)
+		t.Errorf("reconcile count = %v, wanted %v", got, want)
 	}
 	if got, want := impl.WorkQueue.NumRequeues(types.NamespacedName{Namespace: "foo", Name: "bar"}), 0; got != want {
-		t.Errorf("Count = %v, wanted %v", got, want)
+		t.Errorf("requeues = %v, wanted %v", got, want)
 	}
 
 	checkStats(t, reporter, 1, 0, 1, trueString)
@@ -1052,6 +1058,7 @@ func TestStartAndShutdownWithErroringWork(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", reporter)
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	impl.EnqueueKey(types.NamespacedName{Namespace: "", Name: "bar"})
@@ -1106,6 +1113,7 @@ func TestStartAndShutdownWithPermanentErroringWork(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", reporter)
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	impl.EnqueueKey(types.NamespacedName{Namespace: "foo", Name: "bar"})
@@ -1197,6 +1205,7 @@ func TestImplGlobalResync(t *testing.T) {
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
 
 	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
 	doneCh := make(chan struct{})
 
 	go func() {
@@ -1463,6 +1472,7 @@ func TestRunInformersFinished(t *testing.T) {
 	}()
 
 	ctx, cancel := context.WithCancel(TestContextWithLogger(t))
+	t.Cleanup(cancel)
 
 	waitInformers, err := RunInformers(ctx.Done(), fi)
 	if err != nil {
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -714,7 +714,7 @@ func TestEnqueues(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			defer ClearAll()
+			t.Cleanup(ClearAll)
 			impl := NewImplWithStats(&NopReconciler{}, TestLogger(t), "Testing", &FakeStatsReporter{})
 			test.work(impl)
 
@@ -731,7 +731,7 @@ func TestEnqueues(t *testing.T) {
 }
 
 func TestEnqeueAfter(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	impl := NewImplWithStats(&NopReconciler{}, TestLogger(t), "Testing", &FakeStatsReporter{})
 	impl.EnqueueAfter(&Resource{
 		ObjectMeta: metav1.ObjectMeta{
@@ -767,7 +767,7 @@ func TestEnqeueAfter(t *testing.T) {
 }
 
 func TestEnqeueKeyAfter(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	impl := NewImplWithStats(&NopReconciler{}, TestLogger(t), "Testing", &FakeStatsReporter{})
 	impl.EnqueueKeyAfter(types.NamespacedName{Namespace: "waiting", Name: "for"}, 5*time.Second)
 	impl.EnqueueKeyAfter(types.NamespacedName{Namespace: "the", Name: "waterfall"}, 500*time.Millisecond)
@@ -800,7 +800,7 @@ func (cr *CountingReconciler) Reconcile(context.Context, string) error {
 }
 
 func TestStartAndShutdown(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	r := &CountingReconciler{}
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
 
@@ -832,16 +832,16 @@ func TestStartAndShutdown(t *testing.T) {
 	}
 }
 
-type CountingLeaderAwareReconciler struct {
+type countingLeaderAwareReconciler struct {
 	reconciler.LeaderAwareFuncs
 
 	m     sync.Mutex
 	Count int
 }
 
-var _ reconciler.LeaderAware = (*CountingLeaderAwareReconciler)(nil)
+var _ reconciler.LeaderAware = (*countingLeaderAwareReconciler)(nil)
 
-func (cr *CountingLeaderAwareReconciler) Reconcile(ctx context.Context, key string) error {
+func (cr *countingLeaderAwareReconciler) Reconcile(ctx context.Context, key string) error {
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		return err
@@ -859,9 +859,9 @@ func (cr *CountingLeaderAwareReconciler) Reconcile(ctx context.Context, key stri
 }
 
 func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	promoted := make(chan struct{})
-	r := &CountingLeaderAwareReconciler{
+	r := &countingLeaderAwareReconciler{
 		LeaderAwareFuncs: reconciler.LeaderAwareFuncs{
 			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) {
 				close(promoted)
@@ -901,9 +901,9 @@ func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
 }
 
 func TestStartAndShutdownWithLeaderAwareWithLostElection(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	promoted := make(chan struct{})
-	r := &CountingLeaderAwareReconciler{
+	r := &countingLeaderAwareReconciler{
 		LeaderAwareFuncs: reconciler.LeaderAwareFuncs{
 			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) {
 				close(promoted)
@@ -968,7 +968,7 @@ func TestStartAndShutdownWithLeaderAwareWithLostElection(t *testing.T) {
 }
 
 func TestStartAndShutdownWithWork(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	r := &CountingReconciler{}
 	reporter := &FakeStatsReporter{}
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", reporter)
@@ -1044,7 +1044,7 @@ func (er *ErrorReconciler) Reconcile(context.Context, string) error {
 }
 
 func TestStartAndShutdownWithErroringWork(t *testing.T) {
-	defer ClearAll()
+	t.Cleanup(ClearAll)
 	r := &ErrorReconciler{}
 	reporter := &FakeStatsReporter{}
 	impl := NewImplWithStats(r, TestLogger(t), "Testing", reporter)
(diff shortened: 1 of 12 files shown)


lionelvillard:
`leader.go` seems to be generic enough to be either in `leaderelection` or somewhere else.

evankanderson:
Matt, do you want to pick out and `/assign` some reviewers? I'm willing to do it, but in the absence of specific ownership, I fear this won't get reviewed in a timely fashion.

mattmoor:
@evankanderson Waiting on the feature track doc to get approvals before landing any of the pieces here, which is why I've left this as WIP.

leaderelection/context.go
```diff
@@ -0,0 +1,225 @@
...
+	// The resource name is the lowercase:
+	//   {component}.{workqueue}.{index}-of-{total}
+	return strings.ToLower(fmt.Sprintf("%s.%s.%02d-of-%02d", b.component, b.name, b.index, b.total))
+}
+
+// Has implements reconciler.Bucket
+func (b *bucket) Has(nn types.NamespacedName) bool {
+	h := fnv.New32a()
+	h.Write([]byte(nn.Namespace + "." + nn.Name))
+	ii := h.Sum32() % b.total
```
vagababov, line 223: This is going to be bad when `total` is power of 2 — you basically discard upper `32-log(total)` bits.
We can probably specialize that case
```
if b.total & (b.total-1) == 0 {
  hash = hash ^ (hash >> bits(b.total))
}
```

vagababov: Alternatives are also if we want to do some kind of consistent hashing, so during resize the objects don't migrate much from bucket to bucket — especially important for autoscaler. Less for others.
If we're doing online resizing.

reconciler/leader.go
```diff
@@ -0,0 +1,122 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Bucket is an opaque type used to scope leadership.
+type Bucket interface {
+	// Name returns a string representing this bucket, which uniquely
+	// identifies the bucket and is suitable for use as a resource lock name.
+	Name() string
+
+	// Has determines whether this Bucket covers a particular key.
```
vagababov, line 31:
```suggestion
	// Has determines whether this Bucket contains a particular key.
```
:)

reconciler/leader.go
```diff
@@ -0,0 +1,122 @@
...
+	Has(key types.NamespacedName) bool
+}
+
+// AllBuckets returns a Bucket that "Has()" all keys.
+func AllBuckets() Bucket {
```
vagababov, line 36: UniversalBucket :D

reconciler/leader.go
```diff
@@ -0,0 +1,122 @@
...
+func (laf *LeaderAwareFuncs) IsLeaderFor(key types.NamespacedName) bool {
+	laf.RLock()
+	defer laf.RUnlock()
+
+	for _, bkt := range laf.buckets {
+		if bkt.Has(key) {
+			return true
+		}
+	}
+	return false
```
vagababov, line 76: This is o(n) loop for each reconciliation.
May be we should invert the relationship and say "key x belong to bucket i" — do I own bucket i?
Buckets can be then map[int]bucket — to ensure O(1) lookup here.

leaderelection/context.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (b *standardBuilder) BuildElector(ctx context.Context, la reconciler.LeaderAware, name string, enq func(reconciler.Bucket, types.NamespacedName)) (Elector, error) {
+	logger := logging.FromContext(ctx)
+
+	id, err := UniqueID()
+	if err != nil {
+		return nil, err
+	}
+
+	buckets := make([]Elector, 0, b.lec.Buckets)
```
vagababov, line 90: we can probably optimize and use `AllBuckets` here if bucket count = 1.

reconciler/leader.go
mattmoor: I think of the bucket factor as a generally small constant factor, so I'm not super worried about this right now.

Perhaps we can introduce a BucketSet construct that we can use to optimize this later, since I expect we'll control all of the implementations of Bucket for the foreseeable future.

mattmoor:
> leader.go seems to be generic enough to be either in leaderelection or somewhere else

Perhaps, but the concepts are intended to apply to Reconciliation, and the bucketing is intended to partition the key-space (`types.NamespacedName`).  I think the `reconciler.Foo` will help frame some of the names in ways that contextualize them and make them clearer.

The nice thing about Go interfaces is you can just copy them elsewhere later (if you want) and they still apply! 😉

mattmoor:
/hold

I'm going to change the hold type here.  Dave is the last approver on the feature track document, but also a reviewer here.  I've stopped editing these (other than rebases and review comments) a few weeks ago, so the main thing blocking this is the feature track approvals and code review.

controller/controller_test.go
```diff
@@ -825,6 +832,141 @@ func TestStartAndShutdown(t *testing.T) {
...
+	}) {
+		cr.m.Lock()
+		defer cr.m.Unlock()
+		cr.Count++
+	}
+	return nil
+}
+
+func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
+	defer ClearAll()
```
vagababov, line 862:
```suggestion
	t.Cleanup(ClearAll)
```
👅 
So in other places.

leaderelection/context.go
```diff
@@ -0,0 +1,225 @@
...
+			defer sg.Done()
+			le.Run(ctx)
+		}(le)
+	}
+}
+
+// runUntilCancelled wraps a single-term Elector into one that runs until
+// the passed context is cancelled.
+type runUntilCancelled struct {
+	le Elector
```
vagababov, line 177:
```suggestion
 Elector
```
inline perhaps?

leaderelection/context.go
vagababov: As discussed I have a more complicated solution, which we'll put in once this merges.

reconciler/leader_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestLeaderAwareFuncs(t *testing.T) {
+
```
vagababov, line 27:
```suggestion
```

leaderelection/context.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// Run implements Elector
+func (ruc *runUntilCancelled) Run(ctx context.Context) {
+	// Run executes a single iteration of an election:
+	// 1. Wait to become leader
+	// 2. Call OnStartedLeading
+	// 3. Wait for ctx cancel or loss of leader.
+	// 4. Call OnStoppedLeading
```
yanweiguo, line 186: This is only correct when the field `le` is a LeadElector of k8s.io/client-go/tools/leaderelect. Maybe make it clear?

leaderelection/context.go
```diff
@@ -0,0 +1,225 @@
...
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/reconciler"
+	"knative.dev/pkg/system"
+)
+
+// WithLeaderElectorBuilder infuses a context with the ability to build
+// LeaderElectors with the provided component configuration acquiring resource
+// locks via the provided kubernetes client.
+func WithLeaderElectorBuilder(ctx context.Context, kc kubernetes.Interface, cc ComponentConfig) context.Context {
```
yanweiguo, line 38: How about naming it `WithStandardLeaderElectorBuilder` to distinguish from the builder for statefulset in the coming future?

reconciler/leader.go
yanweiguo: We need "key x belong to bucket i" relationship looking up for StatefulSet case.

reconciler/leader_test.go
```diff
@@ -0,0 +1,83 @@
...
+	laf := LeaderAwareFuncs{}
+
```
yanweiguo, line 29:
```suggestion

```

reconciler/leader_test.go
```diff
@@ -0,0 +1,83 @@
...
+	wantBkt := UniversalBucket()
+
```
yanweiguo, line 31:
```suggestion
```

reconciler/leader_test.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	laf.PromoteFunc = func(bkt Bucket, gotFunc func(Bucket, types.NamespacedName)) {
+		gotFunc(bkt, wantKey)
+		if !called {
+			t.Error("gotFunc didn't call wantFunc!")
+		}
+
+		// Check that we're not called while the lock is held,
```
yanweiguo, line 53: I don't understand the first half of this sentence. Looks like we only check that "we are already the Leader" here.

mattmoor: rephrased

leaderelection/context.go
mattmoor: done

reconciler/leader.go
vagababov: https://gist.github.com/vagababov/47e5f62c5865dbcc9deb1cd48655d7d2

vagababov: This is the algorithm I am going to swap in.

leaderelection/config.go
```diff
@@ -30,6 +30,10 @@ import (
 
 const ConfigMapNameEnv = "CONFIG_LEADERELECTION_NAME"
 
+// MaxBuckets is the maximum number of buckets to allow users to define.
+// This is a variable so that it may be customized in the binary entrypoint.
+var MaxBuckets uint32 = 10
```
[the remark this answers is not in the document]
vagababov: Also why uint32: here and in the CM.
wouldn't int be just as good?  We validate the value anyway (against maxbuckets, so can just as well check > 0). And this would allow us not to add one more method to the cm parsing library.

or `for i := uint32(0); ` 🌮

leaderelection/context.go
```diff
@@ -0,0 +1,219 @@
...
+					la.Promote(bkt, enq)
+				},
+				OnStoppedLeading: func() {
+					logger.Infof("%q has stopped leading %q", rl.Identity(), bkt.Name())
+					la.Demote(bkt)
+				},
+			},
+			ReleaseOnCancel: true,
+
+			// TODO: use health check watchdog, knative/pkg#1048
```
vagababov, line 129: Do we need both comments? (5 lines below references same stuff).

reconciler/leader.go
```diff
@@ -0,0 +1,118 @@
...
+
+// Name implements Bucket
+func (b *bucket) Name() string {
+	return ""
+}
+
+// Has implements Bucket
+func (b *bucket) Has(nn types.NamespacedName) bool {
+	return true
+}
```
vagababov, line 118: should those be next to the `UniversalBucket`?

controller/controller_test.go
```diff
@@ -825,8 +832,143 @@ func TestStartAndShutdown(t *testing.T) {
 	}
 }
 
+type countingLeaderAwareReconciler struct {
+	reconciler.LeaderAwareFuncs
+
+	m     sync.Mutex
+	Count int
```
vagababov, line 839: `Count` -> `count`?

leaderelection/config.go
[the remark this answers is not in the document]
mattmoor: > this would allow us not to add one more method to the cm parsing library

I always get a little annoyed with our helpers (e.g. `ptr`) when certain basic types are missing, and my choice is either patching upstream (heavy) or casting downstream.

It's less annoying here because I'm not calling an API that requires `uint32`, but it's also easy enough to add the helper while I'm here 😇

leaderelection/context.go
mattmoor: good catch

reconciler/leader.go
mattmoor: 🤩

reconciler/leader.go
mattmoor: done

reconciler/leader.go
```diff
@@ -0,0 +1,118 @@
...
+// Promote implements LeaderAware
+func (laf *LeaderAwareFuncs) Promote(b Bucket, enq func(Bucket, types.NamespacedName)) {
+	promote := func() func(Bucket, func(Bucket, types.NamespacedName)) {
+		laf.Lock()
+		defer laf.Unlock()
+		if laf.buckets == nil {
+			laf.buckets = make(map[string]Bucket, 1)
+		}
+		laf.buckets[b.Name()] = b
+		return laf.PromoteFunc
```
dprotaso, line 98: returning the `PromoteFunc` seems unnecessary

reconciler/leader.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+// Demote implements LeaderAware
+func (laf *LeaderAwareFuncs) Demote(b Bucket) {
+	demote := func() func(Bucket) {
+		laf.Lock()
+		defer laf.Unlock()
+		delete(laf.buckets, b.Name())
+		return laf.DemoteFunc
```
dprotaso, line 112: returning the `DemoteFunc ` seems unnecessary

reconciler/leader.go
```diff
@@ -0,0 +1,118 @@
...
+package reconciler
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Bucket is an opaque type used to scope leadership.
+type Bucket interface {
```
dprotaso, line 26: Do you think the types in this file are better suited for the `leaderelection` package?

leaderelection/context_test.go
```diff
@@ -0,0 +1,143 @@
+// +build !race
```
dprotaso, line 1: is the race not avoidable? this seems like a smell

reconciler/leader.go
```diff
@@ -0,0 +1,118 @@
...
+// Has implements Bucket
+func (b *bucket) Has(nn types.NamespacedName) bool {
+	return true
+}
+
+// LeaderAware is implemented by Reconcilers that are aware of their leader status.
+type LeaderAware interface {
+	// Promote is called when we become the leader of a given Bucket.  It must be
+	// supplied with an enqueue function through which a Bucket resync may be triggered.
+	Promote(b Bucket, enq func(Bucket, types.NamespacedName))
```
dprotaso, line 58: I think we should allow the `Promote` callback to signal if there were any errors. This should allow the leader election runners to cancel the lease. Looking at the changes in the `genreconciler` this would happen if we failed to list all objects in order to re-sync.

```suggestion
	Promote(b Bucket, enq func(Bucket, types.NamespacedName)) error
```

dprotaso, line 58: I'm still on the fence about the `enq` callback 

`NewReconciler` generated code seems to use it because they're no higher awareness of a workqueue. That could potentially be solved by passing in a `EnqueueKey` function. `controller.NewImpl` is a similar story.

All this feels like a work around because our controller.Impl construction is coupled with the workqueue. It's a chicken and egg problem.  Reconcilers want an enqueue func, but controllers need a reconciler but only when creating a controller do we create a work queue.

If we moved the responsibility of creating the queue outside of the controller I wonder if this gives us the flexibility we want.  Then `MaybeEnqueueBucketKey(reconciler.Bucket, types.NamespacedName)` would only need to a helper and not passed around everywhere.

leaderelection/context.go
```diff
@@ -0,0 +1,218 @@
...
+			Lock:          rl,
+			LeaseDuration: b.lec.LeaseDuration,
+			RenewDeadline: b.lec.RenewDeadline,
+			RetryPeriod:   b.lec.RetryPeriod,
+			Callbacks: leaderelection.LeaderCallbacks{
+				OnStartedLeading: func(context.Context) {
+					logger.Infof("%q has started leading %q", rl.Identity(), bkt.Name())
+					la.Promote(bkt, enq)
+				},
+				OnStoppedLeading: func() {
```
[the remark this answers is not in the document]
mattmoor: 1. The only return path from `Run` is [here](https://github.com/kubernetes/client-go/blob/18c1ef0354fe489b5d9a58088676f854be243aa2/tools/leaderelection/leaderelection.go#L204) which I understood to only happen when context is cancelled.  This should also prevent iteration in the loop below.

2. My understanding is that `defer runtime.HandleCrash()` is intended to keep panics from killing things.

3. Yes, but the `le.renew` call is not and blocks the return.

leaderelection/context.go
```diff
@@ -0,0 +1,218 @@
...
+					logger.Infof("%q has stopped leading %q", rl.Identity(), bkt.Name())
+					la.Demote(bkt)
+				},
+			},
+			ReleaseOnCancel: true,
+
+			Name: b.lec.Component,
```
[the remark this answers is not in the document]
mattmoor:
```
	// Name is the name of the resource lock for debugging
	Name string
```

Looks like it is non-semantic, so happy to change this.  I don't understand why they don't use `rl.Identity()` for this though...

reconciler/leader.go
mattmoor: No because there are non-LE ways of sharding up a Reconciler's key-space, e.g. what we're exploring with statefulset ordinals.

reconciler/leader.go
mattmoor: The leader election package doesn't provide us with any facility to abdicate, and technically the only way for the lister to return an error is for `meta.Accessor` to fail, which seems virtually impossible.

mattmoor: I'm open to signature changes, but I think the more general conversation of the workqueue's relationship to controller.Impl is probably an orthogonal conversation (I'll reserve further comment for that convo 😉 ).

I think my preference would be to land this and iterate rather than trying to explode everything here/now. 😬

reconciler/leader.go
mattmoor: done

reconciler/leader.go
mattmoor: done

leaderelection/context.go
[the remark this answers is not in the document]
dprotaso: >My understanding is that defer runtime.HandleCrash() is intended to keep panics from killing things.

It runs handlers and then rethrows  - I'm surprised it's in that Run method

https://github.com/kubernetes/apimachinery/blob/v0.17.6/pkg/util/runtime/runtime.go#L55

leaderelection/context_test.go
dprotaso: ahhh.. link the issue?

reconciler/leader.go
dprotaso: > I think my preference would be to land this and iterate rather than trying to explode everything here/now. 😬

Agreed - given this is an alpha API we have flexibilty

dprotaso: > The leader election package doesn't provide us with any facility to abdicate

Cancelling the context passed to `LeaderElector` allows you to release the lock.

reconciler/leader.go
dprotaso: fair point

leaderelection/context_test.go
mattmoor: I thought I did somewhere, but lemme grep and update if not.

reconciler/leader.go
mattmoor: sunk up on slack, this returns an error, but we still log fatal for now (w/ TODO)

controller/controller_test.go
```diff
@@ -820,13 +827,150 @@ func TestStartAndShutdown(t *testing.T) {
...
+		defer close(doneCh)
+		StartAll(ctx, impl)
+	}()
+
+	select {
+	case <-promoted:
+		// We expect to be promoted immediately, since there is no
+		// ElectorBuilder attached to the context.
+	case <-doneCh:
+		t.Error("StartAll finished early.")
```
dprotaso, line 887:
```suggestion
		t.Fatal("StartAll finished early.")
```

Should we also add a timeout case here so the test doesn't hang if someone breaks something?

controller/controller_test.go
```diff
@@ -820,13 +827,150 @@ func TestStartAndShutdown(t *testing.T) {
...
+
+	select {
+	case <-time.After(1 * time.Second):
+		t.Error("Timed out waiting for controller to finish.")
+	case <-doneCh:
+		// We expect the work to complete.
+	}
+
+	if got, want := r.count, 0; got != want {
+		t.Errorf("count = %v, wanted %v", got, want)
```
dprotaso, line 900: more descriptive error?

controller/controller_test.go
```diff
@@ -820,13 +827,150 @@ func TestStartAndShutdown(t *testing.T) {
...
+		LeaderAwareFuncs: reconciler.LeaderAwareFuncs{
+			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) error {
+				close(promoted)
+				return nil
+			},
+		},
+	}
+	impl := NewImplWithStats(r, TestLogger(t), "Testing", &FakeStatsReporter{})
+
+	ctx, cancel := context.WithCancel(context.Background())
```
dprotaso, line 874:
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)
```

I think we should call cancel as part of the cleanup in case tests error out/fail to stop any go routines

Should probably do this for all tests

controller/controller_test.go
dprotaso: likewise for other places that have a similar assertion

leaderelection/context_test.go
```diff
@@ -0,0 +1,145 @@
...
+	case <-demoted:
+		t.Error("Got demoted, want no actions.")
+	case <-created:
+		t.Error("Got created, want no actions.")
+	case <-updated:
+		t.Error("Got updated, want no actions.")
+	default:
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
dprotaso, line 114: ``
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	t.Cleanup(cancel)
```

leaderelection/context.go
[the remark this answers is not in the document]
dprotaso: circling back for others

1)  is inaccurate & @mattmoor is correct - so the tight loop is in the leader election code in client-go

2) because 1 was inaccurate this means 2 (`OnStoppedLeading blocks LeaderElector.Run`) isn't a 'big' problem

3) my intent here was to highlight the difference on how the callback `Promote` & `Demote` behave different (since one blocks Run and the other doesn't). It's not a huge issue because of my bad assumption in 1)

dprotaso:
wave

PR #1302, opened 2020-05-05
Implement the second wave of per-reconciler leaderelection

mattmoor:
Detailed design: https://docs.google.com/document/d/1i_QHjQO2T3SNv49xjZLWlivcc0UvZN1Tbw2NKxThkyM/edit

Issue: https://github.com/knative/pkg/issues/1181

WIP until #1301 lands

### DIFF
--- a/webhook/certificates/certificates.go
+++ b/webhook/certificates/certificates.go
@@ -23,10 +23,12 @@ import (
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
+	pkgreconciler "knative.dev/pkg/reconciler"
 	"knative.dev/pkg/system"
 	certresources "knative.dev/pkg/webhook/certificates/resources"
 )
@@ -37,6 +39,8 @@ const (
 )
 
 type reconciler struct {
+	pkgreconciler.LeaderAwareFuncs
+
 	client       kubernetes.Interface
 	secretlister corelisters.SecretLister
 	secretName   string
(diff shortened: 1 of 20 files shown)


webhook/certificates/certificates.go
```diff
@@ -37,17 +39,27 @@ const (
...
+var _ pkgreconciler.LeaderAware = (*reconciler)(nil)
 
 // Reconcile implements controller.Reconciler
 func (r *reconciler) Reconcile(ctx context.Context, key string) error {
-	return r.reconcileCertificate(ctx)
+	if r.IsLeaderFor(types.NamespacedName{
+		Namespace: system.Namespace(),
+		Name:      r.secretName,
+	}) {
+		// only reconciler the certificate when we are leader.
```
vagababov, line 59: looks like this is immutable? May be make it a member? Or perhaps it should use `key` instead?

webhook/configmaps/configmaps.go
```diff
@@ -58,13 +61,19 @@ type reconciler struct {
...
 var _ controller.Reconciler = (*reconciler)(nil)
+var _ pkgreconciler.LeaderAware = (*reconciler)(nil)
 var _ webhook.AdmissionController = (*reconciler)(nil)
 var _ webhook.StatelessAdmissionController = (*reconciler)(nil)
 
 // Reconcile implements controller.Reconciler
 func (ac *reconciler) Reconcile(ctx context.Context, key string) error {
 	logger := logging.FromContext(ctx)
 
+	if !ac.IsLeaderFor(types.NamespacedName{Name: ac.name}) {
```
vagababov, line 72: Namespace?

webhook/configmaps/configmaps.go
```diff
@@ -58,13 +61,19 @@ type reconciler struct {
...
+		logger.Debugf("Skipping key %q, not the leader.", key)
```
vagababov, line 73: Key is not part of the leader selection per code.

webhook/certificates/certificates.go
mattmoor: I moved `key` to be a member of all of the singleton reconcilers.

webhook/configmaps/configmaps.go
mattmoor: This is reconciling the validating webhook, so no namespace.

webhook/configmaps/configmaps.go
mattmoor: All of these should now print `ac.key`

webhook/certificates/controller.go
```diff
@@ -68,7 +70,7 @@ func NewController(
 
 	// Reconcile when the cert bundle changes.
 	secretInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
-		FilterFunc: controller.FilterWithNameAndNamespace(system.Namespace(), wh.secretName),
+		FilterFunc: controller.FilterWithNameAndNamespace(key.Namespace, key.Name),
```
dprotaso, line 73: lol the func name `FilterWithNameAndNamespace` doesn't match the args order

mattmoor: Yeah, and it'd be nice if it took a `types.NamespacedName` 🙄

webhook/configmaps/configmaps.go
vagababov: we're really abusing the type, whose comment says `mandatory namespace` 😁

PR #1307, opened 2020-05-06
Switch KRShaped interface to retrieve entire condition set

whaught:
Issue #1226
- Switch interface to retrieve the entire conditions set

As I looked at integrating this downstream I've realized this will be more flexible so that we can have generic awareness of all the conditions that exist for a resource. This will also allow us to ensure they are initialized.

### DIFF
--- a/apis/duck/v1/kresource_type.go
+++ b/apis/duck/v1/kresource_type.go
@@ -34,7 +34,7 @@ type KRShaped interface {
 
 	GetStatus() *Status
 
-	GetTopLevelConditionType() apis.ConditionType
+	GetConditionSet() apis.ConditionSet
 }
 
 // Asserts KResource conformance with KRShaped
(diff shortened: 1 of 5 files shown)


apis/duck/v1/kresource_type.go
```diff
@@ -34,7 +34,7 @@ type KRShaped interface {
 
 	GetStatus() *Status
 
-	GetTopLevelConditionType() apis.ConditionType
```
n3wscott, line 37: Here is a hitch in this plan: for things like alternate broker implementations, the reconciler for the broker gets to redefine what the ConditionSet for the resource is, and it ignores the type's conditions set, it also avoids all the `Mark*` methods. So it might not be true that _this_ condition set is the one in play for the resource under reconciliation.

n3wscott: cc: @grantr

whaught: We we could either say alternate brokers are not KRShaped and need to manage things on their own or if they share the same top-level condition, define a set with only that.

grantr: For the alternate broker in https://github.com/google/knative-gcp we have a separate API type that defines its own ConditionSet specific to the alternate implementation. There's a separate generated reconciler for that type that the alternate broker controller uses to reconcile. I'm not sure if this is the right pattern long-term.

grantr: @n3wscott do you have an example of a single type that works with multiple ConditionSets? I don't know of any.

mattmoor: Assuming "Top Level" means Ready/Succeeded, then generally conformance dictates that this should not vary.  I think that the rest can vary at all levels of severity, assuming our semantics are adhered to.

n3wscott: OK, just note we have an issue if we ever use any other method on the ConditionSet other than the toplevel method.

PR #1308, opened 2020-05-06
Ensure that all needed imports are tracked in reconciler generator

ian-mi:
Currently it is possible for the generator to omit these imports, although this behavior appears to depend on the generator dependency versions.

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -145,6 +145,10 @@ func (g *reconcilerReconcilerGenerator) GenerateType(c *generator.Context, t *ty
 			Package: "context",
 			Name:    "Context",
 		}),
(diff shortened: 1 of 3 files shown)


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -321,7 +325,7 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 	}
 
 	// Synchronize the status.
-	if equality.Semantic.DeepEqual(original.Status, resource.Status) {
+	if {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status) {
```
n3wscott, line 328: nice catch... yeah I thought I had got them all but there are always some more 🙄

n3wscott:
/lgtm
/approve

Thanks for the cleanup!!

ian-mi:
Thanks for reviewing! Reassigning to Matt for approval.
/assign @mattmoor

PR #1309, opened 2020-05-06
Add custom object encoders for the set of string and namespaced name

vagababov:
In future we might add more (endpoints is a really good candidate)

/assign @yanweiguo mattmoor

### DIFF
--- a/logging/object_encoders.go
+++ b/logging/object_encoders.go
@@ -44,9 +44,9 @@ func StringSet(s sets.String) zapcore.ObjectMarshalerFunc {
 
 // NamespacedName returns a marshaler for NamespacedName.
 // To use this in sugared logger do:
-//	logger.Infow("Enqueueing", zap.Object("key", logging.NamespacedName(n)))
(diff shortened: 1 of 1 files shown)


logging/object_encoders.go
```diff
@@ -0,0 +1,45 @@
...
+	return func(enc zapcore.ObjectEncoder) error {
+		enc.AddString("keys", strings.Join(s.UnsortedList(), ","))
+		return nil
+	}
+}
+
+// NamespacedName returns a marshaler for NamespacedName.
+func NamespacedName(n types.NamespacedName) zapcore.ObjectMarshalerFunc {
+	return func(enc zapcore.ObjectEncoder) error {
+		enc.AddString("key", n.Namespace+"/"+n.Name)
```
mattmoor, line 42: We use this to hold cluster scoped resources with no Namespace, so perhaps handle that?

vagababov: This is done.

PR #1311, opened 2020-05-08
Generate common logic for krshaped resources

whaught:
Issue #1226

The plan here is to opt-in downstream repos by adding the `krshapedlogic=true` tag to genreconciler. Later we can invert this and make it default (which will nudge future implementers to implement krshaped for all resources and use common logic) and opt out with 'krshapedlogic=false'

Alternatively we could instead code-generate the following with a runtime check which would compile for non-krshaped resources:

```
krshaped, isKRShaped := interface{}(resource).(duckv1.KRShaped)
if (isKRShaped) {
    reconciler.PreProcessReconcile(ctx, resource)
}
```

However I slightly prefer compile-time checks. Thoughts from reviewers appreciated.

### DIFF
--- a/injection/README.md
+++ b/injection/README.md
@@ -460,7 +460,7 @@ Reconcilers can handle common logic for resources that conform to the KRShaped i
 the generated code to automatically increment ObservedGeneration.
 
 ```go
-// +genreconciler:krshaped=true
+// +genreconciler:krshapedlogic=true
 ```
 
 Setting this annotation will emit the following in the generated reconciler.
--- a/injection/README.md
+++ b/injection/README.md
@@ -456,8 +456,9 @@ for that resource.
 
 **krshaped=true may become the default if omitted in the future**
 
-Reconcilers can handle common logic for resources that conform to the KRShaped interface. This allows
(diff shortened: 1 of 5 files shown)


injection/README.md
```diff
@@ -452,6 +452,27 @@ value of the `<key>` annotation on a resource must match the value provided to
 `NewImpl` (or `NewReconcile`) for `ReconcileKind` or `FinalizeKind` to be called
 for that resource.
 
+#### Annotation based common logic
+
+**krshaped=true may become the default if omitted in the future**
+
+Reconcilers can handle common logic for resources that conform to the KRShaped interface. This allows
+the generated code to automatically increment ObservedGeneration.
```
mattmoor, line 460: Format markdown:
```suggestion
Reconcilers can handle common logic for resources that conform to the KRShaped
interface. This allows the generated code to automatically increment
ObservedGeneration.
```

mattmoor, line 460: Format markdown:
```suggestion
Reconcilers can handle common logic for resources that conform to the KRShaped
interface. This allows the generated code to automatically increment
ObservedGeneration.
```

PR #1317, opened 2020-05-09
add Status.Annotations field

vaikas:
https://docs.google.com/document/d/1LFE4IwQ7yj6zjO7qwMSXDRfPk8ll9hSk6uCh2Xx--AE/edit#

### DIFF
--- a/apis/duck/v1/status_types.go
+++ b/apis/duck/v1/status_types.go
@@ -45,6 +45,12 @@ type Status struct {
 	// +patchMergeKey=type
 	// +patchStrategy=merge
 	Conditions Conditions `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// Annotations is additional Status fields for the Resource to save some
+	// additional State as well as convey more information to the user. This is
+	// roughly akin to Annotations on any k8s resource, just the reconciler conveying
+	// richer information outwards.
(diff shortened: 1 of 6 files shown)


duglin:
No strong opinion on the name, but I like the idea of controllers, and 3rd party extensions, having a spot to store data in the resource that won't potentially overlap with what some people think of as "user data" (meaning spec & meta).  

And since Ksvc and Revison both use `duckv1.Status` then it means we get to use it in serving too - so that's cool.

evankanderson:
Do we want to enforce shape requirements on the conditions similar to k8s conditions? (annotation-like names, arbitrary string values)

vaikas:
@evankanderson I'd rather not enforce it but provide guidance? We could say that folks embedding, should use something like:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L110
to validate it? But the onus is really on the embedder (or I'm missing something) about how to enforce it at this level.

evankanderson:
If we're providing a default implementation, could our implemention be `Validatable`?

It wouldn't enforce it for all objects that match the duck, but it might provide sensible default behavior for most of Knative.

vaikas:
I don't see where we'd be able to enforce this without the actual caller would do something, since this is not user controlled field and is only used by the controller, not sure what you have in mind for default implementation?

vagababov:
ugh, needed a hold here 🙄

PR #1316, opened 2020-05-08
[WIP] Add a new "Goose" e2e test tool for leaderelection.

mattmoor:
This adds a new main package under knative.dev/pkg/leaderelection/goose, which
loads the leader election configuration, and periodically enumerates the leases
in the system namespace and kills a leader pod for each sharded reconciler
under a component.

Based on https://github.com/knative/pkg/pull/1304

PR #1318, opened 2020-05-09
Implement the K8s lifecycle in webhook.

mattmoor:
The webhook never properly implemented the Kubernetes SIGTERM/SIGKILL lifecycle, and doesn't even really support readiness probes today.  This change enables folks to use a block like this on their webhook container:

```yaml
        readinessProbe: &probe
          periodSeconds: 1
          httpGet:
            scheme: HTTPS
            port: 8443
            httpHeaders:
            - name: k-kubelet-probe
              value: "webhook"
        livenessProbe: *probe
```

With this, the webhook won't report as `Ready` until a probe has succeeded, and when the SIGTERM is received, we will start failing probes for a grace period (so our Endpoint drops) before shutting down the webhook's HTTP Server.

This was uncovered by running the webhook across 10 replicas in Serving with the "Goose" (https://github.com/knative/pkg/pull/1316) enabled for the e2e tests.  The failure mode I saw was conversion webhook requests failing across random tests.

This also moves the Serving probe-detection function into PKG.

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -17,6 +17,8 @@ limitations under the License.
 package network
 
 import (
+	"net/http"
+	"strings"
 	"time"
 )
 
@@ -42,4 +44,18 @@ const (
 	// with this header will not be passed to the user container or
 	// included in request metrics.
 	ProbeHeaderName = "K-Network-Probe"
+
+	// Since K8s 1.8, prober requests have
+	//   User-Agent = "kube-probe/{major-version}.{minor-version}".
+	KubeProbeUAPrefix = "kube-probe/"
+
+	// Istio with mTLS rewrites probes, but their probes pass a different
+	// user-agent.  So we augment the probes with this header.
+	KubeletProbeHeaderName = "K-Kubelet-Probe"
 )
+
+// IsKubeletProbe returns true if the request is a Kubernetes probe.
+func IsKubeletProbe(r *http.Request) bool {
+	return strings.HasPrefix(r.Header.Get("User-Agent"), KubeProbeUAPrefix) ||
+		r.Header.Get(KubeletProbeHeaderName) != ""
+}
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -222,7 +222,7 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
 	select {
 	case <-stop:
 		// Start failing readiness probes immediately.
-		logger.Debug("Starting to fail readiness probes...")
+		logger.Info("Starting to fail readiness probes...")
 		close(wh.stopCh)
 
 		// Wait for a grace period for the above to take effect and this Pod's
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	// Injection stuff
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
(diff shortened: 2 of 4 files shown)


network/network.go
```diff
@@ -42,4 +44,18 @@ const (
 	// with this header will not be passed to the user container or
 	// included in request metrics.
 	ProbeHeaderName = "K-Network-Probe"
+
+	// Since K8s 1.8, prober requests have
```
vagababov, line 48: Do we need the version comment? :) 1.8 is so far in the past

webhook/webhook.go
```diff
@@ -71,6 +73,13 @@ const (
 	Connect Operation = admissionv1beta1.Connect
 )
 
+var (
+	// GracePeriod is the duration that the webhook will wait after it's
+	// context is cancelled (and probes are failing) before shutting down
+	// the http server.
+	GracePeriod = 30 * time.Second
```
vagababov, line 80: Given this is not in data path and it's basically API server calling, should 5-6s suffice?
Also if we inject it, it can be a const :)

webhook/webhook.go
```diff
@@ -206,16 +221,38 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
 
 	select {
 	case <-stop:
-		if err := server.Close(); err != nil {
+		// Start failing readiness probes immediately.
+		logger.Debug("Starting to fail readiness probes...")
```
vagababov, line 225:
```suggestion
		logger.Info("Starting to fail readiness probes...")
```
This is once in a lifetime event, can as well log it, might help with debugging.

webhook/webhook.go
```diff
@@ -206,16 +221,38 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
...
+		close(wh.stopCh)
+
+		// Wait for a grace period for the above to take effect and this Pod's
+		// endpoint to be removed from the webhook service's Endpoints.
+		// For this to be effective, it must be greater than the probe's
+		// periodSeconds times failureThreshold by a margin suitable to
+		// propagate the new Endpoints data across the cluster.
+		time.Sleep(wh.gracePeriod)
+
+		if err := server.Shutdown(context.Background()); err != nil {
```
vagababov, line 235: I think this is (was) incorrect. We probably need to `eg.Wait()` in any case.

network/network.go
mattmoor: This is copy/pasta from Serving.  I don't think it's harming anyone to know the history 🤓

webhook/webhook.go
mattmoor: I made it a var so that folks can customize it from their `main` method as needed, as we do for tests.

When I first added this I thought the tests were hanging 😬

webhook/webhook.go
mattmoor: We `eg.Wait()` below (hasn't changed), which seems right because the go routine we launched won't return before Shutdown (or Close) is called.

webhook/webhook.go
mattmoor: done

webhook/webhook.go
mattmoor: Clarified on chat that this is about the error path.  I tweaked the code so that the shutdown is done via `eg.Go` so that the trailing `eg.Wait` happens on all paths and we wait for all outstanding go routines to terminate.

network/network.go
vagababov: It probably made sense when there were two constants, but now there's just one :)
🤷

webhook/webhook.go
```diff
@@ -206,16 +221,40 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
...
+			// endpoint to be removed from the webhook service's Endpoints.
+			// For this to be effective, it must be greater than the probe's
+			// periodSeconds times failureThreshold by a margin suitable to
+			// propagate the new Endpoints data across the cluster.
+			time.Sleep(wh.gracePeriod)
+
+			return server.Shutdown(context.Background())
+		})
+
+		// Wait for all outstanding go routined to terminate, including our new one.
```
vagababov, line 239: Sneaky :-)

PR #1323, opened 2020-05-11
Don't record data points when metric config is not initialized yet

yanweiguo:
A follow up PR of #1263 to a race condition that recording the data points before the metric config is initialized, at which point whether to record is unknown.

This is a breaking change for unit tests of downstream repos. Changes necessary have been made to knative/serving: https://github.com/knative/serving/pull/7911. Verified that unit tests won't break with this PR.


metrics/config.go
```diff
@@ -153,7 +154,20 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 // record applies the `ros` Options to each measurement in `mss` and then records the resulting
 // measurements in the metricsConfig's designated backend.
 func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
-	if mc == nil || mc.recorder == nil {
+	if mc == nil {
+		log.Println(`The metricsConfig has not been initialized yet.
```
mdemirhan, line 158: Logging this seems excessive. Shouldn't this be a comment instead? Also, should we not return an error here instead if the metrics config is not initialized?

yanweiguo: We are not returning an error here. This only happens when some code triggers `record` while the metrics config hasn't been initialized via `UpdateExporterFromConfig` in another go runtime (race condition).

The log here is to throw error message for downstream repos. Otherwise it will be hard to debug when someone, who is not familiar with this lib, working on adding new metrics and correspond unit tests.

PR #1327, opened 2020-05-12
Use our new tag parser to find genreconciler

whaught:
Previously we used the k8s tag parser which splits on '=' to look for
"+genreconciler" or "+genreconciler:class"

Now we have a parser that splits on ':' and expects a comma separated list so we can simplify the search.

The old parser leads to a bug if we try to do things in this order without more hardcoding:
"+genreconciler:krshaped=true,class=foo"


whaught:
I've already been messing with this change downstream to confirm. It works!
/unhold

PR #1329, opened 2020-05-13
Add a library to parse config map values.

markusthoemmes:
We keep copy and pasting code that handles this, so why not make a reusable library for all our projects? :slightly_smiling_face: 

Proof it works: https://github.com/knative/serving/pull/7935

### DIFF
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


configmap/parse.go
```diff
@@ -0,0 +1,109 @@
...
+*/
+
+package configmap
+
+import (
+	"strconv"
+	"strings"
+	"time"
+)
+
```
vagababov, line 24: I'd give `func(map[string]string)` a name `ParseFunc` sounds like fun :)

markusthoemmes: Done.

PR #1332, opened 2020-05-13
Library for writing interactive "shell script" in Go

coryrc:
Advance knative/test-infra#2063

Already used elsewhere

/cc @chaodaiG 
/cc @chizhg

### DIFF
--- a/test/interactive/command.go
+++ b/test/interactive/command.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Helper functions for running interactive CLI sessions from Go
+package interactive
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// could be public so extensions of this can test themselves
+var defaultRun func(*exec.Cmd) error
+
+func init() {
+	defaultRun = func(c *exec.Cmd) error { return c.Run() }
+}
+
+// Command represents running an executable, to make it easy to write interactive "shell scripts" in Go.
+// This routes standard in, out, and error from and to the calling program, intended usually to be a login shell run by a person.
+type Command struct {
+	Name string
+	Args []string
+	// If LogFile is set, anything send to stdout and stderr is tee'd to the file named in LogFile when .Run() is called.
+	LogFile string
+	run     func(*exec.Cmd) error
+}
+
+// NewCommand creates an Command, breaking out .Name and .Args for you
+func NewCommand(cmdAndArgs ...string) Command {
+	if len(cmdAndArgs) < 1 {
+		log.Fatal("NewCommand must have a least the command given")
+	}
+	return Command{cmdAndArgs[0], cmdAndArgs[1:], "", defaultRun}
+}
+
+// Run executes the command with exec.Command(), sets standard in/out/err, and returns the result of exec.Cmd.Run()
+// Will tee stdout&err to .LogFile if set
+func (c Command) Run() error {
+	cmd := exec.Command(c.Name, c.Args...)
+	cmd.Stdin = os.Stdin
+	if len(c.LogFile) > 0 {
+		f, err := os.OpenFile(c.LogFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+		if err != nil {
+			log.Fatal(err)
+		}
+		cmd.Stdout = io.MultiWriter(os.Stdout, f)
+		cmd.Stderr = io.MultiWriter(os.Stderr, f)
+	} else {
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+	}
+	return c.run(cmd)
+}
+
+// String conforms to Stringer
+func (c Command) String() string {
+	s := c.Name + " " + strings.Join(c.Args, " ")
+	return fmt.Sprintf("Command to run: %s", s)
+}
+
+// AddArgs appends arguments to the current list of args
+func (c *Command) AddArgs(args ...string) {
+	for _, a := range args {
+		c.Args = append(c.Args, a)
+	}
+}
--- a/test/interactive/command.go
+++ b/test/interactive/command.go
@@ -18,7 +18,6 @@ limitations under the License.
 package interactive
 
 import (
-	"fmt"
 	"io"
 	"log"
 	"os"
@@ -73,7 +72,7 @@ func (c Command) Run() error {
 // String conforms to Stringer
 func (c Command) String() string {
 	s := c.Name + " " + strings.Join(c.Args, " ")
-	return fmt.Sprintf("Command to run: %s", s)
+	return "Command to run: " + s
 }
 
 // AddArgs appends arguments to the current list of args
--- a/test/interactive/command.go
+++ b/test/interactive/command.go
@@ -78,7 +78,5 @@ func (c Command) String() string {
 
 // AddArgs appends arguments to the current list of args
 func (c *Command) AddArgs(args ...string) {
-	for _, a := range args {
-		c.Args = append(c.Args, a)
-	}
+	c.Args = append(c.Args, args...)
 }
--- a/test/interactive/docker.go
+++ b/test/interactive/docker.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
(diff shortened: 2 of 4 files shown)


test/interactive/command.go
```diff
@@ -0,0 +1,84 @@
...
+// could be public so extensions of this can test themselves
+var defaultRun func(*exec.Cmd) error
+
+func init() {
+	defaultRun = func(c *exec.Cmd) error { return c.Run() }
+}
+
+// Command represents running an executable, to make it easy to write interactive "shell scripts" in Go.
+// This routes standard in, out, and error from and to the calling program, intended usually to be a login shell run by a person.
+type Command struct {
```
chizhg, line 38: I feel the functionality is a bit overlapping with https://github.com/knative/pkg/blob/master/test/cmd/command.go, any chance we can move it there?

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+	for k, v := range e {
+		d.AddArgs("-e", fmt.Sprintf("%s=%s", k, v))
+	}
+}
+
+// AddMount add arguments for the --mount command
+func (d *Docker) AddMount(typeStr, source, target string, optAdditionalArgs ...string) {
+	addl := ""
+	for _, arg := range optAdditionalArgs {
+		addl = addl + "," + arg
```
chizhg, line 73: This will be `addl := strings.Join(arg, ",")`

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+	overlayDir := path.Join(tmpDir, "overlay")
+	// The options for overlay mount are confusing
+	// You need empty directories for upper and work (and overlay, though you can mount over directories that have files in them if you *want* to...)
+	mount := NewCommand("sudo", "mount", "-t", "overlay", "-o",
+		fmt.Sprintf("lowerdir=%s,upperdir=%s/upper,workdir=%s/work", externalDirectory, tmpDir, tmpDir),
+		"none", overlayDir)
+	// Print command to run so user knows why it is asking for sudo password (if it does)
+	fmt.Println(mount)
```
chizhg, line 102: Use `log` and be a little bit verbose?

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+	if err = mount.Run(); err != nil {
+		log.Fatalf("Unable to create overlay mount, so giving up: %v", err)
+	}
+	d.AddMount("bind", overlayDir, internalDirectory)
+	return func() {
+		// Print command to run so user knows why it is asking for sudo password (if it does)
+		umount := NewCommand("sudo", "umount", overlayDir)
+		fmt.Println(umount)
```
chizhg, line 110: Ditto?

test/interactive/docker_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"os/exec"
+	"testing"
+)
+
+func TestEnv(t *testing.T) {
+	e := make(Env)
+	err := e.PromoteFromEnv("PWD")
+	if err != nil {
+		t.Error(err)
+		fmt.Printf("The Environ is:\n%+v\n", os.Environ())
```
chizhg, line 31: `t.Logf`

test/interactive/docker_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	if e["PWD"] != os.Getenv("PWD") {
+		t.Errorf("Promotion did not occur correctly: %v;%v", e["PWD"], os.Getenv("PWD"))
```
chizhg, line 34:
```suggestion
		t.Errorf("Promotion for $PWD did not occur correctly: %s != %s", e["PWD"], os.Getenv("PWD"))
```

test/interactive/docker_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	badName := "GEEZ_I_REALLY_H0PE_TH1S_DOES_NOT_EXIST"
```
chizhg, line 36: 👍

test/interactive/command.go
```diff
@@ -0,0 +1,84 @@
...
+	Name string
+	Args []string
+	// If LogFile is set, anything send to stdout and stderr is tee'd to the file named in LogFile when .Run() is called.
+	LogFile string
```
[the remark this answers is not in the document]
coryrc: if `c := NewCommand()`, then `c.LogFile = "..."` is the way I use it

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// PromoteFromEnv pulls the named environment variables from the environment and puts them in the Env.
+// It does not stop on error and returns an error listing all the failed values
+func (e Env) PromoteFromEnv(envVars ...string) error {
+	var err error
+	for _, env := range envVars {
+		v := os.Getenv(env)
+		if v == "" {
+			err = fmt.Errorf("Environment Variables '%s' is not set; %v", env, err)
```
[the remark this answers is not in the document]
coryrc: Huh, will change. I like it including all.

test/interactive/command.go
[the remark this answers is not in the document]
coryrc: but could use a Setter instead if you want

test/interactive/command.go
```diff
@@ -0,0 +1,84 @@
...
+
+// String conforms to Stringer
+func (c Command) String() string {
+	s := c.Name + " " + strings.Join(c.Args, " ")
+	return fmt.Sprintf("Command to run: %s", s)
+}
+
+// AddArgs appends arguments to the current list of args
+func (c *Command) AddArgs(args ...string) {
+	for _, a := range args {
```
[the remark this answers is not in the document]
coryrc: d'oh, I always forget that

test/interactive/command.go
coryrc: That library explicitly runs commands in a different environment than the caller's environment, while this library is intended for commands running in the same environment (like a shell script). Given it's just a tiny wrapper around exec.Command(), there's nothing else to combine.

test/interactive/docker.go
coryrc: the leading `,` is necessary with arguments, which rules out your first example, or necessary that the string is empty when no arguments, ruling out the second example unless we add `if len(optAdditionaArgs) > 0` first. Your call

test/interactive/docker_test.go
coryrc: actually, e["PWD"] is not guaranteed either, so will make a couple changes

test/interactive/docker_test.go
coryrc: oops

test/interactive/docker.go
coryrc: I use log elsewhere so should be consistent.

test/interactive/command.go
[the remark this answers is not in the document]
chizhg: I would prefer two functions - `NewCommand` and `NewCommandWithLogFile`, but up to you.

test/interactive/docker.go
chizhg: I think below is more understandable.
```
if len(optAdditionaArgs) != 0 {
    addl = "," + strings.Join(optAdditionalArgs, ",")
}
```

test/interactive/command.go
chizhg: SG

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+func init() {
+	defaultDockerCommands = []string{"docker", "run", "-it", "--rm", "--entrypoint", "bash"}
+}
+
+// Env represents a collection of environment variables and their values
+type Env map[string]string
+
+// Docker is mostly an Command preloaded with arguments which setup Docker for running an image interactively.
+type Docker struct {
+	Command
```
chizhg, line 39: Nit: should `Command` be private so change to `command`?

coryrc: then .Run and .AddArgs aren't available

test/interactive/command.go
```diff
@@ -0,0 +1,82 @@
...
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+	}
+	return c.run(cmd)
+}
+
+// String conforms to Stringer
+func (c Command) String() string {
+	s := c.Name + " " + strings.Join(c.Args, " ")
+	return fmt.Sprintf("Command to run: %s", s)
```
vagababov, line 76:
```suggestion
	"Command to run: " + s
```

test/interactive/docker.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// PromoteFromEnv pulls the named environment variables from the environment and puts them in the Env.
+// It does not stop on error and returns an error listing all the failed values
+func (e Env) PromoteFromEnv(envVars ...string) error {
+	var err error
+	for _, env := range envVars {
+		v := os.Getenv(env)
+		if v == "" {
+			err = fmt.Errorf("environment variable '%s' is not set; %v", env, err)
```
vagababov, line 49:
```suggestion
			err = fmt.Errorf("environment variable %q is not set; %w", env, err)
```

test/interactive/command.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Command represents running an executable, to make it easy to write interactive "shell scripts" in Go.
+// This routes standard in, out, and error from and to the calling program, intended usually to be a login shell run by a person.
+type Command struct {
+	Name string
+	Args []string
+	// If LogFile is set, anything send to stdout and stderr is tee'd to the file named in LogFile when .Run() is called.
+	LogFile string
+	run     func(*exec.Cmd) error
```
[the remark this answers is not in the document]
coryrc: If we were to have a complicated function like AddRWOverlay() with nested calls to NewCommand() that we could actually test in a useful way, defaultRun allows you to test those nested calls without actually running commands.

test/interactive/docker.go
coryrc: I don't wish to support unwrap so I'll just take the first

test/interactive/command.go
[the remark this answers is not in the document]
coryrc: The second leads to having lots of combinations; I could do the "Option" like the runcommands library

test/interactive/command.go
coryrc: done

PR #1335, opened 2020-05-15
Use config map parsing lib and enhance it further.

markusthoemmes:
- Use pkg/configmap parser for observability and leaderelection config. Logging config ain't worth it and I didn't yet dare to touch metric config.
- Adds a new helper for parsing StringSets. That one is useful in multiple places.
- Enhance the error reporting to always include the parsed key to be useful in error messages.

### DIFF
--- a/configmap/parse_test.go
+++ b/configmap/parse_test.go
@@ -19,15 +19,19 @@ package configmap
 import (
 	"testing"
 	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 type testConfig struct {
-	str string
-	boo bool
-	i32 int32
-	i64 int64
-	f64 float64
-	dur time.Duration
+	Str string
+	Boo bool
+	I32 int32
+	I64 int64
+	F64 float64
+	Dur time.Duration
+	Set sets.String
 }
 
 func TestParse(t *testing.T) {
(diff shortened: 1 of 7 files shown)


markusthoemmes:

Git checkout.

configmap/parse_test.go
```diff
@@ -19,15 +19,19 @@ package configmap
...
 )
 
 type testConfig struct {
-	str string
-	boo bool
-	i32 int32
-	i64 int64
-	f64 float64
-	dur time.Duration
+	Str string
```
yanweiguo, line 28: Why do these fields need to be exported?

markusthoemmes: to make `cmp.Equal` work.

vagababov: https://godoc.org/github.com/google/go-cmp/cmp#AllowUnexported

markusthoemmes: Oh neat, I thought there was only an option to ignore unexported!

markusthoemmes: fixed!

tracing/config/tracing.go
```diff
@@ -98,39 +99,29 @@ func NewTracingConfigFromMap(cfgMap map[string]string) (*Config, error) {
 		}
 	}
 
-	if endpoint, ok := cfgMap[zipkinEndpointKey]; !ok && tc.Backend == Zipkin {
+	if err := cm.Parse(cfgMap,
```
slinkydeveloper, line 102: Why don't you parse `backendKey` and `enable` here in a local variable and then you move all the validation algorithm below?

markusthoemmes: Lemme see how that feels like.

markusthoemmes: Ah: It's special cased logic that's in an either/or fashion. I can't properly express that in the logic here so I figured I'd just keep it as-is.

tracing/config/tracing.go
```diff
@@ -98,39 +99,29 @@ func NewTracingConfigFromMap(cfgMap map[string]string) (*Config, error) {
...
-		sampleRateFloat, err := strconv.ParseFloat(sampleRate, 64)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse sampleRate in tracing config: %w", err)
-		}
-		if sampleRateFloat < 0 || sampleRateFloat > 1 {
-			return nil, fmt.Errorf("sampleRate = %v must be in [0, 1] range", sampleRateFloat)
-		}
-		tc.SampleRate = sampleRateFloat
+	if tc.SampleRate < 0 || tc.SampleRate > 1 {
+		return nil, fmt.Errorf("sampleRate = %v must be in [0, 1] range", tc.SampleRate)
```
vagababov, line 124:
```suggestion
		return nil, fmt.Errorf("sample-rate = %v must be in [0, 1] range", tc.SampleRate)
```

markusthoemmes: Fixed.

PR #1337, opened 2020-05-15
Clean context.TODO in pkg and other minor nits

vagababov:
- TODO is for migration, we know we want background there
- other nits

/assign mattmoor


vagababov:
L33T! :)

PR #1342, opened 2020-05-16
reconcile owners

vagababov:
/assign mattmoor

🍻  for Dan 😢


vagababov:
yeah, perhaps we can just put toc+leads in communal pkg ownership

PR #1343, opened 2020-05-19
add nil check to postprocess

whaught:
We are expecting that pre-process will have added a top-level condition, but if for some reason the reconciler clears conditions we want to  log a warning. As-is this might panic.


reconciler/reconcile_common.go
```diff
@@ -49,7 +49,9 @@ func PostProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 	newStatus.ObservedGeneration = resource.GetGeneration()
 
 	rc := mgr.GetTopLevelCondition()
-	if rc.Reason == failedGenerationBump {
+	if rc == nil {
```
vagababov, line 52: Fold the assignment in?
```suggestion
	if rc := mgr.GetTopLevelCondition(); rc == nil {
```

PR #1344, opened 2020-05-19
Do not hardcode knative in test/prow library

chizhg:
Do not hardcode knative in test/prow library, this is part of the fixes for https://github.com/knative/test-infra/issues/2092

/cc @chaodaiG

### DIFF
--- a/test/prow/prow.go
+++ b/test/prow/prow.go
@@ -34,9 +34,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


chizhg:
> Is there a callsite change somewhere?

This function is only referenced by test-infra, and it's used to fix an issue in test-infra, so it won't break things.

PR #1345, opened 2020-05-19
Retry EOF errors by the spoofing client

mgencur:
This is for connection/network errors. For example, when a gateway is temporarily down, like in HA tests. In that case it can throw an error like this: 
```
The service at http://kourier-gateway-h-a-hhqgngvs.example.com didn't return success: response: <nil> did not pass checks: Get "http://kourier-gateway-h-a-hhqgngvs.example.com": EOF
```

### DIFF
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -51,3 +51,7 @@ func isConnectionRefused(err error) bool {
 func isConnectionReset(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "connection reset by peer")
 }
+
+func isEOF(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "EOF")
+}
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -51,7 +51,3 @@ func isConnectionRefused(err error) bool {
 func isConnectionReset(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "connection reset by peer")
 }
-
-func isEOF(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "EOF")
-}
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -246,22 +246,22 @@ func (sc *SpoofingClient) Poll(req *http.Request, inState ResponseChecker, error
 // DefaultErrorRetryChecker implements the defaults for retrying on error.
 func DefaultErrorRetryChecker(err error) (bool, error) {
 	if isTCPTimeout(err) {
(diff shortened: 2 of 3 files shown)


test/spoof/error_checks.go
```diff
@@ -51,3 +51,7 @@ func isConnectionRefused(err error) bool {
 func isConnectionReset(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "connection reset by peer")
 }
+
+func isEOF(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "EOF")
```
vagababov, line 56: can we use: https://golang.org/pkg/io/#pkg-variables EOF from here and just equality compare/

mgencur: Sounds good.

test/spoof/error_checks.go
```diff
@@ -19,6 +19,8 @@ limitations under the License.
 package spoof
 
 import (
+	"errors"
+	"io"
```
mattmoor, line 23: Format Go code:
```suggestion

```

test/spoof/error_checks.go
mgencur: Pushed. Actually had to use a little different approach because of this error that I was getting on `cmp.Equal()`: 
```
panic: cannot handle unexported field at {*errors.errorString}.s:
	"errors".errorString
```
But anyway, I think this approach is even better. I also removed the test because it probably doesn't make sense anymore. We're using a built-in function now.

test/spoof/spoof.go
```diff
@@ -257,6 +259,10 @@ func DefaultErrorRetryChecker(err error) (bool, error) {
 	if isConnectionReset(err) {
 		return true, fmt.Errorf("Retrying for connection reset: %w", err)
 	}
+	// Retry on connection/network errors.
+	if errors.Is(err, io.EOF) {
+		return true, fmt.Errorf("Retrying for: %w", err)
```
vagababov, line 264:
```suggestion
		return true, fmt.Errorf("retrying for: %w", err)
```

mgencur: The other errors in this function start with a capital. Do you want to change them all? Otherwise I don't think it makes sense to have this single one different.

vagababov: all should be :) we can clean up separately, I can do that, if you don't feel like it. :)

mgencur: Alright. Done in a new commit.

mgencur:
/assign @chizhg

Please, review.

mgencur:
@vagababov @chizhg thanks for reviews!

PR #1348, opened 2020-05-22
Add the key of the object to the log context

vagababov:
We don't log _what_ we convert, but only _what type_ it is.
And it's not very useful
So log all the stuff

/assign @dprotaso @mattmoor

### DIFF
--- a/webhook/resourcesemantics/conversion/conversion.go
+++ b/webhook/resourcesemantics/conversion/conversion.go
@@ -27,10 +27,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
 	"knative.dev/pkg/logging"
+	"knative.dev/pkg/logging/logkey"
 )
 
 // Convert implements webhook.ConversionController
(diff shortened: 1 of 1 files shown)


webhook/resourcesemantics/conversion/conversion.go
```diff
@@ -109,17 +110,26 @@ func (r *reconciler) convert(
...
 		zap.String("hubType", formatGVK(hubGVK)),
-	))
+	)
 
 	// TODO(dprotaso) - potentially error on unknown fields
 	if err = json.Unmarshal(inRaw.Raw, &in); err != nil {
 		return ret, fmt.Errorf("unable to unmarshal input: %w", err)
 	}
 
+	if acc, err := kmeta.DeletionHandlingAccessor(in); err == nil {
```
dprotaso, line 125: I wonder if the convertible interface should implement 
`metav1.Object` which would simplify all of this

but that's another issue

vagababov: yeah, that'd help stuff.
But I guess different PR, yes :)

webhook/resourcesemantics/conversion/conversion.go
```diff
@@ -109,17 +110,26 @@ func (r *reconciler) convert(
...
+		logger = logger.With("key", acc.GetNamespace()+"/"+acc.GetName())
```
dprotaso, line 126: note our controllers have a private `safeKey` method in case they're operating against clustered scope objects

dprotaso, line 126: do we want to use `logkey.Key` ?

webhook/resourcesemantics/conversion/conversion.go
dprotaso: sure

webhook/resourcesemantics/conversion/conversion.go
vagababov: I don't think knative resources (which are only ones here) can be cluster scoped?

vagababov: Done.

vagababov: Went the fancy way :)

vagababov:
PTAL

webhook/resourcesemantics/conversion/conversion.go
vagababov: Also I can ignore the error, since I guess anything that can be in the webhook version conversions has to be a k8s object :)

PR #1349, opened 2020-05-25
Make onAfterStore synchronous to the update itself.

markusthoemmes:
Generally, the `onAfterStore` function is used to kick off a global resync, which should be a very quick operation. Having `onAfterStore` being called asynchronously on a goroutine makes it very hard or straight out impossible to control it, especially in tests.

If there's a costly operation being done `onAfterStore`, it should be up to the caller to decide to make that asynchronous and thus put it into a goroutine.

This contributes to a lot of test flakes, at least in Knative Serving.

/assign @vagababov


vagababov:
Since it's not gonna make to release, we have lots of time to see if it is not a good idea (I.e. we're timing out because of this).
So some "It took %v to update all CM watchers" would be nice.

PR #1353, opened 2020-05-26
Allow to specify service name and endpoint IP respectively for zipkin tracing

nak3:
This patch allows to specify service name and hostPort for zipkin tracing.

Currently the serviceName(display in the console) and
host(endpoint) must be same name.

This patch allows to specify both serviceName and host. 

Please also refer to https://github.com/knative/serving/pull/8069.

### DIFF
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -119,10 +119,10 @@ func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
 }
 
 // WithExporterFull supports hostPort argument, which is used for the endpoint.
-// The hostPort allows HOST:PORT or HOST. If no port is specified, use 80 port.
+// The hostPort allows HOST:PORT or HOST. If no port is specified, port 80 will be used.
 func WithExporterFull(name, hostPort string, logger *zap.SugaredLogger) ConfigOption {
 	return func(cfg *config.Config) error {
-		const defaultPort = "80"
+		const defaultPort = ":80"
 		var (
 			exporter trace.Exporter
 			closer   io.Closer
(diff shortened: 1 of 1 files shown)


tracing/opencensus.go
```diff
@@ -130,17 +138,22 @@ func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
...
 				if err != nil {
 					return fmt.Errorf("unable to get hostname: %w", err)
 				}
-				name = n
+				hostPort = n
+			}
+			if name == "" {
+				name = hostPort
+			}
+			if strings.IndexByte(hostPort, ':') < 0 {
```
markusthoemmes, line 154: Why not `!strings.Contains(hostPort, ":")`?

markusthoemmes: If you're worried about performance: It does the exact same thing :). It optimizes `len(substr) == 1` to an IndexByte call.

vagababov: Also `LastIndex`?

nak3:
/assign @mdemirhan 

Could you please take a look?

tracing/opencensus.go
```diff
@@ -114,7 +115,14 @@ func createOCTConfig(cfg *config.Config) *trace.Config {
...
 // it to export traces based on the configuration read from config-tracing.
 func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
+	return WithExporterFull(name, name, logger)
+}
+
+// WithExporterFull supports hostPort argument, which is used for the endpoint.
+// The hostPort allows HOST:PORT or HOST. If no port is specified, port 80 will be used.
+func WithExporterFull(name, hostPort string, logger *zap.SugaredLogger) ConfigOption {
 	return func(cfg *config.Config) error {
+		const defaultPort = ":80"
```
markusthoemmes, line 125: What is the significance of this `80`? Is it ever correct in the context of a pod?

nak3: The `:80` is for compatibility. Current client code calls `WithExporterFull()` without using port but [fixed :80 is appended](https://github.com/knative/pkg/pull/1353/files#diff-a95c97ba65048dbcfd15b2d2b54b4a0dL143) so I thought we should keep using it.

nak3:
@mattmoor Could you please take a look at this? Current zipkin/jaeger tracing does not register the endpoint IP, so it is not possible to find the endpoint pod we want to investigate. This patch can change the IP tag from service's to endpoint's.

Current output:

ip

nak3:
@mattmoor friendly ping.

PR #1355, opened 2020-05-26
Kubelet probes would result in the webhook writing the HTTP status twice

dprotaso:
Doesn't seem like it affected anything - just writes out some extra
log messages

### DIFF
--- a/webhook/webhook_test.go
+++ b/webhook/webhook_test.go
@@ -129,7 +129,7 @@ func TestWebhookKubeletProbe(t *testing.T) {
 	webhook.ServeHTTP(&recorder, probeReq)
 
 	if got, want := recorder.Code, http.StatusInternalServerError; got != want {
-		t.Fatalf("probe got HTTP status %d - expected %d", got, want)
+		t.Fatalf("Probe got HTTP status %d - expected %d", got, want)
 	}
 
 	if got, want := recorder.writeCount, 1; got != want {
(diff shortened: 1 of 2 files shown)


webhook/webhook_test.go
```diff
@@ -96,3 +98,51 @@ func TestRegistrationStopChanFire(t *testing.T) {
...
+	defer cancel()
+
+	recorder := bombRecorder{ResponseRecorder: httptest.NewRecorder()}
+	probeReq := httptest.NewRequest("GET", "/", nil)
+	probeReq.Header.Set("User-Agent", "kube-probe/1.16")
+
+	webhook.ServeHTTP(&recorder, probeReq)
+
+	if got, want := recorder.Code, http.StatusOK; got != want {
+		t.Fatalf("probe got HTTP status %d - expected %d", got, want)
```
vagababov, line 114:
```suggestion
		t.Fatalf("Probe got HTTP status %d - expected %d", got, want)
```
Same below.

webhook/webhook_test.go
```diff
@@ -96,3 +98,51 @@ func TestRegistrationStopChanFire(t *testing.T) {
...
+	if got, want := recorder.writeCount, 1; got != want {
+		t.Errorf("HTTP status was written %d times - expected only one write", got)
+	}
+
+	// Stop the webhook - which means probes should fail
+	//
+	// The steps below aren't obvious and requires you to
+	// know the implementation details
+	cancel()
+	webhook.Run(ctx.Done())
```
vagababov, line 126: can you just do do `close(webhook.stop)`?

dprotaso: i'd rather not access the internals directly if I can avoid it

PR #1357, opened 2020-05-26
Add Status.Annotations to the conversion workflow

vagababov:
/assign @dprotaso @vaikas

### DIFF
--- a/apis/duck/v1/status_types.go
+++ b/apis/duck/v1/status_types.go
@@ -21,6 +21,7 @@ import (
 
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/apis/duck"
+	"knative.dev/pkg/kmeta"
 )
 
 // +genduck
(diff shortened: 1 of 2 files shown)


apis/duck/v1/status_types.go
```diff
@@ -93,6 +94,8 @@ func (s *Status) GetCondition(t apis.ConditionType) *apis.Condition {
 // return true the condition type will be copied to the sink
 func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ...func(apis.ConditionType) bool) {
 	sink.ObservedGeneration = source.ObservedGeneration
+	// This will deep copy the map.
+	sink.Annotations = kmeta.UnionMaps(source.Annotations, nil)
```
dprotaso, line 98: if only union maps was a variadic ;)

vagababov: I can do that, actually with backward compatibility, I guess.
But again different exercise :D

vaikas:
/lgtm
/approve

Thanks!!!

PR #1361, opened 2020-05-27
User variadic version and remove nil

vagababov:
Update duck/status to use new version of UnionMaps

/assign @vaikas mattmoor

### DIFF
--- a/apis/duck/v1/status_types.go
+++ b/apis/duck/v1/status_types.go
@@ -96,7 +96,7 @@ func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ..
 	sink.ObservedGeneration = source.ObservedGeneration
 	if source.Annotations != nil {
 		// This will deep copy the map.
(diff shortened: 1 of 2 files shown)


apis/duck/v1/status_types.go
```diff
@@ -96,7 +96,7 @@ func (source *Status) ConvertTo(ctx context.Context, sink *Status, predicates ..
 	sink.ObservedGeneration = source.ObservedGeneration
 	if source.Annotations != nil {
 		// This will deep copy the map.
-		sink.Annotations = kmeta.UnionMaps(source.Annotations, nil)
+		sink.Annotations = kmeta.UnionMaps(source.Annotations)
```
vaikas, line 99: nit, but maybe copy the comment about deep copying to the v1beta1 as well?

vagababov: eh, sure.

vagababov:
ping :)

PR #1363, opened 2020-05-28
Implement Unwrap() and Is() for permanentErrors

antoineco:
Without the `Unwrap()` method, `ReconcilerEvents` can't be unwrapped from `permanentError`s (L198), and we jump directly over the block (L209).

https://github.com/knative/pkg/blob/a468cb85692fc17e6abd2677f0b88f3337089723/client/injection/kube/reconciler/core/v1/namespace/reconciler.go#L196-L212

`Is()` allows for `permanentError`s to be wrapped themselves and still be considered as permanent.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -448,7 +448,7 @@ func (c *Impl) processNextWorkItem() bool {
 func (c *Impl) handleErr(err error, key types.NamespacedName) {
 	c.logger.Errorw("Reconcile error", zap.Error(err))
 
-	// Re-queue the key if it's an transient error.
+	// Re-queue the key if it's a transient error.
 	// We want to check that the queue is shutting down here
 	// since controller Run might have exited by now (since while this item was
 	// being processed, queue.Len==0).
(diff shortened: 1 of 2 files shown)


controller/controller.go
```diff
@@ -495,14 +495,10 @@ type permanentError struct {
...
-// IsPermanentError returns true if given error is permanentError
+// IsPermanentError returns true if the given error is a permanentError.
 func IsPermanentError(err error) bool {
-	switch err.(type) {
-	case permanentError:
-		return true
-	default:
-		return false
-	}
+	_, ok := err.(permanentError)
```
vagababov, line 500: can we use errors.Is instead?

antoineco: That's what I also expected, but if I change that function to
```diff
 func IsPermanentError(err error) bool {
-       _, ok := err.(permanentError)
-       return ok
+       return errors.Is(err, permanentError{})
 }
```
my test fails with `Expected type controller.permanentError to be a permanentError`.

antoineco: @vagababov oh, I know why. `Is()` calls `Unwrap()`, which returns the wrapped error instead of the permanentError.

vagababov: I see.
Thanks.

antoineco: Solved in accacc9f9a883fe0793fa3a7b78dc8b6bc9664c6.
It's actually a good pattern because a permanentError can be wrapped itself, and still be permanent.

antoineco:
@vagababov sorry for cancelling your LGTM, I removed the unnecessary `err` var from the `Is()` method.

The PR is ready now.

PR #1367, opened 2020-05-29
Don't swallow errors from stackdriver init.

evankanderson:
Found this while poking at writing an e2e test case for stackdriver.

Without this patch, I get a panic. With the patch, I get:

> unable to init: stackdriver: couldn't initialize trace client: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

/assign @jjzeng-seattle

### DIFF
--- a/metrics/stackdriver_exporter.go
+++ b/metrics/stackdriver_exporter.go
@@ -102,11 +102,12 @@ func init() {
 
 func newOpencensusSDExporter(o stackdriver.Options) (view.Exporter, error) {
 	e, err := stackdriver.NewExporter(o)
-	if err == nil {
-		// Start the exporter.
-		// TODO(https://github.com/knative/pkg/issues/866): Move this to an interface.
-		e.StartMetricsExporter()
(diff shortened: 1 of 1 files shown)


metrics/stackdriver_exporter.go
```diff
@@ -102,11 +102,12 @@ func init() {
...
-	if err == nil {
-		// Start the exporter.
-		// TODO(https://github.com/knative/pkg/issues/866): Move this to an interface.
-		e.StartMetricsExporter()
+	if err != nil {
+		return nil, err
 	}
+	// Start the exporter.
+	// TODO(https://github.com/knative/pkg/issues/866): Move this to an interface.
+	e.StartMetricsExporter()
```
vagababov, line 110: Well then we should consume the error from `StartMetricsExporter` as well.

evankanderson: Good catch, I didn't realize it returned an `error`.

PR #1370, opened 2020-06-02
Junithelper refactor into a function that can be re-used.

peterfeifanchen:
Junithelper was previously just a command, refactoring it as a function so it can be re-used by upstream vendors.

/cc @chizhg @chaodaiG

### DIFF
--- a/test/junit/junit.go
+++ b/test/junit/junit.go
@@ -21,6 +21,8 @@ package junit
 import (
 	"encoding/xml"
 	"fmt"
+	"io/ioutil"
+	"log"
 )
 
 // TestStatusEnum is a enum for test result status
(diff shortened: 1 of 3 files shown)


chaodaiG:
Might be my personal preference, I would prefer
```
- junithelper
  - cmd
    - main.go
  - pkg
    - junit.go
```

peterfeifanchen:
> Might be my personal preference, I would prefer
> 
> ```
> - junithelper
>   - cmd
>     - main.go
>   - pkg
>     - junit.go
> ```

I actually prefer it the other way since when you import you can tell its junit.{something} rather than a very innocuous pkg.{something}

chaodaiG:
> I actually prefer it the other way since when you import you can tell its junit.{something} rather than a very innocuous pkg.{something}

Makes sense to me. I still feel it's a bit weird for a child package to import parent package, normally it's the other way around. The previous suggestion could be:
```
- junithelper
  - cmd
    - main.go
  - pkg
    - junit
      - junit.go
```

Alternatively, both on top level:
```
- junithelper
- junit
```

peterfeifanchen:
> > I actually prefer it the other way since when you import you can tell its junit.{something} rather than a very innocuous pkg.{something}
> 
> Makes sense to me. I still feel it's a bit weird for a child package to import parent package, normally it's the other way around. The previous suggestion could be:
> 
> ```
> - junithelper
>   - cmd
>     - main.go
>   - pkg
>     - junit
>       - junit.go
> ```
> 
> Alternatively, both on top level:
> 
> ```
> - junithelper
> - junit
> ```

I think it should be fine where it is. Probably something like pkg/junit would be better to keep the level the same. But things are already so messy, I think it's nice to keep all the mains in testutils as a place to put executables. The rest should eventually be moved into a. pkg dir which we can start here.

PR #1371, opened 2020-06-02
Add the flag on the config-observability, for enabling the tag label on the request metrics

igsong:
To toggle the feature in https://github.com/knative/serving/pull/8103, this PR introduces new flag in `config-observability`. 

The name of the flag is `metrics.enable-tag-label-on-request-metrics` and if it is `true`, all the metrics `requests-*` exposed by the queue-proxy will have additional label `tag`, which is the value of the tag header.

The default value is `false`.


tcnghia:
/project "Networking: Routing Based on Tag Header" "In progress"

ZhiminXiang:
friendly ping @yanweiguo :) need your approval

yanweiguo:
I think this one is no longer necessary according to the discussion in https://github.com/knative/serving/pull/8103

igsong:
Right. I'll close. Thanks

PR #1372, opened 2020-06-02
Initialize conditions in preprocess

whaught:
Initialize conditions in preprocess

This is an oft-repeated piece of code that we have in reconcilers. We should ensure conditions are initialized before we reconcile.

### DIFF
--- a/reconciler/reconcile_common.go
+++ b/reconciler/reconcile_common.go
@@ -32,6 +32,8 @@ func PreProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 	if newStatus.ObservedGeneration != resource.GetGeneration() {
 		condSet := resource.GetConditionSet()
 		manager := condSet.Manage(newStatus)
+		// Ensure conditions are initialized before we modify.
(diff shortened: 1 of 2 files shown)


reconciler/reconcile_common.go
```diff
@@ -32,6 +32,8 @@ func PreProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 	if newStatus.ObservedGeneration != resource.GetGeneration() {
 		condSet := resource.GetConditionSet()
 		manager := condSet.Manage(newStatus)
+		// Ensure conditions are initialized before we modify.
+		manager.InitializeConditions()
```
grantr, line 36: Just to confirm my understanding, this behavior is opt-in per controller?

grantr: Synced with @whaught online. This is opt-in and only a few controllers are using it. Those controllers already initialize conditions at the start of reconcile.

whaught: https://github.com/knative/serving/pull/8169

Tested in serving to make sure.

PR #1373, opened 2020-06-03
Throw the exception of loggingconfig

Gallardot:
Refer to https://github.com/knative/operator/issues/119 . The root cause is lost. So I think throw the exception of loggingconfig is ok.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"fmt"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -96,7 +96,13 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
...
 		var err error
 		loggingConfigMap, err = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(logging.ConfigMapName(), metav1.GetOptions{})
-		return err == nil || apierrors.IsNotFound(err), nil
+		if err == nil {
+			return true, nil
+		}
+		if apierrors.IsNotFound(err) || apierrors.IsTimeout(err) || apierrors.IsServerTimeout(err) {
+			return false, nil
+		}
+		return false, err
```
vagababov, line 105: So what are we trying to achieve here? Retry on NotFound/Timeout?
And you want to pass the error in the other cases?

Gallardot: > So what are we trying to achieve here? Retry on NotFound/Timeout?
> And you want to pass the error in the other cases?

Yes, exactly！

Previously, we would only get the `timed out waiting for the condition` error message. Retry on NotFound. But sometimes that's not the root cause.

Refer to https://github.com/knative/operator/issues/119#issuecomment-638010432. In fact, What I encounter is forbidden Error. We lost the root cause.

vagababov (commented):
/lgtm
/approve

injection/sharedmain/main.go
```diff
@@ -105,7 +105,13 @@ func GetLoggingConfig(ctx context.Context) (*logging.Config, error) {
 	if err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
 		var err error
 		loggingConfigMap, err = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(ctx, logging.ConfigMapName(), metav1.GetOptions{})
-		return err == nil || apierrors.IsNotFound(err), nil
+		if err == nil {
+			return true, nil
+		}
+		if apierrors.IsNotFound(err) || apierrors.IsTimeout(err) || apierrors.IsServerTimeout(err) {
```
markusthoemmes, line 111: This makes me a little nervous tbh as I'm not sure if these are the only errors happening on this path.

Could we just store the error and return it eventually like so:

```golang
var lastErr error
if err := wait.PollImmediate(1*time.Second, 5*time.Second, func() (bool, error) {
  loggingConfigMap, lastErr = kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(ctx, logging.ConfigMapName(), metav1.GetOptions{})
  return lastErr == nil || apierrors.IsNotFound(lastErr), nil
}); err != nil {
  return nil, fmt.Errorf("timed out waiting for the condition: %w", lastErr)
}

Gallardot: Thanks @markusthoemmes .

I agree with you. I will fix it ASSAP.

markusthoemmes (approved):
/lgtm
/approve

Thanks!

PR #1374, opened 2020-06-03
Use a hexadecimal representation for the example checksum.

markusthoemmes:
As per @julz comment on https://github.com/knative/serving/pull/8123 this is more usual. I like it better anyway as the `Checksum` function now controls the representation.

/assign @julz


julz:
👍 love it.

/lgtm

PR #1378, opened 2020-06-03
Update boskos dependency to sigs.k8s.io/boskos

ixdy:
Boskos has moved into kubernetes-sigs/boskos, so I'm preemptively updating the dep here so that future updates are painless. There shouldn't be any notable boskos functionality changes.


ixdy:
Running `hack/update-deps.sh` seems to have removed all of the `## explicit` comments, and I'm not sure why.

chaodaiG:
@chizhg 

> Running `hack/update-deps.sh` seems to have removed all of the `## explicit` comments, and I'm not sure why.

@chizhg , any idea?

ixdy:
It looks like `hack/update-k8s-deps.sh v0.17.6` has the desired result, so that should unblock me, but I think there might be something broken in `hack/update-deps.sh`.

chizhg:
> @chizhg
> 
> > Running `hack/update-deps.sh` seems to have removed all of the `## explicit` comments, and I'm not sure why.
> 
> @chizhg , any idea?

Not sure about it. The `vendor/modules.txt` file is auto-generated by Go mod, and normally we don't touch it.

chizhg:
> It looks like `hack/update-k8s-deps.sh v0.17.6` has the desired result, so that should unblock me, but I think there might be something broken in `hack/update-deps.sh`.

That's weird, it looks the k8s dependencies are already in v0.17.6

ixdy:
hm, I reset everything and ran `hack/update-deps.sh` again, and this time it did the right thing. So I'm not really sure what's going on.

ixdy:
all tests are passing; I believe this is ready for review.

ixdy:
rebased

PR #1380, opened 2020-06-04
Use annotation rather than label for configmap checksum

julz:
Slightly more correct to use an annotation rather than a label since this isn't something one would select on.

/assign @markusthoemmes

### DIFF
--- a/configmap/hash-gen/main.go
+++ b/configmap/hash-gen/main.go
@@ -68,19 +68,24 @@ func process(data []byte) ([]byte, error) {
 		return nil, nil
 	}
 
-	labels := traverse(content, "metadata", "labels")
-	if labels == nil {
-		// TODO(markusthoemmes): Potentially handle missing metadata and labels?
-		return nil, errors.New("'metadata.labels' not found")
+	metadata := traverse(content, "metadata")
+	if metadata == nil {
+		return nil, errors.New("'metadata' not found")
+	}
+
+	annotations := traverse(metadata, "annotations")
+	if annotations == nil {
+		annotations = &yaml.Node{Kind: yaml.MappingNode}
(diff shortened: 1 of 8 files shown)


configmap/hash-gen/main.go
```diff
@@ -68,19 +68,24 @@ func process(data []byte) ([]byte, error) {
...
-	if labels == nil {
-		// TODO(markusthoemmes): Potentially handle missing metadata and labels?
-		return nil, errors.New("'metadata.labels' not found")
+	metadata := traverse(content, "metadata")
+	if metadata == nil {
+		return nil, errors.New("'metadata' not found")
+	}
+
+	annotations := traverse(metadata, "annotations")
+	if annotations == nil {
```
markusthoemmes, line 77: Can you add a test for this, that adds the mapping node for annotations **and** the annotation itself?

julz: Yeah good idea, will do

julz: (actually the existing tests do that, but I’ll add one for the annotation node already existing)

julz: done, PTAL :)

julz: (actually, now I think about it, this was technically already covered - `add.yaml` covers the case where annotations don't already exist, `update.yaml` covers the case where they don't).

markusthoemmes: You're right, I misread the diff. Feel free to drop the commit :joy:

julz: hehe ok :)

julz:
tested out a bump locally and it looks like this should Just Work with a regular `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh` without any staging needed \o/. Unholding.

/unhold

PR #1381, opened 2020-06-05
Remove the trailing \n

vagababov:
Due to the parser difference we had 2 different values in WH and in the hash gen

Fixes: https://github.com/knative/serving/issues/8211

/assign @markusthoemmes

### DIFF
--- a/configmap/example.go
+++ b/configmap/example.go
@@ -26,11 +26,12 @@ const (
 	// ExampleKey signifies a given example configuration in a ConfigMap.
 	ExampleKey = "_example"
 
-	// ExampleChecksumLabel is the label that stores the computed checksum.
-	ExampleChecksumLabel = "knative.dev/example-checksum"
+	// ExampleChecksumAnnotation is the annotation that stores the computed checksum.
+	ExampleChecksumAnnotation = "knative.dev/example-checksum"
 )
 
-// Checksum generates a checksum for the example value to be compared against a respective label.
+// Checksum generates a checksum for the example value to be compared against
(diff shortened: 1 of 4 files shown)


vagababov:
It is important to cut only trailing spaces.

configmap/hash-gen/main.go
```diff
@@ -74,7 +76,7 @@ func process(data []byte) ([]byte, error) {
 		return nil, errors.New("'metadata.labels' not found")
 	}
 
-	checksum := configmap.Checksum(example.Value)
+	checksum := configmap.Checksum(strings.TrimRightFunc(example.Value, unicode.IsSpace))
```
markusthoemmes, line 79: Should we trim the value in the checksum generator itself and trim it left and right for good measure? Seems like no harm done if whitespace is added and I wonder if that could accidentally happen during edits :thinking:

vagababov: Sure sounds good.

configmap/example.go
```diff
@@ -29,7 +30,9 @@ const (
 	ExampleChecksumAnnotation = "knative.dev/example-checksum"
 )
 
-// Checksum generates a checksum for the example value to be compared against a respective annotation.
+// Checksum generates a checksum for the example value to be compared against
+// a respective annotation.
+// Heading and trailing spaces are ignored.
```
julz, line 35: pedantry:

```suggestion
// Leading and trailing spaces are ignored.
```

vagababov: Done.

Issue #1319, opened 2020-05-09
Move leaderelection away from enableComponents


mattmoor:
/area API
/kind cleanup

In hindsight, this was a poor choice for how to enable leaderelection, since it doesn't follow our paradigm of supporting hot reloads.

A few problems top-of-mind:
 - Enabling it for components doesn't restart the components (this leads to fragile trickery like scaling down/up deployments to force restarts after patching the configmap).
 - The field validation for this hinders our extensibility (e.g. imagine validating the component names within `knative-sources` 😬)

cc @mgencur @pmorie

mattmoor:
I'm thinking that a flag defined by `sharedmain` would solve things, but wondering what others think.

mgencur:
How would the user enable/disable leader election with the flag? By editing the controller's Deployment and changing the `args` section?
I'm thinking that it's not necessary to give the user options for enabling/disabling each controller (autoscaler, autoscaler-hpa, controller, etc.) separately. Having an option for setting this for `all` components that we support seems sufficient to me.
The ConfigMap `config-leaderelection` seems a natural place for me to have these settings. I believe it's possible to enforce pod restarts when this config is changed, something like this??:
```
logger.Info("Leader-elected mode changed. Restarting controller...)
os.Exit(1)
```
There are more configuration options for leader election, not just enabling/disabling it. Right now I think none of them can be changed at runtime because we don't "watch" the ConfigMap. So we'd need  more flags to be able to configure this? Or we'd need the flag and also the ConfigMap.

mgencur:
OTOH, having leader election enabled by default with only one pod for the given deployment would be an option too. With just one pod it doesn't have effect so this could be safe. I can try a test PR with this to let all test suites run.

mattmoor:
> having leader election enabled by default with only one pod for the given deployment would be an option too.

IMO we should move the currently supported components to this model, but we need a way for components that don't yet support it (e.g. autoscaler) to be able to opt-out.

pmorie:
I agree the current implementation isn't optimal - a CLI flag sounds okay to me.

mattmoor:
This is done

mgencur:
OK. This is closed without even looking at the code in the linked PR? Sigh.

PR #1382, opened 2020-06-05
Define flag --leader-election-enabled for sharedmain

mgencur:
Fixes https://github.com/knative/pkg/issues/1319

Preview of knative-serving using this flag is at https://github.com/knative/serving/pull/8205 (this preview is slightly out-of-date. The flag name changed from `--leader-election-disabled` to `--leader-election-enabled`

When/if this is merged , there will be changes needed in knative/serving, net-istio, net-kourier, net-certmanager because the common flags master and kubeconfig are now at "sharedmain" package level. These are in fact common to all the repositories so no need to duplicate them everywhere. See the knative-serving PR for details

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -59,6 +59,15 @@ import (
 	"knative.dev/pkg/webhook"
 )
 
+var (
+	masterURL = flag.String("master", "",
+		"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	kubeconfig = flag.String("kubeconfig", "",
+		"Path to a kubeconfig. Only required if out-of-cluster.")
+	leaderElectionEnabled = flag.Bool("leader-election-enabled", false,
+		"Set this flag to true to enable leader election.")
+)
+
 // GetConfig returns a rest.Config to be used for kubernetes client creation.
 // It does so in the following order:
 //   1. Use the passed kubeconfig/masterURL.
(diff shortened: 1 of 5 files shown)


n3wscott:
I would rather see this in a config map, in-fact, there is already a leader election config map: https://github.com/knative/serving/blob/master/config/core/configmaps/leader-election.yaml

n3wscott:
ah in the linked issue this was talked about. The flag is about forcing a re-roll out for the change to the pods. 

Ignore me :D

mgencur:
Sure. it's possible to keep default as it is now. I'm just curious what will be the signal to flip the behavior and enable this by default. Because it's not clear right now. See my comment on https://github.com/knative/pkg/issues/1319. All tests pass if the Leader Election is enabled and the controllers have just a single pod - that could be default behavior even now, IMO.
If this is not desired let me know if it's better to have `--leader-election-disabled` with true as default or `--leader-election-enabled` with false as default.

mgencur:
@mattmoor I've rebased it to the latest changes from you. And also changed the default to leader election disabled. Can you review, pls?

injection/sharedmain/main.go
```diff
@@ -187,36 +200,38 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
...
 
-	// Set up leader election config
-	leaderElectionConfig, err := GetLeaderElectionConfig(ctx)
-	if err != nil {
-		logger.Fatalw("Error loading leader election configuration", zap.Error(err))
-	}
-	leConfig := leaderElectionConfig.GetComponentConfig(component)
-
-	if !leConfig.LeaderElect {
+	if *leaderElectionEnabled {
```
n3wscott, line 203: what is going on with the `*`? is that required?

mgencur: Yes. Because `flag.Bool(...)` returns a pointer to bool. So we dereference it.

matzew:
@mgencur what's the status here ?

mgencur:
@matzew hey, please look at my comment in the linked issue. Closing this.

Issue #1387, opened 2020-06-07
package apis/duck/v1beta1 imports a client, forcing consumers onto the same kube version


alvaroaleman:
/area api
/kind bug

The `apis/duck/v1beta1` imports the `apis/duck` package which contains a custom build informer (`apis/duck/typed.go`). This means that each and every consumer of this package is forced to use the same version of the kubernetes dependencies as this project and that its impossible to use a downstream generated client or dynamic client and then freely choose whatever kube version is desired. This is especially bad since this is an api package.

PR #1388, opened 2020-06-07
Avoid importing apis/duck from apis/duck/<<version>>

alvaroaleman:
Fixes https://github.com/knative/pkg/issues/1387

* Moves some type definitions from apis/duck into apis/duck/ducktypes
* Creates aliases from apis/duck to the new location to stay compatible
* Moves a set of var definitions from the apis/duck/{v1beta1,v1} into apis/duck


apis/duck/v1_tests.go
```diff
@@ -0,0 +1,78 @@
...
+
+package duck
+
+import (
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+
+	"knative.dev/pkg/apis/duck/v1"
```
mattmoor, line 25: Format Go code:
```suggestion

```

apis/duck/v1beta1/addressable_types.go
```diff
@@ -24,8 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"knative.dev/pkg/apis"
-	"knative.dev/pkg/apis/duck"
-	v1 "knative.dev/pkg/apis/duck/v1"
+	"knative.dev/pkg/apis/duck/ducktypes"
+	"knative.dev/pkg/apis/duck/v1"
```
mattmoor, line 28: Format Go code:
```suggestion

```

alvaroaleman:
@mattmoor sorry, I don't understand your comments. According to your suggestions I should remove imports which would break the code.

The code in this PR is properly formatted via `go fmt -s`:
```
$ gofmt -s -w ./apis/..
alvaro@t470s-[2020-06-06-22:09]:[try1][1].../tmp.BhWuy7JwT9/pkg
$ git status|grep -v vendor
On branch try1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)

no changes added to commit (use "git add" and/or "git commit -a"
```

apis/duck/v1_tests.go
alvaroaleman: This would break compilation

apis/duck/v1beta1/addressable_types.go
alvaroaleman: This would break compilation

vagababov: why would it?

alvaroaleman: Because it removes an import that is being referenced?

alvaroaleman:
@markusthoemmes is there anything I have to do regarding the failing presubmit? I just moved code around so the test coverage shouldn't change to how it was before

alvaroaleman:
@markusthoemmes @mattmoor @vagababov the issue this PR fixes is pretty problematic for downstream projects that import the knative api. The PR is apicompatible thanks to the alias so it shouldn't cause any issue.

Can you please take another look?

apis/duck/v1beta1/addressable_types.go
vagababov: I guess we need to call it `duckv1` to conform to our repository conventions.
I think at some point it saw both named and unnamed import.

alvaroaleman: there is already a `duckv1`. My PR does not change anything about the existing imports other than removing the useless `v1` prefix.

apis/duck/v1_tests.go
mattmoor, line 25: Format Go code:
```suggestion

```

apis/duck/v1beta1/addressable_types.go
mattmoor, line 28: Format Go code:
```suggestion

```

alvaroaleman:
Can someone please have a look? The fact that we import these types blocks us from being able to upgrade kube dependencies. There is zero reason to import apimachinery into an api package.

vaikas:
/lgtm
/approve
/hold
Adding a hold so folks that had comments can look. I'll remove by EOD if nobody complains. @n3wscott @vagababov 
My most sincere apologies, this totally fell through the cracks :(

vaikas:
@alvaroaleman re: coverage, you're fine nothing else to do, not your fault :)

apis/duck/v1_tests.go
mattmoor: This is the output of `goimports`, you need:
```suggestion

	v1 "knative.dev/pkg/apis/duck/v1"
```

n3wscott:
I don't understand why this is needed. the stuff in duck/ defines interfaces that are used in the versions. Now ducktype is a pseudo version but the code is the same. Why is this better?

alvaroaleman:
@n3wscott the problem is that `knative.dev/pkg/apis/duck/v1` imports  `knative.dev/pkg/apis/duck1` which in turn imports `k8s.io/apimachinery/pkg{runtime,runtime/schema,watch}`.


Because there are implementations in `duck/v1` for `duck` interfaces that themselves return an interface which is also defined in `duck`, the latter had to be moved out because golang apparently does not allow an interface implementation to have a return type that differs from what is in that implemented interface, even if that is another interface that gets implemented by the return type. i.E.:

```diff
diff --git a/apis/duck/v1/addressable_types.go b/apis/duck/v1/addressable_types.go
index 3cfcac33..e53daaf0 100644
--- a/apis/duck/v1/addressable_types.go
+++ b/apis/duck/v1/addressable_types.go
@@ -24,7 +24,6 @@ import (
        "k8s.io/apimachinery/pkg/runtime"
 
        "knative.dev/pkg/apis"
-       "knative.dev/pkg/apis/duck/ducktypes"
 )
 
 // +genduck
@@ -68,7 +67,7 @@ var (
 )
 
 // GetFullType implements duck.Implementable
-func (*Addressable) GetFullType() ducktypes.Populatable {
+func (*Addressable) GetFullType() *AddressableType {
        return &AddressableType{}
 }
```
fails to compile:
```
apis/duck/v1_tests.go:48:37: cannot use &"knative.dev/pkg/apis/duck/v1".Addressable literal (type *"knative.dev/pkg/apis/duck/v1".Addressable) as type ducktypes.Implementable in field value:
	*"knative.dev/pkg/apis/duck/v1".Addressable does not implement ducktypes.Implementable (wrong type for GetFullType method)
		have GetFullType() *"knative.dev/pkg/apis/duck/v1".AddressableType
		want GetFullType() ducktypes.Populatable
```

Hence, those interfaces had to be moved to a new package that is imported by both `api/duck` and `api/duck/v1`

apis/duck/v1_tests.go
alvaroaleman: I have no idea what you are talking about, the change you are suggesting looks like you have some broken tooling

n3wscott:
Sounds like you have a dep problem and the repo is not the problem. Duck/v1 also imports api machinery so it is not clear how moving the interfaces changes anything. 

Are you using a very very new version of golang? Use compile this arrangement all the time in Knative and numerous other projects. 

What is the exact compile issue? And are you using go mod?

alvaroaleman:
@n3wscott Thanks for your feedback.

> Sounds like you have a dep problem and the repo is not the problem

Yes I have a dep problem _because_ of this repo.

> Duck/v1 also imports api machinery so it is not clear how moving the interfaces changes anything.

Yes it imports some from apimachinery but much less, just `k8s.io/apimachinery/pkg/apis/meta/v1` and `k8s.io/apimachinery/pkg/runtime` of which it just uses metav1.ObjectMeta, metav1.TypeMeta and runtime.Object, all of which are extremely stable types/interfaces. 

> Are you using a very very new version of golang?

1.13 but that should be irrelevant for this

>  Use compile this arrangement all the time in Knative and numerous other projects.

I don't understand what you mean by this, can you elaborate?

> What is the exact compile issue?

https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/test-infra/17869/pull-test-infra-bazel/1269424108739760128#1:build-log.txt%3A4846 
From https://github.com/kubernetes/test-infra/pull/17869

That pull does not fail on knative anymore with the changes in this PR.

> And are you using go mod?

Yes, but I doubt that is relevant.

alvaroaleman:
@n3wscott does that explanation make sense to you? Is there anything further you would like to know? If not, I would appreciate a `/hold cancel` so we can proceed with updating the kube dependencies :)

alvaroaleman:
@n3wscott ping

alvaroaleman:
@n3wscott Can you please respond? This is still blocking us from upgrading the version of our kube dependencies even though the absolute only thing we want from this repo are the types. It is not sustainable for downstream projects if their dependencies force them to use a specific kube client-go version.

vagababov:
Scott's on vacation
/cc @mattmoor @vaikas

n3wscott:
/unhold 

Sorry looks like you are trying to cherry pick a single class and it is at a package that pulls a ton of other stuff.

Sorry for the delay, my GitHub notifications are a... challenge 😭

Thanks for the ping!

/Lgtm
/Approve

PR #1391, opened 2020-06-08
Use CRD & Webhook v1 APIs

dprotaso:
Addresses: https://github.com/knative/serving/issues/6584


hack/update-codegen.sh
```diff
@@ -41,7 +41,7 @@ EXTERNAL_INFORMER_PKG="k8s.io/client-go/informers" \
   ${REPO_ROOT_DIR}/hack/generate-knative.sh "injection" \
     k8s.io/client-go \
     k8s.io/api \
-    "admissionregistration:v1beta1 apps:v1 autoscaling:v1,v2beta1 batch:v1,v1beta1 core:v1 rbac:v1" \
+    "admissionregistration:v1beta1,v1 apps:v1 autoscaling:v1,v2beta1 batch:v1,v1beta1 core:v1 rbac:v1" \
```
vagababov, line 44: do we still need the beta?

dprotaso: No - but I don't want to break anyone consuming it downstream so I left it

dprotaso: We can drop with when the API version is dropped by K8s or it becomes a pain to maintain

vagababov: it's probably file. It's just more files :) If they're unused might not even be linked, so should not matter

dprotaso:
Integration and upgrade tests passed in downstream https://github.com/knative/serving/pull/8259

I was going to swap serving's CRDs to v1 but that can be a separate PR and doesn't need to block the webhook infra using v1 APIs

matzew:
started looking at eventing in https://github.com/knative/eventing/pull/3360

PR #1399, opened 2020-06-11
Allows for webhooks to exclude certain namespaces

### LINKED ISSUE #1379: allow for webhooks to exclude certain namespaces
**Problem**
Some webhooks such as: https://github.com/knative/eventing-contrib/blob/master/kafka/source/config/500-webhook-configuration.yaml#L18 do not limit themselves to just certain namespaces. This can be problematic if that webhook is acting up (e.g. the service is not responding), and the request that is being processed is not related to Knative at all but instead could be something critical such as being part of a system upgrade.

One options we discussed today is to define a label that can be added to any namespace such that the Knative webhooks could then be defined to exclude any namespace that has that label. For example, an admin could add the label to all Kube system namespaces to ensure that knative webhooks are never called for system related resource operations.

This technically isn't just an eventing issue, but since this is the first webhook I found with this problem I picked on it as a testcase and if we can find a solution that works for this one then we can consider applying it to all Knative webhooks.

( https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements )

/cc @mattmoor @mpetason

mpetason:
Added a namespaces selector to the mutating webhook configuration which
allows for excluding namespaces from the webhook

Fixes #1379

### DIFF
--- a/webhook/resourcesemantics/defaulting/defaulting.go
+++ b/webhook/resourcesemantics/defaulting/defaulting.go
@@ -30,6 +30,7 @@ import (
 	jsonpatch "gomodules.xyz/jsonpatch/v2"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 	admissionlisters "k8s.io/client-go/listers/admissionregistration/v1beta1"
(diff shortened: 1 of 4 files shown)


webhook/resourcesemantics/defaulting/defaulting.go
```diff
@@ -172,6 +173,12 @@ func (ac *reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
 			continue
 		}
 		webhook.Webhooks[i].Rules = rules
+		webhook.Webhooks[i].NamespaceSelector = &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{{
+				Key:      "pkg.knative.dev/skipWebhooks",
```
julz, line 178: Think k8s labels are normally hyphen-case, so this may be better as:

```suggestion
				Key:      "pkg.knative.dev/skip-webhooks",
```

mpetason: I'm looking for feedback on what we should name it, if we want to name it something different. I'll update with the hyphen for now.

mpetason:
Forgot to update the table tests! Updating now.

webhook/resourcesemantics/defaulting/defaulting.go
julz: ah ok - I didn’t realise there’s already a conversation on the name.. I’ll let others bike-shed on that, as long as it’s hyphen-case 😅

webhook/resourcesemantics/defaulting/defaulting.go
```diff
@@ -172,6 +173,12 @@ func (ac *reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
 			continue
 		}
 		webhook.Webhooks[i].Rules = rules
+		webhook.Webhooks[i].NamespaceSelector = &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{{
+				Key:      "pkg.knative.dev/skip-webhooks",
```
mattmoor, line 178: Can we name this closer to what the binding equivalent is called?

PR #1402, opened 2020-06-12
Always initialize conditions and set defaults when reconciling

### LINKED ISSUE #1401: We should always InitializeConditions when reconciling
## Expected Behavior

We need to initialize conditions prior to reconciling in case new conditions are added/defaulted across releases

## Actual Behavior

We actually skip initializing conditions if gen == observed gen 
https://github.com/knative/pkg/blob/2ddcfef739a2e0ad5af446bffa0ecd6ce49b760e/reconciler/reconcile_common.go#L29-L41

whaught:
Fixes #1401

We always want to initialize resources in case the defaults have changed or conditions have been added as we change versions (even if they haven't been written to).

We already do this for most reconcilers, but it will be easier to do in the common spot:
https://github.com/knative/serving/blob/d9d5ae4393ef31f1aa639be5bb6e9c0079e39bce/pkg/reconciler/route/route.go#L112

In the future if we can default-on-read or run this on upgrade-only it would be nice to not need to do it for every single reconciliation.

### DIFF
--- a/reconciler/reconcile_common.go
+++ b/reconciler/reconcile_common.go
@@ -52,12 +52,12 @@ func PreProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 // PostProcessReconcile contains logic to apply after reconciliation of a resource.
 func PostProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 	logger := logging.FromContext(ctx)
-	newStatus := resource.GetStatus()
-	mgr := resource.GetConditionSet().Manage(newStatus)
+	status := resource.GetStatus()
+	mgr := resource.GetConditionSet().Manage(status)
 
 	// Bump observed generation to denote that we have processed this
 	// generation regardless of success or failure.
-	newStatus.ObservedGeneration = resource.GetGeneration()
+	status.ObservedGeneration = resource.GetGeneration()
 
 	if rc := mgr.GetTopLevelCondition(); rc == nil {
 		logger.Warn("A reconciliation included no top-level condition")
--- a/reconciler/reconcile_common.go
+++ b/reconciler/reconcile_common.go
@@ -19,6 +19,7 @@ package reconciler
 import (
 	"context"
 
+	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	"knative.dev/pkg/logging"
 )
@@ -29,12 +30,20 @@ const failedGenerationBump = "NewObservedGenFailure"
 func PreProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
 	newStatus := resource.GetStatus()
 
-	if newStatus.ObservedGeneration != resource.GetGeneration() {
-		condSet := resource.GetConditionSet()
-		manager := condSet.Manage(newStatus)
-		// Ensure conditions are initialized before we modify.
-		manager.InitializeConditions()
+	// We may be reading a version of the object that was stored at an older version
+	// and may not have had all of the assumed defaults specified.  This won't result
+	// in this getting written back to the API Server, but lets downstream logic make
+	// assumptions about defaulting.
+	if d, ok := resource.(apis.Defaultable); ok {
+		d.SetDefaults(ctx)
+	}
 
+	// Ensure conditions are initialized before we modify.
+	condSet := resource.GetConditionSet()
+	manager := condSet.Manage(newStatus)
+	manager.InitializeConditions()
+
+	if newStatus.ObservedGeneration != resource.GetGeneration() {
 		// Reset Ready/Successful to unknown. The reconciler is expected to overwrite this.
 		manager.MarkUnknown(condSet.GetTopLevelConditionType(), failedGenerationBump, "unsuccessfully observed a new generation")
 	}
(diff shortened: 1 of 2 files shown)


reconciler/reconcile_common.go
```diff
@@ -29,12 +30,20 @@ const failedGenerationBump = "NewObservedGenFailure"
...
+	// in this getting written back to the API Server, but lets downstream logic make
+	// assumptions about defaulting.
+	if d, ok := resource.(apis.Defaultable); ok {
+		d.SetDefaults(ctx)
+	}
 
+	// Ensure conditions are initialized before we modify.
+	condSet := resource.GetConditionSet()
+	manager := condSet.Manage(newStatus)
+	manager.InitializeConditions()
```
n3wscott, line 44: Are you sure the webhook will store the status changes? I thought there was something about the webhook only allowing spec changes? Because of the fact we don't post to a status update on the subresource, @mattmoor ?

whaught: This is the reconciler, not the webhook?

reconciler/reconcile_common.go
```diff
@@ -29,12 +30,20 @@ const failedGenerationBump = "NewObservedGenFailure"
...
-		manager := condSet.Manage(newStatus)
-		// Ensure conditions are initialized before we modify.
-		manager.InitializeConditions()
+	// We may be reading a version of the object that was stored at an older version
+	// and may not have had all of the assumed defaults specified.  This won't result
+	// in this getting written back to the API Server, but lets downstream logic make
+	// assumptions about defaulting.
+	if d, ok := resource.(apis.Defaultable); ok {
+		d.SetDefaults(ctx)
+	}
```
antoineco, line 39: Even with the comment I fail to understand that part. Why does a function that alters the status need to care about applying defaults? Is the intention to implement some sort of webhook-less defaulting mechanism?

reconciler/reconcile_common.go
```diff
@@ -29,12 +30,20 @@ const failedGenerationBump = "NewObservedGenFailure"
...
 
+	// Ensure conditions are initialized before we modify.
+	condSet := resource.GetConditionSet()
+	manager := condSet.Manage(newStatus)
```
antoineco, line 43: nit: by the look of it it seems like this variable should be named `status` instead of `newStatus` because most objects will want to return their _existing_ status here. Am I right?

reconciler/reconcile_common.go
whaught: No. The webhook is still responsible for defaulting. This is for when defaults change but the user never writes to the resource, the global resync would still default them during reconcile.

You'll see this line repeated in the serving reconcilers:
https://github.com/knative/serving/blob/fcfbb56753d17575969a6f58c432090c8c69733c/pkg/reconciler/configuration/configuration.go#L61

https://github.com/knative/serving/blob/fcfbb56753d17575969a6f58c432090c8c69733c/pkg/reconciler/revision/revision.go#L136

https://github.com/knative/serving/blob/fcfbb56753d17575969a6f58c432090c8c69733c/pkg/reconciler/route/route.go#L112

https://github.com/knative/serving/blob/fcfbb56753d17575969a6f58c432090c8c69733c/pkg/reconciler/service/service.go#L70

I just want to either agree we need this and unify it here if its something we want all reconcilers to be doing. In the future if we can default-on-read instead it may not be nessecary.

reconciler/reconcile_common.go
whaught: sure. I think in a previous version there was both newResource and oldResource as params to this function - that's gone.

reconciler/reconcile_common.go
antoineco: Got it thanks 👍  I read "revision" instead of "version", that's where my confusion came from.

reconciler/reconcile_common.go
```diff
@@ -29,12 +30,20 @@ const failedGenerationBump = "NewObservedGenFailure"
...
-		condSet := resource.GetConditionSet()
-		manager := condSet.Manage(newStatus)
-		// Ensure conditions are initialized before we modify.
-		manager.InitializeConditions()
+	// We may be reading a version of the object that was stored at an older version
+	// and may not have had all of the assumed defaults specified.  This won't result
+	// in this getting written back to the API Server, but lets downstream logic make
+	// assumptions about defaulting.
+	if d, ok := resource.(apis.Defaultable); ok {
+		d.SetDefaults(ctx)
```
vagababov, line 38: This might be a problem for Autoscaler, because those are pluggable and can be a third party implementation. The implementation is chosen based on the annotation. And one of the important debugging signals here is uninitialized conditions if this annotations is set incorrectly.
Now we do filter (https://github.com/knative/serving/blob/master/pkg/reconciler/autoscaling/kpa/controller.go#L101) those, so I _think_ this should be alright, but want to double check that this won't default if we're not actually going into the `ReconcileKind`.

whaught: This is _only_ called immediately before `ReconcileKind` https://github.com/knative/pkg/blob/22662525751b007a039ef72c05193f6a0d3602ee/codegen/cmd/injection-gen/generators/reconciler_reconciler.go#L315

Issue #1287, opened 2020-05-04
update-codegen.sh doesn't work outside GOPATH


ian-mi:
## Expected Behavior
hack/update-codegen.sh should work outside of GOPATH

## Actual Behavior
update-codegen either does nothing or uses an incorrect pkg version.

## Steps to Reproduce the Problem

1. Move repo outside of GOPATH.
2. Make a change to the injection generators.
3. Run ./hack/update-codegen.sh
4. Observe no change to generated code.

antoineco:
I'd love to see this happen too. Unfortunately this has to be tackled in the Kubernetes generators directly. The issue is that those generators always output to `{{output-base}}/{{output-package}}/`.

Technically, it _does_ work outside of the GOPATH, but your repo needs to be cloned is a directory matching `-output-package` _exactly_. For example, if your code is inside `~/my-knative-eventing` and you set `-output-base` to `../../`, all generated files will be written to
```
~/knative.dev/eventing/pkg/client/...
```
instead of
```
~/my-knative-eventing/pkg/client/...
```
because `-output-package` is set to `knative.dev/eventing`.

Cynocracy:
/assign Cynocracy

I didn't see this before hitting the corner case myself :) I have a hacky attempt at a fix in knative.dev/networking, if it works I can upstream it into library.

mattmoor:
cc @n3wscott @ImJasonH

n3wscott:
This is a bug in gengo upstream in Kubernetes.

mattmoor:
/lifecycle frozen

I believe there's a hack @scothis had that created a temporary GOPATH in `./hack/update-codegen.sh` that we could incorporate.

cc @Cynocracy

scothis:
I first saw this pattern in kpack. It basically involves setting up a fake GOPATH for gengo

Cynocracy:
Does this still repro with the new k8s version? I was really hoping it did not :(

Fake GOPATH sounds good, we've done similar things internally when we anger the great and powerful Go Mod.

PR #1403, opened 2020-06-12
Groom conditions LastTransitionTime in postprocess

### LINKED ISSUE #1394: Move LastTransitionTime grooming to reconcile post-process
## Expected Behavior
When updating conditions, we want to ensure that LastTransitionTime does not change if a condition is set but otherwise keeps the same value. Currently we use the ConditionManager to do so on each modification of a condition.

## Actual Behavior
Some implementers might reset the conditions or switch between conditions within the same reconciliation. On write-back this can cause churn of the LastTransitionTime field which may result in unneeded watch events / resource versions.

Example: https://github.com/knative/serving/blob/baace2803b2ce4ff17078db8410ad884c2a2a6f5/pkg/reconciler/autoscaling/hpa/hpa.go#L64

In the case where this reconciler is Inactive we get here: https://github.com/knative/serving/blob/baace2803b2ce4ff17078db8410ad884c2a2a6f5/pkg/reconciler/autoscaling/hpa/hpa.go#L106

We'll keep flipping from Active to Inactive. The user will just see Inactive with a marching timestamp.

## Proposed Solution
In the post-process step for reconciliation, compare the original resource to the written one and preserve the old condition if nothing has changed (exempting timestamps for equality). This will make it so implementers don't need to worry about changing the resource more than once with unintended side-effects.

At the moment its expected that all writes to conditions go through the ConditionManager, but this will also make it so that direct changes to the condition struct won't have unintended side effects either (in practice I don't think people are doing this, but could potentially make future implementations more readable)

whaught:
Fixes #1394

* This fixes behavior if a condition is flip/flopped in a single reconciliation
* All changed conditions share the same time


reconciler/reconcile_common.go
```diff
@@ -55,4 +60,24 @@ func PostProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
...
+// groomConditionsTransitionTime ensures that the LastTransitionTime only advances for resources
+// where the condition has changed during reconciliation. This also ensures that all advanced
+// conditions share the same timestamp.
+func groomConditionsTransitionTime(resource, oldResource duckv1.KRShaped) {
+	now := apis.VolatileTime{Inner: metav1.NewTime(time.Now())}
+	for _, cond := range resource.GetStatus().Conditions {
+
+		if oldCond := oldResource.GetStatus().GetCondition(cond.Type); oldCond != nil {
+			cond.LastTransitionTime = oldCond.LastTransitionTime
+			if reflect.DeepEqual(&cond, &oldCond) {
```
vagababov, line 76: Why do we need to take addresses?

whaught: we do not

reconciler/reconcile_common.go
```diff
@@ -55,4 +60,26 @@ func PostProcessReconcile(ctx context.Context, resource duckv1.KRShaped) {
...
 		logger.Warn("A reconciler observed a new generation without updating the resource status")
 	}
+
+	groomConditionsTransitionTime(resource, oldResource)
+}
+
+// groomConditionsTransitionTime ensures that the LastTransitionTime only advances for resources
+// where the condition has changed during reconciliation. This also ensures that all advanced
+// conditions share the same timestamp.
+func groomConditionsTransitionTime(resource, oldResource duckv1.KRShaped) {
```
mattmoor, line 70: 🤔  maybe I'm misunderstanding, but iiuc the entire point of VolatileTime is to avoid what this is doing 😅

whaught: The VolatileTime strategy has a few weaknesses I've outlined in #1394

- Requires use of k8s specific comparison libraries
- The snapshot of now() is taken multiple times during reconciliation, but all written back at the same time. This can make it hard to tell which conditions are changed in a single reconciliation step (although this is optional, we could keep them).
- It fails to keep the LastTransitionTime stable if a single reconciliation sets conditions more than once (eg. flip-flopping but winds up in the same place).

This obviates the VolatileTime and condition manager strategy so that implementers don't really need to worry about how they are modifying the status struct - it just works.

mattmoor: Let me comment on the issue, but it may be worth discussing more synchronous Weds (put on the agenda for API WG?).

whaught: Yea let's just discuss in WG - that'll be easiest

mattmoor:
Yeah, this breaks forward edges, so it'll have to land in a few waves downstream.  Thanks @Cynocracy for doing this for networking 😍

mattmoor:
To clarify: manual changes aren't needed, but the sequencing of updates will need to follow the graph, so it'll likely take a couple applications of the automation before things are running smoothly again.

Cynocracy:
Gotcha, I made a manual change in networking and am now poking around to
see why it doesn't seem to regenerate on its own.

Cynocracy:
Ended up being  https://github.com/knative/pkg/issues/1287, I have a
potential fix that is very much in the spirit of the /hack/.

PR #1410, opened 2020-06-16
Add a passthrough method RegisterResourceView

jjzeng-seattle:
During the testing of stackdriver exporter with the dynamic labels in PR https://github.com/knative/pkg/pull/1392, we found out that we need to introduce a breaking change.

With the recorder extract the metric tags into resource labels, and send the measurement to the meter associated with the resource.  The problem is that currently `view.Register` is used, and the view only registered to the default meter, not the corresponding meter to the resource.  If we use `RegisterResourceView` instead, then it will work because it register the views to all meters.

If we merge that PR in its current form, we also need to fix all callers at the same time.  That will be too risky.  Therefore @evankanderson and I came up with this approach, by introducing the change up front.  After this one we will fix all callers(serving, eventing, eventing-contrib) to use this new method, so https://github.com/knative/pkg/pull/1392 won't be breaking change.

This will be replaced with the real implementation in https://github.com/knative/pkg/pull/1392.

### DIFF
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -85,13 +85,24 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
 	return ConfigMapWatcher(component, nil, logger)
 }
 
-// RegisterResourceView is placeholder for real implementation in https://github.com/knative/pkg/pull/1392.
+// RegisterResourceView is similar to view.Register(), except that it will
+// register the view across all Resources tracked by the system, rather than
+// simply the default view.
+// this is a placeholder for real implementation in https://github.com/knative/pkg/pull/1392.
 // That PR will introduce a breaking change, as we need to convert some view.Register to RegisterResourceView in
 // all callers, in serving, eventing, and eventing-contrib.
 // Since that PR is huge, a better approach is to introduce the breaking change up front, by creating this
 // passthrough method, and fixing all the callers before merging that PR.
 func RegisterResourceView(views ...*view.View) error {
 	return view.Register(views...)
+}
+
+// UnregisterResourceView is similar to view.unRegister(), except that it will
+// unregister the view across all Resources tracked by the system, rather than
+// simply the default view.
+// this is a placeholder for real implementation in https://github.com/knative/pkg/pull/1392.
+func UnregisterResourceView(views ...*view.View) {
+	view.Unregister(views...)
 }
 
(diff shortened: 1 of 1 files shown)


metrics/exporter.go
```diff
@@ -85,6 +85,15 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
 	return ConfigMapWatcher(component, nil, logger)
 }
 
+// RegisterResourceView is placeholder for real implementation in https://github.com/knative/pkg/pull/1392.
+// That PR will introduce a breaking change, as we need to  all view.Register to RegisterResourceView in
```
mattmoor, line 89: Remove trailing whitespace:
```suggestion
// That PR will introduce a breaking change, as we need to  all view.Register to RegisterResourceView in
```

mattmoor, line 89: Format Go code:
```suggestion
// That PR will introduce a breaking change, as we need to  all view.Register to RegisterResourceView in
```

metrics/exporter.go
```diff
@@ -85,6 +85,15 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
...
```
antoineco, line 89: > we need to  all view.Register to RegisterResourceView

Something is missing from that sentence

evankanderson:
```suggestion
// That PR will introduce a breaking change, as we need to convert some view.Register to RegisterResourceView in
```

We don't need to change all calls, only those that report on behalf of additional resources. (i.e. GC stats don't need to be changed, but per-Trigger, per-Broker, or per-Revision views will need to be updated.)

metrics/exporter.go
```diff
@@ -85,6 +85,15 @@ func UpdateExporterFromConfigMap(component string, logger *zap.SugaredLogger) fu
...
+// all callers, in serving, eventing, and eventing-contrib.
+// Since that PR is huge, a better approach is to introduce the breaking change up front, by creating this
+// passthrough method, and fixing all the callers before merging that PR.
```
evankanderson, line 92: Why not just copy the comment from `resource_view.go` in the other PR, and then add this implementation note as a comment inside the function?

```golang
// RegisterResourceView is similar to view.Register(), except that it will
// register the view across all Resources tracked by the system, rather than
// simply the default view.
```

PR #1411, opened 2020-06-16
Make Prometheus exporter port configurable via env

antoineco:
## Proposed Changes

- Add support for a new environment variable `METRICS_PROMETHEUS_PORT` that can override the default Prometheus port in the metrics exporter.

Rational: the default port 9090 clashes with Serving's queue proxy when a receive adapter is deployed as a Knative Service. There is currently no way to make that work without explicitly disabling metrics by setting the `K_METRICS_CONFIG` env var to `""`.

Second tentative after closing knative/eventing#3316.

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -67,6 +67,7 @@ const (
 	CollectorAddressKey = "metrics.opencensus-address"
 	CollectorSecureKey  = "metrics.opencensus-require-tls"
 
+	prometheusPortEnvName = "METRICS_PROMETHEUS_PORT"
 	defaultPrometheusPort = 9090
 	maxPrometheusPort     = 65535
 	minPrometheusPort     = 1024
@@ -217,11 +218,18 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
 	if mc.backendDestination == Prometheus {
 		pp := ops.PrometheusPort
 		if pp == 0 {
-			pp = defaultPrometheusPort
+			var err error
+			pp, err = prometheusPort()
+			if err != nil {
+				return nil, fmt.Errorf("failed to determine Prometheus port: %w", err)
+			}
 		}
+
 		if pp < minPrometheusPort || pp > maxPrometheusPort {
-			return nil, fmt.Errorf("invalid port %v, should between %v and %v", pp, minPrometheusPort, maxPrometheusPort)
+			return nil, fmt.Errorf("invalid port %d, should be between %d and %d",
+				pp, minPrometheusPort, maxPrometheusPort)
 		}
+
 		mc.prometheusPort = pp
 	}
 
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
/lgtm

That should solve your use case :wink:

antoineco:
I do hope I'm not the only person who deploys Knative Services that expose metrics via the Kn shared main :)

lionelvillard:
We need this for the mt github adapter.

lionelvillard:
related to https://github.com/knative/serving/issues/8047

metrics/config.go
```diff
@@ -327,6 +328,24 @@ import (
...
+// PrometheusPort returns the TCP port number configured via the environment
+// for the Prometheus metrics exporter if it's set, a default value otherwise.
+// No validation is performed on the port value.
+func PrometheusPort() int {
+	ppStr := os.Getenv(prometheusPortEnvName)
+	if ppStr == "" {
+		return defaultPrometheusPort
+	}
+
+	pp, err := strconv.Atoi(ppStr)
```
vagababov, line 340: You probably want more validation
1. strconv.ParseUint(x, 10, 16) — would catch all but 0.
2. 0 is a valid port in unix, but you won't be able to use generally, since system will assign a random port — so probably want to exclude that as well.

metrics/config_test.go
```diff
@@ -1006,3 +1031,19 @@ func TestStackdriverRecord(t *testing.T) {
...
+	if err := os.Setenv(name, val); err != nil {
+		t.Errorf("Failed to set env var %s: %v", name, err)
+	}
+
+	return func() {
+		if err := os.Unsetenv(name); err != nil {
+			t.Logf("Failed to unset env var %q: %s", name, err)
+		}
+	}
+}
```
vagababov, line 1049: honestly, I'd just go with
```
setenv
defer unsetenv
```
in the test. There is 0 chance this is likely to fail :)

metrics/config.go
antoineco: My intention was really just to read an int here. Not saying your idea is not good, but the validation happens in the caller right now and I didn't want to make this PR a bigger refactoring PR.

Would it be better if I just make this function private?

metrics/config_test.go
antoineco: The reason why I chose that approach is because the golang-ci linter dislikes unhandled errors.

If you feel strongly about it I can revert, but if we ever run linters in the CI those small quality improvements will matter.

metrics/config.go
vagababov: yeah, I don't see it being called elsewhere.
But you know changing strconv.Atoi to strconv.ParseUint is not a bigger refactoring :)
We can skip 0 check for existing code to take care of.

metrics/config_test.go
vagababov: eh, sure. :)

metrics/config.go
antoineco: Alright, let me revisit this. Good suggestion.

yanweiguo:
You can set `PrometheusPort` in the `ExporterOptions`, for example: https://github.com/knative/pkg/blob/099343d49e12915275bcd946fc97716f981aa6e2/metrics/exporter.go#L107

I'm curious why it doesn't work for your case.

antoineco:
@yanweiguo Because ConfigMap watchers do not expose the `ExporterOptions`, only the `ConfigMap` field can be modified by changing the actual ConfigMap data. Examples:

* [eventing/pkg/reconciler/source/config_watcher.go](https://github.com/knative/eventing/blob/master/pkg/reconciler/source/config_watcher.go)
* [pkg/injection/sharedmain/main.go](https://github.com/knative/pkg/blob/0d967ff0d8789d19096d03062e8c2fd961c95407/injection/sharedmain/main.go#L367-L379)

If I want to access the `ExporterOptions`, I have to write my own main, which defies the purpose of the shared main, especially when writing event sources.

metrics/config.go
yanweiguo: +1 to private func.

antoineco:
@vagababov I addressed both of your comments.

metrics/config.go
antoineco: Done
https://github.com/knative/pkg/blob/0ec08e2b0f1d605c10ca5eba3fed4523e12dca35/metrics/config.go#L338-L355

metrics/config_test.go
antoineco: Reverted in 0ec08e2b, since we don't run linters in Knative.
https://github.com/knative/pkg/blob/0ec08e2b0f1d605c10ca5eba3fed4523e12dca35/metrics/config_test.go#L620-L624

PR #1416, opened 2020-06-17
Add a ShouldEmitMetrics function and use it

Cynocracy:
Specifically, this is useful to transform the Metrics passed into kubemetrics.
Without this gating the emission of kubeclient metrics, we emit some custom
stackdriver metrics while we wait for the config to load / be applied.

### DIFF
--- a/controller/stats_reporter.go
+++ b/controller/stats_reporter.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"errors"
+	"net/url"
 	"time"
 
 	"go.opencensus.io/stats"
@@ -49,6 +50,31 @@ var (
 	keyTagKey        = tag.MustNewKey("key")
 	successTagKey    = tag.MustNewKey("success")
 )
+
+type configAwareLatencyMetric struct {
+	delegate kubemetrics.LatencyMetric
+}
+
+func (m configAwareLatencyMetric) Observe(verb string, u url.URL, latency time.Duration) {
+	if !metrics.ShouldEmitMetrics() {
+		return
+	}
+	m.delegate.Observe(verb, u, latency)
+}
+
+type configAwareResultMetric struct {
+	delegate kubemetrics.ResultMetric
+}
+
+func (m configAwareResultMetric) Increment(code string, method string, host string) {
+	if !metrics.ShouldEmitMetrics() {
+		return
+	}
+	m.delegate.Increment(code, method, host)
+}
+
+var _ kubemetrics.LatencyMetric = (*configAwareLatencyMetric)(nil)
+var _ kubemetrics.ResultMetric = (*configAwareResultMetric)(nil)
 
 func init() {
 	// Register to receive metrics from kubernetes workqueues.
@@ -156,7 +182,7 @@ func init() {
 			stats.UnitNone,
 		),
 	}
-	kubemetrics.Register(cp.NewLatencyMetric(), cp.NewResultMetric())
+	kubemetrics.Register(configAwareLatencyMetric{delegate: cp.NewLatencyMetric()}, configAwareResultMetric{delegate: cp.NewResultMetric()})
 
 	views := []*view.View{{
 		Description: "Depth of the work queue",
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"path"
 	"strconv"
@@ -151,16 +152,17 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
 
 // record applies the `ros` Options to each measurement in `mss` and then records the resulting
 // measurements in the metricsConfig's designated backend.
-func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) error {
-	if mc == nil {
-		// Don't record data points if the metric config is not initialized yet.
-		// At this point, it's unclear whether should record or not.
-		return nil
-	}
+func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) {
+	var err error = nil
 	if mc.recorder == nil {
-		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+		err = stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+	} else {
+		err = mc.recorder(ctx, mss, ros...)
 	}
-	return mc.recorder(ctx, mss, ros...)
+	if err != nil {
+		// TODO: DO NOT SUBMIT We were ignoring this
+		log.Print(err)
+	}
 }
 
 func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metricsConfig, error) {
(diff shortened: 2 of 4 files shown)


metrics/config.go
```diff
@@ -151,16 +152,17 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) {
+	var err error = nil
 	if mc.recorder == nil {
-		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+		err = stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+	} else {
+	        err = mc.recorder(ctx, mss, ros...)
+        }
+	if err != nil {
+	  // TODO: DO NOT SUBMIT We were ignoring this
```
mattmoor, line 163: Found "DO NOT SUBMIT".

metrics/config.go
```diff
@@ -151,16 +152,17 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
-		return nil
-	}
+func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) {
+	var err error = nil
 	if mc.recorder == nil {
-		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+		err = stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+	} else {
+	        err = mc.recorder(ctx, mss, ros...)
+        }
```
mattmoor, line 161: Format Go code:
```suggestion
		err = mc.recorder(ctx, mss, ros...)
	}
```

metrics/config.go
```diff
@@ -151,16 +152,17 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+	if err != nil {
+	  // TODO: DO NOT SUBMIT We were ignoring this
+	  log.Print(err)
```
mattmoor, line 164: Format Go code:
```suggestion
		// TODO: DO NOT SUBMIT We were ignoring this
		log.Print(err)
```

metrics/record_test.go
```diff
@@ -125,6 +125,9 @@ func testRecord(t *testing.T, measure *stats.Int64Measure, shouldReportCases []c
 			setCurMetricsConfig(test.metricsConfig)
 			Record(ctx, test.measurement)
 			metricstest.CheckLastValueData(t, test.measurement.Measure().Name(), map[string]string{}, test.measurement.Value())
+			if !ShouldEmitMetrics() {
+			   t.Fatal("Metrics were recorded when ShouldEmitMetrics() == false")
```
mattmoor, line 129: Format Go code:
```suggestion
				t.Fatal("Metrics were recorded when ShouldEmitMetrics() == false")
```

metrics/record_test.go
```diff
@@ -154,6 +157,9 @@ func testRecord(t *testing.T, measure *stats.Int64Measure, shouldReportCases []c
 			Record(ctx, test.measurement)
 			metricstest.CheckLastValueData(t, test.measurement.Measure().Name(), map[string]string{},
 				float64(len(shouldReportCases))) // The value is still the last one of shouldReportCases
+			if ShouldEmitMetrics() {
+			   t.Fatal("Metrics were NOT recorded when ShouldEmitMetrics() == true")
```
mattmoor, line 161: Format Go code:
```suggestion
				t.Fatal("Metrics were NOT recorded when ShouldEmitMetrics() == true")
```

metrics/config.go
Cynocracy: yup

metrics/config.go
```diff
@@ -151,16 +152,17 @@ func NewStackdriverClientConfigFromMap(config map[string]string) *StackdriverCli
...
+func (mc *metricsConfig) record(ctx context.Context, mss []stats.Measurement, ros ...stats.Options) {
+	var err error = nil
 	if mc.recorder == nil {
-		return stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+		err = stats.RecordWithOptions(ctx, append(ros, stats.WithMeasurements(mss...))...)
+	} else {
+		err = mc.recorder(ctx, mss, ros...)
+	}
+	if err != nil {
+		// TODO: DO NOT SUBMIT We were ignoring this
```
mattmoor, line 163: Found "DO NOT SUBMIT".

mattmoor, line 163: Found "DO NOT SUBMIT".

metrics/record.go
```diff
@@ -25,14 +25,24 @@ import (
 // TODO should be properly refactored and pieces should move to eventing and serving, as appropriate.
 // 	See https://github.com/knative/pkg/issues/608
 
+// Takes a function that would record a metric, and returns one that only does so if
+// the metric config has been configured.
+func ShouldEmitMetrics() bool {
```
vagababov, line 30: is this required to be public?

Cynocracy: I could move ConfigAware*Metric here, but then they'd need to be public. Or we'd need to add a public method "ConfigureKubemetrics" or something.

metrics/config.go
mattmoor, line 163: Found "DO NOT SUBMIT".

metrics/record.go
vagababov: But where is it used outside the package? Or it's for future use?

Cynocracy: It's used in controlller/stats_reporter.go, which I may have added after your first round of review.

controller/stats_reporter.go
```diff
@@ -50,6 +51,31 @@ var (
...
+	}
+	m.delegate.Observe(verb, u, latency)
+}
+
+type configAwareResultMetric struct {
+	delegate kubemetrics.ResultMetric
+}
+
+func (m configAwareResultMetric) Increment(code string, method string, host string) {
+	if !metrics.ShouldEmitMetrics() {
```
yanweiguo, line 70: Could you explain why calling `Increment` when configmap hasn't been loaded will result in unexpected custom metrics?

Cynocracy: I assume for the same reason that we don't emit metrics when we Record() when the config isn't loaded.

Cynocracy: https://github.com/knative/pkg/blob/3c828cf99f67f8066e956490f634ebe90abd6892/metrics/client.go#L33 doesn't seem to call Record.

I can make the method private if I move the configAwareProxy there, though, so I'll do that.

controller/stats_reporter.go
```diff
@@ -50,6 +51,31 @@ var (
 	successTagKey    = tag.MustNewKey("success")
 )
 
+type configAwareLatencyMetric struct {
+	delegate kubemetrics.LatencyMetric
+}
+
+func (m configAwareLatencyMetric) Observe(verb string, u url.URL, latency time.Duration) {
```
vagababov, line 58: should this be *URL?

Cynocracy:
After speaking with Yanwei, this may not be the explanation for what I was seeing, so closing this (it was likely caused by a separate cluster in my project and the fact that we can't filter custom metrics by cluster)

PR #1417, opened 2020-06-17
Make things private that should be so

vagababov:
Make some constants and enums private.
The ones I found to be used elsewhere, I kept public.

/assign @markusthoemmes @yanweiguo

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -40,32 +40,32 @@ const (
 	// The following keys are used to configure metrics reporting.
 	// See https://github.com/knative/serving/blob/master/config/config-observability.yaml
 	// for details.
-	AllowStackdriverCustomMetricsKey = "metrics.allow-stackdriver-custom-metrics"
+	allowStackdriverCustomMetricsKey = "metrics.allow-stackdriver-custom-metrics"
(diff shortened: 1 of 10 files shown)


metrics/config.go
```diff
@@ -40,32 +40,32 @@ const (
 	// The following keys are used to configure metrics reporting.
 	// See https://github.com/knative/serving/blob/master/config/config-observability.yaml
 	// for details.
-	AllowStackdriverCustomMetricsKey = "metrics.allow-stackdriver-custom-metrics"
+	allowStackdriverCustomMetricsKey = "metrics.allow-stackdriver-custom-metrics"
 	BackendDestinationKey            = "metrics.backend-destination"
-	ReportingPeriodKey               = "metrics.reporting-period-seconds"
+	reportingPeriodKey               = "metrics.reporting-period-seconds"
 	// Stackdriver client configuration keys
```
matzew, line 46: Are we fine w/ mixing private/public in one "block" ?

or should the be "ordered" (private/public and alphabetically?)

vagababov: :shrug: I can do that.

vagababov: ok sorted and reordered and split enum alone.

PR #1419, opened 2020-06-17
Reconcile the settings with net/http

vagababov:
No need to diverge from the timeout setting.
Also note that maxidle conns are also different from the upstream

/assign @tcnghia @julz


tcnghia:
Thanks @julz.  This looks good to me.

/unhold

PR #1421, opened 2020-06-18
Update config_test.go and metricstest to support multiple Meters.

evankanderson:
This is in preparation for #1392 ; it also cleans up some odd non-go-like test code in `metricstest` (though not all of it, because I didn't want to make a breaking change to the exposed interface).

/assign @yanweiguo


metrics/metricstest/metricstest.go
```diff
@@ -166,19 +181,29 @@ func lastRow(t test.T, name string) *view.Row {
...
-	t.Helper()
-	d, err := view.RetrieveData(name)
-	if err != nil {
-		t.Error("Reporter.Report() error", "metric", name, "error", err)
-		return nil
+func checkExactlyOneRow(t test.T, name string) (*view.Row, error) {
+	// view.Meter implements (and is exposed by) metricproducer.GetAll. Since
+	// this is a test, reach around and cast these to view.Meter.
+	var retval *view.Row
+	for _, producer := range metricproducer.GlobalManager().GetAll() {
```
yanweiguo, line 188: Will `GetAll` return only the default view currently?

evankanderson: Yes.

I didn't check this cast because this was test code, and I thought it was better to blow up than to try to handle it gracefully, but I can make this a graceful cast if you prefer.

evankanderson:
Fixed the build; not sure why it picked up the `go 1.13` mod info.

PR #1422, opened 2020-06-18
Set relevant golang env variables in update-codegen.sh.

markusthoemmes:
The same as https://github.com/knative/serving/blob/master/hack/update-codegen.sh

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -19,6 +19,12 @@ set -o nounset
 set -o pipefail
 
 export GO111MODULE=on
+# If we run with -mod=vendor here, then generate-groups.sh looks for vendor files in the wrong place.
+export GOFLAGS=-mod=
+
+if [ -z "${GOPATH:-}" ]; then
+  export GOPATH=$(go env GOPATH)
+fi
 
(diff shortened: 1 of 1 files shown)


hack/update-codegen.sh
```diff
@@ -19,6 +19,12 @@ set -o nounset
 set -o pipefail
 
 export GO111MODULE=on
+# If we run with -mod=vendor here, then generate-groups.sh looks for vendor files in the wrong place.
```
chizhg, line 22: Just out of curiosity, can you explain a bit more on this comment? Where would the `generate-groups.sh` look for vendor files if we have `-mod=vendor` here?

markusthoemmes: Honestly: Dunno. I only copied this in from Serving :joy:.

matzew: same in eventing: 
https://github.com/knative/eventing/blob/master/hack/update-codegen.sh#L22

evankanderson: It looks like we don't actually invoke `generate-groups.sh` in `pkg`, possibly because we don't include any concrete CRDs here.

PR #1423, opened 2020-06-18
Abort reconcilation on finalizer patch errors.

markusthoemmes:
This aborts reconcilation if finalizers could not be patched correctly with an error and thus it retries.

We shouldn't start a reconcilation if we haven't been able to correctly add the finalizer first. Otherwise we could get into a weird spot where the resources are created before the finalizer and in a very degenerate case the finalizer wouldn't even be executed. Also, the current code has a bug where if the patch fails the resource coming back from the patch (essentially an empty object) is passed into ReconcileKind, causing headaches.

The same imo goes for removing a finalizer. Failing to remove a finalizer blocks the resource forever, thus we should retry if that happens too.

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -147,6 +147,7 @@ func (g *reconcilerReconcilerGenerator) GenerateType(c *generator.Context, t *ty
 			Package: "context",
 			Name:    "Context",
 		}),
+		"fmtErrorf":           c.Universe.Package("fmt").Function("Errorf"),
 		"reflectDeepEqual":    c.Universe.Package("reflect").Function("DeepEqual"),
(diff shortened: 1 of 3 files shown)


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -330,7 +331,7 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		// and reconciled cleanly (nil or normal event), remove the finalizer.
 		reconcileEvent = fin.FinalizeKind(ctx, resource)
 		if resource, err = r.clearFinalizer(ctx, resource, reconcileEvent); err != nil {
-			logger.Warnw("Failed to clear finalizers", zap.Error(err))
+			return {{.fmtErrorf|raw}}("failed to clear finalizers: %w", err)
```
matzew, line 334: /lgtm

PR #1424, opened 2020-06-18
adding constant for annotation

### LINKED ISSUE #1407: Add constants for supported duck labels and annotations
**Problem**
We don't have any constant set for common labels and annotations such as
* duck.knative.dev/addressable
* duck.knative.dev/source
* ...

This can lead to repetition in the code base, especially in projects that use Knative as a dependency.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Contributors

**Exit Criteria**
Supported labels and annotations are defined as public constants inside `pkg/apis/duck`

Example: https://github.com/knative/serving/blob/3b5be63/pkg/apis/serving/register.go

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?
1/2

matzew:

fixes #1407

### DIFF
--- a/apis/duck/register.go
+++ b/apis/duck/register.go
@@ -24,6 +24,6 @@ const (
 	AddressableDuckVersionLabel = "duck.knative.dev/addressable"
 
 	// SourceDuckVersionLabel is the label we use to declare
-	// which Source duck version type we conform to.
+	// that a type conforms to the Source duck type.
 	SourceDuckVersionLabel = "duck.knative.dev/source"
 )
--- a/apis/duck/register.go
+++ b/apis/duck/register.go
@@ -20,7 +20,7 @@ const (
 	GroupName = "duck.knative.dev"
 
 	// AddressableDuckVersionLabel is the label we use to declare
-	// which Addressable duck version type we conform to.
+	// that a type conforms to the Addressable duck type.
 	AddressableDuckVersionLabel = "duck.knative.dev/addressable"
 
 	// SourceDuckVersionLabel is the label we use to declare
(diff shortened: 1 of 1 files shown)


apis/duck/register.go
```diff
@@ -18,4 +18,12 @@ package duck
 
 const (
 	GroupName = "duck.knative.dev"
+
+	// AddressableDuckVersionAnnotation is the annotation we use to declare
```
Abd4llA, line 22: I think it's a label?

antoineco: Agree, "Label" is a more term.

apis/duck/register.go
```diff
@@ -18,4 +18,12 @@ package duck
 
 const (
 	GroupName = "duck.knative.dev"
+
+	// AddressableDuckVersionLabel is the label we use to declare
+	// which Addressable duck version type we conform to.
```
antoineco, line 23: It's only used with the value `true` (could even be an empty string) so it's not about the version AFAIK. Or did I miss an update?
```suggestion
	// that a type conforms to the Addressable duck type.
```

matzew:
@mattmoor @vaikas mind taking a looksie here ? thanks!

apis/duck/register.go
```diff
@@ -18,4 +18,12 @@ package duck
...
+	AddressableDuckVersionLabel = "duck.knative.dev/addressable"
+
+	// SourceDuckVersionLabel is the label we use to declare
+	// which Source duck version type we conform to.
```
antoineco, line 27: Same comment.
```suggestion
	// that a type conforms to the Source duck type.
```

PR #1425, opened 2020-06-18
Fix gen-reconciler stub - remove TODOs

whaught:
- Remove the TODOs for ObservedGeneration and intializing conditions in the reconciler generated stub

  - These are not needed in ReconcileKind, they are handled outside of that in generated logic


client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/stub/reconciler.go
```diff
@@ -48,13 +48,9 @@ var _ customresourcedefinition.Interface = (*Reconciler)(nil)
 
 // ReconcileKind implements Interface.ReconcileKind.
 func (r *Reconciler) ReconcileKind(ctx context.Context, o *v1beta1.CustomResourceDefinition) reconciler.Event {
-	// TODO: use this if the resource implements InitializeConditions.
-	// o.Status.InitializeConditions()
```
n3wscott, line 52: niiiiccceee

PR #1429, opened 2020-06-19
Add tracecontextb3 HTTPFormats

Harwayne:
Helps with google/knative-gcp#1147.

### Changes
- Add a `tracecontextb3` package which contains three `propagation.HTTPFormat`s.
    - All formats read both TraceContext and B3 information during ingress, preferring TraceContext.
    - Formats differ in what they write for egress, based on their name.
        - `B3Egress` sends only b3.
        - `TraceContextEgress` sends only TraceContext
        - `TraceContextB3Egress` sends both.

---

This is being created because we found that serving relied on B3 tracing headers. While eventing relied on TraceContext tracing headers. So an HTTP request sent from a Broker to a Knative Service would break into two distinct traces.


ian-mi:
Maybe it would be better to only set the outgoing headers which are present in the incoming request? This could be a little tricky to implement but it should be possible to wrap the handler with one which inserts a format key into ctx. On the other hand there may be little downside to always writing both formats.

tracing/b3tracecontext/b3_traceparent.go
```diff
@@ -0,0 +1,53 @@
...
+	traceContext = &tracecontext.HTTPFormat{}
+)
+
+// HTTPFormat is a propagation.HTTPFormat that reads both B3 and TraceContext tracing headers,
+// preferring TraceContext. It will write both formats.
+type HTTPFormat struct{}
+
+var _ propagation.HTTPFormat = (*HTTPFormat)(nil)
+
+func (H *HTTPFormat) SpanContextFromRequest(req *http.Request) (trace.SpanContext, bool) {
```
vagababov, line 40:
```suggestion
func (h *HTTPFormat) SpanContextFromRequest(req *http.Request) (trace.SpanContext, bool) {
```

tracing/b3tracecontext/b3_traceparent.go
```diff
@@ -0,0 +1,53 @@
...
+	if sc, ok := b3Format.SpanContextFromRequest(req); ok {
+		return sc, true
+	}
+	return trace.SpanContext{}, false
+}
+
+func (H *HTTPFormat) SpanContextToRequest(sc trace.SpanContext, req *http.Request) {
+	traceContext.SpanContextToRequest(sc, req)
+	b3Format.SpanContextToRequest(sc, req)
+}
```
vagababov, line 53: so we set both headers?

Harwayne: Yes.

tracing/b3tracecontext/b3_traceparent.go
Harwayne: Removed entirely.

Harwayne:
> Maybe it would be better to only set the outgoing headers which are present in the incoming request? This could be a little tricky to implement but it should be possible to wrap the handler with one which inserts a format key into ctx. On the other hand there may be little downside to always writing both formats.

I think there is little downside in always writing both formats. Given that, I think writing a new Handler is overkill for this situation. 

And even if we did write that Handler, I think we would still need this (write both formats) propagator for situations where something other than HTTP was used as the input. For example an eventing piece reading from Kafka or Pub/Sub. A wrapping handler would not help, as we don't have an incoming HTTP request. We would need to wrap an `ochttp.Transport`. It seems far easier, and with little downside, to just use the same propagator for both.

tracing/b3tracecontext/b3_traceparent.go
```diff
@@ -0,0 +1,55 @@
...
+
+// The propagation.HTTPFormats that this is built on.
+var (
+	b3Format     = &b3.HTTPFormat{}
+	traceContext = &tracecontext.HTTPFormat{}
+)
+
+// HTTPFormat is a propagation.HTTPFormat that reads both B3 and TraceContext tracing headers,
+// preferring TraceContext. It will write both formats.
+type HTTPFormat struct{}
```
ian-mi, line 36: I would prefer the more general type HTTPFormats []propagation.HTTPFormat. This would make it easier to change the supported formats or change format preference.

Harwayne: Done.

slinkydeveloper:
Yeah i think that could be interesting for kafka tracing too!

/assign

tracing/propagation/http_format_sequence.go
```diff
@@ -0,0 +1,63 @@
...
+		&b3.HTTPFormat{},
+	},
+}
+
+// HTTPFormatSequence is a propagation.HTTPFormat that applies multiple other propagation formats.
+// For incoming requests, it will use the first SpanContext it can find, checked in the order of
+// HTTPFormatSequence.Formats.
+// For outgoing requests, it will apply all the formats to the outgoing request, in the order of
+// HTTPFormatSequence.Formats.
+type HTTPFormatSequence struct {
```
ian-mi, line 42: nit: The struct isn't necessary here.

Harwayne: Done.

Harwayne:
> Similar to the issue @ian-mi raised, I don't get why we should write always in both formats, to me it sounds pretty overkill (and adds unnecessary overhead to the message). We could make it configurable for the egress maybe? Like b3/tracecontext/both and default to tracecontext? I would prefer to stick with one format on egress side, and support both on ingress.

My main hesitation of only outputting one format is that we are more likely to hit a component that speaks the other form of tracing, thus stopping the existing trace. That is what is happening in Knative today (before this PR). Every hop between eventing and serving cuts the trace because they use different tracing formats. Once this PR is in and all the downstream repos are updated and the user has updated to all those new versions, then it would no longer be a problem for Knative itself. But I'm worried about any software downstream of Knative that happens to rely on the existing format. 

If we go the route of configurable egress, then I think we need to have serving use B3 output and eventing use TraceContext output, to match what is already present in both places. As both may already have downstream software that relies on it.

I think the additional overhead on the message is small enough (either ~70 or ~100 bytes), that its existence is justified by the increased chance of interoperating with other software.

> Also, before merging this one, can you run the e2e tracing tests on eventing with this patch?

Will do.

slinkydeveloper:
> Every hop between eventing and serving cuts the trace because they use different tracing formats. 

To me this sounds like a different problem, because we don't have an alignment between their tracing propagation and our tracing propagation. Also as far as I understood the b3 propagation is specific to zipkin (which, by the way, can understand also traces propagated with trace context), so that's another reason for me to prefer making it configurable (same on serving).

Also as far as I understood we should make the `config-tracing` the unique source of truth both for serving and eventing, and configure on it the format to use. So all the knative installation will use the same format, and user won't have to worry about which one uses which format. This way, we won't need to use both formats but we can just stick with one of the two (and eventually, as opt-in feature, emit both)

> I think the additional overhead on the message is small enough (either ~70 or ~100 bytes)

Well, it's always additional bytes that are potentially not used downstream.

Harwayne:
> > Every hop between eventing and serving cuts the trace because they use different tracing formats.
> 
> To me this sounds like a different problem, because we don't have an alignment between their tracing propagation and our tracing propagation. Also as far as I understood the b3 propagation is specific to zipkin (which, by the way, can understand also traces propagated with trace context), so that's another reason for me to prefer making it configurable (same on serving).
> 
> Also as far as I understood from the problem, we should make the `config-tracing` the unique source of truth both for serving and eventing, and configure on it the format to use. So all the knative installation will use the same format, and user won't have to worry about which one uses which format. This way, we won't need to use both formats but we can just stick with one of the two (and eventually, as opt-in feature, emit both)

Makes sense. This PR will lay the ground work for allowing that option. This repo will need to extend the `config-tracing` format to include that option. Each downstream option will need to ensure all places that trace have the value plumbed through. 

> > I think the additional overhead on the message is small enough (either ~70 or ~100 bytes)
> 
> Well, it's always additional bytes that are potentially not used downstream.

Done. Although I still believe it is better to send the extra bytes than to have the trace broken.

PR #1430, opened 2020-06-19
Expect resources are krshaped by default

whaught:
- Switches to always generating krshapedlogic by default
    -krshapedlogic=false can be used to disable

/hold will test-integrate this with downstream
Tested and waiting on:
https://github.com/knative/eventing/pull/3356
https://github.com/knative/serving/pull/8388

### DIFF
--- a/codegen/cmd/injection-gen/generators/packages.go
+++ b/codegen/cmd/injection-gen/generators/packages.go
@@ -201,7 +201,7 @@ func isKRShaped(tags map[string]map[string]string) bool {
 	if !has {
 		return false
 	}
-	return vals["krshapedlogic"] == "true"
+	return vals["krshapedlogic"] != "false"
 }
 
 func isNonNamespaced(tags map[string]map[string]string) bool {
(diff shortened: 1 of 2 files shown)


codegen/cmd/injection-gen/generators/packages.go
```diff
@@ -201,7 +201,7 @@ func isKRShaped(tags map[string]map[string]string) bool {
 	if !has {
 		return false
 	}
-	return vals["krshapedlogic"] == "true"
```
n3wscott, line 204: the gotchya here is we generate the reconcilers for things like namespace and unless the generated code is also testing the shape of the thing, we are gonna break the namespace and endpoints reconcilers. We don't and can't update the annotations in that code. 

See https://github.com/knative/pkg/blob/master/hack/update-codegen.sh#L52

and look at the logic behind `--force-genreconciler-kinds`

I am not opposed to to always generating the krshaped integrations. We might need some extra guards on the controller side, and forgive me if they are already there.

n3wscott: it looks like the diff above has no real changes, so seems we are ok. Or the `krshapedlogic` code ignores the forced path.

Issue #1427, opened 2020-06-18
Why not use coordination.Lease v1 API exclusively for leader election locks


dprotaso:
https://github.com/knative/pkg/blob/d5043a43329217814d1e3e2095c969ef890120df/leaderelection/config.go#L33

I saw the leader election allows using different resource locks - I was curious why this needs to be exposed to operators and why we don't just use coordination.Lease v1 exclusively.

/assign @pmorie

dprotaso:
Found out that client-go lease locks for leader election weren't around until 1.16.

So we should be able to switch to leases exclusively for our next release since we've bump our k8s deps to 1.17.6

dprotaso:
Meaning we should be good to exclusively use leases only

/unassign pmorie
/good-first-issue

PR #1439, opened 2020-06-23
use coordination.Leases exclusively for leader election

dprotaso:
Prior locks were for K8s version <1.14

Fixes: #1427

I left the plumbing of the lock setting through everything since there may be a new lock in the future we would want to switch to - if folks feel like this is unnecessary let me know and I can rip it out

### DIFF
--- a/leaderelection/config.go
+++ b/leaderelection/config.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 
 	cm "knative.dev/pkg/configmap"
 )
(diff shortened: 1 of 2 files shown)


markusthoemmes:
Is this a "breaking" change? Do we need to check if anybody is using something different than leases first?

dprotaso:
> Is this a "breaking" change? Do we need to check if anybody is using something different than leases first?

Technically this feature is alpha so breaking changes are allowed

mattmoor:
According to @pmorie (in slack) the Operator doesn't reference this field, and the docs I just updated don't mention configuring this.  I believe we've also established that all supported K8s versions have had the coordination API (probably since the inception of this feature).

I agree that this is breaking, but also that it's alpha and now's the time to make breaking changes.  @markusthoemmes beyond this simply being available in prior releases, do you have reason to think this might present problems?

markusthoemmes:
Nope, just pointing out it's potentially breaking and making sure we though about it. That seems to be the case :yum: . I got no objections beyond that.

dprotaso:
cool - can someone approve/lgtm this change and I'll go to the various repos and drop the lock types from the configs

PR #1441, opened 2020-06-23
Ingore files under benchmark folder

chizhg:
Do not consider files under the benchmark folder as benchmarks.
This PR will fix the Prow job errors after `.ko.yaml` was added to the `benchmarks` folder in serving, see https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-recreate-clusters/1275330518623195136

/cc @vagababov


vagababov:
Did it try to create a benchmark with `.ko.data` name?

chizhg:
> Did it try to create a benchmark with `.ko.data` name?

No, it only tried to create a cluster with the `serving--.ko.yaml` name.

PR #1442, opened 2020-06-23
remove krshapedlogic tag, this is default now

whaught:
- remove krshapedlogic tag, this is default now

### DIFF
--- a/apis/test/example/v1alpha1/foo_types.go
+++ b/apis/test/example/v1alpha1/foo_types.go
@@ -27,7 +27,7 @@ import (
 )
 
 // +genclient
-// +genreconciler:krshapedlogic=true
+// +genreconciler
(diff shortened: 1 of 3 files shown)


tom24d:
Is this safe? I see generated code on eventing-contrib is still not krshaped enabled, as well as not enabled on tekton side https://github.com/tektoncd/pipeline/pull/2807.

matzew:
well, I updated the contrib, and set it as disabled - we can pick by source, no ?

tom24d:
It is safe if we use false tag. sorry for interrupting 🤭

whaught:
the false tag is safe, and should generate the same as before when the tag was omitted. I was trying to discover which downstreams were missed. I will follow up and investigate eventing-contrib.

PR #1444, opened 2020-06-24
Return TypeMeta as part of the admission response

dprotaso:
Unblocks https://github.com/knative/serving/pull/8461

This has similar semantics to conversion reviews.


dprotaso:
/hold

just validating in that serving PR

PR #1447, opened 2020-06-25
Use TraceContextB3Egress for spoof client

nak3:
Serving's TestShouldHaveHeadersSet expects `x-b3-spanid` and
`x-b3-traceid` headers from client, but it was removed by https://github.com/knative/pkg/commit/5658d93fb07ee641a4e0155f3f9c262da7199035.

Due to this, non-Istio ingress like Kourier and Contour started failing the conformance test:

https://testgrid.knative.dev/serving#kourier-stable&include-filter-by-regex=TestShouldHaveHeadersSet
https://testgrid.knative.dev/serving#contour-latest&include-filter-by-regex=TestShouldHaveHeadersSet

So this patch changes to use TraceContextB3Egress for spoof client.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -135,7 +135,7 @@ func New(
 	// Enable Zipkin tracing
 	roundTripper := &ochttp.Transport{
 		Base:        transport,
-		Propagation: tracecontextb3.TraceContextEgress,
+		Propagation: tracecontextb3.TraceContextB3Egress,
 	}
 
 	sc := SpoofingClient{


matzew:
I think that @slinkydeveloper has some good tracing experience ;-)

nak3:
/cc @Harwayne 

Would you mind taking a look?

nak3:
@Harwayne gentle reminder. Would you mind taking a look?

nak3:
@mattmoor @Harwayne  Could you please take a look?

The new change in https://github.com/knative/pkg/commit/5658d93fb07ee641a4e0155f3f9c262da7199035 removed `&b3.HTTPFormat{}` from spoof client.
But serving's conformance test (`TestShouldHaveHeadersSet`) still expects the tracing headers:

https://github.com/knative/serving/blob/bb1b0d3b48934614f731c9990f3bc5ab629bed03/test/conformance/runtime/header_test.go#L86-L92

Can we add back the header to spoof client? Or we should update the serving's conformance test?

nak3:
ping @Harwayne

PR #1450, opened 2020-06-25
[release-0.14] Have the webhook responses return TypeMeta

dprotaso:
This is required for v1 APIs and it's blocking smooth upgrades from 0.14 to 0.16

knative/eventing#3409


dprotaso:
Actually - we don't really test upgrades from N-2 so I'm going to close this

PR #1451, opened 2020-06-27
Enable use of StatefulSet in sharedmain.

mattmoor:
This change allows for (just webhook for now) controllers going through sharedmain to opt into Yanwei's logic by setting several environment variables.

I was able to pull this change in downstream and change the webhook to use a StatefulSet with the following environment:
```
+        # These settings are used for statefulset-based
+        # leader selection.
+        - name: CONTROLLER_ORDINAL
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: STATEFUL_SERVICE_NAME
+          value: "webhook"
```

Running the above with 10 replicas and 10 buckets worked as intended (keys were evenly distributed across the replicas).

### DIFF
--- a/leaderelection/config.go
+++ b/leaderelection/config.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/kelseyhightower/envconfig"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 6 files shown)


leaderelection/config.go
```diff
@@ -129,9 +130,22 @@ type ComponentConfig struct {
...
-	Port            string
-	Protocol        string
+	ServiceName     string `envconfig:"STATEFUL_SERVICE_NAME" required:"true"`
+	Port            string `envconfig:"STATEFUL_SERVICE_PORT" default:"80"`
+	Protocol        string `envconfig:"STATEFUL_SERVICE_PROTOCOL" default:"http"`
+}
+
+func NewStatefulSetConfig() (*StatefulSetConfig, error) {
+	var ssc StatefulSetConfig
+	ssn, _, err := ParseControllerOrdinal()
```
vagababov, line 140: This seems strange we parse the value manually and delegate the rest to the envconfig.
It seems envconfig has methods for fancier parsing, which we might use, but this can be done later.

yanweiguo: I wonder whether we still need `StatefulSetName` in `StatefulSetConfig`. It's only used in `statefulSetBuilder. BuildElector`, where `ControllerOrdinal` is called. However `ControllerOrdinal` calls `ParseControllerOrdinal` which returns stateful set name as well. I think we can just use `ParseControllerOrdinal`.

PR #1452, opened 2020-06-27
Some cleanups around LE

vagababov:
- make funcs and types not expected to be used outside private (we can always make them public, should the need arise)
- add some more checks in tests
- fix some strings

/assign @yanweiguo @mattmoor

### DIFF
--- a/leaderelection/config.go
+++ b/leaderelection/config.go
@@ -145,9 +145,11 @@ func (ssID *statefulSetID) Decode(v string) error {
 	return fmt.Errorf("%q is not a valid stateful set controller ordinal", v)
 }
 
+var _ envconfig.Decoder = (*statefulSetID)(nil)
+
 // statefulSetConfig represents the required information for a StatefulSet service.
 type statefulSetConfig struct {
-	StatefulSetID statefulSetID `envconfig:"CONTROLLER_ORDINAL" required:"true"`
+	StatefulSetID statefulSetID `envconfig:"STATEFUL_CONTROLLER_ORDINAL" required:"true"`
 	ServiceName   string        `envconfig:"STATEFUL_SERVICE_NAME" required:"true"`
 	Port          string        `envconfig:"STATEFUL_SERVICE_PORT" default:"80"`
 	Protocol      string        `envconfig:"STATEFUL_SERVICE_PROTOCOL" default:"http"`
--- a/leaderelection/config.go
+++ b/leaderelection/config.go
@@ -19,6 +19,8 @@ package leaderelection
 import (
 	"fmt"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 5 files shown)


leaderelection/config.go
```diff
@@ -127,25 +129,37 @@ type ComponentConfig struct {
...
+	ssName  string
+	ordinal int
 }
 
-func NewStatefulSetConfig() (*StatefulSetConfig, error) {
-	var ssc StatefulSetConfig
-	ssn, _, err := ParseControllerOrdinal()
-	if err != nil {
-		return nil, err
+func (ssID *statefulSetID) Decode(v string) error {
```
mattmoor, line 138: Wanna add `var _ envconfig.Decoder = (*statefulSetID)(nil)`?

leaderelection/config.go
```diff
@@ -127,25 +129,37 @@ type ComponentConfig struct {
...
+		return err
 	}
-	ssc.StatefulSetName = ssn
-	if err := envconfig.Process("", &ssc); err != nil {
+	return fmt.Errorf("%q is not a valid stateful set controller ordinal", v)
+}
+
+// statefulSetConfig represents the required information for a StatefulSet service.
+type statefulSetConfig struct {
+	StatefulSetID statefulSetID `envconfig:"CONTROLLER_ORDINAL" required:"true"`
```
mattmoor, line 150: I wonder if we should make this `STATEFUL_ORDINAL` for consistency? 🤔

leaderelection/ordinal.go
```diff
@@ -27,20 +27,17 @@ import (
 // should be set to the pod name via the downward API.
 const controllerOrdinalEnv = "CONTROLLER_ORDINAL"
 
-// ParseControllerOrdinal returns the parts of the StatefulSet's pod name.
-func ParseControllerOrdinal() (string, uint64, error) {
-	v := os.Getenv(controllerOrdinalEnv)
+// ControllerOrdinal tries to get ordinal from the pod name of a StatefulSet,
+// which is provided from the environment variable CONTROLLER_ORDINAL.
+func ControllerOrdinal() (int, error) {
```
mattmoor, line 32: you really hate `uint` 🤣 

Why does this have to be public still?

leaderelection/config.go
vagababov: Yeah, it was a sore in my eye a well. let's do it. Went for `STATEFUL_CONTROLLER_ORDINAL`.

leaderelection/ordinal.go
```diff
@@ -27,20 +27,17 @@ import (
...
+	v, ok := os.LookupEnv(controllerOrdinalEnv)
+	if !ok {
+		return 0, fmt.Errorf("%s envvar is not set", controllerOrdinalEnv)
+	}
```
mattmoor, line 36: Can we dedup the implementation with `newStatefulSetConfig`?

leaderelection/context.go
```diff
@@ -53,10 +56,10 @@ func WithStandardLeaderElectorBuilder(ctx context.Context, kc kubernetes.Interfa
 	})
 }
 
-// WithStatefulSetLeaderElectorBuilder infuses a context with the ability to build
+// withStatefulSetElectorBuilder infuses a context with the ability to build
 // Electors which are assigned leadership based on the StatefulSet ordinal from
 // the provided component configuration.
-func WithStatefulSetLeaderElectorBuilder(ctx context.Context, cc ComponentConfig, ssc StatefulSetConfig) context.Context {
+func withStatefulSetElectorBuilder(ctx context.Context, cc ComponentConfig, ssc statefulSetConfig) context.Context {
```
yanweiguo, line 62: I think we might still want to make this public as autoscaler binary is not using the sharedmain. And only StatefulSet mode works for autoscaler. Falling back to Standard mode via `WithDynamicLeaderElectorBuilder` doesn't work in autoscaler case.

leaderelection/context.go
```diff
@@ -36,10 +36,13 @@ import (
 // WithDynamicLeaderElectorBuilder sets up the statefulset elector based on environment,
 // falling back on the standard elector.
 func WithDynamicLeaderElectorBuilder(ctx context.Context, kc kubernetes.Interface, cc ComponentConfig) context.Context {
-	ssc, err := NewStatefulSetConfig()
+	logger := logging.FromContext(ctx)
+	ssc, err := newStatefulSetConfig()
 	if err == nil {
-		return WithStatefulSetLeaderElectorBuilder(ctx, cc, *ssc)
+		logger.Info("Running with StatefulSet leader election")
```
yanweiguo, line 42: I think the logs here are redundant as in `buildElector` we have similar logs.

leaderelection/ordinal.go
vagababov: done.

leaderelection/context.go
vagababov: I talked with Matt about it a bit. We can just call WithDynaminLeaderElectorBuilder and make it fail if SS builder is expected.

leaderelection/context.go
vagababov: 🤷  it's once per start log, so should be ok.

leaderelection/ordinal.go
vagababov: autoscaler will call it per @yanweiguo  :)

Issue #1426, opened 2020-06-18
Need to add json tag in VolatileTime struct


mattcui:
## Expected Behavior
Our controller is based on operator-sdk, and we use duck condition in our controller. We expect to upgrade to operator-sdk 0.17.1 without any problem.

## Actual Behavior
But it failed at `operator-sdk generate crds` step
```
operator-sdk generate crds
INFO[0000] Running CRD generator.
/usr/local/go/src/time/time.go:139:2: encountered struct field "wall" without JSON tag in type "Time"
/usr/local/go/src/time/time.go:140:2: encountered struct field "ext" without JSON tag in type "Time"
/usr/local/go/src/time/time.go:147:2: encountered struct field "loc" without JSON tag in type "Time"
/Users/mattcui/myspace/tenant-domain-controller/pkg/mod/knative.dev/pkg@v0.0.0-20191107185656-884d50f09454/apis/volatile_time.go:27:2: encountered struct field "Inner" without JSON tag in type "VolatileTime"
FATA[0001] error generating CRDs from APIs in pkg/apis: error generating CRD manifests: error running CRD generator: not all generators ran successfully
```

## How to fix
I think an easy way to fix it is to add json tag for VolatileTime struct at https://github.com/knative/pkg/blob/master/apis/volatile_time.go#L32, it should be changed to 

```
type VolatileTime struct {
	Inner metav1.Time `json:"inner"`
}
```

I could submit a PR for it. Thanks.

whaught:
I think adding the json tag will work for your purpose. That said I have a related PR to remove VolatileTime which will also make this go away, but there's a little bit of legwork to get that done. If it's urgent we can just add in the json tag.

whaught:
This struct uses a custom marshal/unmarshal override to inline this field, what are you using that is expecting every struct field to contain a json tag? Your suggesting would name the field 'inner' but we are expecting this type to be transparently omitted from json.

It would be nice to be able to have `json:",inline"` but I don't think that's a mainline go feature

mattcui:
@whaught I think it should work for me if you add inline json tag. Thanks.

raushan2016:
We also have similar requirement where we want to leverage the `Status`. We have a controller generated using kubebuilder 2.3 

```
import (
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	duckv1 "knative.dev/pkg/apis/duck/v1"
)

[...]

PR #1453, opened 2020-06-29
[WIP/DO NOT REVIEW] Add inline json tag to VolatileTime inner in conditions stuct

whaught:
https://github.com/knative/pkg/issues/1426

-Adds inline to the .inner structure for VolatileTime
     - Note marshall/unmarshall are implemented already to bypass the default, but this will help users who may be using the tags for codegen


whaught:
This would break downstream too much. We should have a discussion on pkg versioning someday to do this sort of thing.

PR #1455, opened 2020-06-30
bump k8s min version to 1.16.0

dprotaso:


mattmoor:
😬 I thought we landed this weeks ago 😲

dprotaso:
Ideally we'd bump to 1.16.9 but that it's not available on EKS

https://github.com/knative/serving/issues/6984#issuecomment-610984091

PR #1456, opened 2020-06-30
Add an option to skip automated status updates in a reconciler.

markusthoemmes:
This option is necessary to be able to create reconcilers like Serving's labeler, that is purely adding labels to resources. If that fails, the new automated observed generation handling changes the status and that gets written to the API currently, which is not desired.

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
@@ -108,6 +108,10 @@ type reconcilerImpl struct {
 
 	// finalizerName is the name of the finalizer to reconcile.
 	finalizerName string
+
+	// skipStatusUpdates configures whether or not this reconciler automatically updates
+	// the status of the reconciled resource.
+	skipStatusUpdates bool
 }
 
 // Check that our Reconciler implements controller.Reconciler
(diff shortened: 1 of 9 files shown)


client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
```diff
@@ -254,20 +258,22 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
-		// to status with this stale state.
-	} else if !isLeader {
-		logger.Warn("Saw status changes when we aren't the leader!")
-		// TODO: Consider logging the diff at Debug?
-	} else if err = r.updateStatus(original, resource); err != nil {
-		logger.Warnw("Failed to update resource status", zap.Error(err))
-		r.Recorder.Eventf(resource, corev1.EventTypeWarning, "UpdateFailed",
-			"Failed to update status for %q: %v", resource.Name, err)
-		return err
+	if r.skipStatusUpdates {
```
mattmoor, line 261: the bool is flipped

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -426,20 +430,22 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
-		// to status with this stale state.
-	} else if !isLeader {
-		logger.Warn("Saw status changes when we aren't the leader!")
-		// TODO: Consider logging the diff at Debug?
-	} else if err = r.updateStatus(original, resource); err != nil {
-		logger.Warnw("Failed to update resource status", zap.Error(err))
-		r.Recorder.Eventf(resource, {{.corev1EventTypeWarning|raw}}, "UpdateFailed",
-			"Failed to update status for %q: %v", resource.Name, err)
-		return err
+	if r.skipStatusUpdates {
```
mattmoor, line 433: bool is flipped (commenting again on visible file)

markusthoemmes: ugh, rushing...

client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
```diff
@@ -254,20 +258,22 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
-		// to status with this stale state.
-	} else if !isLeader {
-		logger.Warn("Saw status changes when we aren't the leader!")
-		// TODO: Consider logging the diff at Debug?
-	} else if err = r.updateStatus(original, resource); err != nil {
-		logger.Warnw("Failed to update resource status", zap.Error(err))
-		r.Recorder.Eventf(resource, corev1.EventTypeWarning, "UpdateFailed",
-			"Failed to update status for %q: %v", resource.Name, err)
-		return err
+	if !r.skipStatusUpdates {
```
whaught, line 261: I wonder if this would read better as:

```
switch {
case !r.skipStatusUpdates:
   // The reconciler has explicitly opted-out of updates
case !isLeader:
   // This reconciler is on an observing replica. Only the leader-elected pod makes updates.
case equality.Semantic.DeepEqual(original.Status, resource.Status):
   // No write-back to make.
default:
  if err = r.updateStatus(original, resource); err != nil {
  }
}
```

PR #1457, opened 2020-06-30
Update generated reconciler code with a switch

whaught:
Slight readability update to https://github.com/knative/pkg/pull/1456

I think this better highlights the checks we do before enacting the status writeback.
Added comments for the additional cases.

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -435,15 +435,15 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 	case r.skipStatusUpdates:
 		// This reconciler implementation is configured to skip resource updates.
 		// This may mean this reconciler does not observe spec, but reconciles external changes.
-	case !isLeader:
-		// High-availability reconcilers may have many replicas watching the resource, but only
-		// the elected leader is expected to write modifications.
-		logger.Warn("Saw status changes when we aren't the leader!")
 	case {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status):
 		// If we didn't change anything then don't call updateStatus.
 		// This is important because the copy we loaded from the injectionInformer's
 		// cache may be stale and we don't want to overwrite a prior update
 		// to status with this stale state.
+	case !isLeader:
+		// High-availability reconcilers may have many replicas watching the resource, but only
+		// the elected leader is expected to write modifications.
+		logger.Warn("Saw status changes when we aren't the leader!")
 	default:
 		if err = r.updateStatus(original, resource); err != nil {
 			logger.Warnw("Failed to update resource status", zap.Error(err))
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -430,17 +430,22 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 		reconcileEvent = rof.ObserveFinalizeKind(ctx, resource)
 	}
 
-	if !r.skipStatusUpdates {
-		// Synchronize the status.
-		if {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status) {
-			// If we didn't change anything then don't call updateStatus.
-			// This is important because the copy we loaded from the injectionInformer's
-			// cache may be stale and we don't want to overwrite a prior update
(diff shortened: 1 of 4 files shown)


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -430,17 +430,22 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
-			// If we didn't change anything then don't call updateStatus.
-			// This is important because the copy we loaded from the injectionInformer's
-			// cache may be stale and we don't want to overwrite a prior update
-			// to status with this stale state.
-		} else if !isLeader {
-			logger.Warn("Saw status changes when we aren't the leader!")
-			// TODO: Consider logging the diff at Debug?
-		} else if err = r.updateStatus(original, resource); err != nil {
+	// Synchronize the status.
+	switch {
```
n3wscott, line 434: I would change the order of these to the following:

```
	case !isLeader:
	case r.skipStatusUpdates:
	case {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status):
```

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -430,17 +430,22 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
+		// This may mean this reconciler does not observe spec, but reconciles external changes.
+	case {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status):
+		// If we didn't change anything then don't call updateStatus.
+		// This is important because the copy we loaded from the injectionInformer's
+		// cache may be stale and we don't want to overwrite a prior update
+		// to status with this stale state.
+	case !isLeader:
+		// High-availability reconcilers may have many replicas watching the resource, but only
+		// the elected leader is expected to write modifications.
+		logger.Warn("Saw status changes when we aren't the leader!")
```
n3wscott, line 446: this dropped the todo comment: ```// TODO: Consider logging the diff at Debug?``` maybe just zap debug log the diff? idk

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -430,17 +430,22 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
...
-		} else if !isLeader {
-			logger.Warn("Saw status changes when we aren't the leader!")
-			// TODO: Consider logging the diff at Debug?
-		} else if err = r.updateStatus(original, resource); err != nil {
+	// Synchronize the status.
+	switch {
+	case r.skipStatusUpdates:
+		// This reconciler implementation is configured to skip resource updates.
+		// This may mean this reconciler does not observe spec, but reconciles external changes.
+	case {{.equalitySemantic|raw}}.DeepEqual(original.Status, resource.Status):
```
n3wscott, line 438: I moved this to the last case to attempt to avoid the cost of the DeepEqual on an object we don't care if we are not the leader or want to skip status updates.

n3wscott:
/lgtm
/approve
/hold

Hold for nit on switch ordering to try to move some expensive logic out of the  short circuit path.

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
whaught: I had it that way originally thinking the deepEqual is computationally more expensive, but that actually changes the functionality.

Doing so would print: `logger.Warn("Saw status changes when we aren't the leader!")` for all non-leaders all the time and is expecting the equality check to happen first

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
n3wscott: they are dependent. Ignore me.

n3wscott:
/unhold
I did not understand the ordering need. 
/lgtm

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
mattmoor: We should warn when folks are mutating status and shouldn't be.  If it is noisy, then good 😅

whaught: Right but if we made scotts' suggestion it would warn even on no-change (which is not what we want)

mattmoor: Right!

/approve

PR #1463, opened 2020-07-01
Auto-update deps when --upgrade is provided.

### LINKED ISSUE #1462: Knative doesn't pull in security fixes
## Expected Behavior
Running update-deps.sh --upgrade pulls in minor version updates for all dependencies (or at least patches)

## Actual Behavior
Dependencies are only updated if strictly necessary by the use of new packages.

1. Find a dependency with a security patch recently
2. Run update deps
3. Confused pikachu face
4. Realize you need to manually pull in the patch.

Cynocracy:
This is preferable to requiring devs to manually curate all
dependencies, but has some unfortunate fallout in that we
will need to pin all dependencies with breaking changes
and not just those that are updated by test-infra changes.

Fixes #1462


Cynocracy:
Ugh, I definitely just messed this up, one moment.

Cynocracy:
/assign @evankanderson
:wave: :smiley:

n3wscott:
We currently use the `--upgrade` to mean just Knative components.

Cynocracy:
Ah, I see.

Do we routinely update deps outside of the autogen PRs?
I'm a little nervous only any process that requires a human to remember to intervene.

Maybe a different flag could be used to avoid confusion? --update (alias for --update=knative) and --update=all, which we can use from the PR generator?

Cynocracy:
boo

evankanderson:
Sorry for the delayed comment; I think Scott addressed the "current usage" a bit, but I'd like to also talk about the *why* of the current setup.

I think most of the following should be non-controversial:

1. We'd like to pick up fixes in our dependencies
1. Changes that do one focused thing are better than changes that do many things
1. Humans are bad at following rules (distraction, laziness, confusion, never learned them, etc)

Given item 2, I think we want the standard flow for developers to avoid "sometimes picking up a ball of updates in unrelated libraries". Given item 3 and item 1, I think we want automation for bumping dependencies that we don't explicitly hold back, since doing otherwise requires a human to pay attention to the repos (and our testing should allow us to upgrade other dependencies). Given 2 and 3, I think the default tool behavior we have now that pulls items from Knative but not elsewhere is probably correct.

I think we need a new flag (`--upgrade-all` or `--all-deps` + `--upgrade`) that updates all the dependencies, which we can then fold into the Knobots automation that produces these PRs: https://github.com/knative/pkg/pull/1486

If you poke me tomorrow after clearing the merge conflicts and rerunning `./hack/update-deps.sh`, I'll be happy to approve this unless @mattmoor has concerns.

Cynocracy:
Makes sense, happy to modify the PR with that flag tomorrow.

We'll probably need to learn roughly by trial and error which dependencies
end up being problematic (typically those which make breaking changes to
APIs in a minor version jump, or I suppose where we rely on something 'not
part of the API' yet public), but overall I'm hopeful this will be
beneficial.

Cynocracy:
Updated (fingers crossed on the tests :)

Cynocracy:
TestMetricsExport/Prometheus: resource_view_test.go:317: Unexpected prometheus output (-want +got):
          (
          )
:confused:

vagababov:
Trim space? :)

Cynocracy:
Prow was misleading me, actual error is:

```                """
            -   # HELP testComponent_global_export_counts Count of exports via standard OpenCensus view.
            -   # TYPE testComponent_global_export_counts counter
            -   testComponent_global_export_counts 2
            -   # HELP testComponent_resource_global_export_count Count of exports via RegisterResourceView.
            -   # TYPE testComponent_resource_global_export_count counter
            -   testComponent_resource_global_export_count 2
            -   # HELP testComponent_testing_value Test value
            -   # TYPE testComponent_testing_value gauge
            -   testComponent_testing_value{project="p1",revision="r1"} 0
            -   testComponent_testing_value{project="p1",revision="r2"} 1
            +   An error has occurred while serving metrics:
            + 
            +   collected metric "testComponent_testing_value" { gauge:<value:1 > } was collected before with the same name and label values```

Cynocracy:
We were using an unversioned commit, which update=all was trying to replace with the older, labeled version.

This is a case where if we want to keep pinned for an extended period we should be using the replace section.

Evan, do you know if the prometheus exporter changes we need will land soon? (Basically, should I update the replace section or assume that by the time we automate update=all we're OK to update that as well?)

evankanderson:
I don't know when the Prometheus changes will be part of a release; they've been committed upstream, but the release schedule for the prometheus plugin seems very ad-hoc.

evankanderson:
Which is to say that we should probably intentionally pin it, and add a comment next to the pin explaining that we're waiting for an upstream release past 0.2.0.

n3wscott:
close?

markusthoemmes:
@Cynocracy any news on this one? Seems vastly outdated.

PR #1464, opened 2020-07-01
Remove resource_name tag from webhook stats

tragiclifestories:
Proposed in relation to https://github.com/tektoncd/pipeline/issues/2872.

Motivation: Common use cases for this webhook involve using Kubernetes's generateName API to randomise resource names (this is a good idea in Tekton pipelines, for example, where there are uniqueness constraints. That means that the webhook metrics here end up with very high cardinality, which makes Prometheus fall over. Even without generateName, it is possible to shoot oneself in the foot this way. 

This commit just removes the resource_name label altogether.

### DIFF
--- a/webhook/stats_reporter.go
+++ b/webhook/stats_reporter.go
@@ -55,7 +55,6 @@ var (
 	resourceGroupKey     = tag.MustNewKey("resource_group")
 	resourceVersionKey   = tag.MustNewKey("resource_version")
 	resourceResourceKey  = tag.MustNewKey("resource_resource")
-	resourceNameKey      = tag.MustNewKey("resource_name")
 	resourceNamespaceKey = tag.MustNewKey("resource_namespace")
 	admissionAllowedKey  = tag.MustNewKey("admission_allowed")
 )
@@ -93,7 +92,6 @@ func (r *reporter) ReportRequest(req *admissionv1.AdmissionRequest, resp *admiss
 		tag.Insert(resourceGroupKey, req.Resource.Group),
 		tag.Insert(resourceVersionKey, req.Resource.Version),
 		tag.Insert(resourceResourceKey, req.Resource.Resource),
-		tag.Insert(resourceNameKey, req.Name),
 		tag.Insert(resourceNamespaceKey, req.Namespace),
 		tag.Insert(admissionAllowedKey, strconv.FormatBool(resp.Allowed)),
 	)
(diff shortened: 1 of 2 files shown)


evankanderson:
Would converting this to report actual different opencensus.Resource values (rather than tags) solve the issue, or is the issue that the cardinality of `resource_name` is so high that it's not actually a good candidate for a metric.

I kind of suspect the latter -- right now it seems like we could be reporting a metric for every single custom object that goes through the webhook, which could be thousands of names...

tragiclifestories:
> is the issue that the cardinality of resource_name is so high that it's not actually a good candidate for a metric.

In my experience running tekton, resource_name is unambiguously a metric of unworkable cardinality. It is possible that this is the tekton folks 'doing it wrong' by randomising resource names, but if they are, I expect many others are too.

tragiclifestories:
@evankanderson any advance on this? Just got indirectly reminded of its existence ... 

I understand if there's more serious reorganisation of this metrics code afoot, but this change should be orthogonal to all that and will disarm a big old footgun.

mattmoor:
/lgtm
/approve
/hold

I think you have to remove the "draft" bit and we need to make sure the CLA bot is happy with you 😉

dprotaso:
@tragiclifestories have you filled out the CLA? Details here: https://knative.dev/community/contributing/contributing/#contributor-license-agreements

dprotaso:
bump @tragiclifestories - were you able to sign the CLA?

vagababov:
Given it's been more than a month, I'd vote to just re-do the work.

tragiclifestories:
@dprotaso just signed it now - though I thought I already had for some Tekton thing long ago.

dprotaso:
>  though I thought I already had for some Tekton thing long ago.

The Tekton CLA is distinct from the Google CLA

vagababov:
@googlebot rescan

dprotaso:
🎉

vagababov:
somanyclas

tragiclifestories:
> > though I thought I already had for some Tekton thing long ago.
> 
> The Tekton CLA is distinct from the Google CLA

It didn't use to be, though - I remember because there was a hassle about signing the Google one with work ...

dprotaso:
I think the latest unit failure is legit

/retest

dprotaso:
or not - weird

tragiclifestories:
_runs away_

PR #1469, opened 2020-07-08
Move the hashing from the serving

vagababov:
/assign @mattmoor

### DIFF
--- a/hash/hash.go
+++ b/hash/hash.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


hash/hash.go
```diff
@@ -0,0 +1,164 @@
...
+	})
+	return hd
+}
+
+// ChooseSubset consistently chooses n items from `from`, using
+// `target` as a seed value.
+// ChooseSubset is an internal function and presumes sanitized inputs.
+// TODO(vagababov): once initial impl is ready, think about how to cache
+// the prepared data.
+func ChooseSubset(from []string, n int, target string) sets.String {
```
mattmoor, line 127: I didn't realize that our input order mattered, but it does.  Should we have this take a sets.String (has better symmetry too)

hash/hash.go
```diff
@@ -0,0 +1,164 @@
...
+		hd.nameLookup[hs] = f
+	}
+	// Sort for consistent mapping later.
+	sort.Slice(hd.hashPool, func(i, j int) bool {
+		return hd.hashPool[i] < hd.hashPool[j]
+	})
+	return hd
+}
+
+// ChooseSubset consistently chooses n items from `from`, using
```
mattmoor, line 122: We should really be better about documenting stuff in PKG.  Could I bother you for a detailed `doc.go` package comment with usage?

vagababov: Done. Not sure about the usage in doc...

PR #1472, opened 2020-07-09
Use quotation marks for generated checksums

whaught:
* If you win the hashcode lottery and get a checksum that is all numeric, this parses as an int instead of a string and fails.
    * Quotes ensure proper parsing
    * This was mysterious and confusing in: #8582

### DIFF
--- a/configmap/hash-gen/testdata/update.yaml
+++ b/configmap/hash-gen/testdata/update.yaml
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: bla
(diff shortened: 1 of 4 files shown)


configmap/hash-gen/testdata/update.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
   labels:
     serving.knative.dev/release: devel
   annotations:
-    knative.dev/example-checksum: bla
+    knative.dev/example-checksum: "bla"
```
vagababov, line 23: this one is redundant, I think, since it's the input :) but 🤷

whaught:
@googlebot check that cla again

PR #1477, opened 2020-07-10
Add BucketSet type which implements consistent bucketing

vagababov:
BucketSet permits mapping keys to buckets in a consistent manner, permitting us to use this type
in the HA applications. BucketSet implements the Bucket interface so it can be used in reconciler/leaderelection
pieces and has an additional method which returns the owner for the key. This is necessary in autoscaler
where requests need to be forwarded to an appropriate autoscaler.

/assign mattmoor @yanweiguo


vagababov:
Irrelevant
/retest

vagababov:
Patched the gomod stuff

hash/bucketer.go
```diff
@@ -0,0 +1,93 @@
...
+	lru "github.com/hashicorp/golang-lru"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// BucketSet answers to what bucket does key X belong in a
+// consistent manner (consistent as in consistent hashing).
+// In addition Bucket implements reconciler.Bucket interface, so it
+// can be used both in leader election and in routing applications.
+type BucketSet struct {
```
yanweiguo, line 33: Personally I'd prefer to split bucket interface and buckeset into two structs because bucketset sounds like a set of multiple buckets. Maybe we can give it a better name?

hash/bucketer.go
```diff
@@ -0,0 +1,93 @@
...
+// Scientifically inferred preferred cache size.
+const cacheSize = 4096
+
+func newCache() *lru.Cache {
+	c, _ := lru.New(cacheSize)
+	return c
+}
+
+// NewBucketSet creates a new bucket set with the given name.
+func NewBucketSet(name string, bucketList sets.String) *BucketSet {
```
yanweiguo, line 55: I guess we can define a construct func with this func and `newStatefulSetConfig` for StatefulSet case in leaderelection package?

hash/bucketer_test.go
```diff
@@ -0,0 +1,98 @@
...
+		t.Errorf("Cache[%s] = %q, want: %q", knownKey, n, thisBucket)
+	}
+	// unknownKey should've migrated.
+	if got := b.Owner(unknownKey); got == otherBucket {
+		t.Errorf("Owner = %q, don't want: %q", got, otherBucket)
+	}
+}
+
+func TestBucketHas(t *testing.T) {
+	b := NewBucketSet(thisBucket, buckets)
```
yanweiguo, line 93: Add `NewBucketSet(otherBucket, buckets)` then test `Has` returns false?

hash/bucketer.go
```diff
@@ -0,0 +1,93 @@
...
+	return &BucketSet{
+		name:    name,
+		cache:   newCache(),
+		buckets: bucketList,
+	}
+}
+
+// Has returns true if this bucket owns the key and
+// implements reconciler.Bucket interface.
```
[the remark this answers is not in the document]
vagababov: why do we need name?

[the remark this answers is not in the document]
vagababov: Ah.. for bucket interface. Sure.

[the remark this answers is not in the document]
vagababov: Done.

hash/bucketer.go
vagababov: Not sure I follow.

hash/bucketer.go
vagababov: Not sure... it kind of takes care of multiple buckets as well.

hash/bucketer_test.go
vagababov: Done.

hash/bucketer.go
yanweiguo: I mean how we can use this in leaderelection package.

vagababov: Ah, yea in SS elector we can just call new and wrap the return

hash/bucketer.go
mattmoor: My understanding was that we were talking about a BucketSet interface to complement and aggregate multiple Buckets lead by the same replica.  The intention being to avoid the loop over the buckets a replica leads each time we enqueue.

Also, must this be public?  😉   Why not return the interface?

hash/bucketer.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"knative.dev/pkg/reconciler"
+)
+
+var _ = (reconciler.Bucket)(nil)
```
mattmoor, line 30: As written this does nothing 🤣 
```suggestion
var _ reconciler.Bucket = (*BucketSet)(nil)
```

hash/bucketer.go
vagababov: Besides go philosophy being return types accept interfaces, we have Owner method.

hash/bucketer.go
vagababov: duh thanks

vagababov: Done

hash/bucketer.go
```diff
@@ -0,0 +1,112 @@
...
+	cache *lru.Cache
+
+	// mu guards buckets.
+	mu sync.RWMutex
+	// All the bucket names. Needed for building hash universe.
+	// `name` must be in this set.
+	buckets sets.String
+}
+
+// Bucket implements reconciler.Bucket and wraps around BuketSet
```
mattmoor, line 45:
```suggestion
// Bucket implements reconciler.Bucket and wraps around BucketSet
```

hash/doc.go
```diff
@@ -17,6 +17,11 @@ limitations under the License.
...
 //
 // - ChooseSubset is a consistent hashing/mapping function providing
-//   a  consistent selection of N keys from M (N<=M) keys for a given
+//   a consistent selection of N keys from M (N<=M) keys for a given
 //   target.
+// - BucketSet is a bucketer library which uses ChooseSubset under the
+//   the hood in order to implement consistent mapping between keys and
+//   set of buckets, indentified by unique names. Compared to basic bucket
+//   implementtion which just does hash%num_buckets, when the number of
+//   buckets change only a small subset of keys are supposed to migrate.
```
mattmoor, line 26: Do you want to mention Bucket here now?

hash/bucketer.go
vagababov: Done.

hash/doc.go
vagababov: Done.

PR #1482, opened 2020-07-13
Remove the ResourceLock field from the pkg/LE

vagababov:
we only use a single possible way, so no need to have, parse and validate
field that is in effect a constant.

/assign @dprotaso mattmoor


mattmoor:
/lgtm
/approve
/hold

Let's wait for coordination with downstream repos.

vagababov:
I got eventing to compile, so I'll merge this in the morning.

PR #1483, opened 2020-07-14
Promoting Binding to v1beta1

nachocano:
- Moved Binding to v1beta1
- This generated a circle dependency as apis/duck was importing v1beta1 in v1beta1_tests.go. I ended up adding implements_test as the coverage was failing, and moved the compile time checks to the non-test files. Did the same for v1.

/assign @vaikas


nachocano:
Thanks for reviewing @aliok !!!
@vaikas I think I still your approval powers for getting this one in.

PR #1487, opened 2020-07-14
Add flags to disable chaos per-component and configure period.

mattmoor:


kflag/set.go
```diff
@@ -0,0 +1,42 @@
...
+
+var _ flag.Value = (*StringSet)(nil)
+
+func (i *StringSet) String() string {
+	return fmt.Sprintf("%v", i.Value)
+}
+
+func (i *StringSet) Set(value string) error {
+	if i.Value == nil {
+		i.Value = sets.NewString()
```
vagababov, line 38:
```suggestion
		i.Value = make(sets.String,1)
```

mattmoor: done

PR #1490, opened 2020-07-14
Use hash.BucketSet for standard leader election mode

yanweiguo:
- Use `hash.Bucket` as the bucket in standard leader election mode.
- Remove the no longer used `leaderelection.bucket` structure.


leaderelection/context.go
```diff
@@ -168,9 +160,27 @@ func (b *standardBuilder) buildElector(ctx context.Context, la reconciler.Leader
...
+	bkts := make([]reconciler.Bucket, 0, cc.Buckets)
+	for name := range names {
+		bkts = append(bkts, hash.NewBucket(name, bs))
+	}
+	return bkts
+}
+
+func standardBucketName(ordinal uint32, queueName string, cc ComponentConfig) string {
+	return strings.ToLower(fmt.Sprintf("%s.%s.%02d-of-%02d", cc.Component, queueName, ordinal, cc.Buckets))
 }
```
vagababov, line 184: This will not work well if the number of buckets were to change dynamically. I don't think we're planning for that, but should be called out in a comment just in case.

yanweiguo: I don't think dynamical change of bucket size is supported now. Not only this func, the whole stuff doesn't work well in bucket size change case. I don't think we need call out for this func. Instead we can call out in the user-guide or something similar for customers.

evankanderson: Add a comment to `config.go` around lines 78 and 111 that changing `Buckets` dynamically is not supported?

leaderelection/context.go
```diff
@@ -168,9 +160,27 @@ func (b *standardBuilder) buildElector(ctx context.Context, la reconciler.Leader
...
+	names := make(sets.String, cc.Buckets)
+	for i := uint32(0); i < cc.Buckets; i++ {
+		names.Insert(standardBucketName(i, queueName, cc))
 	}
-	return &runAll{les: buckets}, nil
+	bs := hash.NewBucketSet(names)
+
+	bkts := make([]reconciler.Bucket, 0, cc.Buckets)
+	for name := range names {
+		bkts = append(bkts, hash.NewBucket(name, bs))
```
evankanderson, line 177: I'm slightly confused here why the `hash.NewBucket` method isn't a method on the `hash.BucketSet`, but I don't think that needs to be fixed in this review.

vagababov: Didn't think of that API twist. Let me add it.

mattmoor: 🤣  it's what I originally suggested, but orthogonal to this.

With fresh eyes, I wonder whether it would make sense to also have a method that returns this entire list, which I think the autoscaler would also need for metrics forwarding to the appropriate shard, right?

mattmoor:
/hold

Just make sure you have Evan's feedback incorporated.

yanweiguo:
Will follow up based on https://github.com/knative/pkg/pull/1500 since both standard mode and statefulset mode need change.

/hold cancel

leaderelection/context.go
vagababov: I think Yanwei builds that directly in the AS main,

yanweiguo: There's already a method `hash.BucketSet` returning the entire list.

PR #1491, opened 2020-07-14
Read from all meters instead of just he default one

jjzeng-seattle:
This is necessary once we move into using resource instead of metric labels, as we don't deal with only the default meter any more.

A lot of tests in serving will break without this change.

### DIFF
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -30,7 +30,7 @@ import (
 func CheckStatsReported(t test.T, names ...string) {
 	t.Helper()
 	for _, name := range names {
-		d, err := view.RetrieveData(name)
+		d, err := readRowsFromAllMeters(name)
 		if err != nil {
 			t.Error("For metric, Reporter.Report() error", "metric", name, "error", err)
 		}
@@ -45,7 +45,7 @@ func CheckStatsReported(t test.T, names ...string) {
 func CheckStatsNotReported(t test.T, names ...string) {
 	t.Helper()
 	for _, name := range names {
-		d, err := view.RetrieveData(name)
+		d, err := readRowsFromAllMeters(name)
 		// err == nil means a valid stat exists matching "name"
 		// len(d) > 0 means a component recorded metrics for that stat
 		if err == nil && len(d) > 0 {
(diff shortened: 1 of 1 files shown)


metrics/metricstest/metricstest.go
```diff
@@ -166,9 +166,12 @@ func CheckSumData(t test.T, name string, wantTags map[string]string, wantValue f
 //
 // In normal process shutdown, metrics do not need to be unregistered.
 func Unregister(names ...string) {
-	for _, n := range names {
-		if v := view.Find(n); v != nil {
-			view.Unregister(v)
+	for _, producer := range metricproducer.GlobalManager().GetAll() {
+		meter := producer.(view.Meter)
```
vagababov, line 170: Is it guaranteed to successfully cast

evankanderson:
(To be clear, the tests in serving will break once the records are reported using Resource, not with the state of `pkg` before thes PR.)

jjzeng-seattle:
> (To be clear, the tests in serving will break once the records are reported using Resource, not with the state of `pkg` before this PR.)

Thanks for clearing.

metrics/metricstest/metricstest.go
```diff
@@ -195,28 +198,32 @@ func lastRow(t test.T, name string, meter view.Meter) *view.Row {
...
 		d, err := meter.RetrieveData(name)
 		if err != nil || len(d) == 0 {
 			continue
 		}
-		if len(d) > 1 {
-			return nil, fmt.Errorf("expected 1 row for metric %q got %d", name, len(d))
-		}
-		if retval != nil {
-			return nil, fmt.Errorf("got 2 rows from different meters: %+v, %+v", *retval, d[0])
+		if rows != nil {
```
vagababov, line 221: it might be useful to store the initial `meter` as well

metrics/metricstest/metricstest.go
evankanderson: Yes, in practice.

metrics/metricstest/metricstest.go
evankanderson: If we find this is having a problem, the `meter` pointers probably won't help that much.

Better would be to use the [`Read` method from `metricproducer.Producer`](https://godoc.org/go.opencensus.io/metric/metricproducer#Producer), but that has a different signature, so it doesn't help for current code.

evankanderson: I'm planning to put together a PR to introduce a new set of interfaces to `metricstest` that will read out the metrics from `metricproducer` instead of the `RetrieveData` method.

PR #1493, opened 2020-07-15
Add component name as the metric prefix for opencensus exporter

jjzeng-seattle:
And update the ocagent version to include the functionality.


metrics/opencensus_exporter.go
```diff
@@ -35,6 +35,9 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
 	if config.collectorAddress != "" {
 		opts = append(opts, ocagent.WithAddress(config.collectorAddress))
 	}
+	if config.component != "" {
```
evankanderson, line 38: Do we need `config.domain` as well?

How does this match with `WithServiceName` on line 34?

Stackdriver sets the metric prefix to `path.Join(mc.domain, mc.component) on line 256 of `config.go`.

evankanderson: It looks like `WithServiceName` is used to fill in [`ServiceInfo` on the Node information representing the system that's reporting the metrics](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/agent/common/v1/common.proto#L44).

evankanderson:
It looks like you need to run `update-codegen.sh` as well.

metrics/opencensus_exporter.go
jjzeng-seattle: We don't want to add domain here, which is a stackdriver specific here.  If we add it here, all other metric backends in the opentelemetry metrics collector service will get the domain prefix, which is undesired.

Instead the domain prefix will be added by the stackdriver exporter  in opentelemetry metrics collector service.

jjzeng-seattle: We could use the `ServiceInfo`, but then we need to modify the exporters in opentelemetry metrics collector service to be aware of it and add it as the prefix.

The prefix here is more straightforward.

jjzeng-seattle:
> It looks like you need to run `update-codegen.sh` as well.

Done.

metrics/opencensus_exporter.go
evankanderson: I agree that adding the prefix here:

1. Makes sense
2. Is more straightforward than extracting from `ServiceInfo`

My question was around avoiding collisions in the resource names; right now Stackdriver (which has a resource-path name) names the exports something like:

* `knative.dev/serving/autoscaler/actual_pods`
* `custom.googleapis.com/knative.dev/autoscaler/not_ready_pods`

I'm wondering if the `serving` (and `knative.dev`) part of the first name should be passed along via OpenCensus as well as the `autoscaler` part. Passing along those first parts of the domain makes it easier to avoid namespace collisions (for example, if there were "controller" components in both serving and eventing).

evankanderson:
One comment -- if you think that we don't need the additional prefixes on the OpenCensus exported variable names, I'm happy to let that be, but now that I see what's going on, I feel like we should include the information we have, and Stackdriver can strip off the `mc.domain` part for custom metric names if needed.

metrics/opencensus_exporter.go
evankanderson: In particular, I've found a metrics named collision between serving and eventing which is resolved via the `mc.domain` field, which is extracted from the `METRICS_DOMAIN` environment variable:

* Serving: [`config/core/deployments/controller.yaml`](https://github.com/knative/serving/blob/master/config/core/deployments/controller.yaml#L60) with name [`controller`](https://github.com/knative/serving/blob/master/cmd/controller/main.go#L45)
* Eventing: [`config/core/deployments/controller.yaml`](https://github.com/knative/eventing/blob/master/config/core/deployments/controller.yaml#L55) with name [`controller`](https://github.com/knative/eventing/blob/master/cmd/controller/main.go#L37)

I'd prefer to clean up the `METRICS_DOMAIN` in serving to all be under the same domain, but we'll need to see if that's breaking for anyone.

evankanderson:
Note that this will cause a change in the opencensus-generated names from components, from (for example):

`request_count`

to

`knative.dev/serving/activator/request_count`

I think this is correct, because it makes it easier to select out metrics that are Knative-related, as opposed to needing to hunt under other names.

Note that we're not doing this for the legacy Prometheus exporter, which never included this information in its names.

evankanderson:
Also a note of record; I'm not sure that doing this with a (process-wide) prefix rather than in the actual metric definition sites is the right way to go, but I'd prefer not to upend the whole apple cart at once, no matter what it looks like.

evankanderson:
If we decide that we do want `knative.dev/serving/controller/work_queue_depth` and `knative.dev/eventing/controller/work_queue_depth` to end up with the same metric name, we can remove or empty the `METRICS_DOMAIN` later. My preference would be not to remove that information until we're sure that those two should be the same metric name.

metrics/opencensus_exporter.go
```diff
@@ -35,6 +35,16 @@ func newOpenCensusExporter(config *metricsConfig, logger *zap.SugaredLogger) (vi
 	if config.collectorAddress != "" {
 		opts = append(opts, ocagent.WithAddress(config.collectorAddress))
 	}
+	metrixPrefix := ""
+	if config.domain != "" {
+		metrixPrefix = config.domain
+	}
+	if config.domain != "" {
+		metrixPrefix += "/" + config.component
+	}
```
evankanderson, line 44: What about:

```golang
config.domain = path.Join(config.domain, config.component)
```

?

evankanderson: Whoops, that should be:

```golang
metricsPrefix := path.join(config.domain, config.component)
```

PR #1494, opened 2020-07-15
Remove key in tags to reduce metrics count

lanceliuu:
Issue: https://github.com/knative/serving/issues/8609

Signed-off-by: Lance Liu <xuliuxl@cn.ibm.com>

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -452,7 +452,7 @@ func (c *Impl) processNextWorkItem() bool {
 		if err != nil {
 			status = falseString
 		}
-		c.statsReporter.ReportReconcile(time.Since(startTime), keyStr, status)
+		c.statsReporter.ReportReconcile(time.Since(startTime), status)
 	}()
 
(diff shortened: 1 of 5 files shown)


evankanderson:
You probably also want to remove `keyTagKey` from line 170 and 175 (which is where we instruct OpenCensus to aggregate divided by `key`), and the definition on line 49.

lanceliuu:
I removed the key tag accordingly. Thank you @evankanderson.
After apply this patch, the total metrics count for controller drops significantly.
Before:
```
~ wc -l metrics-controller.txt
   68628 metrics-controller.txt
```
After:
```
~ wc -l metrics-controller.txt
     570 metrics-controller.txt
```
Also the size of metrics drops from ~11M to ~66K
@vagababov

evankanderson:
Wow, the cardinality for `key` must have been really high (which will cause problems with Stackdriver, at least, but many monitoring systems struggle with high-cardinality labels).

/lgtm
/approve

PR #1495, opened 2020-07-15
Update Stackdriver custom exporter to support RecordBatch where some records are special and others are not.

evankanderson:
/assign @jjzeng-seattle 

This addresses the usage at https://github.com/knative/serving/blob/master/pkg/reconciler/autoscaling/kpa/kpa.go#L220 where `requested_pod_count` is promoted to a "native" Stackdriver type, but `not_ready_pod_count` and `terminating_pod_count` are not.

I *really hope* that this is the last customization we need in `pkg` that corresponds to a Stackdriver backend change somewhere in Google's source control systems (and once we move to OpenCensus exporting, this translation can be done in an OpenCensus collector-compatible binary shipped by the Google Cloud Run team).

### DIFF
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"sort"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


evankanderson:
/assign @yanweiguo 

Sorry that the diff in `stackdriver_exporter.go` is so wonky; GitHub's diff computer seems to have made some bad choices when I indented the block and renamed a few variables.

Issue #1488, opened 2020-07-14
logstream doesn't work with HPA or chaos


mattmoor:
/area API
/area monitoring
/area test-and-release
/kind bug

## Expected Behavior

`logstream` should pick up new system pods and start streaming their logs.

## Actual Behavior

`logstream` doesn't pick up new logs (HPA) and often stops logging altogether when run with chaos.

## Crux

The crux of the issue is that pods are only listed at startup, so as pods are retired and new ones brought up over the course of the test the logging simply falls off.

See here: https://github.com/knative/pkg/blob/16eea5bd5b12aeef02a696d73f3661da627e2726/test/logstream/kubelogs.go#L61

vagababov:
/close
This is done and should be merged to serving here https://github.com/knative/serving/pull/8643

PR #1496, opened 2020-07-15
Make logstream watch the pods as they change.

vagababov:
- watch the pods and track what we already watch
- start watchers for the new pods

For #1488

/assign @markusthoemmes mattmoor

### DIFF
--- a/test/logstream/kubelogs.go
+++ b/test/logstream/kubelogs.go
@@ -36,9 +36,8 @@ import (
 )
 
 type kubelogs struct {
-	namespace   string
-	kc          *test.KubeClient
-	watchedPods sets.String
+	namespace string
+	kc        *test.KubeClient
 
 	once sync.Once
 	m    sync.RWMutex
@@ -54,23 +53,22 @@ var _ streamer = (*kubelogs)(nil)
 const timeFormat = "15:04:05.000"
 
 func (k *kubelogs) startForPod(eg *errgroup.Group, pod *corev1.Pod) {
-	k.watchedPods.Insert(pod.Name)
 	// Grab data from all containers in the pods.  We need this in case
 	// an envoy sidecar is injected for mesh installs.  This should be
 	// equivalent to --all-containers.
 	for _, container := range pod.Spec.Containers {
 		// Required for capture below.
-		pod, container := pod, container
+		psn, pn, cn := pod.Namespace, pod.Name, container.Name
 		eg.Go(func() error {
 			options := &corev1.PodLogOptions{
-				Container: container.Name,
+				Container: cn,
 				// Follow directs the api server to continuously stream logs back.
 				Follow: true,
 				// Only return new logs (this value is being used for "epsilon").
 				SinceSeconds: ptr.Int64(1),
 			}
 
-			req := k.kc.Kube.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, options)
+			req := k.kc.Kube.CoreV1().Pods(psn).GetLogs(pn, options)
 			stream, err := req.Stream()
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
 const timeFormat = "15:04:05.000"
 
+func (k *kubelogs) startForPod(eg *errgroup.Group, pod *corev1.Pod) {
+	k.watchedPods.Insert(pod.Name)
+	// Grab data from all containers in the pods.  We need this in case
+	// an envoy sidecar is injected for mesh installs.  This should be
+	// equivalent to --all-containers.
+	for _, container := range pod.Spec.Containers {
+		// Required for capture below.
+		pod, container := pod, container
```
evankanderson, line 63: We really only need to capture `pod.Name` and `container.Name`, right?

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
+				if p.Status.Phase == corev1.PodRunning && p.DeletionTimestamp == nil {
+					for _, cond := range p.Status.Conditions {
+						if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+							t.Log("The new knative pod is ready", p.Name)
+							k.startForPod(eg, p)
+							break
+						}
+					}
+				}
+				t.Log("Not ready yet pod", p.Name)
```
markusthoemmes, line 118:
```suggestion
				t.Log("Pod is not ready yet", p.Name)
```

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
+			case watch.Added, watch.Modified:
+				if k.watchedPods.Has(p.Name) {
+					t.Log("Already watching pod", p.Name)
+					continue
+				}
+				// If pod is ready then start the logs.
+				if p.Status.Phase == corev1.PodRunning && p.DeletionTimestamp == nil {
+					for _, cond := range p.Status.Conditions {
+						if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
+							t.Log("The new knative pod is ready", p.Name)
```
markusthoemmes, line 112:
```suggestion
							t.Log("New pod is ready", p.Name)
```

test/logstream/kubelogs.go
markusthoemmes, line 63: Why do we need to capture `pod` though?

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
+			// Pods get killed with chaos duck, so logs might end
+			// before the test does.
+		})
+	}
+}
+
+func (k *kubelogs) watchPods(t test.TLegacy, eg *errgroup.Group) {
+	wi, err := k.kc.Kube.CoreV1().Pods(k.namespace).Watch(metav1.ListOptions{})
+	if err != nil {
+		t.Logf("Logstream knative pod watch failed, logs might be missing: %v", err)
```
markusthoemmes, line 93: Should we surface this error similar to the `List` error in `init`?

evankanderson:
/retest

I sent #1497 to fix the workqueue_test flakiness.

test/logstream/kubelogs.go
markusthoemmes: Right, but it's not a loop-variable :thinking:

test/logstream/kubelogs.go
mattmoor, line 112: I think this would flow/read better with a bit more encapsulation and redrawing the line between these functions.  Something like:

```
if watched.Has(p.Name) {
  continue
}
if !isReady(pod) {
  continue
}
watched.Insert(pod.Name)
k.startForPod(pod)

```

This also lets you make `watchedPods` a local variable instead of a field.

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
-				scanner := bufio.NewScanner(stream)
-				for scanner.Scan() {
-					k.handleLine(scanner.Bytes())
-				}
-				return fmt.Errorf("logstream completed prematurely for %s/%s: %w",
-					pod.Name, container.Name, scanner.Err())
-			})
-		}
+		pod := pod
+		k.startForPod(&eg, &pod)
```
mattmoor, line 145: Do we still need this loop, or can it be subsumed by watchPods?

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
-				// Read this container's stream.
-				scanner := bufio.NewScanner(stream)
-				for scanner.Scan() {
-					k.handleLine(scanner.Bytes())
-				}
-				return fmt.Errorf("logstream completed prematurely for %s/%s: %w",
-					pod.Name, container.Name, scanner.Err())
-			})
-		}
+		pod := pod
```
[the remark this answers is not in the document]
vagababov: more or less standard go practice for for loops that spawn goroutines (had to add it actually based on debugging, I thought the passing to function would work too).

test/logstream/kubelogs.go
vagababov: and pod namespace. I can reduce this to `pns, pn, cn :=...`, given than containers are copied as a whole.

test/logstream/kubelogs.go
```diff
@@ -49,54 +53,99 @@ var _ streamer = (*kubelogs)(nil)
...
+func (k *kubelogs) watchPods(t test.TLegacy, eg *errgroup.Group) {
+	wi, err := k.kc.Kube.CoreV1().Pods(k.namespace).Watch(metav1.ListOptions{})
+	if err != nil {
+		t.Logf("Logstream knative pod watch failed, logs might be missing: %v", err)
+		return
+	}
+	go func() {
+		for ev := range wi.ResultChan() {
+			p := ev.Object.(*corev1.Pod)
+			t.Logf("Pod event %v for %s", ev.Type, p.Name)
```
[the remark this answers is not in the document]
vagababov: helped in this debugging, probably useless in the tests. So i'll just remove it.

test/logstream/kubelogs.go
vagababov: I tried to avoid formatting call... this sounds a bit off, but I think I can go with colon,

test/logstream/kubelogs.go
vagababov: escapes in go goroutine?

test/logstream/kubelogs.go
vagababov: I debated that... I am not sure failing a test, because of logging failure is a valid concern?

test/logstream/kubelogs.go
vagababov: No this does not work because we also start pods in init. But I think along the lines you left below — that I can remove the loop in init.

test/logstream/kubelogs.go
vagababov: yeah, I think so.

test/logstream/kubelogs.go
vagababov: But I'll go along with what we have now.

vagababov:
Fixed the build and updated according to reviews.

vagababov:
Why my new commits don't show? 
The last two from here: https://github.com/vagababov/pkg/commits/2020714-logs don't show at all.

PR #1497, opened 2020-07-15
Remove delay-based flakiness from workqueue_test.go

evankanderson:
I noticed that `metrics/workqueue_test.go` seemed to be fairly flaky:

https://testgrid.knative.dev/pkg#continuous&sort-by-failures=&include-filter-by-regex=metrics.TestWorkqueueMetrics

It turns out that as we've added some latency to extracting metrics via metricstest, the fairly low latency of `AddRateLimited` with the default rate limiter has started to exceed the latency of extracting 5 metrics from the backend and comparing them. (I'm a little surprised; I was unable to duplicate this locally. I *was* able to trigger failures in the `CheckCountData` on line 130 after the sleep by reducing the sleep to <4ms, but the failure rate wasn't consistent.)

I choose a much bigger delay interval (50ms), and was then able to cut the sleep to 100ms and this passed 100/100 times locally. (Of course, the old test passed every time locally, too, but I feel more confident about this timing.)

### DIFF
--- a/metrics/workqueue_test.go
+++ b/metrics/workqueue_test.go
@@ -35,6 +35,20 @@ func newFloat64(name string) *stats.Float64Measure {
 	return stats.Float64(name, "bar", "wtfs/s")
 }
 
+type FixedRateLimiter struct {
+	delay time.Duration
+}
+
+func (f *FixedRateLimiter) When(item interface{}) time.Duration {
+	return f.delay
+}
+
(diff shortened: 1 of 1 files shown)


evankanderson:
Hmm, the same failure. I'm not quite sure what's driving the latency of the main code path vs elsewhere, but trying increasing the delay to 200ms.

PR #1500, opened 2020-07-15
Add a new method to permit creation of buckets from the bucketset

vagababov:
This is a natural factory.

I don't want to mess with Yanwei's PR, so I didn't remove the old method.
But we can reconcile that in the end.

/assign @yanweiguo @evankanderson

### DIFF
--- a/hash/bucketer.go
+++ b/hash/bucketer.go
@@ -19,6 +19,7 @@ limitations under the License.
 package hash
 
 import (
+	"errors"
 	"sync"
 
 	lru "github.com/hashicorp/golang-lru"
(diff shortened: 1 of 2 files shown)


hash/bucketer.go
```diff
@@ -86,36 +86,44 @@ func (b *Bucket) Has(nn types.NamespacedName) bool {
 	return b.buckets.Owner(nn.String()) == b.name
 }
 
+// NewBucket creates a new Bucket with given name based on this bucketset.
+func (bs *BucketSet) NewBucket(name string) *Bucket {
+	return &Bucket{
+		name:    name,
```
yanweiguo, line 92: Since you're here, I think it's chance to add a check that the given name is in the bucketSet.buckets.

evankanderson: +1

vagababov: Sounds good!
Done.

PR #1502, opened 2020-07-15
Generating reconciler for Deployments

nachocano:
- Adding a generated reconciler with leader-election capabilities for Deployments.
- Need this for knative-gcp https://github.com/google/knative-gcp/issues/1466

/assign @vaikas


markusthoemmes:
Why not generate this downstream if it's only needed there? @nachocano

nachocano:
I guess I saw the namespace and crd ones in pkg so I went ahead and added
it there.
I see the namespace one is being used by eventing and serving. The crd one
seems only eventing...
Irrespective of that, do you want me to remove it? Or are you OK with
leaving it for now?

markusthoemmes:
I'm okay leaving it for now, I was just curious if there's a downstream blocker to generate it there.

nachocano:
Nope, I think there is no blocker. It was my bad I guess.
OK, will leave it for now then, and we can revisit later on. thanks!

PR #1503, opened 2020-07-15
Use the new method of BucketSet to create Bucket

yanweiguo:
A follow up of https://github.com/knative/pkg/pull/1490.

Introduce `ActualBucektList` func.

/assign @mattmoor @vagababov


leaderelection/context.go
```diff
@@ -174,9 +178,13 @@ func newStandardBuckets(queueName string, cc ComponentConfig) []reconciler.Bucke
 
 	bkts := make([]reconciler.Bucket, 0, cc.Buckets)
 	for name := range names {
-		bkts = append(bkts, hash.NewBucket(name, bs))
+		bkt, err := bs.NewBucket(name)
+		if err != nil {
+			return nil, err
```
mattmoor, line 183: IIUC this error path is impossible to reach because the only error condition for NewBucket is when `name` is not in `names`, which we know cannot happen.

If this entire loop were a method on `BucketSet` to produce the list of buckets, then it could know this and we could avoid the unreachable error path everywhere, right?  cc @vagababov 

Unrelated to the above, it seems like the list of buckets we produce here is going to be randomly ordered, I'm not sure that's a good thing. 🤔

yanweiguo: You're right. Added a comment and ignore the error.

Also use a ordered list.

leaderelection/context.go
```diff
@@ -173,8 +173,10 @@ func newStandardBuckets(queueName string, cc ComponentConfig) []reconciler.Bucke
...
 
 	bkts := make([]reconciler.Bucket, 0, cc.Buckets)
-	for name := range names {
-		bkts = append(bkts, hash.NewBucket(name, bs))
+	for _, name := range names.List() {
+		// The error shouldn't happen as we make sure that `name` is in the BucketSet.
+		bkt, _ := bs.NewBucket(name)
+		bkts = append(bkts, bkt)
 	}
 	return bkts
```
mattmoor, line 181: If there's an error path, I'd rather we checked the error.  My point was that if we had a method like `BucketList` that returned `[]Bucket` instead of `[]string` then this could simply be:

```
return bs.ActualBucketList()
```

mattmoor: This can naturally avoid the error path because it knows the bucket `.Has()` the name.  To ignore the error is to know the details behind the abstraction, which limits our ability to change that implementation in a sane way 😅

leaderelection/context.go
vagababov: IT sounds like an OK method to add. :) Iteration is the way to go.

vagababov:
queue tests should be fixed here: https://github.com/knative/pkg/pull/1516

leaderelection/context.go
yanweiguo: Introduced `ActualBucketList`. PTAL

hash/bucketer.go
```diff
@@ -87,17 +77,24 @@ func (b *Bucket) Has(nn types.NamespacedName) bool {
 	return b.buckets.Owner(nn.String()) == b.name
 }
 
-// NewBucket creates a new Bucket with given name based on this bucketset.
-func (bs *BucketSet) NewBucket(name string) (*Bucket, error) {
+// ActualBucketList creates a new list of all possible Bucket based on this bucketset
+// in the order of sorted bucket name.
+func (bs *BucketSet) ActualBucketList() []reconciler.Bucket {
```
vagababov, line 82: I'd just go with `(All)Buckets`.

hash/bucketer.go
```diff
@@ -87,17 +77,24 @@ func (b *Bucket) Has(nn types.NamespacedName) bool {
...
+	bkts := make([]reconciler.Bucket, 0, len(bs.buckets))
+	for _, n := range bs.sortedBucketNames() {
+		bkts = append(bkts, &Bucket{
+			name:    n,
+			buckets: bs,
+		})
+	}
```
vagababov, line 89:
```suggestion
	bkts := make([]reconciler.Bucket, len(bs.buckets))
	for i, n := range bs.sortedBucketNames() {
		bkts[i] = &Bucket{
			name:    n,
			buckets: bs,
		}
	}
```

hash/bucketer.go
yanweiguo: done

hash/bucketer.go
```diff
@@ -87,17 +77,24 @@ func (b *Bucket) Has(nn types.NamespacedName) bool {
 	return b.buckets.Owner(nn.String()) == b.name
 }
 
-// NewBucket creates a new Bucket with given name based on this bucketset.
-func (bs *BucketSet) NewBucket(name string) (*Bucket, error) {
+// Buckets creates a new list of all possible Bucket based on this bucketset
+// in the order of sorted bucket name.
```
vagababov, line 81:
```suggestion
// ordered by bucket name.
```

yanweiguo:
Closed in favor of https://github.com/knative/pkg/pull/1530

PR #1506, opened 2020-07-16
Remove logging from pod watcher to avoid race conditions.

markusthoemmes:
These cause a race condition as can be seen in https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-continuous/1283612572876017666.

They capture `t` far beyond the test's lifetime (this is `t` of the test that first initializes the logstream).

If we alternatively think that those logs are valuable, we should probably run through the set of active tests and log it to each of them. I didn't think they're all too important to the test output though.


vagababov:
Stupid logger:-(

vagababov:
They are useful to debug why the logs aren't there. But 🤷 rare situation.

Issue #1505, opened 2020-07-16
Implement a workqueue "slow lane".


mattmoor:
/area API
/kind feature

Today our shared controller infrastructure uses a `workqueue.RateLimitingInterface` to manage keys ([here](https://github.com/knative/pkg/blob/096656a2d4c88b8d6030feb2384da1d347fb3ccf/controller/controller.go#L189-L194)).

Most of the stuff that we stick onto the queue is driven by informer events, which point to specific resources and result in a single key being queued, e.g. via [EnqueueKey](https://github.com/knative/pkg/blob/096656a2d4c88b8d6030feb2384da1d347fb3ccf/controller/controller.go#L346-L350), and occasionally we'll snooze a resource through a mechanism like [EnqueueKeyAfter](https://github.com/knative/pkg/blob/096656a2d4c88b8d6030feb2384da1d347fb3ccf/controller/controller.go#L360-L365).

However, there are global events, which may prompt us to reconsider *everything* in our key space, e.g. changing a controller's configmap.  We refer to these as "global resyncs" and we have a [GlobalResync](https://github.com/knative/pkg/blob/096656a2d4c88b8d6030feb2384da1d347fb3ccf/controller/controller.go#L495-L499) method that helps us manage these.

One of the problems we hit with global resyncs was that they "just" queued all resources for reconciliation, which is mostly what we want, but the problem is that many of these resyncs may be benign and when we're talking about 100s or 1000s of keys they clog up the workqueue preventing other work that's queued due to informer events from being processed in a timely fashion.

Historically we've worked around this by having global resyncs `EnqueueAfter` with a jitter factor based on the size of the resync, which helps distribute the keys of the global resync over some period of time.  However, this means that it could take us 100s to resync 100 keys even if no other work is being done, and the adequately jitter these we need a pretty global view of the resync being performed.

---

The basic idea behind this issue is that instead of jittering, we would simply place keys knowingly enqueued due to a resync event onto a "slow lane", either internal to the workqueue or as a second workqueue.

Traditional queuing events would push work onto the "fast lane", and explicit resync events (e.g. global resync, [failover](https://github.com/knative/pkg/pull/1504)) would push work onto the "slow lane".

This enables us to keep processing "slow lane" keys when the "fast lane" is empty, while also doing a better job of keeping "slow lane" keys from getting in the way of keys that have actually seen an informer event.

mattmoor:
[...]

PR #1512, opened 2020-07-17
Implement Slowlane/FastLane queues

vagababov:
The suggested implementation is basically a Y joint where the data flows from 2 queues into the third queue, with the caveat that one of the paths has priority over the other.
This serves as a drop-in replacement in the controller.
Which will be in a separate pr.

/assign mattmoor

For #https://github.com/knative/pkg/issues/1505

### DIFF
--- a/controller/two_lane_queue.go
+++ b/controller/two_lane_queue.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"k8s.io/client-go/util/workqueue"
+)
+
+// twoLaneQueue is a rate limited queue that wraps around two queues
+// -- fast queue (anonymously aliased), whose contents are processed with priority.
+// -- slow queue (slowLane queue), whose contents are processed if fast queue has no items.
+// All the default methods operate on the fast queue, unless noted otherwise.
+type twoLaneQueue struct {
+	workqueue.RateLimitingInterface
+	slowLane workqueue.RateLimitingInterface
+	// consumerQueue is necessary to ensure that we're not reconciling
+	// the same object at the exact same time (e.g. if it had been enqueued
+	// in both fast and slow and is the only object there).
+	consumerQueue workqueue.Interface
+
+	name string
+
+	fastChan chan interface{}
+	slowChan chan interface{}
+}
+
+// Creates a new twoLaneQueue.
+func newTwoLaneWorkQueue(name string) *twoLaneQueue {
+	rl := workqueue.DefaultControllerRateLimiter()
+	tlq := &twoLaneQueue{
+		RateLimitingInterface: workqueue.NewNamedRateLimitingQueue(
+			rl,
+			name+"-fast",
+		),
+		slowLane: workqueue.NewNamedRateLimitingQueue(
+			rl,
+			name+"-slow",
+		),
+		consumerQueue: workqueue.NewNamed(name + "-consumer"),
+		name:          name,
+		fastChan:      make(chan interface{}),
+		slowChan:      make(chan interface{}),
+	}
+	tlq.runConsumer()
+	tlq.runProducers()
+	return tlq
+}
+
+func process(q workqueue.Interface, ch chan interface{}) {
+	// Sender closes the channel
+	defer close(ch)
+	for {
+		i, d := q.Get()
+		// If the queue is empty and we're shutting down — stop the loop.
+		if d {
+			break
+		}
+		ch <- i
+		q.Done(i)
+	}
+}
+
+func (tlq *twoLaneQueue) runProducers() {
+	go process(tlq.RateLimitingInterface, tlq.fastChan)
+	go process(tlq.slowLane, tlq.slowChan)
+}
+
+func (tlq *twoLaneQueue) runConsumer() {
+	tlq.consumerQueue = workqueue.NewNamed(tlq.name + "consumer")
+	// Shutdown flags.
+	fast, slow := true, true
+	go func() {
+		// When both producer queues are shutdown stop the consumerQueue.
+		defer tlq.consumerQueue.ShutDown()
+		// While any of the queues is still running, try to read off of them.
+		for fast || slow {
+			// By default drain the fast lane.
+			// Channels in select are picked random, so first
+			// we have a select that only looks at the fast lane queue.
+			if fast {
+				select {
+				case item, ok := <-tlq.fastChan:
+					if !ok {
+						// This queue is shutdown and drained. Stop looking at it.
+						fast = false
+						continue
+					}
+					tlq.consumerQueue.Add(item)
+					continue
+				default:
+					// This immediately exits the wait if the fast chan is empty.
+				}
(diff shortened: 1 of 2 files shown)


controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+func (tlq *twoLaneQueue) runConsumer() {
+	tlq.consumerQueue = workqueue.NewNamed(tlq.name + "consumer")
+	// Shutdown flags.
+	fast, slow := true, true
+	go func() {
+		for fast || slow {
+			// By default drain the fast lane.
+			if fast {
+				select {
+				case item, ok := <-tlq.fastChan:
```
whaught, line 98: Is this switch necessary?

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+// -- fast queue, whose contents are processed with priority.
+// -- slow queue, whose contents are processed if fast queue has no items.
+// All the default methods operate on the fast queue, unless noted otherwise.
+type twoLaneQueue struct {
+	workqueue.RateLimitingInterface
+	slowLane workqueue.RateLimitingInterface
+	// consumerQueue is necessary to ensure that we're not reconciling
+	// the same object at the exact same time (e.g. if it had been enqueued
+	// in both fast and slow and is the only object there).
+	consumerQueue workqueue.Interface
```
mattmoor, line 39: field names don't line up with comment

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		// Sender closes the channel
+		close(ch)
+	}(q)
+}
+
+func (tlq *twoLaneQueue) runProducers() {
+	process(tlq.RateLimitingInterface, tlq.fastChan)
+	process(tlq.slowLane, tlq.slowChan)
+}
```
mattmoor, line 87:
```suggestion
func process(q workqueue.Interface, ch chan item) {
	// Sender closes the channel
	defer close(ch)
	for {
		i, d := q.Get()
		// If the queue is empty and we're shutting down — stop the loop.
		if d {
			return
		}
		ch <- item{i: i, sd: d}
	}
}

func (tlq *twoLaneQueue) runProducers() {
	go process(tlq.RateLimitingInterface, tlq.fastChan)
	go process(tlq.slowLane, tlq.slowChan)
}
```

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+package controller
+
+import (
+	"k8s.io/client-go/util/workqueue"
+)
+
+// item wraps workqueue.Interface.Get() return
+type item struct {
+	i  interface{}
+	sd bool
```
mattmoor, line 26: This struct is overkill because `sd` is always false.  When it would be true we close the channel, which is actually a better signal overall, so why not just use `chan interface{}`?

controller/two_lane_queue.go
mattmoor: IMO when this processor has sent the item from `Get()` along the channel, that workqueue is `Done()` with that item.

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+			if fast {
+				select {
+				case item, ok := <-tlq.fastChan:
+					if !ok {
+						fast = false
+						continue
+					}
+					tlq.consumerQueue.Add(item.i)
+					continue
+				default:
```
mattmoor, line 105: It's probably worth documenting the hell out of the flow of control here for folks that don't understand all of the magic around Go channels.

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+			}
+			// Fast lane is empty, so wait for either channel to trigger. Since
+			// channels are picked randomly any of the lanes might trigger.
+			select {
+			case item, ok := <-tlq.fastChan:
+				if !ok {
+					fast = false
+					continue
+				}
+				tlq.consumerQueue.Add(item.i)
```
mattmoor, line 116: When `!fast` and there are items on `slowChan` are we sure that we will drain `slowChan`, or will the closed `fastChan` case dominate?

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Done implements workqueue.Interface.
+// Done marks the item as completed in all the queues.
+// NB: this will just re-enqueue the object on the queue that
+// didn't originate the object.
+func (tlq *twoLaneQueue) Done(i interface{}) {
+	tlq.consumerQueue.Done(i)
+	tlq.RateLimitingInterface.Done(i)
+	tlq.slowLane.Done(i)
```
mattmoor, line 144: IMO this should just be:
```suggestion
	tlq.consumerQueue.Done(i)
```

controller/two_lane_queue.go
```diff
@@ -0,0 +1,163 @@
...
+		tlq.consumerQueue.ShutDown()
+	}()
+}
+
+// Shutdown implements workqueue.Interace.
+// Shutdown shuts down both queues.
+func (tlq *twoLaneQueue) ShutDown() {
+	tlq.RateLimitingInterface.ShutDown()
+	tlq.slowLane.ShutDown()
+}
```
mattmoor, line 135: I wonder whether we should block until `tlq.consumerQueue` is shutting down?

controller/two_lane_queue.go
vagababov: it was more of a descriptive comment, but updated.

controller/two_lane_queue.go
vagababov: Sure that makes sense.

controller/two_lane_queue.go
vagababov: I guess from the original version. removed.

controller/two_lane_queue.go
vagababov: Yes it will drain. The order test is the opposite (slow is closed, but fast has on my machine a few hundred items that are drained).

controller/two_lane_queue.go
vagababov: Done.

controller/two_lane_queue.go
vagababov: Shutdown is a non blocking call on the queue, so why would it be here?

controller/two_lane_queue.go
```diff
@@ -91,39 +86,47 @@ func (tlq *twoLaneQueue) runConsumer() {
 	// Shutdown flags.
 	fast, slow := true, true
 	go func() {
+		// When both producer queues are shutdown stop the consumerQueue.
+		defer tlq.consumerQueue.ShutDown()
+		// While any of the queues is still running, try to read off of them.
 		for fast || slow {
 			// By default drain the fast lane.
+			// Channels in select are picked random, so first
+			// we have a select that only looks at the fast lane queue.
```
mattmoor, line 93: my point about documenting things was that I don't think it's clear to folks that `default:` allows you to escape if no channels are ready to receive.

mattmoor:
/lgtm
/approve
/hold

It would be good to import this into serving / eventing and make sure things don't blow up.

vagababov:
It's in serving: https://github.com/knative/serving/pull/8671

vagababov:
Meh, the latest one I removed controller.
I am going to merge this, since it's just a library for now and stage another to import to serving.

vaikas:
Do we get any metrics out of these queues by default? It would be nice to understand if we get into a state where the slow queue is for example never drained, or starts backing up, etc. I just took a quick look and didn't see any, do we get them from somewhere else?

mattmoor:
We register a metrics provider with the K8s `workqueue.SetProvider` hook here: https://github.com/knative/pkg/blob/e863db03445ff6094e8f81b8ede724116d11dfdd/controller/stats_reporter.go#L91

We should be able to track the "slow" queue with that.

vagababov:
Yep, we'll get automatically metrics from all 3 queues...

vaikas:
Cool, that's what I figured but just wanted to check 👍

PR #1516, opened 2020-07-17
Add more time for the checks

vagababov:
On the prow clusters which are usually strapped for CPU this fails due to wrong order of things.
So the final check should just wait for the channel to close rather than race a timer.

### DIFF
--- a/controller/two_lane_queue_test.go
+++ b/controller/two_lane_queue_test.go
@@ -49,7 +49,6 @@ func TestDoubleKey(t *testing.T) {
 	// Verifies that we don't get double concurrent processing of the same key.
 	q := newTwoLaneWorkQueue("live-in-the-fast-lane")
 	q.Add("1")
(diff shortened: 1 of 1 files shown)


controller/two_lane_queue_test.go
```diff
@@ -75,17 +75,12 @@ func TestDoubleKey(t *testing.T) {
...
+	case <-time.After(500 * time.Millisecond):
 		// Expected.
 	}
 	// This should permit the re-reading of the same key.
 	q.Done(k)
-	select {
-	case <-sentinel:
-		// Expected.
-	case <-time.After(200 * time.Millisecond):
-		t.Error("The item was not processed as expected")
```
mattmoor, line 87: I don't understand why we'd stop checking for this within a reasonable period of time?

vagababov: Yeah, I think the problem is differnet.

Issue #1509, opened 2020-07-16
webhook EOF errors


mattmoor:
/area API
/area test-and-release
/kind bug

## Expected Behavior

When we run our e2e tests with chaos there are no failures due to the webhook shutting down. 

## Actual Behavior

We intermittently see failures like this: `Post https://eventing-webhook.knative-eventing-qh1fjbnng8.svc:443/resource-conversion?timeout=30s: EOF` ever since we enabled webhook chaos.

mattmoor:
cc @vaikas

mattmoor:
I already made two changed to try and mitigate this:
1. Bumped the terminationGracePeriodSecond in our webhook pods to `300` to give the webhook itself greater control over the drain duration.
2. Bumped the `network.DefaultDrainTimeout`: https://github.com/knative/pkg/pull/1501 to give the API Server more time to observe the endpoints change.

mattmoor:
From the Go `http.Server` `Shutdown` documentation:

> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.

I wondered if this might be our problem, but we call `ListenAndServeTLS` here: https://github.com/knative/pkg/blob/f1b82401dc8a5951d6034be93a3a2423556a3104/webhook/webhook.go#L207-L210 and seem to properly check for `ErrServerClosed`

mattmoor:
I started a thread in `#sig-api-machinery` here: https://kubernetes.slack.com/archives/C0EG7JC6T/p1594910493127400

mattmoor:
Alright, so one of the things I have been wondering is around Keep-Alives and whether that might be a reason the API server takes so long to realize an endpoint is no longer good.  I noticed the following comment on [Go's `SetKeepAlivesEnabled`](https://golang.org/pkg/net/http/#Server.SetKeepAlivesEnabled):

> SetKeepAlivesEnabled controls whether HTTP keep-alives are enabled. By default, keep-alives are always enabled. Only very resource-constrained environments **or servers in the process of shutting down should disable them**.

... **emphasis** mine.


Now it turns out that Go's http logic is already smart about disabling keep-alive when it is shutting down:

```go
func (s *Server) doKeepAlives() bool {
	return atomic.LoadInt32(&s.disableKeepAlives) == 0 && !s.shuttingDown()
}
```

[...]

PR #1517, opened 2020-07-17
Implement a new shared "Drainer" handler.

mattmoor:
This implements a new `http.Handler` called `Drainer`, which is intended to wrap some inner `http.Handler` business logic with a new outer handler that can respond to Kubelet probes (successfully until told to "Drain()").

This takes over the webhook's relatively new probe handling and lame duck logic with one key difference.  Previously the webhook waited for a fixed period after SIGTERM before exitting, but the new logic waits for this same grace period AFTER THE LAST REQUEST.  So if the handler keeps getting (non-probe) requests, the timer will continually reset, and once it stops receiving requests for the configured grace period, "Drain()" will return and the webhook will exit.

The goal of this work is to try to better cope with what we believe to be high tail latencies of the API server seeing that a webhook replica is shutting down.

Related: https://github.com/knative/pkg/issues/1509

### DIFF
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handlers
+
+import (
+	"net/http"
+	"sync"
+	"time"
+
+	"knative.dev/pkg/network"
+)
+
+// Drainer wraps an inner http.Handler to support responding to kubelet
+// probes with a "200 OK" until the handler is told to Drain.
+// When the Drainer is told to Drain, it will immediately start to fail
+// probes with a "500 shutting down", and the call will block until no
+// requests have been received for QuietPeriod (defaults to
+// network.DefaultDrainTimeout).
+type Drainer struct {
+	// Mutex guards the initialization and resets of the timer
+	sync.Mutex
+
+	// Inner is the http.Handler to which we delegate actual requests.
+	Inner http.Handler
+
+	// QuietPeriod is the duration that must elapse without any requests
+	// after Drain is called before it may return.
+	QuietPeriod time.Duration
+
+	// once is used to initialize timer
+	once sync.Once
+
+	// timer is used to orchestrate the drain.
+	timer *time.Timer
+}
+
+// Ensure Drainer implements http.Handler
+var _ http.Handler = (*Drainer)(nil)
+
+// ServeHTTP implements http.Handler
+func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if network.IsKubeletProbe(r) { // Respond to probes regardless of path.
+		if d.draining() {
+			http.Error(w, "shutting down", http.StatusInternalServerError)
+		} else {
+			w.WriteHeader(http.StatusOK)
+		}
+		return
+	}
+
+	d.reset()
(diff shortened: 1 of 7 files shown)


network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+	// Mutex guards the initialization and resets of the timer
+	sync.Mutex
+
+	// Inner is the http.Handler to which we delegate actual requests.
+	Inner http.Handler
+
+	// QuietPeriod is the duration that must elapse without any requests
+	// after Drain is called before it may return.
+	QuietPeriod time.Duration
+
```
vagababov, line 43: I'd hide those as well and add a ctor

network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Ensure Drainer implements http.Handler
+var _ http.Handler = (*Drainer)(nil)
+
+// ServeHTTP implements http.Handler
+func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if network.IsKubeletProbe(r) { // Respond to probes regardless of path.
+		if d.draining() {
+			http.Error(w, "shutting down", http.StatusInternalServerError)
```
vagababov, line 58: 503?

network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Drainer wraps an inner http.Handler to support responding to kubelet
+// probes with a "200 OK" until the handler is told to Drain.
+// When the Drainer is told to Drain, it will immediately start to fail
+// probes with a "500 shutting down", and the call will block until no
+// requests have been received for QuietPeriod (defaults to
+// network.DefaultDrainTimeout).
+type Drainer struct {
+	// Mutex guards the initialization and resets of the timer
+	sync.Mutex
```
vagababov, line 35:
```suggestion
	sync.RWMutex
```

network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if d.timer.Stop() {
+		d.timer.Reset(d.QuietPeriod)
+	}
+}
+
+// draining returns whether we are draining the handler.
+func (d *Drainer) draining() bool {
+	d.Lock()
+	defer d.Unlock()
```
vagababov, line 102:
```suggestion
	d.RLock()
	defer d.RUnlock()
```
With RLock is definitely better, than the write lock above, but still atomic would probably be faster.

network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Drain blocks until QuietPeriod has elapsed since the last request,
+// starting when this is invoked.
+func (d *Drainer) Drain() {
+	d.once.Do(func() {
+		t := func() *time.Timer {
+			d.Lock()
+			defer d.Unlock()
+			if d.QuietPeriod == 0 {
```
vagababov, line 76:
```suggestion
			if d.QuietPeriod <= 0 {
```

network/handlers/drain.go
```diff
@@ -0,0 +1,105 @@
...
+	})
+}
+
+// reset resets the drain timer to the full amount of time.
+func (d *Drainer) reset() {
+	d.Lock()
+	defer d.Unlock()
+	if d.timer == nil {
+		return
+	}
```
vagababov, line 93: On every request lock/unlock is basically serializing everything.
Probably not the best idea.
You can probably replace it with a atomic 0/1 toggle, that's much cheaper to check.

network/handlers/drain.go
mattmoor: done

network/handlers/drain.go
mattmoor: Makes it more complicated to have an optional QuietPeriod 🤷

network/handlers/drain.go
mattmoor: With the tweaks I added it should be exclusively using the read locks until Drain is called, which should be cheap enough.  It also only wraps part of the request.

network/handlers/drain.go
mattmoor: done

network/handlers/drain.go
vagababov: just pass 0 as QP? ;-)

mattmoor: 🤢

PR #1521, opened 2020-07-19
Harden and reaarange the test for drain handler

vagababov:
/assign mattmoor


vagababov:
```
>rt -count=50
PASS
ok      knative.dev/pkg/network/handlers        11.396s

```
`rt == go test -race`

PR #1524, opened 2020-07-20
Not working queue metrics

evankanderson:
I noticed a couple bits of flakiness after #1392 went into `pkg/metrics`; one is in the new tests, which seem to lose the reporting race more often on Prow than on my computer, and one is in existing tests that seemed to be losing a race in the opposite direction and seeing an "add" based on a background timer before the test expected it.

/assign @jjzeng-seattle


evankanderson:
/close

Oh look, I sent that PR already.

PR #1525, opened 2020-07-20
Harden the drain tests more

vagababov:
- introduce synthetic timer that gives us the capability to actually trigger events in the drain itself
- unfortunately we still need time.After calls in the test due to the timing for go routines to finish (they race the default: case and lose).

OTOH, the happy runs take virtually no time now.
```
>rt -count=2200
PASS
ok      knative.dev/pkg/network/handlers        1.414s
```

/assign mattmoor

### DIFF
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -24,6 +24,26 @@ import (
 	"knative.dev/pkg/network"
 )
 
+type timer interface {
+	Stop() bool
+	Reset(time.Duration) bool
+	tickChan() <-chan time.Time
+}
+
+type sysTimer struct {
+	*time.Timer
+}
+
+func (s *sysTimer) tickChan() <-chan time.Time {
+	return s.Timer.C
+}
+
+var newTimer = func(d time.Duration) timer {
+	return &sysTimer{
+		time.NewTimer(d),
+	}
+}
+
 // Drainer wraps an inner http.Handler to support responding to kubelet
 // probes with a "200 OK" until the handler is told to Drain.
 // When the Drainer is told to Drain, it will immediately start to fail
@@ -45,7 +65,7 @@ type Drainer struct {
 	once sync.Once
 
 	// timer is used to orchestrate the drain.
-	timer *time.Timer
+	timer timer
 }
 
 // Ensure Drainer implements http.Handler
@@ -69,18 +89,20 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // Drain blocks until QuietPeriod has elapsed since the last request,
 // starting when this is invoked.
 func (d *Drainer) Drain() {
+	// Note: until the first caller exits, the others
+	// will wait blocked as well.
 	d.once.Do(func() {
-		t := func() *time.Timer {
+		t := func() timer {
 			d.Lock()
 			defer d.Unlock()
 			if d.QuietPeriod <= 0 {
 				d.QuietPeriod = network.DefaultDrainTimeout
 			}
-			d.timer = time.NewTimer(d.QuietPeriod)
+			d.timer = newTimer(d.QuietPeriod)
 			return d.timer
 		}()
 
-		<-t.C
+		<-t.tickChan()
 	})
 }
 
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -38,6 +38,8 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
+// This constructor is overridden in tests to control the progress
+// of time in the test.
 var newTimer = func(d time.Duration) timer {
 	return &sysTimer{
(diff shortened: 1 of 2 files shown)


network/handlers/drain.go
```diff
@@ -24,6 +24,26 @@ import (
...
+
+type sysTimer struct {
+	*time.Timer
+}
+
+func (s *sysTimer) tickChan() <-chan time.Time {
+	return s.Timer.C
+}
+
+var newTimer = func(d time.Duration) timer {
```
mattmoor, line 41: comment that we overwrite in tests?

vagababov: Done.

network/handlers/drain_test.go
```diff
@@ -55,41 +109,56 @@ func TestDrainMechanics(t *testing.T) {
...
 
 	select {
-	case <-time.After(40 * time.Millisecond):
-		// Drain is blocking.
 	case <-done:
 		t.Error("Drain terminated prematurely.")
+	case <-init:
+		// OK.
 	}
+	mt.advance(40 * time.Millisecond)
```
mattmoor, line 125: Given your fancy new mocking, why not 99 :)

network/handlers/drain_test.go
```diff
@@ -55,41 +109,56 @@ func TestDrainMechanics(t *testing.T) {
...
+	}
 
 	// Check for 503 as a probe response when shutting down.
 	resp = httptest.NewRecorder()
 	drainer.ServeHTTP(resp, probe)
 	if got, want := resp.Code, http.StatusServiceUnavailable; got != want {
 		t.Errorf("Probe status = %d, wanted %d", got, want)
 	}
+	// Verify no reset was called.
+	if mt.resetCalls != rc+1 {
```
mattmoor, line 141: I think this would read better if you captured it before the call, even if this is technically correct.

network/handlers/drain_test.go
```diff
@@ -55,41 +109,56 @@ func TestDrainMechanics(t *testing.T) {
...
 			t.Error("Drain terminated prematurely.")
+		default:
+			// OK
 		}
 		// For the last one we don't want to reset the drain timer.
 		if i < 2 {
 			drainer.ServeHTTP(w, req)
 		}
 	}
+	if mt.resetCalls != rc+3 {
```
mattmoor, line 158: If we're going to just use `+n`, why not start at zero and just say `n`?  We are partially assuming total knowledge :)

network/handlers/drain_test.go
vagababov: Why not even closer.

network/handlers/drain_test.go
vagababov: but `rc` is captured before the call? Or you want before the initial Drain?

network/handlers/drain_test.go
vagababov: Natural progress of things 😊 ?

network/handlers/drain_test.go
```diff
@@ -155,8 +163,9 @@ func TestDrainMechanics(t *testing.T) {
 			drainer.ServeHTTP(w, req)
 		}
 	}
-	if mt.resetCalls != rc+3 {
-		t.Errorf("ResetCalls = %d, want: %d", mt.resetCalls, rc+1)
+	// Two more drains should have been called.
+	if got, want := mt.resetCalls, rc+2; got != want {
+		t.Errorf("ResetCalls = %d, want: %d", got, want)
```
mattmoor, line 168: Move this into the `if` in the loop and do it in the same style?

network/handlers/drain_test.go
```diff
@@ -193,7 +202,7 @@ func TestDrainMechanics(t *testing.T) {
 		// Expected.
 	}
 
-	mt.advance(61 * time.Millisecond)
+	mt.advance(60 * time.Millisecond)
```
mattmoor, line 205: 60 seems arbitrary now.  Perhaps use `timeout` and add a comment noting the significance of this advance?  e.g. moving past our trigger point?

network/handlers/drain_test.go
```diff
@@ -74,6 +74,11 @@ func TestDrainMechanics(t *testing.T) {
 		}
 	)
 
+	const (
+		timeout   = 100 * time.Millisecond
+		razorThin = timeout - time.Nanosecond
```
mattmoor: I think having `epilon = ` here and `timeout-epsilon` below would probably be clearest, assuming the type system allows it.

vagababov: ok done.

network/handlers/drain_test.go
vagababov: sure.

network/handlers/drain_test.go
vagababov: just used `epsilon`.

PR #1527, opened 2020-07-20
Make the chaosduck jitter

mattmoor:
/hold

I want to import this downstream in serving (and maybe eventing) before I land this.

### DIFF
--- a/leaderelection/chaosduck/main.go
+++ b/leaderelection/chaosduck/main.go
@@ -30,6 +30,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/controller"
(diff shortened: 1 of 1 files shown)


leaderelection/chaosduck/main.go
```diff
@@ -45,13 +46,15 @@ type components map[string]sets.String
...
+	tributePeriod      time.Duration = 20 * time.Second
+	tributeFactor                    = 2.0
 )
 
 func init() {
 	// Note that we don't explicitly call flag.Parse() because ParseAndGetConfigOrDie below does this already.
 	flag.Var(&disabledComponents, "disable", "A repeatable flag to disable chaos for certain components.")
-	flag.DurationVar(&tributePeriod, "period", tributePeriod, "How frequently to terminate a leader pod per component.")
+	flag.DurationVar(&tributePeriod, "period", tributePeriod, "How frequently to terminate a leader pod per component (this is the base duration used with the jitter factor from -factor).")
+	flag.Float64Var(&tributeFactor, "factor", tributeFactor, "The jitter factor to apply to the period.")
```
vagababov, line 57: why not just `flag.Duration`, etc?

mattmoor: AFAIK can't do that for for `flag.Var`, so we needed `init()` anyways, so I avoided the annoying pointers you usually get 😉

mattmoor:
The downstream eventing PR is fine with this: https://github.com/knative/eventing/pull/3665
The serving PR was unhappy with what I based this on (included istio), but net-contour was happy: https://github.com/knative/serving/pull/8721

I backed out istio chaos from the baseline PR, so 🤞 this is good to go.

PR #1530, opened 2020-07-21
Another try to use hash.BucketSet for standard leader election mode

yanweiguo:
https://github.com/knative/pkg/pull/1490 was rolled back.

This PR fixes the bug that the bucket changes across for loop which is used in a callback asynchronously. Downstream CI/CD test: https://github.com/knative/serving/pull/8718 

/assign @mattmoor @vagababov


leaderelection/context.go
```diff
@@ -107,24 +106,20 @@ type standardBuilder struct {
...
-			index: i,
-			total: b.lec.Buckets,
-		}
-
+	bkts := newStandardBuckets(queueName, b.lec)
+	electors := make([]Elector, 0, b.lec.Buckets)
+	for i := range bkts {
+		// Use a local var which won't change across the for loop since it is
+		// used in a callback asynchronously.
+		bkt := bkts[i]
```
mattmoor, line 122: You can also use:

```suggestion
	for _, bkt := range bkts {
		// Use a local var which won't change across the for loop since it is
		// used in a callback asynchronously.
		bkt := bkt
```

That is a nasty one, good catch!

yanweiguo:
Merged https://github.com/knative/pkg/pull/1503 as well.

hash/bucketer_test.go
```diff
@@ -101,34 +101,47 @@ func TestBucketSetUpdate(t *testing.T) {
...
+	got := make([]string, 4)
+	for i, b := range bkts {
+		got[i] = b.Name()
 	}
-	_, err = bs.NewBucket("giboulée")
-	if err == nil {
-		t.Error("NewBucket unexpectedly succeeded")
+
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("got %v, want %v", got, want)
```
vagababov, line 116:
```suggestion
		t.Errorf("Got %v, want %v", got, want)
```
also print diff?

hash/bucketer_test.go
```diff
@@ -101,34 +101,47 @@ func TestBucketSetUpdate(t *testing.T) {
...
+func TestBucketSetBuckets(t *testing.T) {
 	bs := NewBucketSet(buckets)
-	_, err := bs.NewBucket(thisBucket)
-	if err != nil {
-		t.Error("NewBucket = ", err)
+	bkts := bs.Buckets()
+
+	// Sorted
+	want := []string{"aguacero", "chaparrón", "chubasco", "monsoon"}
+	got := make([]string, 4)
```
vagababov, line 110:
```suggestion
	got := make([]string, len(bkts))
```

hash/bucketer_test.go
yanweiguo: I think it's easy to get the diff from the output as the array isn't too big.

vagababov: 🤷 why make me work, when we can make the prow do the work :)

yanweiguo: fine. done

yanweiguo:
@googlebot rescan
@googlebot cla

PR #1531, opened 2020-07-21
Added logging of port forwarding stdout/stderr

slinkydeveloper:
This change modifies the signature of `PortForwarding` to include an stdout/stderr `io.Writer`. It also provides a writer that can be built from a logging function

### DIFF
--- a/test/monitoring/monitoring.go
+++ b/test/monitoring/monitoring.go
@@ -18,6 +18,7 @@ package monitoring
 
 import (
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


test/monitoring/monitoring.go
```diff
@@ -59,10 +60,10 @@ func Cleanup(pid int) error {
 }
 
 // PortForward sets up local port forward to the pod specified by the "app" label in the given namespace
-func PortForward(logf logging.FormatLogger, podList *v1.PodList, localPort, remotePort int, namespace string) (int, error) {
+func PortForward(logf logging.FormatLogger, podList *v1.PodList, localPort, remotePort int, namespace string, stdout io.Writer, stderr io.Writer) (int, error) {
```
vaikas, line 63: Maybe add a comment that stdout/stderr are where the results of the command are written. As in, why it's different from logf.

vaikas: Actually, can't we use the getcombinedoutput?
https://golang.org/src/os/exec/exec.go?s=15990:16036#L547

vaikas: is it because of the background bits?

vaikas: Since the kubectl should be fairly fast, I wonder if it is necessary to run it in the background instead of waiting for it to return and fail fast instead of later?

slinkydeveloper: It has to run in background because the process remains open

vaikas: yeah, I realized that, forgot to resolve the conversation, but stamped it :)

Issue #1523, opened 2020-07-20
Add metrics for leader election


yanweiguo:
/area API
/area monitoring

/kind feature

As this [feature track doc](https://docs.google.com/document/d/1emXelQiY0-cItq78QF5ZPvfpT6LD3KwXT0sCWM5cLy0/edit?ts=5ee82603#heading=h.n8a530nnrb) proposes, we need to add some metrics about leader election for observability.

PR #1535, opened 2020-07-22
Add owned bucket count metric for leader election

yanweiguo:
For #1523.

- Introduce a metric called `controller_owned_bucket_count` (welcome better name)
- Will report  that metric if both pod name and reconciler work queue name are available.
- Add work queue name to `LeaderAwareFuncs` for webhook controllers.

Verified in Knative serving.


reconciler/leader.go
```diff
@@ -109,9 +115,31 @@ func (laf *LeaderAwareFuncs) Demote(b Bucket) {
...
+		laf.reportBucketCount(len(laf.buckets))
 	}()
 
 	if demote := laf.DemoteFunc; demote != nil {
 		demote(b)
 	}
 }
+
+func (laf *LeaderAwareFuncs) reportBucketCount(count int) {
+	if laf.metricsCtx == nil {
```
vagababov, line 127: you have a race here since demote/promote can be independently called in parallel.
I recommend to put it either in ctor, inject it during construction or use a `sync.Once`.

reconciler/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	podNameKey        = tag.MustNewKey("pod_name")
+	reconcilerNameKey = tag.MustNewKey("reconciler_name")
+
+	metricKeys = []tag.Key{podNameKey, reconcilerNameKey}
+
+	podName = os.Getenv(podNameEnvKey)
+)
+
+func init() {
+	register()
```
vagababov, line 47: Any reason why we don't just inline it right now?

webhook/certificates/controller.go
```diff
@@ -33,6 +33,10 @@ import (
 	"knative.dev/pkg/webhook"
 )
 
+const (
+	workQueueName = "WebhookCertificates"
+)
+
```
vagababov, line 39:
```suggestion
const workQueueName = "WebhookCertificates"
```

webhook/configmaps/controller.go
```diff
@@ -35,6 +35,10 @@ import (
 	"knative.dev/pkg/webhook"
 )
 
+const (
+	workQueueName = "ConfigMapWebhook"
+)
+
 // NewAdmissionController constructs a reconciler
```
vagababov, line 42:
```suggestion
const workQueueName = "ConfigMapWebhook"

// NewAdmissionController constructs a reconciler
```

webhook/resourcesemantics/conversion/controller.go
```diff
@@ -34,6 +34,10 @@ import (
 	"knative.dev/pkg/webhook"
 )
 
+const (
+	workQueueName = "ConversionWebhook"
+)
+
```
vagababov, line 40:
```suggestion
const workQueueName = "ConversionWebhook"
```

webhook/resourcesemantics/defaulting/controller.go
```diff
@@ -35,6 +35,10 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics"
 )
 
+const (
+	workQueueName = "DefaultingWebhook"
+)
```
vagababov, line 40:
```suggestion
const workQueueName = "DefaultingWebhook"
```
Man, WH has many queues....

webhook/resourcesemantics/validation/controller.go
```diff
@@ -35,6 +35,10 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics"
 )
 
+const (
+	workQueueName = "ValidationWebhook"
+)
```
vagababov, line 40:
```suggestion
const workQueueName = "ValidationWebhook"
```
The torture never stops (c)

reconciler/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	podNameEnvKey = "POD_NAME"
+)
+
+var (
+	controllerOwnedBucketCountM = stats.Int64(
+		"controller_owned_bucket_count",
+		"Number of buckets a controller Pod owns for a reconciler",
+		stats.UnitDimensionless)
+
+	podNameKey        = tag.MustNewKey("pod_name")
```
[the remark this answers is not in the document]
yanweiguo: not in pkg repo.

reconciler/leader.go
yanweiguo: We have write locks in promote/demote which can prevent the race?

What's ctor? Even though with construction or `sync.Once`, we need to check whether it is nil every time, right?

reconciler/metrics.go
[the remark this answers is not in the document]
vagababov: hm, I thought they were in pkg, but 🤷

reconciler/leader.go
vagababov: Not really. After `sync.Once` call, which is the only place where you set the value you can check for it to be nil without mutex, since you know it will always be nil.

yanweiguo:
> Do we actually serialize promote/demote now?

Part of them are serialized by lock, where I added the metric reporting: https://github.com/knative/pkg/blob/1127f59d7b73b75db42fed5ba9acce04ce4e1d77/reconciler/leader.go#L92

reconciler/leader.go
yanweiguo: Added Once.

reconciler/leader.go
```diff
@@ -65,10 +68,13 @@ type LeaderAware interface {
...
-	buckets map[string]Bucket
+	buckets    map[string]Bucket
+	metricsCtx context.Context
+	once       sync.Once
 
-	PromoteFunc func(b Bucket, enq func(Bucket, types.NamespacedName)) error
-	DemoteFunc  func(b Bucket)
+	WorkQueueName string
+	PromoteFunc   func(b Bucket, enq func(Bucket, types.NamespacedName)) error
+	DemoteFunc    func(b Bucket)
```
mattmoor, line 77: This (how much we have to add here) is pretty rough, and I don't really think it's worth it if we have to pollute this so much 😬

I was thinking that this logic would be in `controller.Impl`, which already has the `Name`.  Ack that the LeaderAware interface likely doesn't expose enough metadata for metrics as-is, but perhaps instead we can:
1. Expose an additional interface that `controller.Impl` can consume to emit metrics, and
2. Have `LeaderAwareFuncs` implement that interface with the metadata it has.

If that doesn't make sense, I'd be happy to jump on a quick call and we can hash it out.

yanweiguo: Yeah that makes sense. In that way, I think we need a timer in `controller.Impl` to report metrics periodically instead of reporting when promoting/demoting. Is that right?

yanweiguo: Ping.

mattmoor: Maybe we should dig into https://godoc.org/k8s.io/client-go/tools/leaderelection#SetProvider ?

yanweiguo: https://godoc.org/k8s.io/client-go/tools/leaderelection#SwitchMetric only has a parameter for `On` and `Off` which is the  identity of the leader. Only the first call of `SetProvider` has an effect. IIUC, we'd like to count owned buckets on reconciler level instead of pod level. Unless we wrap reconciler name into the identity, we can't distinguish among reconcilers. @mattmoor WDYT?

vagababov:
Is this still ongoing?

markusthoemmes:
@yanweiguo any news?

reconciler/leader.go
yanweiguo: @mattmoor ping.

I don't think https://godoc.org/k8s.io/client-go/tools/leaderelection#SetProvider provides a way to set useful labels. Should we just implement our own logic?

mattmoor: File an issue upstream?

Issue #1541, opened 2020-07-22
Turning on logging will give twice the output.


vaikas:
## Expected Behavior
Enabling logging for controllers using -v 7 (for example) to get HTTP logs would give me logs.

## Actual Behavior
It does, but it gives them twice. Looks like this change was unnecessary and now we wrap things twice so you get twice the logs:
https://github.com/knative/pkg/pull/1515

## Steps to Reproduce the Problem

1. Start your controller with a -v 7 (for example)
2. look at the logs
3. you'll see every request there twice.

PR #1542, opened 2020-07-22
kube already gives us this, do not double log http

vaikas:
Fixes: https://github.com/knative/pkg/issues/1541


mattmoor:
I guess we only needed the `klog.Init`?

vaikas:
yeah. I left the decorate blog there also since that seemed handy.

PR #1543, opened 2020-07-22
Allow creating controller with custom WorkQueue.

Cynocracy:
Which was possible before via field modification.
Not switching to a builder pattern mostly for speed of resolution.
Happy to consider alternatives (including renaming the field back).


Cynocracy:
Need to think about this more, the interface isn't even public.

vagababov:
if we need custom queue we will need to inject it via ctor.
But mostly: why?

Cynocracy:
Rate limiting.

PR #1545, opened 2020-07-23
Fix the race in the test

vagababov:
The test was inherently racy and I saw it flake more than once (e.g. https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/1526/pull-knative-pkg-unit-tests/1286086935722332160)
So this fixes it.
Also other minor nits around the test to make it more readable.

/assign mattmoor @yanweiguo


yanweiguo:
Thanks for the fix.

PR #1546, opened 2020-07-23
Allow creating controller with custom RateLimiter.

Cynocracy:
Which was possible before via field modification.
Not switching to a builder pattern mostly for speed of resolution.
Happy to consider alternatives.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -205,34 +205,37 @@ type Impl struct {
 	statsReporter StatsReporter
 }
 
+// Encapsulates options for creating a new controller, including
+// throttling and stats behavior.
+type GenericOptions struct {
+	reporter StatsReporter
+        rateLimiter workqueue.RateLimiter
+}
+
 // NewImpl instantiates an instance of our controller that will feed work to the
 // provided Reconciler as it is enqueued.
 func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string) *Impl {
-	return NewImplFull(r, logger, workQueueName, MustNewStatsReporter(workQueueName, logger), nil)
-}
-
-func NewImplWithRateLimiter(r Reconciler, logger *zap.SugaredLogger, workQueueName string, rateLimiter *workqueue.RateLimiter) *Impl {
-	return NewImplFull(r, logger, workQueueName, nil, rateLimiter)
+	return NewImplFull(r, logger, workQueueName, GenericOptions{})
 }
 
 func NewImplWithStats(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter) *Impl {
-	return NewImplFull(r, logger, workQueueName, reporter, nil)
+	return NewImplFull(r, logger, workQueueName, GenericOptions{reporter: reporter})
 }
 
-func NewImplFull(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter, rateLimiter *workqueue.RateLimiter) *Impl {
+func NewImplFull(r Reconciler, logger *zap.SugaredLogger, workQueueName string, options GenericOptions) *Impl {
 	logger = logger.Named(workQueueName)
-	var rl workqueue.RateLimiter
-	if rateLimiter == nil {
-		rl = workqueue.DefaultControllerRateLimiter()
-	} else {
-		rl = *rateLimiter
+	if options.rateLimiter == nil {
+		options.rateLimiter = workqueue.DefaultControllerRateLimiter()
+	}
+	if options.reporter == nil {
+		options.reporter = MustNewStatsReporter(workQueueName, logger)
 	}
 	return &Impl{
 		Name:          workQueueName,
 		Reconciler:    r,
-		workQueue:     newTwoLaneWorkQueue(workQueueName, rl),
+		workQueue:     newTwoLaneWorkQueue(workQueueName, options.rateLimiter),
 		logger:        logger,
-		statsReporter: reporter,
+		statsReporter: options.reporter,
 	}
 }
 
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -211,12 +211,12 @@ func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string) *Imp
 	return NewImplFull(r, logger, workQueueName, MustNewStatsReporter(workQueueName, logger), nil)
 }
 
-func NewImplWithStats(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter) *Impl {
-        return NewImplFull(r, logger, workQueueName, reporter, nil)
+func NewImplWithRateLimiter(r Reconciler, logger *zap.SugaredLogger, workQueueName string, rateLimiter *workqueue.RateLimiter) *Impl {
+	return NewImplFull(r, logger, workQueueName, nil, rateLimiter)
 }
 
-func NewImplWithLimiter(r Reconciler, logger *zap.SugaredLogger, workQueueName string, ratelimiter *workqueue.RateLimiter) *Impl {
-        return NewImplFull(r, logger, workQueueName, MustNewStatsReporter(workQueueName, logger), ratelimiter)
+func NewImplWithStats(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter) *Impl {
+	return NewImplFull(r, logger, workQueueName, reporter, nil)
 }
 
 func NewImplFull(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter, rateLimiter *workqueue.RateLimiter) *Impl {
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -208,15 +208,29 @@ type Impl struct {
 // NewImpl instantiates an instance of our controller that will feed work to the
 // provided Reconciler as it is enqueued.
 func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string) *Impl {
-	return NewImplWithStats(r, logger, workQueueName, MustNewStatsReporter(workQueueName, logger))
+	return NewImplFull(r, logger, workQueueName, MustNewStatsReporter(workQueueName, logger), nil)
 }
 
 func NewImplWithStats(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter) *Impl {
+        return NewImplFull(r, logger, workQueueName, reporter, nil)
+}
+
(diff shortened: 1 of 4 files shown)


Cynocracy:
Covering the new lines sounds hard, so I'd prefer not to. Unless you feel very strongly about it D:

controller/controller.go
```diff
@@ -208,15 +208,29 @@ type Impl struct {
...
+}
+
+func NewImplFull(r Reconciler, logger *zap.SugaredLogger, workQueueName string, reporter StatsReporter, rateLimiter *workqueue.RateLimiter) *Impl {
 	logger = logger.Named(workQueueName)
+	var rl workqueue.RateLimiter
+	if rateLimiter == nil {
+		rl = workqueue.DefaultControllerRateLimiter()
+	} else {
+		rl = *rateLimiter
+	}
```
vagababov, line 229: I'd actually push this to the queue, but 🤷 
This should work too.

controller/controller.go
```diff
@@ -208,15 +208,29 @@ type Impl struct {
 // NewImpl instantiates an instance of our controller that will feed work to the
 // provided Reconciler as it is enqueued.
 func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string) *Impl {
```
evankanderson, line 210: This is starting to look like a good candidate for [the options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

Cynocracy: +1, I can switch.

vagababov: They should share the RL I think? So just one of them?

Cynocracy: Not sure I understand your comment.

My issues is that there is no func(blah *blah) that can modify the ratelimiter, as it's private on the rate limiting queue interface.

So really the only 'options' pattern I can think of is

1. a struct with the ratelimiter on it
2. an interface that can return a ratelimiter

and I don't see any reason to prefer those over this, unless y'all agree 1 is better (with an Options that can contain a statslogger and/or ratelimiter)

vagababov: Not sure I follow, just NewImpl(opts Options).
We should not modify rate limiter. We should either pass one created from outside, or just use the default (what we do right now).

Cynocracy:
Thanks, sounds good.

I wasn't sure if that was as preferable as the function passing style if
that pattern didn't work.

controller/controller.go
```diff
@@ -205,20 +205,37 @@ type Impl struct {
 	statsReporter StatsReporter
 }
 
+// Encapsulates options for creating a new controller, including
+// throttling and stats behavior.
+type GenericOptions struct {
+	reporter StatsReporter
+        rateLimiter workqueue.RateLimiter
```
mattmoor, line 212: Format Go code:
```suggestion
	reporter    StatsReporter
	rateLimiter workqueue.RateLimiter
```

controller/controller_test.go
```diff
@@ -424,6 +424,15 @@ func (nr *NopReconciler) Reconcile(context.Context, string) error {
 	return nil
 }
 
+type testRateLimiter struct {
+	t *testing.T
+        delay time.Duration
```
mattmoor, line 429: Format Go code:
```suggestion
	t     *testing.T
	delay time.Duration
```

controller/controller_test.go
```diff
@@ -424,6 +424,15 @@ func (nr *NopReconciler) Reconcile(context.Context, string) error {
...
+}
+func (t testRateLimiter) When(interface {}) time.Duration {  return t.delay }
+func (t testRateLimiter) Forget(interface {}) {}
+func (t testRateLimiter) NumRequeues(interface{}) int { return 0 }
```
mattmoor, line 433: Format Go code:
```suggestion

func (t testRateLimiter) When(interface{}) time.Duration { return t.delay }
func (t testRateLimiter) Forget(interface{})             {}
func (t testRateLimiter) NumRequeues(interface{}) int    { return 0 }

```

controller/two_lane_queue_test.go
```diff
@@ -22,10 +22,73 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/workqueue"
 )
 
+type chanRateLimiter struct {
+	t *testing.T
+	// Called when this ratelimiter is consulted for when to process a value.
+	whenCalled chan interface{}
+	retryCount map[interface{}]int
+}
```
mattmoor, line 33: Format Go code:
```suggestion
}

```

controller/two_lane_queue_test.go
```diff
@@ -22,10 +22,73 @@ import (
...
+func (r chanRateLimiter) When(item interface{}) time.Duration {
+	r.whenCalled<- item
```
mattmoor, line 35: Format Go code:
```suggestion
	r.whenCalled <- item
```

controller/two_lane_queue_test.go
```diff
@@ -22,10 +22,73 @@ import (
...
+	return 0 * time.Second
+}
+func (r chanRateLimiter) Forget(item interface{}) {
+	r.t.Fatalf("Forgetting item %+v, we should not be forgetting any items.", item)
+}
+func (r chanRateLimiter) NumRequeues(item interface{}) int {
+	return r.retryCount[item]
+}
```
mattmoor, line 43: Format Go code:
```suggestion
}

```

controller/controller.go
Cynocracy: Added an options at the end, wdyt?

Anything further and we risk signficant backward incompatible changes, which I don't want to take the responsibility for rolling out across the federated repos.

Cynocracy:
I disagree with that, mostly because go modules bite back when you make backward incompatible changes.

If noone uses NewImplFull though, I'd be fine with that. Let me check.

controller/controller.go
Cynocracy: Ah, Lol, NewImplFull is one I made up. Sure, Options can include the other stuff

vagababov:
😁

Cynocracy:
:rofl:

controller/controller_test.go
```diff
@@ -424,6 +424,17 @@ func (nr *NopReconciler) Reconcile(context.Context, string) error {
...
+type testRateLimiter struct {
+	t     *testing.T
+	delay time.Duration
+}
+
+func (t testRateLimiter) When(interface{}) time.Duration { return t.delay }
+func (t testRateLimiter) Forget(interface{})             {}
+func (t testRateLimiter) NumRequeues(interface{}) int    { return 0 }
+
+var _ workqueue.RateLimiter = &testRateLimiter{}
```
vagababov, line 436:
```suggestion
var _ workqueue.RateLimiter = (*testRateLimiter)(nil)
```
is usually how we roll.

controller/controller.go
evankanderson: I was trying to suggest

```go
type Option func(*Impl)

func NewImpl(r Reconciler, logger *zap.SugaredLogger, workQueueName string, opts ...Option) {
  ret := &Impl{
    Name: workQueueName,
    Reconciler: r,
    logger: logger.Named(workQueueName),
    statsReporter: MustNewStatsReporter(workQueueName, logger),  // assumes statsreporter is cheap to create
  }
  for _, o := range opts {
    o.Apply(ret)
  }
  if ret.workQueue == nil {  // Avoid creating a new workqueue if that is expensive
    ret.workQueue: newTwoLaneWorkQueue(workQueueName, workqueue.DefaultControllerRateLimiter())
  }
  return ret
}

func WithLimit(rateLimiter *workqueue.RateLimiter) Option {
  return func(i *Impl) {
    i.workQueue = newTwoLaneWorkQueue(i.Name, rateLimiter)
  }
}

func WithReporter(reporter StatsReporter) Option {
  return func(i *Impl) {
    i.statsReporter = reporter
  }
}
```

But I think this form is fine, too.

Cynocracy:
Ohh. That makes sense. I don't see a good reason to prefer this over the
other right now but thank you for explaining how you saw it working!

I guess the trick is to use the fact that you're within the package to
mutate the object. Doesn't that get confusing when the same field could
easily be modified multiple times?

evankanderson:
> Ohh. That makes sense. I don't see a good reason to prefer this over the other right now but thank you for explaining how you saw it working! I guess the trick is to use the fact that you're within the package to mutate the object. Doesn't that get confusing when the same field could easily be modified multiple times?

The idea is to build up a series of modifications, so you could write:

```golang
c := NewImpl(r, logger, "foo", WithLimit(small), WithReporter(s))
```

If you call "WithReporter" more than once, you'll only get the last reporter, but you could have WithReporter fail and change the Option signature to include an error return if you wanted.

vagababov:
This usually means you still need an `options` struct to apply these modifications to (not necessarily, but usually). Also this might lead to exposing some fields to be public, but 🤷 
I am game either way.

evankanderson:
I used `Impl` as the options struct here, but Victor is right. And the functional options pattern tends to provide a sealed interface by using package visibility to provide functions to set package-private fields on the options or implementation struct.

In any case, you currently only have 2x2 = 4 simple methods vs 1 complicated + 2 simple methods, so the cost is not too high. Once you get 3 optional arguments, the cost is 3x3 = 9 simple methods or 1 complicated + 3 simple = 4 methods, and that's probably the tipping point.

Cynocracy:
I see. Thanks, that makes sense. Keeping the interface sealed seems like a good reason to prefer that, though as you mentioned, there's not much difference unless you want to avoid needing to expose fields on an Option/the underlying struct unnecessarily.

Anyway, I'm happy with the PR now. Anything you'd prefer to see changed before merging?

Cynocracy:
Ah right forgot about the lint. Will do

controller/controller_test.go
Cynocracy: Cool! I like it.

Cynocracy:
/retest

I'm a little concerned this means my test is flaky :/

controller/controller.go
```diff
@@ -205,20 +205,40 @@ type Impl struct {
 	statsReporter StatsReporter
 }
 
+// ControllerOptions encapsulates options for creating a new controller,
+// including throttling and stats behavior.
+type ControllerOptions struct {
```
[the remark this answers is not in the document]
Cynocracy: There's a conflict with controller.Options, so I reverted this part.

Cynocracy:
/retest

Metrics that time.

evankanderson:
/hold cancel

I was fine with any interface, I just wanted to point out the interface alternatives to the java "chained constructor" pattern.

PR #1554, opened 2020-07-24
Make EnsureRecorded part of metricstest

evankanderson:
Call EnsureRecorded automatically in the "Assert" methods.

This should make testing serving/eventing easier.

Increase test coverage.

/assign @jjzeng-seattle

### DIFF
--- a/metrics/metricstest/resource_metrics_test.go
+++ b/metrics/metricstest/resource_metrics_test.go
@@ -21,12 +21,16 @@ limitations under the License.
 package metricstest
 
 import (
+	"context"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"go.opencensus.io/metric/metricdata"
 	"go.opencensus.io/resource"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
 )
 
 func TestMetricEqual(t *testing.T) {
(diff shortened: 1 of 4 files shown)


metrics/metricstest/resource_metrics_test.go
```diff
@@ -315,3 +319,53 @@ func TestMetricShortcuts(t *testing.T) {
...
+	view.Register(countView)
+
+	ctx, err := tag.New(context.Background(), tag.Upsert(tagKey, "alpha"))
+	if err != nil {
+		t.Errorf("Unable to create context: %v", err)
+	}
+	stats.Record(ctx, count.M(5))
+	stats.Record(ctx, count.M(3))
+
+	AssertMetric(t, Metric{Name: "count"})
```
yanweiguo, line 341: Can we use `AssertMetricExists(t, "count")` here?

metrics/metricstest/resource_metrics_test.go
```diff
@@ -315,3 +319,53 @@ func TestMetricShortcuts(t *testing.T) {
...
+	}
+
+	view.Register(countView)
+
+	ctx, err := tag.New(context.Background(), tag.Upsert(tagKey, "alpha"))
+	if err != nil {
+		t.Errorf("Unable to create context: %v", err)
+	}
+	stats.Record(ctx, count.M(5))
+	stats.Record(ctx, count.M(3))
```
yanweiguo, line 339: Does the second `Record` ensures the first `Record` got recorded so we don't need to call `EnsureRecorded` here?

evankanderson: Record pushes items into a `chan` of commands to a background goroutine managed by the Meter, but doesn't await a return message on the channel.

`EnsureRecorded` calls a function on the Meter that pushes a command onto the same `chan` and then waits for a response message on a return `chan`. Therefore, this acts as a barrier to ensure that all earlier commands on the `chan` have been processed, including all the previous `Record` calls. Since each Meter has its own background thread, we need to send a command on each (therefore the loop across all `metricproducers` on line 111 of `resource_metrics.go`).

metrics/metricstest/resource_metrics_test.go
evankanderson: Yep; I picked AssertMetrics because the code seemed like it would be slightly more complex, but now I realize I could make AssertMetricExists call through to AssertMetric (done).

evankanderson: This does make the error slightly more opaque ("Wrong metric <comparison>", rather than "couldn't find metric %q"), but that seems acceptable for a test helper.

evankanderson:
Actually, I'm going to sneak in a `metricstest.EnsureRecorded()` in the prometheus case, since it seems like it's hitting the same error.

PR #1560, opened 2020-07-27
Remove our copy of the proxy watcher.

markusthoemmes:
We've long bumped the K8s dependencies. Deleted code is good code.

/assign @mattmoor


vagababov:
matt's out for the week and it seems unused in the main repos anyway, so yolo
/unhold

PR #1566, opened 2020-07-28
Make sure metrics tests are re-entrant

vagababov:
Currently they are not. 'go test -count=2' will miserably fail.
So ensure they are. Also add a helper to return current value (this might be a better way than trying to jump through hoops to erase everything).

Other minor nits.

/assign @evankanderson @yanweiguo

### DIFF
--- a/metrics/metricstest/metricstest.go
+++ b/metrics/metricstest/metricstest.go
@@ -118,9 +118,31 @@ func CheckDistributionCount(t test.T, name string, wantTags map[string]string, e
 
 }
 
+// GetLastValueData returns the last value for the given metric, verifying tags.
+func GetLastValueData(t test.T, name string, tags map[string]string) float64 {
+	t.Helper()
+	return GetLastValueDataWithMeter(t, name, tags, nil)
+}
+
+// GetLastValueDataWithMeter returns the last value of the given metric using meter, verifying tags.
+func GetLastValueDataWithMeter(t test.T, name string, tags map[string]string, meter view.Meter) float64 {
+	t.Helper()
+	if row := lastRow(t, name, meter); row != nil {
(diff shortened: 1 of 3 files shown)


metrics/metricstest/metricstest.go
```diff
@@ -118,9 +118,31 @@ func CheckDistributionCount(t test.T, name string, wantTags map[string]string, e
...
 }
 
+// GetLastValueData returns the last value for the given metric, verifying tags.
+func GetLastValueData(t test.T, name string, tags map[string]string) float64 {
+	t.Helper()
+	return GetLastValueDataWithMeter(t, name, tags, nil)
+}
+
+// GetLastValueDataWithMeter returns the last value of the given metric using meter, verifying tags.
+func GetLastValueDataWithMeter(t test.T, name string, tags map[string]string, meter view.Meter) float64 {
```
yanweiguo, line 128: Could we use this func in `CheckLastValueDataWithMeter`? Looks like we can just

```
func CheckLastValueDataWithMeter(t test.T, name string, wantTags map[string]string, wantValue float64, meter view.Meter) {
  t.Helper()
  if GetLastValueDataWithMeter(t, name, tags, meter) != wantValue {
    t.Error("Reporter.Report() wrong value", "metric", name, "got", s.Value, "want", wantValue)
  }
}
```

vagababov: Sure.

vagababov: DOne.

evankanderson:
/lgtm

I can look at converting this to the new helpers and maybe adding better support for "incremented by X" in general.

vagababov:
Yeah those would be much more helpful to check the diffs rather than absolute values — will make most of the metrics testing easier.

PR #1568, opened 2020-07-29
use const when gen reconciler

runzexia:


whaught:
> This reference to the constant is in the generated code, but the declaration of the constant is not (and private scoped). Does this work when we build generated reconcilier from it?

I take it back, the constants are also generated. I wonder if we want to move the const declaration to 	"knative.dev/pkg/reconciler" and reference it instead?

runzexia:
> > This reference to the constant is in the generated code, but the declaration of the constant is not (and private scoped). Does this work when we build generated reconcilier from it?
> 
> I take it back, the constants are also generated. I wonder if we want to move the const declaration to "knative.dev/pkg/reconciler" and reference it instead?
> 
> Would you run `./hack/update-codegen.sh` on this change to generate this change for /pkg?

ok, i move it to `knative.dev/pkg/reconciler` package and run `./hack/update-codegen.sh`

whaught:
> > > This reference to the constant is in the generated code, but the declaration of the constant is not (and private scoped). Does this work when we build generated reconcilier from it?
> > 
> > 
> > I take it back, the constants are also generated. I wonder if we want to move the const declaration to "knative.dev/pkg/reconciler" and reference it instead?
> > Would you run `./hack/update-codegen.sh` on this change to generate this change for /pkg?
> 
> ok, i move it to `knative.dev/pkg/reconciler` package and run `./hack/update-codegen.sh`

I think we need to also update "pkg/codegen/cmd/injection-gen/generators/reconciler_state.go" now that the constant isn't local?

whaught:
https://gist.github.com/whaught/fd236804c4155dbf22657e778ac2b637

I think you'll want to reference the external const like this

runzexia:
> https://gist.github.com/whaught/fd236804c4155dbf22657e778ac2b637
> 
> I think you'll want to reference the external const like this

cool👍

PR #1570, opened 2020-07-30
Allow reconcilers to watch more than once config store

whaught:
## Proposed Change
- Create a list of config stores in Controller Options
    - Implementers may snapshot many config stores and attach to config
    - Use case: serving garbage collection has its own config and will listen to the features config

- The existing single option is left for backwards-compatibility for consumers of /pkg. We can update downstream over time to use the new field before removing.


whaught:
/close

Looks like we instead want to duplicate configs into store structs for each reconciler

PR #1575, opened 2020-08-01
Set reporting period for all meters

jjzeng-seattle:
In addition, set it for all newly-created meters.


metrics/config.go
```diff
@@ -296,6 +296,8 @@ func createMetricsConfig(ops ExporterOptions, logger *zap.SugaredLogger) (*metri
 		mc.reportingPeriod = 60 * time.Second
 	} else if mc.backendDestination == prometheus {
 		mc.reportingPeriod = 5 * time.Second
+	} else if mc.backendDestination == openCensus {
+		mc.reportingPeriod = 60 * time.Second
 	}
```
yanweiguo, line 301: Change this `else if` section to a `switch`?

metrics/resource_view.go
```diff
@@ -186,6 +199,11 @@ func meterExporterForResource(r *resource.Resource) *meterExporter {
 	mE.m = view.NewMeter()
 	mE.m.SetResource(r)
 	mE.m.Start()
+
+	mc := getCurMetricsConfig()
+	if mc != nil {
+		mE.m.SetReportingPeriod(mc.reportingPeriod)
```
yanweiguo, line 205: Is it possible to add an unit test for it?

metrics/config_test.go
```diff
@@ -281,6 +281,7 @@ var (
 			domain:             servingDomain,
 			component:          testComponent,
 			backendDestination: openCensus,
+			reportingPeriod:    60 * time.Second,
```
vagababov, line 284:
```suggestion
			reportingPeriod:    time.Minute,
```

metrics/config.go
jjzeng-seattle: done.

metrics/resource_view.go
jjzeng-seattle: I thought about it.  It's hard to verify the internal of a meter.  Testing the effect(waiting for the reports) is not suitable for unit tests.

metrics/config_test.go
jjzeng-seattle: done.

yanweiguo: no change here.

PR #1577, opened 2020-08-03
Add field to observability config for enabling request logging

skonto:
This is required in order to fix this issue: https://github.com/knative/serving/issues/8644.

### DIFF
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -59,6 +59,9 @@ type ObservabilityConfig struct {
 	// EnableProfiling indicates whether it is allowed to retrieve runtime profiling data from
 	// the pods via an HTTP server in the format expected by the pprof visualization tool.
 	EnableProfiling bool
+
+	// EnableRequestLogging enables activator/queue-proxy to write request logs.
+	EnableRequestLogging bool
 }
 
 func defaultConfig() *ObservabilityConfig {
@@ -77,6 +80,7 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		cm.AsString("logging.revision-url-template", &oc.LoggingURLTemplate),
 		cm.AsString("logging.request-log-template", &oc.RequestLogTemplate),
 		cm.AsBool("logging.enable-probe-request-log", &oc.EnableProbeRequestLog),
+		cm.AsBool("logging.enable-request-logging", &oc.EnableRequestLogging),
 		cm.AsString("metrics.request-metrics-backend-destination", &oc.RequestMetricsBackend),
 		cm.AsBool("profiling.enable", &oc.EnableProfiling),
 	); err != nil {
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -60,14 +60,15 @@ type ObservabilityConfig struct {
 	// the pods via an HTTP server in the format expected by the pprof visualization tool.
 	EnableProfiling bool
 
-	// EnableRequestLogging enables activator/queue-proxy to write request logs.
-	EnableRequestLogging bool
+	// EnableRequestLog enables activator/queue-proxy to write request logs.
+	EnableRequestLog bool
 }
 
 func defaultConfig() *ObservabilityConfig {
 	return &ObservabilityConfig{
 		LoggingURLTemplate:    DefaultLogURLTemplate,
 		RequestMetricsBackend: defaultRequestMetricsBackend,
+		EnableRequestLog:      true,
 	}
 }
 
@@ -80,7 +81,7 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		cm.AsString("logging.revision-url-template", &oc.LoggingURLTemplate),
 		cm.AsString("logging.request-log-template", &oc.RequestLogTemplate),
 		cm.AsBool("logging.enable-probe-request-log", &oc.EnableProbeRequestLog),
-		cm.AsBool("logging.enable-request-logging", &oc.EnableRequestLogging),
+		cm.AsBool("logging.enable-request-log", &oc.EnableRequestLog),
 		cm.AsString("metrics.request-metrics-backend-destination", &oc.RequestMetricsBackend),
 		cm.AsBool("profiling.enable", &oc.EnableProfiling),
 	); err != nil {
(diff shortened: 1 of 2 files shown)


metrics/config_observability.go
```diff
@@ -77,6 +80,7 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		cm.AsString("logging.revision-url-template", &oc.LoggingURLTemplate),
 		cm.AsString("logging.request-log-template", &oc.RequestLogTemplate),
 		cm.AsBool("logging.enable-probe-request-log", &oc.EnableProbeRequestLog),
+		cm.AsBool("logging.enable-request-logging", &oc.EnableRequestLogging),
```
markusthoemmes, line 83: Will we need to do "smart" defaulting here? If the template is set I guess we should default to `true`, otherwise to `false` to be able to upgrade safely?

yanweiguo, line 83: should we use `logging.enable-request-log` to be consistent with `logging.enable-probe-request-log`?

skonto: @markusthoemmes depends on the semantics. In other frameworks eg. log4j you might set LEVEL.OFF to disable output, that does not mean that configuration does not exist. I am not sure if we want to externalize this `http` convention about the template. I might want to keep the template but only turn the logging off. The next time I might want to turn it by applying the existing, no need to re-apply the template.

skonto: @yanweiguo could be done.

metrics/config_observability.go
```diff
@@ -83,6 +87,16 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		return nil, err
 	}
 
+	if raw, ok := configMap.Data["logging.enable-request-log"]; ok {
+		if strings.EqualFold(raw, "true") && oc.RequestLogTemplate != "" {
```
skonto, line 91: I could start with `true` in the default config above for the flag and then parse it along with the others but this way it is more clear.

metrics/config_observability_test.go
```diff
@@ -48,6 +49,7 @@ func TestObservabilityConfiguration(t *testing.T) {
 			"logging.request-log-template":                `{"requestMethod": "{{.Request.Method}}"}`,
 			"logging.revision-url-template":               "https://logging.io",
 			"logging.write-request-logs":                  "true",
```
yanweiguo, line 51: This flag seems not used. Could you remove it?

metrics/config_observability.go
```diff
@@ -83,6 +87,16 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
...
+			oc.EnableRequestLog = true
+		}
+	} else {
+		if oc.RequestLogTemplate != "" {
+			oc.EnableRequestLog = true
+		}
+	}
```
yanweiguo, line 98:
```suggestion
	} else if oc.RequestLogTemplate != "" {
		// If logging.enable-request-log is not explicitly set, defaults to true if the template is not empty.
		oc.EnableRequestLog = true
	}
```

metrics/config_observability.go
yanweiguo, line 91: Could you add unit tests for 1) flag set to true but template not set 2) flag not set but template set?

metrics/config_observability.go
yanweiguo: I agree with what you proposed.

metrics/config_observability.go
skonto: I didnt like to put the if there because usually I would expect an `else` somewhere and it helps readability imho, but I can optimize it.

metrics/config_observability.go
skonto: sure

metrics/config_observability.go
markusthoemmes: Yep, that indeed sounds good!

metrics/config_observability.go
```diff
@@ -83,6 +87,14 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		return nil, err
 	}
 
+	if raw, ok := configMap.Data["logging.enable-request-log"]; ok {
+		if strings.EqualFold(raw, "true") && oc.RequestLogTemplate != "" {
+			oc.EnableRequestLog = true
+		}
+	} else if oc.RequestLogTemplate != "" {
+		oc.EnableRequestLog = true
+	}
```
markusthoemmes, line 96: What's our plan to distinguish from the default value here? Eventually, this will always be true because we'll set an actual example template that's not empty.

One possibility is to leave this in for 0.17 and then in 0.18, remove these lines and only work off of the switch, calling out in 0.17s release-notes that the switch **must** be set to true for request logging to work come 0.18.

If we want to do it that way, let's add a comment here stating we're going to remove this implicit defaulting after 0.17 cuts (and that we'll provide a proper default for the template once 0.17 cuts too).

skonto: @markusthoemmes this follows the suggestion stated above:

> If the template is set I guess we should default to true, otherwise to false to be able to upgrade safely?

> What's our plan to distinguish from the default value here? Eventually, this will always be true because we'll set an actual example template that's not empty.

The default value is `false` and we start with that assumption of not having logging enabled. So the user either sets the flag to true and then we check if a template exists or if he sets the template we turn the flag to true.

skonto: @markusthoemmes alternatively we could avoid changing the default `true` if a template is empty and emit a warning saying that this requires a template to be set or we could use a default template if none is used so logging can work by default. 
This way downstream, if people have a template, things will not break but also we will not need to remove anything in the future. This way if user wants to disable logging needs to set the flag to `false` in any version (current or future). It is the opposite of what we have right now. WDYTH?

skonto: Added a comment btw.

skonto:
@yanweiguo @markusthoemmes I think its ready for a final review.

metrics/config_observability.go
yanweiguo: Can we keep the template empty by default and add a template example in the comments in `_example` instead? If we go this way, we can enable-request-log defaults to false if template is empty and defaults to true if template is not empty.

markusthoemmes: @yanweiguo for step one, yes. Eventually we don't want that though. The whole point of this is to be able to have _example == default so we'll have to move to a mode where that works eventually.

markusthoemmes:
/assign @vagababov 

to keep me honest here.

PR #1579, opened 2020-08-03
Remove resourceToKey map

ian-mi:
There is currently no mechanism to resolve a resource to a unique pointer causing this map to leak
memory.

### DIFF
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -50,10 +50,7 @@ type ResourceExporterFactory func(*resource.Resource) (view.Exporter, error)
 type meters struct {
 	meters  map[string]*meterExporter
 	factory ResourceExporterFactory
-	// Cache of Resource pointers from metricskey to Meters, to avoid
-	// unnecessary stringify operations
-	resourceToKey map[*resource.Resource]string
-	lock          sync.Mutex
+	lock    sync.Mutex
 }
 
 // Lock regime: lock allMeters before resourceViews. The critical path is in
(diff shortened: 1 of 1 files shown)


grantr:
In my testing this has fixed the memory leak observed in https://github.com/google/knative-gcp/issues/1540.

Before:
Screen Shot 2020-07-31 at 11 43 22 AM
The sawtooth pattern in the middle is the leak causing repeated OOM kills of the pod.

After:
Screen Shot 2020-08-03 at 1 56 53 PM
This chart is a longer timescale to show the difference; the above sawtooth is in the middle of the graph. The end of the sawtooth pattern coincided with disabling metrics collection in the pod to test if that was causing the leak. After this PR was created, I re-enabled metrics collection with this PR applied and observed continued flat memory usage (no change is visible on the graph). I verified that metrics were being collected again.

Here's a similar effect in a different broker workload that also uses the Stackdriver exporter to record metrics. The fix contained in this PR was deployed at about 2:15pm:
Screen Shot 2020-08-03 at 3 53 34 PM

jjzeng-seattle:
As the comments said, mapping pointer to string is for performance reason, to avoid serialization of resource on every calls.

But this line https://github.com/knative/pkg/blob/master/metrics/stackdriver_exporter.go#L231 defeated the purpose, by creating a new resource every time.

This fix is necessary for now.  We should rethink the performance improvement and find a better solution.

/lgtm

metrics/resource_view.go
```diff
@@ -169,12 +165,7 @@ func flushResourceExporters() {
 }
 
 func meterExporterForResource(r *resource.Resource) *meterExporter {
-	key, ok := allMeters.resourceToKey[r]
-	if !ok {
-		key = resourceToKey(r)
-		allMeters.resourceToKey[r] = key
```
evankanderson, line 175: If we need the additional performance, we could write this as:

```golang
key, ptrFound := allMeters.resourceToKey[r]
if !ok {
  key = resourceToKey(r)
}

mE := allMeters.meters[key]
if mE == nil {
  mE = &meterExporter{}
  allMeters.meters[key] = mE
  // Only store the first pointer -> key mapping
  if ptrFound {
    allMeters.resourceToKey[r] = key
  }
}
```

PR #1585, opened 2020-08-05
Automatically set GOMAXPROCS using automaxprocs library

### LINKED ISSUE #1583: Automatically set GOMAXPROCS in sharedmain
## Expected Behavior

GOMAXPROCS for controllers is correctly set when deployed in a container with a cpu limit.

## Actual Behavior

Container GOMAXPROCS are set as if they have access to all the cpus on the host

1. Deploy e.g. knative serving
2. Check GOMAXPROCS of e.g. autoscaler (which has [cpu limit of 300m](https://github.com/knative/serving/blob/master/config/core/deployments/autoscaler.yaml#L47) = less than 1 cpu)
3. See that GOMAXPROCS is set to number of CPUs on host :-( (on my Minikube with 10 CPUs, it's set to 10).

## Some context/other threads:

- https://github.com/golang/go/issues/33803 (Golang issue)
- https://github.com/prometheus/prometheus/issues/6413
- https://github.com/istio/istio/issues/20264

## Proposed solution:

We could use something like https://github.com/uber-go/automaxprocs to fix this. Or we could explicitly set GOMAXPROCS env variable in our yamls to match set quotas, but that won't work well if operators change the limits. Or we could wait for upstream Go to fix this (See linked issue above), but it looks like this is no longer landing soon.

**I think personally https://github.com/uber-go/automaxprocs seems like a reasonable approach here since it copes better with someone using the operator to bump cpu quotas, but would like to see if anyone has concerns with that before PRing :-)**

/assign @vagababov @markusthoemmes mattmoor

julz:
Fixes #1583.

Was only a one-liner and a go get to add, so here's a PR if no-one hates the idea.

/assign @vagababov @markusthoemmes mattmoor


dprotaso:
Out of curiosity what happens when the resource limits on our control plane pods changes?

julz:
> Out of curiosity what happens when the resource limits on our control plane pods changes?

As far as I know you can't change resource limits for pods in a Deployment/StatefulSet without restarting the pods, so any resource change would trigger a restart, and then GOMAXPROCS would be (properly) calculated again.

mattmoor:
Yeah, VPA creates pods with new resource blocks.

julz:
Yeps, [“Whenever VPA updates the pod resources the pod is recreated, which causes all running containers to be restarted.”](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/README.md)

slinkydeveloper:
I love this idea /lgtm

julz:
rebased

julz:
hm. Unlikely to be related since these units don't go anywhere near sharedmain. Looks to be flaky in continuous ci too: e.g. https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-pkg-continuous/1292718278833082368 :-(.

/retest

PR #1586, opened 2020-08-06
Fixes a race in the logstream.

mattmoor:
`eg.Wait` cannot be called before `eg.Go` without tripping Go's race detection.

Internally errgroup uses sync.Wait, which has [the following](https://godoc.org/sync#WaitGroup.Add) warning:

> Note that calls with a positive delta that occur when the counter is zero must happen before a Wait

When `eg.Go` isn't clearly called prior to `eg.Wait` the read of the error field in `eg.Wait` races with the
assignment of the error field in the once of `eg.Go`.

### DIFF
--- a/test/logstream/kubelogs.go
+++ b/test/logstream/kubelogs.go
@@ -112,7 +112,7 @@ func (k *kubelogs) watchPods(t test.TLegacy) {
 		return
 	}
 	eg := errgroup.Group{}
-	go func() {
+	eg.Go(func() error {
 		watchedPods := sets.NewString()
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Doesn't this still race because the inner loop calls `eg.Go` while the outer goroutine is already in the Wait? :thinking:

vagababov:
It shouldn't since first goroutine will already call wg.Add(1) on the internal WG. I think.

markusthoemmes:
Ah it says "when the counter is 0", didn't read properly.

PR #1587, opened 2020-08-06
Replace init with sync.Once

mattmoor:
This makes things insensitive to link ordering.

Also add clarity when the null logstream is being used.

### DIFF
--- a/test/logstream/interface.go
+++ b/test/logstream/interface.go
@@ -18,6 +18,7 @@ package logstream
 
 import (
 	"os"
+	"sync"
 
 	"knative.dev/pkg/system"
 	"knative.dev/pkg/test"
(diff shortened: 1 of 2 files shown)


test/logstream/interface.go
```diff
@@ -31,22 +32,26 @@ type Canceler func()
 // `test.ObjectNameForTest(t)` to `t.Log`.  It returns a Canceler, which must
 // be called before the test completes.
 func Start(t test.TLegacy) Canceler {
+	// Do this lazily to make import ordering less important.
+	once.Do(func() {
+		ns := os.Getenv(system.NamespaceEnvKey)
+		if ns != "" {
```
vagababov, line 38:
```suggestion		
		if ns := os.Getenv(system.NamespaceEnvKey); ns != "" {
```

mattmoor: done

PR #1588, opened 2020-08-06
Remove ErrGroup from kubelogs altogether

vagababov:
The latest change makes eg never to exit. In reality we don't really need EG there at all
and can just store the error.
This will return the _latest_ error, but :shrug:

/assign mattmoor

### DIFF
--- a/test/logstream/kubelogs.go
+++ b/test/logstream/kubelogs.go
@@ -24,7 +24,6 @@ import (
 	"sync"
 	"time"
 
-	"golang.org/x/sync/errgroup"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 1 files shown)


test/logstream/kubelogs.go
```diff
@@ -79,7 +78,9 @@ func (k *kubelogs) startForPod(eg *errgroup.Group, pod *corev1.Pod) {
 			req := k.kc.Kube.CoreV1().Pods(psn).GetLogs(pn, options)
 			stream, err := req.Stream()
 			if err != nil {
-				return err
+				k.m.Lock()
+				defer k.m.Unlock()
```
mattmoor, line 82: This isn't what you want

test/logstream/kubelogs.go
```diff
@@ -79,7 +78,9 @@ func (k *kubelogs) startForPod(eg *errgroup.Group, pod *corev1.Pod) {
...
+				k.err = err
```
mattmoor, line 83:
```suggestion
				func() {
					k.m.Lock()
					defer k.m.Unlock()
					k.err = err
				}()
```

test/logstream/kubelogs.go
vagababov: meh rejiggered :(

vagababov: I have Once there so it was fine at first :D

Issue #1590, opened 2020-08-06
Knative is unable to be deployed to Azure Kubernetes Service 1.17+


n3wscott:
Azure seems to have added a mutation webhook for validating and mutating webhooks that adds the following:

```yaml
namespaceSelector:
    matchExpressions:
    - key: control-plane
      operator: DoesNotExist
```

Ref: https://docs.microsoft.com/en-us/azure/aks/faq#can-i-use-admission-controller-webhooks-on-aks

The result is our webhook controller that also creates the COs for Validating and Mutating webhooks fights with this mutation webhook and it reconciles over and over again. 

see:
```yaml
(╯°□°)╯︵   kubectl get MutatingWebhookConfiguration defaulting.webhook.knative-discovery.knative.dev -oyaml
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  creationTimestamp: "2020-08-06T19:16:05Z"
  generation: 14290
  labels:
    discovery.knative.dev/release: devel
  name: defaulting.webhook.knative-discovery.knative.dev
  resourceVersion: "38793"
```

a generation of 14290 is impressive for a cluster that is 5 minutes old 🤣 

## Solution 1

We add:

```yaml
namespaceSelector:
    matchExpressions:
    - key: control-plane
      operator: DoesNotExist
```

To our template that is used to produce the webhook COs.

## Solution 2

We invest in smarter reconciler diffs to avoid re-creating objects that are mutated by additional mutating webhooks.

PR #1592, opened 2020-08-06
filter "control-plane" labeled namespaces out of our webhooks

n3wscott:
fixes: #1590


webhook/resourcesemantics/defaulting/defaulting.go
```diff
@@ -186,6 +186,9 @@ func (ac *reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
 			MatchExpressions: []metav1.LabelSelectorRequirement{{
 				Key:      "webhooks.knative.dev/exclude",
 				Operator: metav1.LabelSelectorOpDoesNotExist,
+			},{
```
evankanderson, line 189: Can we add a comment about where the `control-plane` annotation comes from (AKS), just for posterity?

dprotaso: Maybe link this upstream issue as well
https://github.com/Azure/AKS/issues/1771

PR #1595, opened 2020-08-07
Switch to the atomic bool from uber package for the profiling flag

vagababov:
- makes code easier to follow
- hides the int32->bool conversions inside the library

/assign mattmoor


vagababov:
FFS
/retest

Issue #1597, opened 2020-08-08
log status changes under codegen at debug


mattmoor:
/area API
/kind feature
/kind good-first-issue

I think that it would be a phenomenally useful debugging tool to have the codegen `logger.Debugf` the `kmp.SafeDiff` of the status updated it performs.

With logstream, this should give us a nearly perfect record of the state transitions all test resources make over the course of a test.

cc @n3wscott @whaught @vagababov @markusthoemmes

whaught:
This would be a lot more useful that some of the existing logging that just writes on every execution (and tends to pollute the logs). eg: [dumping the route object on every reconcile](https://github.com/knative/serving/blob/e0107189599f1285db77f18dcafe9060b2f462b1/pkg/reconciler/route/route.go#L108)

mattmoor:
... and consistent and free!  😎 

This used to be in our boilerplate (pre-codegen) I don't recall why we removed it.

n3wscott:
An overall pass of log levels in this code and adding trace and debug levels would be a huge help. They are a little too chatty at normal level and not enough chatty to debug weird stuff. 

Sgtm

mattmoor:
I may send a PR for this shortly, if I don't run into any unexpected problems :)

PR #1598, opened 2020-08-08
Have codegen log status diffs at debug level

mattmoor:
Fixes: https://github.com/knative/pkg/issues/1597

/hold

For some eyeballs

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/reconciler.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/reconciler.go
@@ -37,6 +37,7 @@ import (
 	sets "k8s.io/apimachinery/pkg/util/sets"
 	record "k8s.io/client-go/tools/record"
 	controller "knative.dev/pkg/controller"
+	kmp "knative.dev/pkg/kmp"
 	logging "knative.dev/pkg/logging"
 	reconciler "knative.dev/pkg/reconciler"
 )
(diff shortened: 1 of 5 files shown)


client/injection/apiextensions/reconciler/apiextensions/v1beta1/customresourcedefinition/reconciler.go
```diff
@@ -317,6 +318,10 @@ func (r *reconcilerImpl) updateStatus(existing *v1beta1.CustomResourceDefinition
 			return nil
 		}
 
+		if diff, err := kmp.SafeDiff(existing.Status, desired.Status); err == nil && diff != "" {
+			logging.FromContext(ctx).Debugf("Updating status with: %s", diff)
```
whaught, line 322: We may want to include the GVK or some identifier for the resource in the message?

mattmoor: The logger is already decorated with the key, and this should print sufficient enclosing context about the Go type to determine the type information.  Maybe we should decorate the logger with `gvk: <foo>` independent of this if that's useful...?

mattmoor: FYI the key decoration is generally how logstream works, which is a big reason I want this :)

whaught: /lgtm

mattmoor:
/unhold

I'm gonna try to get this into serving to see how it looks in the logstream 😇

PR #1600, opened 2020-08-10
add helper to parse namespaced names

dprotaso:


configmap/parse.go
```diff
@@ -144,6 +146,45 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
+		v := strings.Split(raw, string(types.Separator))
+		switch len(v) {
+		// No '/' separator.
+		case 1:
+			name = v[0]
+		case 2:
+			name = v[1]
+			namespace = v[0]
+		default:
+			return fmt.Errorf("failed to parse %q: %s ", key, "expected 'namespace/name' format")
```
julz, line 171:
```suggestion
			return fmt.Errorf("failed to parse %q: %s", key, "expected 'namespace/name' format")
```

configmap/parse.go
```diff
@@ -144,6 +146,45 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
 	}
 }
 
+func AsNamespacedName(key string, target *types.NamespacedName) ParseFunc {
```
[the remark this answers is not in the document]
mattmoor: this

configmap/parse.go
mattmoor, line 171: seems like an odd way to build the error message, also trailing space.

configmap/parse.go
```diff
@@ -144,6 +146,45 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
+	return func(data map[string]string) error {
+		var (
+			name      string
+			namespace string
+			raw       string
+			ok        bool
+		)
+
+		if raw, ok = data[key]; !ok {
```
mattmoor, line 158: Drop declaration above and use:
```suggestion
		raw, ok := data[key]
		if !ok {
```

configmap/parse.go
```diff
@@ -144,6 +146,45 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
+			return nil
+		}
+
+		v := strings.Split(raw, string(types.Separator))
+		switch len(v) {
+		// No '/' separator.
+		case 1:
+			name = v[0]
```
mattmoor, line 166: Why not set `target.Name`?

configmap/parse.go
```diff
@@ -144,6 +146,45 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
+		case 2:
+			name = v[1]
+			namespace = v[0]
```
mattmoor, line 169: Why not set `target.Name[space]`

configmap/parse.go
julz, line 166: docs for types.NamespacedName say ["NamespacedName comprises a resource name, with a mandatory namespace"](https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L23) which makes this feel a bit odd? Do we really want to allow returning invalid NamespacedNames?

configmap/parse.go
dprotaso: man - mondays

configmap/parse.go
dprotaso: I want to validate things prior to setting them on target - to avoid any side effects on validation failures

markusthoemmes:
Any reason not to use `cache.SplitMetaNamespaceKey`?

configmap/parse.go
dprotaso: i can move things around to make it a bit cleaner

configmap/parse.go
dprotaso: should be better now

dprotaso: I agree - fixed

dprotaso:
> cache.SplitMetaNamespaceKey

Funny enough I was looking for that and thought it was in `knative.dev/pkg`

I don't think we need to pull in that package for this logic. Though go1.15 is around the corner and has improvements to reduce the size

ok can folks PTAL again

configmap/parse.go
```diff
@@ -144,6 +146,33 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
+			return fmt.Errorf("failed to parse %q: expected 'namespace/name' format", key)
+		}
+
+		for _, val := range v {
+			if errs := validation.ValidateNamespaceName(val, false); len(errs) > 0 {
+				return fmt.Errorf("failed to parse %q: %s", key, strings.Join(errs, ", "))
+			}
+		}
+
+		target.Namespace = v[0]
```
julz, line 169: just to check: I think this requires the target to exist (otherwise this'd blow up), which means this won't support a nil default for target. I don't know if we want to support nil as a default, but it seems possible we might?

configmap/parse.go
vagababov: I had this fight and lost with Matt before :)
knative accepts things that K8s would deem invalid :)

configmap/parse.go
```diff
@@ -144,6 +146,33 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
...
 
+// AsNamespacedName parses the value at key as a types.NamespacedName into the target, if it exists
+func AsNamespacedName(key string, target *types.NamespacedName) ParseFunc {
+	return func(data map[string]string) error {
+		raw, ok := data[key]
+		if !ok {
+			return nil
+		}
+
+		v := strings.Split(raw, string(types.Separator))
```
vagababov, line 157: we can do SplitN(3) — to avoid doing more work if there are say 30 `/` :)

configmap/parse.go
dprotaso: I'd rather we have a separate method to handle the nil target case.

ie. `AsOptionalNamespacedName`

configmap/parse.go
dprotaso: done!

configmap/parse.go
julz: yeah, I mean in an ideal world we create one method that works for both (like the existing AsQuantity method does), but if we don't ever see a use case for not having to supply a valid NamespacedName as the default (I don't really know the use case for passing a name to know if wanting it to be optional makes sense) then cool

configmap/parse.go
```diff
@@ -144,6 +146,33 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
 	}
 }
 
+// AsNamespacedName parses the value at key as a types.NamespacedName into the target, if it exists
```
vagababov, line 149: Probably makes sense to comment that this works more like k8s NN rather than Knative NN, where we permit empty NS (e.g. the cache split one).

configmap/parse.go
dprotaso: Actually in my use case it's optional so it would make sense to make it a *

configmap/parse.go
dprotaso: done

configmap/parse.go
dprotaso: I added the optional method

dprotaso:
alright this blew up - PTAL again

thanks for the reviews!

configmap/parse.go
julz: heh, cool, fwiw I think I'd personally just go with a single `AsNamespacedName` (taking a double pointer) unless we have a particular use case where we want to not support a nil default (we can always add an extra AsRequiredNamespacedName method that doesn't allow a nil default if a use case emerges for that). Right now AsOptionalNamespacedName works exactly the same as AsQuantity except it has an extra "Optional" in the name for no reason and there's an AsNamespacedName that it sounds like won't actually be used so 🤷.

(I definitely don't personally feel strongly about this, though, and this lgtm apart from that)

configmap/parse.go
```diff
@@ -144,6 +146,47 @@ func AsQuantity(key string, target **resource.Quantity) ParseFunc {
 	}
 }
 
+// AsOptionalNamespacedName parses the value at key as a types.NamespacedName into the target, if it exists
+// The namespace and name are both required and expected to be valid DNS labels
+func AsOptionalNamespacedName(key string, target **types.NamespacedName) ParseFunc {
```
vagababov, line 151: Can't we just use empty NN as absent?

julz: empty NN isn't technically valid but maybe 🤷. If we're going to have to check `target != types.NamespacedName{}` seems probably clearer to say `target != nil`?

dprotaso: nil check will be more succinct

julz:
Let's land it

/lgtm
/unhold

PR #1601, opened 2020-08-10
Fix comment to ensure we tell the truth to the world

vagababov:
Otherwise we return the temp object that loop variable points at.

/assign mattmoor

### DIFF
--- a/apis/duck/v1/status_types.go
+++ b/apis/duck/v1/status_types.go
@@ -78,6 +78,8 @@ func (*Conditions) GetFullType() ducktypes.Populatable {
 func (s *Status) GetCondition(t apis.ConditionType) *apis.Condition {
 	for i, cond := range s.Conditions {
 		if cond.Type == t {
+			// Make sure we return the actual object, rather than the pointer to the
+			// temp loop variable.
 			return &s.Conditions[i]
 		}
 	}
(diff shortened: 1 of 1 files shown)


whaught:
I believe the intent here was to return a copy of the condition here to prevent modification. This was to encourage users to use conditionSet.manage to actually write to it (which forces inclusion of logic to update th ready condition and manage timestamps).

apis/duck/v1/status_types.go
```diff
@@ -76,9 +76,11 @@ func (*Conditions) GetFullType() ducktypes.Populatable {
 
 // GetCondition fetches the condition of the specified type.
 func (s *Status) GetCondition(t apis.ConditionType) *apis.Condition {
-	for _, cond := range s.Conditions {
+	for i, cond := range s.Conditions {
 		if cond.Type == t {
-			return &cond
```
mattmoor, line 81: I think returning a copy is intentional to avoid having folks mutate conditions out from under us and violate the condition set semantics 🤔

mattmoor: cc @n3wscott might know, or look at the commit history?

vagababov: This is fine if we intentionally return a copy, but we need to document that.

PR #1602, opened 2020-08-10
Support customizing lease name and identity

yanweiguo:
1. Add a field called `LeaseName` to customize lease name.
1. Add a field called `Identity` to specify the identify to use to gain the lock for leader election.

This is for the autoscaler scaleout work. See the [feature track doc](https://docs.google.com/document/d/1FyjB1GIJW259nMaNOc7nlQVjvgAZHlQ-fD6tbfUsaz8/edit#)

/assign @vagababov @mattmoor


leaderelection/config.go
```diff
@@ -34,13 +34,17 @@ import (
...
 
-// MaxBuckets is the maximum number of buckets to allow users to define.
-// This is a variable so that it may be customized in the binary entrypoint.
-var MaxBuckets uint32 = 10
+var (
+	// MaxBuckets is the maximum number of buckets to allow users to define.
+	// This is a variable so that it may be customized in the binary entrypoint.
+	MaxBuckets uint32 = 10
+	// supportedResourceLocks is the supported resource lock types for Knative.
+	supportedResourceLocks = sets.NewString(defaultKnativeResourceLock, resourcelock.EndpointsResourceLock)
```
vagababov, line 46: Do we need this? This is mot likely going to be used internally?

leaderelection/context.go
```diff
@@ -168,6 +167,32 @@ func (b *standardBuilder) buildElector(ctx context.Context, la reconciler.Leader
...
+// the ComponentConfig the standarBuiler has.
+func (b *standardBuilder) electorConfig(queueName string) (string, string, []reconciler.Bucket, error) {
+	lock := b.lec.LockType
+	if lock == "" {
+		lock = defaultKnativeResourceLock
+	} else if !supportedResourceLocks.Has(lock) {
+		return "", "", nil, fmt.Errorf("unsupported resource lock type: %s", b.lec.LockType)
+	}
+
+	id := b.lec.Identity
```
vagababov, line 180: if we re-arrange the code, we can get make it much nicer

1. get id
2. switch lock {case a: return ... case b: return...}

leaderelection/context.go
```diff
@@ -181,6 +206,25 @@ func standardBucketName(ordinal uint32, queueName string, cc ComponentConfig) st
 	return strings.ToLower(fmt.Sprintf("%s.%s.%02d-of-%02d", cc.Component, queueName, ordinal, cc.Buckets))
 }
 
+// NewEndpointsBucketSet returns a hash.BucketSet consists of the buckets based on
+// the given ComponentConfig.
+func NewEndpointsBucketSet(cc ComponentConfig) *hash.BucketSet {
+	names := make(sets.String, cc.Buckets)
+	for i := uint32(0); i < cc.Buckets; i++ {
+		names.Insert(endpointsBucketName(i, cc))
+	}
```
vagababov, line 215: Do we need to create services or we'll do that elsewhere?

leaderelection/context.go
yanweiguo: done

leaderelection/context.go
yanweiguo: Yeah, autoscaler needs to watch the changes and create the service.

leaderelection/config.go
yanweiguo: removed

mattmoor:
@vagababov looks like a race in the two-lane-queue

vagababov:
> @vagababov looks like a race in the two-lane-queue

It's the test @Cynocracy  added :|

yanweiguo:
Talked with @mattmoor in slack. We still have to do the update endpoints part in AS.

1. If using endpoints as the resource lock, LE package helps us create the endpoints, we need to 1) watch the endpoints 2) update the endpoints.

2. If using lease as the resource lock, we need to 1) watch the leases 2) create/update the endpoints.

There are multiple places reconciling the same endpoints resource, which can result in conflicts.

So we choose solution 2. PTAL.

leaderelection/config.go
```diff
@@ -112,6 +112,14 @@ type ComponentConfig struct {
 	LeaseDuration time.Duration
 	RenewDeadline time.Duration
 	RetryPeriod   time.Duration
+	// LeaseName is a function to customize the lease name given the index i.
+	// If not present, a name in format {Component}.{queue-name}.{i}-of-{Buckets}
+	// will be use.
```
vagababov, line 117: We probably need to comment _why_ would we need a custom naming scheme.

yanweiguo: done.

yanweiguo:
@mattmoor ping

leaderelection/config.go
```diff
@@ -34,8 +34,8 @@ import (
 
 const (
 	configMapNameEnv = "CONFIG_LEADERELECTION_NAME"
-	// KnativeResourceLock is the only supported lock mechanism for Knative.
-	KnativeResourceLock = resourcelock.LeasesResourceLock
+	// knativeResourceLock is the only supported lock mechanism for Knative.
+	knativeResourceLock = resourcelock.LeasesResourceLock
```
n3wscott, line 38: This change breaks eventing: https://github.com/knative/eventing/pull/3941

leaderelection/config.go
```diff
@@ -112,6 +112,18 @@ type ComponentConfig struct {
 	LeaseDuration time.Duration
 	RenewDeadline time.Duration
 	RetryPeriod   time.Duration
+	// LeaseName is a function to customize the lease name given the index i.
+	// If not present, a name in format {Component}.{queue-name}.{i}-of-{Buckets}
+	// will be use.
+	// Autoscaler need to know the Lease names to filter out Leases which are not
+	// used for Autoscaler. Instead of exposing the names from leadelection package,
+	// we let Autoscaler to pass them in.
+	LeaseName func(i uint32) string
```
n3wscott, line 121: This change breaks Eventing.: 
```
=== FAIL: pkg/reconciler/pingsource TestNew (unknown)
    logger.go:130: 2020-08-28T14:20:00.507Z	DPANIC	zap/sugar.go:210	Ignored key without a value.	{"ignored": "required key PING_IMAGE missing value"}
    logger.go:130: 2020-08-28T14:20:00.509Z	FATAL	pingsource/controller.go:76	Error converting leader election configuration to JSON	{"error": "json: unsupported type: func(uint32) string"}
```

vagababov: Why do you serialize it?

lionelvillard: we pass configurations to the source adapters by value.

PR #1603, opened 2020-08-11
Reduce delays in controller tests

### LINKED ISSUE #1364: controller package tests take over 12s to complete
## Expected Behavior

Tests inside `./controller` take less than 5s to complete.

## Actual Behavior

They take about 12s to complete.

```console
$ go test ./controller -count=1
ok      knative.dev/pkg/controller      12.590s
```

## Additional Info

I noticed this while working on #1363, will investigate later.

antoineco:
A series of tweaks to reduce unnecessary pauses in `controller` tests. Most of them related to excessive Sleeps.

Each commit corresponds to 1 test suite. See individual commit messages for details.

Closes #1364

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -29,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	fakekube "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
@@ -812,19 +813,15 @@ func TestEnqueueAfter(t *testing.T) {
 	defer close(queuePopulated)
 	ctx, cancel := context.WithTimeout(context.Background(), queueCheckTimeout)
 	t.Cleanup(cancel)
-	go func() {
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case <-time.After(5 * time.Millisecond):
-				if impl.WorkQueue().Len() > 0 {
-					queuePopulated <- struct{}{}
-					return
-				}
-			}
+
+	var successCheck wait.ConditionFunc = func() (bool, error) {
+		if impl.WorkQueue().Len() > 0 {
+			queuePopulated <- struct{}{}
+			return true, nil
 		}
-	}()
+		return false, nil
+	}
+	go wait.PollImmediateUntil(5*time.Millisecond, successCheck, ctx.Done())
 
 	select {
 	case <-queuePopulated:
@@ -875,19 +872,15 @@ func TestEnqueueKeyAfter(t *testing.T) {
 	const queueCheckTimeout = shortDelay + 500*time.Millisecond
 	ctx, cancel := context.WithTimeout(context.Background(), queueCheckTimeout)
 	t.Cleanup(cancel)
-	go func() {
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case <-time.After(5 * time.Millisecond):
-				if impl.WorkQueue().Len() > 0 {
-					queuePopulated <- struct{}{}
-					return
-				}
-			}
+
+	var successCheck wait.ConditionFunc = func() (bool, error) {
+		if impl.WorkQueue().Len() > 0 {
+			queuePopulated <- struct{}{}
+			return true, nil
 		}
-	}()
+		return false, nil
+	}
+	go wait.PollImmediateUntil(5*time.Millisecond, successCheck, ctx.Done())
 
 	select {
 	case <-queuePopulated:
@@ -1196,23 +1189,19 @@ func TestStartAndShutdownWithErroringWork(t *testing.T) {
 	// Keep checking the number of requeues, send to channel to indicate success.
 	itemRequeued := make(chan struct{})
 	defer close(itemRequeued)
-	go func() {
-		for {
-			select {
-			case <-ctx.Done():
-				return
-			case <-time.After(5 * time.Millisecond):
-				// Check that the work was requeued in RateLimiter, as NumRequeues
-				// can't fully reflect the real state of queue length.
-				// Here we need to wait for NumRequeues to be more than 1, to ensure
-				// the key get re-queued and reprocessed as expect.
-				if impl.WorkQueue().NumRequeues(item) > 1 {
-					itemRequeued <- struct{}{}
-					return
-				}
-			}
+
+	var successCheck wait.ConditionFunc = func() (bool, error) {
+		// Check that the work was requeued in RateLimiter, as NumRequeues
+		// can't fully reflect the real state of queue length.
+		// Here we need to wait for NumRequeues to be more than 1, to ensure
+		// the key get re-queued and reprocessed as expect.
+		if impl.WorkQueue().NumRequeues(item) > 1 {
+			itemRequeued <- struct{}{}
+			return true, nil
 		}
-	}()
+		return false, nil
+	}
+	go wait.PollImmediateUntil(5*time.Millisecond, successCheck, ctx.Done())
 
 	select {
 	case <-itemRequeued:
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -1450,10 +1450,10 @@ func TestStartInformersEventualSuccess(t *testing.T) {
 	}()
 
 	select {
-	case err := <-errCh:
-		t.Errorf("Unexpected send on errCh: %v", err)
-	case <-time.After(1 * time.Second):
+	case <-time.After(50 * time.Millisecond):
 		// Wait a brief period to ensure nothing is sent.
+	case err := <-errCh:
+		t.Fatal("Unexpected send on errCh:", err)
 	}
 
 	// Let the Sync complete.
@@ -1481,10 +1481,10 @@ func TestStartInformersFailure(t *testing.T) {
 	}()
 
 	select {
-	case err := <-errCh:
-		t.Errorf("Unexpected send on errCh: %v", err)
-	case <-time.After(1 * time.Second):
+	case <-time.After(50 * time.Millisecond):
 		// Wait a brief period to ensure nothing is sent.
+	case err := <-errCh:
+		t.Fatal("Unexpected send on errCh:", err)
 	}
 
 	// Now close the stopCh and we should see an error sent.
@@ -1537,10 +1537,10 @@ func TestRunInformersEventualSuccess(t *testing.T) {
 	}()
 
 	select {
-	case err := <-errCh:
-		t.Fatalf("Unexpected send on errCh: %v", err)
-	case <-time.After(1 * time.Second):
+	case <-time.After(50 * time.Millisecond):
 		// Wait a brief period to ensure nothing is sent.
+	case err := <-errCh:
+		t.Fatal("Unexpected send on errCh:", err)
 	}
 
 	// Let the Sync complete.
@@ -1571,10 +1571,10 @@ func TestRunInformersFailure(t *testing.T) {
 	}()
 
 	select {
-	case err := <-errCh:
-		t.Errorf("Unexpected send on errCh: %v", err)
-	case <-time.After(1 * time.Second):
+	case <-time.After(50 * time.Millisecond):
 		// Wait a brief period to ensure nothing is sent.
+	case err := <-errCh:
+		t.Fatal("Unexpected send on errCh:", err)
 	}
 
 	// Now close the stopCh and we should see an error sent.
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -828,9 +828,8 @@ func TestEnqueueAfter(t *testing.T) {
 
 	select {
 	case <-queuePopulated:
-		cancel()
 		if enqueueDelay := time.Since(enqueueTime); enqueueDelay < shortDelay {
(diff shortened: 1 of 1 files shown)


antoineco:
I'm moving this PR to "ready for review". We're down to ~5s versus >20s (both values from a run on my modest laptop) which sounds satisfying to me. There are still a few "slow" tests that were added after my original issue, so I'm gonna leave these untouched for now.

controller/controller_test.go
```diff
@@ -771,59 +767,145 @@ func TestEnqueues(t *testing.T) {
...
-	// Sleep the remaining time.
-	time.Sleep(time.Second)
-	if got, want := impl.WorkQueue().Len(), 1; got != want {
-		t.Errorf("|Queue| = %d, want: %d", got, want)
+	}, shortDelay)
+
+	// Keep checking the queue length until 'to/fall' gets enqueued, send to channel to indicate success.
+	queuePopulated := make(chan struct{})
+	defer close(queuePopulated)
+	const queueCheckTimeout = shortDelay + 100*time.Millisecond
```
vagababov, line 809: put this up next to the other constants?

controller/controller_test.go
```diff
@@ -771,59 +767,145 @@ func TestEnqueues(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), queueCheckTimeout)
+	defer cancel()
```
vagababov, line 811: `t.Cleanup`

controller/controller_test.go
```diff
@@ -771,59 +767,145 @@ func TestEnqueues(t *testing.T) {
...
+					queuePopulated <- struct{}{}
+					return
+				}
+			}
+		}
+	}()
+
+	select {
+	case <-queuePopulated:
+		cancel()
```
vagababov, line 828: Do we need this cancel here? Presumably we left the gofunc already above?

controller/controller_test.go
```diff
@@ -771,59 +767,145 @@ func TestEnqueues(t *testing.T) {
...
+		if enqueueDelay := time.Since(enqueueTime); enqueueDelay < shortDelay {
+			t.Errorf("Item enqueued within %s, expected at least a %s delay", enqueueDelay, shortDelay)
```
vagababov, line 830:
```suggestion
			t.Errorf("Item enqueued within %v, expected at least a %v delay", enqueueDelay, shortDelay)
```

controller/controller_test.go
```diff
@@ -771,59 +767,145 @@ func TestEnqueues(t *testing.T) {
...
+		t.Fatal("Timed out waiting for item to be put onto the workqueue")
 	}
+
 	impl.WorkQueue().ShutDown()
-	if got, want := drainWorkQueue(impl.WorkQueue()), []types.NamespacedName{{Namespace: "the", Name: "waterfall"}}; !cmp.Equal(got, want) {
-		t.Errorf("Queue = %v, want: %v, diff: %s", got, want, cmp.Diff(got, want))
+
+	got, want := drainWorkQueue(impl.WorkQueue()), []types.NamespacedName{{Namespace: "to", Name: "fall"}}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Error("Unexpected workqueue state (-:expect, +:got):\n", diff)
```
vagababov, line 844:
```suggestion
		t.Errorf("Unexpected workqueue state (-:expect, +:got):\n%s", diff)
```
Unfortunately all error calls insert a space, so for printing diffs we have to use formatters :(

controller/controller_test.go
antoineco: You're right, if we don't cancel here the cleanup will take care of it. Removing it.

controller/controller_test.go
antoineco: Done in 874e7058be0adef73f778eebb7a4c3768cd4de24

controller/controller_test.go
antoineco: TIL, thanks for the tip.

controller/controller_test.go
antoineco: Done in 18884e0548736e103758203e86ea0cdde43d02d2

controller/controller_test.go
antoineco: Done in 7e3844a7c315ce4601dfac4be7054a56d647448e

controller/controller_test.go
antoineco: Done in b419b391422e6ba32744e4284f67b570c6fa89ef

controller/controller_test.go
antoineco: Done in 39b02449c4470b1ec1d41c011f452bf5957c1874

controller/controller_test.go
```diff
@@ -771,59 +767,146 @@ func TestEnqueues(t *testing.T) {
...
-		t.Errorf("|Queue| = %d, want: %d", got, want)
+	}, shortDelay)
+
+	// Keep checking the queue length until 'to/fall' gets enqueued, send to channel to indicate success.
+	queuePopulated := make(chan struct{})
+	defer close(queuePopulated)
+	ctx, cancel := context.WithTimeout(context.Background(), queueCheckTimeout)
+	t.Cleanup(cancel)
+	go func() {
+		for {
```
vagababov, line 816: This is a very busy loop now -- basically 100% cpu worth loop :), do we want to add a sleep of a few ms in the default case?
(or rather replace `default:` with `case <-time.After(5*time.Millisecond)` ? Or may be even 3ms.

antoineco: [done](https://github.com/knative/pkg/compare/d7e119b0f8853d30430d7818994e003309ab1965..34bb8fdc86a02d470bfd5b57b3f94e2cdd3cd941)

antoineco:
Rebased on `master`, let's see if it makes the `Build / Build (1.15.x, ubuntu-latest)` job happy.

antoineco:
Would any of the controller approvers mind giving this PR some attention? @grantr @mattmoor @tcnghia

mattmoor:
@vagababov probably has thoughts and has been in here most recently.

controller/controller_test.go
```diff
@@ -771,59 +767,146 @@ func TestEnqueues(t *testing.T) {
...
+			case <-ctx.Done():
+				return
+			case <-time.After(5 * time.Millisecond):
+				if impl.WorkQueue().Len() > 0 {
+					queuePopulated <- struct{}{}
+					return
+				}
+			}
+		}
+	}()
```
mattmoor, line 827: There is a `wait.Until` that takes a context IIRC.

controller/controller_test.go
```diff
@@ -771,59 +767,146 @@ func TestEnqueues(t *testing.T) {
...
```
mattmoor, line 890: same `wait.Until`

controller/controller_test.go
```diff
@@ -1090,51 +1173,57 @@ func (er *ErrorReconciler) Reconcile(context.Context, string) error {
...
+				// can't fully reflect the real state of queue length.
+				// Here we need to wait for NumRequeues to be more than 1, to ensure
+				// the key get re-queued and reprocessed as expect.
+				if impl.WorkQueue().NumRequeues(item) > 1 {
+					itemRequeued <- struct{}{}
+					return
+				}
+			}
+		}
+	}()
```
mattmoor, line 1215: same `wait.Until`

controller/controller_test.go
```diff
@@ -1090,51 +1173,57 @@ func (er *ErrorReconciler) Reconcile(context.Context, string) error {
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(5 * time.Millisecond):
+				// Check that the work was requeued in RateLimiter, as NumRequeues
+				// can't fully reflect the real state of queue length.
+				// Here we need to wait for NumRequeues to be more than 1, to ensure
+				// the key get re-queued and reprocessed as expect.
+				if impl.WorkQueue().NumRequeues(item) > 1 {
+					itemRequeued <- struct{}{}
```
mattmoor, line 1210: If this is only expected to happen once, generally the pattern in Go is to `close()` the channel instead of sending the empty struct.  If it can happen multiple times, then this is cool too.

controller/controller_test.go
antoineco: Sweet, thanks  for the tip!

controller/controller_test.go
antoineco: That's what I would usually do as well, but there is a dangerous `defer close(itemRequeued)` which causes a panic if I close the channel twice.

Do you think I should rather leave the channel open in case of failure and let it die with the process?

controller/controller_test.go
antoineco: See previous comment.

antoineco:
Reviewers, is anything blocking this PR in its current state?

controller/controller_test.go
```diff
@@ -771,59 +768,138 @@ func TestEnqueues(t *testing.T) {
...
+			Namespace: "to",
 		},
-	}, 20*time.Second)
-	time.Sleep(10 * time.Millisecond)
-	if got, want := impl.WorkQueue().Len(), 0; got != want {
-		t.Errorf("|Queue| = %d, want: %d", got, want)
+	}, shortDelay)
+
+	// Keep checking the queue length until 'to/fall' gets enqueued, send to channel to indicate success.
+	queuePopulated := make(chan struct{})
```
vagababov, line 812: Do we want the channel to be buffered?

controller/controller_test.go
```diff
@@ -771,59 +768,138 @@ func TestEnqueues(t *testing.T) {
...
+	defer close(queuePopulated)
```
vagababov, line 813: t.Cleanup as well? It's possible to join them with `func() {...}`.

controller/controller_test.go
```diff
@@ -771,59 +768,138 @@ func TestEnqueues(t *testing.T) {
...
-		t.Errorf("Queue = %v, want: %v, diff: %s", got, want, cmp.Diff(got, want))
+
+	got, want := drainWorkQueue(impl.WorkQueue()), []types.NamespacedName{{Namespace: "to", Name: "fall"}}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("Unexpected workqueue state (-:expect, +:got):\n%s", diff)
 	}
 }
 
-func TestEnqeueKeyAfter(t *testing.T) {
+func TestEnqueueKeyAfter(t *testing.T) {
```
vagababov, line 847: Same here.

controller/controller_test.go
antoineco: What would be the benefit here? I rarely use buffered channels so bear with me.

controller/controller_test.go
antoineco: Would the benefit of using it justify such an overloaded syntax? (again, bear with me, I've never used `t.Cleanup()` before this PR)

controller/controller_test.go
```diff
@@ -771,59 +767,146 @@ func TestEnqueues(t *testing.T) {
 }
 
 func TestEnqueueAfter(t *testing.T) {
+	const (
+		// longDelay is longer than we expect the test to run.
+		longDelay = 1 * time.Minute
```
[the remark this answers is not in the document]
antoineco: Done in 105bdf7

controller/controller_test.go
vagababov: If there's a problem, the test would fail cleanly after timeout, hopefully rather than timeout and panic because of the deadlock because channel consumer can't read

controller/controller_test.go
vagababov: If there's a problem, the test would fail cleanly after timeout, hopefully rather than timeout and panic because of the deadlock because channel consumer can't read

antoineco: We never close the channel during the test execution (only after as a deferred action) so if I'm not mistaken there is no possible deadlock here. What am I missing?

controller/controller_test.go
antoineco: Makes sense, on it.

controller/controller_test.go
vagababov: Actually, I think I am overthinking this. It is not buffered part that will fail. But check below.

controller/controller_test.go
```diff
@@ -771,59 +768,138 @@ func TestEnqueues(t *testing.T) {
...
+	case <-queuePopulated:
+		if enqueueDelay := time.Since(enqueueTime); enqueueDelay < shortDelay {
+			t.Errorf("Item enqueued within %v, expected at least a %v delay", enqueueDelay, shortDelay)
+		}
+		if got, want := impl.WorkQueue().Len(), 1; got != want {
+			t.Errorf("|Queue| = %d, want: %d", got, want)
+		}
+
+	case <-ctx.Done():
+		t.Fatal("Timed out waiting for item to be put onto the workqueue")
```
vagababov, line 836: Here is the problem, where if due to a bug the channel is not filled -- this will never exit. So we might want a timeout here as well.

antoineco: `ctx` is done after `shortDelay + 500*time.Millisecond` due to the `context.WithTimeout()` above, so `select` will find a candidate no matter what.
The test fails rather quickly here in case things go wrong, I've tried to do things like short circuiting the channel writer while refactoring those tests, and I'm pretty confident the edge cases mentioned are all covered here.

vagababov: 😊 
Ah, nice.
Thanks for pointing out. We rarely do with timeout so it elided me.

antoineco: Yeah those tests are very much time-based due to the introduction of arbitrary delays ("EnqueueAfter"). I would love to use something more elegant than a timeout, it's just to prevent those tests from running forever.

antoineco:
Appreciate the final thorough review @vagababov 👍

PR #1622, opened 2020-08-14
found another way to get repo name in gh actions

n3wscott:
also fixed the real vet errors.

### DIFF
--- a/.github/workflows/go-pr-verify.yaml
+++ b/.github/workflows/go-pr-verify.yaml
@@ -48,7 +48,7 @@ jobs:
     - name: Check out code onto GOPATH
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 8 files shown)


.github/workflows/go-pr-verify.yaml
```diff
@@ -48,20 +48,20 @@ jobs:
     - name: Check out code onto GOPATH
       uses: actions/checkout@v2
       with:
-        fetch-depth: 1
-        path: ./src/knative.dev/${{github.event.pull_request.head.repo.name}}
+        fetch-depth: 14
```
mattmoor, line 51: Why 14?

n3wscott: 😬

PR #1623, opened 2020-08-15
Assorted tweaks to actions.

mattmoor:
Drop GOPATH from all except verify (codegen needs it), and change directories to just use workspace.

Switch to Go 1.15 (go.mod actually controls what's used IIUC).

### DIFF
--- a/.github/workflows/go-build.yaml
+++ b/.github/workflows/go-build.yaml
@@ -15,8 +15,10 @@
 name: Build
 
 on:
+  push:
+    branches: [ 'master', 'release-*' ]
   pull_request:
-    branches: [ master ]
+    branches: [ 'master', 'release-*' ]
 
 jobs:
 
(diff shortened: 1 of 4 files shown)


.github/workflows/go-build.yaml
```diff
@@ -15,23 +15,22 @@
...
+    branches: [ 'master', 'release-*' ]
 
 jobs:
 
   build:
     name: Build
     strategy:
       matrix:
-        go-version: [1.14.x]
+        go-version: [1.15.x]
```
n3wscott, line 29: Should we do both?

.github/workflows/go-build.yaml
```diff
@@ -15,23 +15,22 @@
 name: Build
 
 on:
+  push:
```
n3wscott, line 18: Can you drop the -pr from all files? I added it thinking we had to use the pull event but no longer.

mattmoor: I already did :)

.github/workflows/go-build.yaml
mattmoor: I believe that generally the toolchain keys off of the version in go.mod, so it felt like a waste.  I also figure we'll want 1.15 by the time we're using these for realz.

.github/workflows/go-build.yaml
n3wscott: Oh you did. Noce

PR #1625, opened 2020-08-15
don't pre/post process if skipstatusupdate

whaught:
Fixes https://github.com/knative/serving/issues/9073

- if skipStatusUpdates is set for a controler we don't need to process conditions/observed-generation


mattmoor:
/lgtm
/approve

Thanks!

PR #1644, opened 2020-08-24
Remove the extraneous templates from pkg

vagababov:
/assign mattmoor


mattmoor:
/lgtm
/approve

🎉 

Thanks, I thought I'd sorted this out with my renames, but missed these.

PR #1647, opened 2020-08-24
Return an apimachinery error if psbinding subject WithContext fails

lberk:
In the event a subject's `WithContext()` call fails, then
ReconcileSubject fails without catching the proper
error (sinkbinding's WithContext returns its own custom errors).  This
can manifest itself in instances where the sinkbinding subject has
been deleted before the sinkbinding itself, and leaves the reconcile
deletion loop stuck before the finalizer can be removed.  Leaving the
sinkbinding itself stuck, despite the subject already being deleted.

### DIFF
--- a/webhook/psbinding/reconciler.go
+++ b/webhook/psbinding/reconciler.go
@@ -361,7 +361,7 @@ func (r *BaseReconciler) ReconcileSubject(ctx context.Context, fb Bindable, muta
 	if r.WithContext != nil {
 		ctx, err = r.WithContext(ctx, fb)
 		if err != nil {
-			return err
+			return apierrs.NewNotFound(gvr.GroupResource(), subject.Name)
 		}
 	}
 


matzew (approved):
/lgtm
/approve

webhook/psbinding/reconciler.go
```diff
@@ -361,7 +361,7 @@ func (r *BaseReconciler) ReconcileSubject(ctx context.Context, fb Bindable, muta
 	if r.WithContext != nil {
 		ctx, err = r.WithContext(ctx, fb)
 		if err != nil {
-			return err
+			return apierrs.NewNotFound(gvr.GroupResource(), subject.Name)
```
mattmoor, line 364: It seems odd to me that we would turn arbitrary `WithContext` errors into "not found".

I get that SinkBinding's implementation of `WithContext`'s main failure mode is "not found", but why not push that inward?

Seems like this path here: https://github.com/knative/eventing/blob/1dfad10d1a3aab6b5c3884f7dd7e57477c81f8bd/pkg/reconciler/sinkbinding/controller.go#L131

Pulling this back into PKG's responsibility, we could make `URIFromDestinationV1` return appropriate apimachinery errors and probably expand this fix to a whole class of errors like this.  WDYT?

mattmoor:
Awesome sleuthing btw, it sounds like a nasty bug, and it'll be great to land a fix!

webhook/psbinding/reconciler.go
lberk: @mattmoor thanks! yes it was absolutely a somewhat arbitrary decision (based on the return codes catches downward (upward?) in the function call chain looking for `NotFound`.

The reason I put it here, vs, in the sinkbinding controller (as suggested) is when looking at the return type of `WithContext` there's no guarantee of a different `WithContext` implementation wouldn't hit this as well.  While I definitely appreciate the thought (and was my first inclination as well), I went with this approach so that the pkg portion would be more robust in general.  Thoughts?

mattmoor: Done here, it just feels to me like a bad assumption about WithContext.  If we can achieve the same by fixing the URI resolver, that'd be my preference for "maximal fixage" :)

lberk: ok, thanks for the feedback, will close and make the change there.

PR #1648, opened 2020-08-24
Fix the the request logging boolean toggle

vagababov:
Since we have cut the 0.17, we can go for the proper toggle setup:

- if request logging is on — template must be present and the logging will be on
- if the request of logging is off — template can be present, but will not affect whether the request logs will be collected

Really
Fixes https://github.com/knative/serving/issues/8644

/assign @yanweiguo @julz

### DIFF
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -28,6 +28,8 @@ import (
 const (
 	// The following is used to set the default log url template
 	DefaultLogURLTemplate = "http://localhost:8001/api/v1/namespaces/knative-monitoring/services/kibana-logging/proxy/app/kibana#/discover?_a=(query:(match:(kubernetes.labels.knative-dev%2FrevisionUID:(query:'${REVISION_UID}',type:phrase))))"
+	// DefaultRequestLogTemplate is the default format for emitting request logs.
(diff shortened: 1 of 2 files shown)


vagababov:
(ノಠ益ಠ)ノ彡┻━┻

Unrelated
/retest

metrics/config_observability.go
```diff
@@ -92,15 +95,8 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
...
-	if raw, ok := configMap.Data["logging.enable-request-log"]; ok {
-		if strings.EqualFold(raw, "true") && oc.RequestLogTemplate != "" {
-			oc.EnableRequestLog = true
-		}
-	} else if oc.RequestLogTemplate != "" {
-		// TODO: remove this after 0.17 cuts, this is meant only for smooth transition to the new flag.
-		// Once 0.17 cuts we should set a proper default value and users will need to set the flag explicitly
-		// to enable request logging.
-		oc.EnableRequestLog = true
+	if oc.RequestLogTemplate == "" && oc.EnableRequestLog {
```
yanweiguo, line 98: I think the idea is to use the default value which set in `_example` in the configmap?

vagababov: I thought to do this after importing to Serving. But I guess, we can just do it.

vagababov: Done.

metrics/config_observability_test.go
```diff
@@ -49,38 +48,47 @@ func TestObservabilityConfiguration(t *testing.T) {
...
 			"logging.enable-probe-request-log":            "true",
+			"logging.enable-request-log":                  "false",
 			"logging.enable-var-log-collection":           "true",
+			"logging.request-log-template":                "",
+			"logging.revision-url-template":               "https://logging.io",
+			"metrics.request-metrics-backend-destination": "stackdriver",
+			"profiling.enable":                            "true",
+		},
+	}, {
+		name:    "observability configuration with request log set and template not set",
```
yanweiguo, line 84: My understanding is that if request log set but template not set, we use default tempalte. If template set to empty, throw an error. Is my understanding wrong?

vagababov: yep that's why it's zeroed out in the test.

PR #1651, opened 2020-08-25
Switch bool parsing to the real parser

vagababov:
We parse junk as false, but that's not really correct.
So be more precise and more aggressive in validation.

/assign mattmoor @yanweiguo @julz

### DIFF
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -46,7 +46,7 @@ func AsBool(key string, target *bool) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
 			val, err := strconv.ParseBool(raw)
-			*target = val // This is basically false always, if err != nil.
+			*target = val // If err != nil — this is always false.
 			return err
 		}
 		return nil
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -45,7 +45,9 @@ func AsString(key string, target *string) ParseFunc {
 func AsBool(key string, target *bool) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
-			*target = strings.EqualFold(raw, "true")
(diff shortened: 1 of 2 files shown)


configmap/parse.go
```diff
@@ -45,7 +45,9 @@ func AsString(key string, target *string) ParseFunc {
 func AsBool(key string, target *bool) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
-			*target = strings.EqualFold(raw, "true")
+			val, err := strconv.ParseBool(raw)
+			*target = val // This is basically false always.
```
julz, line 49: I can't figure out why this comment is true, am I missing something obvious? 😓

vagababov: Eh, bad wording: it is always true if err != nil :)

julz: I still dont get it.. doesn't `strconv.ParseBool("true") == true, nil`?

vagababov: you're not looking at the last commit, I think :)

julz: ohhh.. I think I get what you mean -- that if err is non-nil, val is false here

julz: latest commit is better, but I think I'd go for "If err != nil, val is always false here", but that might just be me, either way I get what you're saying now :)

julz:
I guess this is _technically_ a breaking change to the configmaps (since you could theoretically have a falsey value that was valid before but that strconv doesn't like). You'll get a nice error in that case though and it's a pretty degenerate case so 🤷 seems fine to me.

/lgtm
/hold so @mattmoor can look

vagababov:
Yeah, I think it's more correct this way, since somebody might have had a breaking config and not even aware of it, just because _false_ was what made sense.

vagababov:
🕐

PR #1658, opened 2020-08-27
Allows utilities easy mode for Knative Injection

n3wscott:
Adding an easy entry method next to sharedmain to allow for utilities to leverage injection without being a controller.

Example usage in chaos duck.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -126,6 +126,39 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
 	return leaderelection.NewConfigFromConfigMap(leaderElectionConfigMap)
 }
 
+// EnableInjectionOrDie enables Knative Injection and starts the informers.
+// Both Context and Config are optional.
+func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) context.Context {
+	if ctx == nil {
+		ctx = signals.NewContext()
+	}
+	if cfg == nil {
+		cfg = ParseAndGetConfigOrDie()
+	}
+
+	// Respect user provided settings, but if omitted customize the default behavior.
+	if cfg.QPS == 0 {
+		cfg.QPS = rest.DefaultQPS
+	}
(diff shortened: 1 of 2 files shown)


n3wscott:
/retest

wish I could run those tests locally...

vagababov:
You should be able to

/cc @evankanderson

injection/sharedmain/main.go
```diff
@@ -126,6 +126,39 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
...
+		cfg.Burst = rest.DefaultBurst
+	}
+	ctx = injection.WithConfig(ctx, cfg)
+
+	ctx, informers := injection.Default.SetupInformers(ctx, cfg)
+
+	// Start the injection clients and informers.
+	logging.FromContext(ctx).Info("Starting informers...")
+	go func(ctx context.Context) {
+		if err := controller.StartInformers(ctx.Done(), informers...); err != nil {
```
grantr, line 153: This change broke reconcilers that set non-default resyncPeriod values because it starts informers before the controllers are set up. Attempting to set the resyncPeriod is ignored and the informer code logs this warning message:

```
resyncPeriod 300000000000 is smaller than resyncCheckPeriod 36000000000000 and the informer has already started. Changing it to 36000000000000
```
```

```

mattmoor: I think that this is just generally broken.  Without the controllers created the informer events won't be set up and the controllers won't properly resync things on startup.  This is likely masked by the resync we get when we are promoted to leader, but generally informers should not be started until we're done setting up events.

@n3wscott @vagababov What was the motivation for this change?  What will rolling it back in the release branch break?

mattmoor: @grantr we should probably make sure that we have test coverage of the scenario where you folks got broken so that we don't make a similar regression in the future.

n3wscott: anyone working on fixing this or should I try?
@grantr

PR #1665, opened 2020-08-31
Make test override pongTimeout correctly.

markusthoemmes:
This was erroneously changed in https://github.com/knative/pkg/commit/34838c4559697a4502141537a368fac1e09fccd4#diff-6c7213643bc6f930dd8cdc16121ccbc8. It's actually supposed to override the global var.

### DIFF
--- a/websocket/connection_test.go
+++ b/websocket/connection_test.go
@@ -379,7 +379,7 @@ func TestDurableConnectionWhenConnectionBreaksDown(t *testing.T) {
 
 func TestDurableConnectionSendsPingsRegularly(t *testing.T) {
 	// Reset pongTimeout to something quite short.
-	const pongTimeout = 100 * time.Millisecond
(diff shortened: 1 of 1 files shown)


websocket/connection_test.go
```diff
@@ -379,7 +379,7 @@ func TestDurableConnectionWhenConnectionBreaksDown(t *testing.T) {
 
 func TestDurableConnectionSendsPingsRegularly(t *testing.T) {
 	// Reset pongTimeout to something quite short.
-	const pongTimeout = 100 * time.Millisecond
+	pongTimeout = 100 * time.Millisecond
```
julz, line 382: do we want to restore this afterwards? (I guess it doesn't matter right now since no other tests are using this, but theoretically that could change 🤷)

markusthoemmes: Done.

julz (commented):
/lgtm

PR #1674, opened 2020-08-31
Fix the reentrance of the github issue tests

vagababov:
right now they panic
with this change they won't
serve cold
enjoy!

/assign mattmoor @chizhg


vagababov:
I am quite unhappy I couldn't move the test helper. But that's for another day :/

PR #1676, opened 2020-09-01
Add a helper for checking for all expected responses.

mattmoor:


mattmoor:
Context: https://github.com/knative/serving/pull/9239#issuecomment-684992531

PR #1678, opened 2020-09-01
Permit table test deletions to arrive in arbitrary order.

### LINKED ISSUE #411: TableTest - Delete in any order
## Expected Behavior

We set the Table Test `WantDeletes` field as a slice. Normally we don't care about the order in which the deletes occur. We should either make the delete comparison unordered, or at least provide an option to make it unordered.

1. Check out knative/eventing at 0a97ffcd0a93861a4e8fa2a7be82356530d1fd02
2. Run `go test ./pkg/reconciler/apiserversource -count 100`

## Additional Info

The client used by the unit tests returns lists in a non-fixed order, causing the unit test to flake when deleting everything in the list, as described in https://github.com/knative/eventing/pull/1230#issuecomment-491963450.

vagababov:
If you haver ever written a table test, perhaps you noticed that when several objects
are deleted you need to list them is a particular order. And if you're deleting things in parallel
it's just impossible.
But what we care is that N objects are deleted -- so verify that.
The key is type-name-ns, which should work most of the time.

I've cleaned the tests in Eventing to work with this change
and there's an open PR in serving that would make the tests pass
with this change there too

Fixes #411

/assign mattmoor

### DIFF
--- a/reconciler/testing/table.go
+++ b/reconciler/testing/table.go
@@ -307,10 +307,10 @@ func (r *TableRow) Test(t *testing.T, factory Factory) {
 	}
 	if !gotDeletes.Equal(wantDeletes) {
 		if extra := gotDeletes.Difference(wantDeletes); len(extra) > 0 {
-			t.Errorf("Extra or unexpected deletes: %v", extra)
+			t.Errorf("Extra or unexpected deletes: %v", extra.UnsortedList())
 		}
 		if missing := wantDeletes.Difference(gotDeletes); len(missing) > 0 {
-			t.Errorf("Missing deletes: %v", missing)
+			t.Errorf("Missing deletes: %v", missing.UnsortedList())
 		}
 	}
 
--- a/reconciler/testing/table.go
+++ b/reconciler/testing/table.go
@@ -29,6 +29,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


reconciler/testing/table.go
```diff
@@ -285,44 +286,31 @@ func (r *TableRow) Test(t *testing.T, factory Factory) {
...
 		}
-		if !r.SkipNamespaceValidation && got.GetNamespace() != expectedNamespace {
-			t.Errorf("Unexpected delete-collection[%d]: %#v, wanted %s", i, got, expectedNamespace)
-		}
-	}
-	if got, want := len(actions.DeleteCollections), len(r.WantDeleteCollections); got > want {
-		for _, extra := range actions.DeleteCollections[want:] {
-			t.Errorf("Extra delete-collection: %#v", extra)
+		if missing := wantDeletes.Difference(gotDeletes); len(missing) > 0 {
+			t.Errorf("Missing deletes: %v", missing)
```
mattmoor, line 313: These are going to have your weird encoding, right?  🤔

vagababov: Something like this
`trigger_test.go:653: Extra or unexpected deletes: map[subscriptions~~test-broker-test-trigger-test-trigger-uid~~test-namespace:{}]`
but I can change that to `missing.UnorderedList()` to streamline.

vagababov: Aaaannd... done.

PR #1680, opened 2020-09-02
Fix metrics test reentrance

vagababov:
if we clean the view we get success.
Also verify initial size is 0

/assign @yanweiguo


yanweiguo (commented):
/lgtm

PR #1681, opened 2020-09-03
Add logger-type enum to logging configmap

ian-mi:
Setting logger-type to stackdriver in the logging configmap will now use zapdriver to create a zap config suitable for exporting to Stackdriver. Any values specified in the zap-logger-config will override those provided by zapdriver. If logger-type is unspecified the default behavior of using zap-logger-config is preserved.


evankanderson:
/assign

Hi Ian,

Do you have a quick high-level description of the desired state? I'd like to make sure that this lines up with eventually moving most of the metrics adapters into a separate process (the opencensus collector).

ian-mi:
Hi Evan, since this affects only logging I think it should be orthogonal to the metrics collector (at least until we think the opentelemetry-collector's log collection functionality can be adopted: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/overview.md#opentelemetry-collector).

The goal of this PR is to introduce the logger-type field which can be used to select useful preconfigured zap loggers such as the zapdriver logger I've added in this PR. These loggers can still be customized using the zap-logger-config field, but it is not required to specify the entire logger config when doing so as it is by default when no logger-type is selected. The preconfigured loggers can also customize aspects of the zap logger which are not possible to encode in zap-logger-config. For example it is not possible to specify the encoder functions such as LevelEncoder as json. I've added this field in such a way that other useful structured log formats can be added as  logger-types.

PR #1684, opened 2020-09-04
Add IsEstablished method to ManagedConnection to expose whether the connection has been established

yanweiguo:
When using a ManagedConnection for unit test, this function can help us know whether the connection has been established so that we can start sending message. Instead of sending message and checking whether it succeeds or not, checking the return of this function is easier.


vagababov (approved):
/lgtm

Issue #1686, opened 2020-09-07
Log line per conversion is very spammy


mattmoor:
/area API
/kind cleanup
/kind good-first-issue

Our test logs are filled with these relatively useless conversion logs:

```
kubelogs.go:197: I 01:01:49.092 webhook-75cccfdcd8-rlgww [conversion/conversion.go:133] [serving-tests/service-with-traffic-split-rqqcjsym] Converting [kind=Route group=serving.knative.dev version=v1] to version serving.knative.dev/v1alpha1
```

I'd suggest we just eliminate this log line.

mattmoor:
Here is the line: https://github.com/knative/pkg/blob/bfc0cee9456123e3660ca7a5fb034a4c0cf5ef0f/webhook/resourcesemantics/conversion/conversion.go#L133

PR #1687, opened 2020-09-07
Remove spammy log line

mattmoor:
Fixes: https://github.com/knative/pkg/issues/1686


vagababov (commented):
We could've gone with debug?
/lgtm

mattmoor:
Then we'd still get them in logstream.  In a better world I'd relegate it to like `V(8)` or something.  Not sure this has ever been useful, so I'm biasing towards killing it.

PR #1689, opened 2020-09-08
Speed up & deflake tests by disabling bundling

### LINKED ISSUE #1672: TestMetricsExport is super flaky
metrics/resource_view_test.go - TestMetricsExport flakes most of the time.

/assign @jjzeng-seattle @evankanderson 

here's a good example: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/1666/pull-knative-pkg-unit-tests/1300475037081407489

evankanderson:
Fixes #1672 

It turns out that both Stackdriver and OpenCensus have rpc Bundlers that hold reports until at least N have been accumulated or for 1-2s. I had to send an upstream PR to OpenCensus to expose the bundler options in order to speed up reporting, which accounts for the `go.mod` update.

It also turns out that gRPC will take (on my machine, going through ??? WSL2 -> Windows DNS -> internet) about 1s to attempt to resolve `external-svc`, so changing that to a name that _will_ resolve sped up the `config_test` by a second or so. It also turns out that gRPC takes an additional 1-2 seconds to set up "fail-fast" channels _on Windows_ but not on a Linux kernel -- caveat testor if you're attempting to debug timing on Windows.

I *think* this has been de-flaked from a timeout perspective; on my machine (i7 mobile processor, plugged in, running WSL2), the entire suite takes about 5s, compared with 10s before these changes.


evankanderson:
/assign @vagababov since I know this was driving him crazy -- any additional thoughts on tests?

```shell
$ go test ./metrics -count 40 -timeout 900s 2>&1 > /tmp/test-log
$ cat /tmp/test-log
ok      knative.dev/pkg/metrics 208.145s
```

(I didn't push too much past `-count 40` because there seems to be an FD leak somewhere in the tests which was causing the ulimit to be exhausted with 2k FDs)

evankanderson:
Fixing th e races now, forgot to run with `-race`.

evankanderson:
Note that at head, `-count 40` does not complete within 15 minutes. After this change, it's consistently less than 6 minutes.

I've noticed that Prow test runs seem to be much slower than my local machine, so it's possible the timeouts are due to the slower processing time on the Prow cluster.

metrics/resource_view_test.go
```diff
@@ -566,25 +582,38 @@ func TestStackDriverExports(t *testing.T) {
...
+							Value:  ts.Points[0].Value.GetInt64Value(),
+						}
+						// Override 'cluster-name' label to reset to a fixed value
+						if extracted.Labels["cluster_name"] != "" {
+							extracted.Labels["cluster_name"] = "test-cluster"
+						}
+						records = append(records, extracted)
+						if strings.HasPrefix(ts.Metric.Type, "knative.dev/") {
+							if diff := cmp.Diff(ts.Resource.Type, metricskey.ResourceTypeKnativeRevision); diff != "" {
+								t.Errorf("Incorrect resource type for %q: (-want +got): %s", ts.Metric.Type, diff)
```
vagababov, line 602:
```suggestion
								t.Errorf("Incorrect resource type for %q: (-want +got):\n%s", ts.Metric.Type, diff)
```

metrics/resource_view_test.go
```diff
@@ -566,25 +582,38 @@ func TestStackDriverExports(t *testing.T) {
...
-					sdFake.srv.GracefulStop()
+					if len(records) >= len(tc.expected) {
+						// There's no way to synchronize on the internal timer used
+						// by metricsexport.IntervalReader, so shut down the
+						// exporter after the first report cycle.
+						FlushExporter()
+						sdFake.srv.GracefulStop()
+						break loop
+					}
+				case <-time.After(5 * time.Second):
```
vagababov, line 614: Should those timeouts be consistent? it's 4s above.

metrics/stackdriver_exporter.go
```diff
@@ -88,6 +88,10 @@ var (
 	// metricToResourceLabels provides a quick lookup from a custom metric name to the set of tags
 	// which should be promoted to Stackdriver Resource labels via opencensus resources.
 	metricToResourceLabels = map[string]*resourceTemplate{}
+
+	// A variable for testing to reduce the size (number of metrics) buffered before
+	// Stackdriver will send a bundled metric report. Only applies if non-zero.
+	TestOverrideBundleCount = 0
```
vagababov, line 94: why is this a public variable?

metrics/resource_view_test.go
evankanderson: There used to be an extra delay on the Stackdriver outputs, but they should be similar now.

metrics/stackdriver_exporter.go
evankanderson: It seemed like e2e tests in Serving or Eventing might want to be able to disable the batching for the same reasons that the metrics tests would want to do so.

I'd like to add tests for important exports to the Serving and Eventing repos once these tests are no longer flaky.

evankanderson:
Updated!

metrics/stackdriver_exporter.go
vagababov: Sure then let's specify that, since just looking at the comment I see no reason why it should be public.
Also let's do this in a linter happy format `TestOverrideBundleCount is a variable...`

metrics/resource_view_test.go
```diff
@@ -225,9 +226,10 @@ func sortMetrics() cmp.Option {
 
 // Begin table tests for exporters
```
vagababov, line 227:
```suggestion
```

metrics/resource_view_test.go
```diff
@@ -341,19 +343,32 @@ testComponent_testing_value{project="p1",revision="r2"} 1
...
-							Name:   m.MetricDescriptor.Name,
-							Labels: labels,
-							Value:  m.Timeseries[0].Points[0].GetInt64Value(),
-						})
 					}
+					if len(records) >= len(expected) {
+						break loop
+					}
+				case <-time.After(4 * time.Second):
+					t.Error("Timeout reading input")
```
vagababov, line 370: I wonder if fatal is more appropriate here?

metrics/resource_view_test.go
```diff
@@ -341,19 +343,32 @@ testComponent_testing_value{project="p1",revision="r2"} 1
...
-					if len(m.Timeseries) > 0 {
-						labels := map[string]string{}
-						if record.Resource != nil {
-							labels = record.Resource.Labels
+		loop:
+			for {
+				select {
+				case record := <-ocFake.published:
+					if record == nil {
+						continue loop
```
vagababov, line 351: I guess it's not important here?
```suggestion
						continue
```

metrics/resource_view_test.go
```diff
@@ -341,19 +343,32 @@ testComponent_testing_value{project="p1",revision="r2"} 1
 			UnregisterResourceView(gaugeView, resourceCounter)
 
 			records := []metricExtract{}
-			for record := range ocFake.published {
-				for _, m := range record.Metrics {
-					if len(m.Timeseries) > 0 {
-						labels := map[string]string{}
-						if record.Resource != nil {
-							labels = record.Resource.Labels
+		loop:
```
vagababov, line 346: Stylistically I'd prefer a lambda with return rather than goto label, but up to you. :)

metrics/resource_view_test.go
```diff
@@ -566,25 +585,38 @@ func TestStackDriverExports(t *testing.T) {
...
+				select {
+				case record := <-sdFake.published:
+					for _, ts := range record.TimeSeries {
+						extracted := metricExtract{
+							Name:   ts.Metric.Type,
+							Labels: ts.Resource.Labels,
+							Value:  ts.Points[0].Value.GetInt64Value(),
+						}
+						// Override 'cluster-name' label to reset to a fixed value
+						if extracted.Labels["cluster_name"] != "" {
```
vagababov, line 599: Should it not be set if empty?

metrics/resource_view_test.go
evankanderson: Actually, you need the label, otherwise the `continue` applies to the `select`, rather than to the loop.

metrics/resource_view_test.go
evankanderson: Done.

metrics/resource_view_test.go
evankanderson: This is using Golang defaults, so `mapVal["doesNotExist"]` will return the default value from the map `map[string]string`, which is `""`.

metrics/resource_view_test.go
evankanderson: That would look like:

```
for {
  breakErr := error.New("intentional break")
  err := func() error {
    select {
    case record := <-ocFake.published:
      if record == nil {
        return nil
      }
      // ...
      if len(records) >= len(expected) {
        return breakErr
      }
    case <-time.After(5 * time.Second):
      return errors.New("timeout")
    }
  }()
  if err != breakErr {
    t.Fatal("Unable to fetch events", err)
  }
  if err == breakErr {
    break
  }
}

evankanderson: I think with the label is better, stylistically.

metrics/resource_view_test.go
vagababov: Strictly theoretically there may be the value in the map with `""` value. So if this happens due to a bug it won't be caught. But 🤷 .

metrics/resource_view_test.go
vagababov: Up to you :-)

metrics/resource_view_test.go
vagababov: No, if it's a lambda you return from the func rather than continue. But since we're keeping the label, it doesn't matter.

PR #1690, opened 2020-09-08
Use the constant in the unit tests that we already have

vagababov:
and make it public so it can be used in serving as well.

/assign mattmoor


mattmoor (commented):
/lgtm
/approve

vagababov:
Alright. Now it is informed watcher that flakes...

PR #1688, opened 2020-09-08
Alter URIFromObjectReference to return apierrs

lberk:
This allows for reconcile deletion to properly remove finalizer and
delete sinkbinding

Followup from #1647

PR #1692, opened 2020-09-09
Sinkbinding Webhook, allow request if WithContext returns NotFound

lberk:
A race condition can exist with yaml if/when the sinkbinding subject
is not created yet at the time of the webhook's Admit call (even if
it's within the yaml you're trying to apply) but allowing the Admit
request if the resource isn't found, we can avoid this in a single
kubectl apply -f foo.yaml call.  Adding the bindable subject later
will still get Updated accordingly

### DIFF
--- a/webhook/psbinding/psbinding.go
+++ b/webhook/psbinding/psbinding.go
@@ -29,6 +29,7 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


webhook/psbinding/psbinding.go
```diff
@@ -221,7 +222,10 @@ func (ac *Reconciler) Admit(ctx context.Context, request *admissionv1.AdmissionR
 	if ac.WithContext != nil {
 		var err error
 		ctx, err = ac.WithContext(ctx, fb)
-		if err != nil {
+		if apierrs.IsNotFound(err) {
+			//Allow request for now as the resource might not exist yet
+			return &admissionv1.AdmissionResponse{Allowed: true}
+		} else if err != nil {
```
mattmoor, line 228: It seems odd to me to key off of `apierrs` since that is generally used for errors returned from API calls, and `WithContext` is an arbitrary callback.

I wonder whether this would be a reasonable use of wrapped errors (paging @n3wscott who seems to love those)?  Essentially what I'm wondering (thinking out loud) is whether we should provide a way for implementors of `WithContext` to wrap the error they return to guide the behavior here in a way that's fitting for the abstraction.

This would give the implementor of `WithContext` the capacity to pick and choose what error types in what contexts result in this being allowed or denied.

matzew: @n3wscott any comments ?

n3wscott: Where is the change that changes WithContext?

lberk: it was merged in https://github.com/knative/pkg/pull/1688

n3wscott: I agree with matt, I think the simple fix could be a better name `WithContext` there is a lot of code that is running there and it is not at all clear that it will be happening because of a `WithContext` call

lberk:
/close
fixed in knative/eventing#4161

PR #1693, opened 2020-09-09
Deprecate --master flag in favor of --server.

markusthoemmes:
This not only gets rid of the master terminology, but is also actually in-line with how `kubectl` itself behaves. Win everywhere!

/assign @mattmoor


mattmoor (commented):
cc @ImJasonH @vdemeester (Tekton)

/lgtm

PR #1694, opened 2020-09-09
Remove branch examples referencing master.

markusthoemmes:
There's no need to reference master in these examples. Let's drop them.


vagababov (commented):
/lgtm

PR #1696, opened 2020-09-09
Exclude the control-plane from our webhooks for psbinding webhook.

n3wscott:
do not fight with azure's webhook for webhooks

Relates to #1590


mattmoor (commented):
/lgtm

🤢

mattmoor:
/hold

Link the Azure issue?

n3wscott:
It is linked from our issue, but it is this one : https://github.com/Azure/AKS/issues/1771

It is not clear it is an issue. They are doing a normal thing, and we are re-reconciling the object back to how we thing it should look. We should be smarter about mutation webhooks in general.

PR #1698, opened 2020-09-10
Enhance error message on failure to update _example.

markusthoemmes:
We still see users frequently struggle with this and the error message seemingly doesn't help them. This gives a more explicit description of what the user can do to fix the situation.

/assign @mattmoor @vagababov


vagababov (commented):
I'm using bottom :) right above meta
/lgtm

Issue #1397, opened 2020-06-10
Change log level to debug in admission webhook response details


toVersus:
I'm building my own mutating webhook based on [knative webhook library](https://github.com/knative/pkg/tree/master/webhook), which injects the sensitive data from external secret store to K8s Secret object. I like the idea of generating/rotating CA and server certificates for webhook on its own and currently it's working as expected. However, I'm unhappy with the admission log output, which contains the raw sensitive data. Can we change the log level from `info` to `debug` only at this part? If so, I will submit a PR.

https://github.com/knative/pkg/blob/fef70cc2616b2937d13a1aed2b9f8b539c43c768/webhook/admission.go#L98-L100

nak3:
I think that the change makes sense.

@evankanderson @dprotaso @antoineco What do you think? It is different reason but similar to another request on slack https://knative.slack.com/archives/C93E33SN8/p1595308110264500

evankanderson:
I think it makes sense to cast a critical eye at our logging in general, which is pretty noisy in the "happy case" when things are stable.

dprotaso:
> Can we change the log level from info to debug only at this part? If so, I will submit a PR.

Feel free to change the level.  Another option may be to reduce the properties being logged on Info and include them on debug.

nak3:
@toVersus would you mind working on it?

/assign @toVersus

PR #1702, opened 2020-09-12
Remove patch detail for webhook being logged on info

toVersus:
It might contain the sensitive data, so print it to debug log.

Fixes: #1397

### DIFF
--- a/logging/logkey/constants.go
+++ b/logging/logkey/constants.go
@@ -62,19 +62,4 @@ const (
 	// GitHubCommitID is the key used to represent the GitHub Commit ID where the
 	// Knative component was built from in logs
 	GitHubCommitID = "commit"
-
-	// AdmissionReviewUID is the key used to represent the admission review
-	// request/response UID in logs
-	AdmissionReviewUID = "admissionreview/uid"
-
-	// AdmissionReviewAllowed is the key used to represent whether or not
-	// the admission request was permitted in logs
-	AdmissionReviewAllowed = "admissionreview/allowed"
-
-	// AdmissionReviewResult is the key used to represent extra details into
-	// why an admission request was denied in logs
-	AdmissionReviewResult = "admissionreview/result"
-
-	// AdmissionReviewPatchType is the key used to represent the type of Patch in logs
-	AdmissionReviewPatchType = "admissionreview/patchtype"
 )
--- a/logging/logkey/constants.go
+++ b/logging/logkey/constants.go
@@ -62,4 +62,19 @@ const (
 	// GitHubCommitID is the key used to represent the GitHub Commit ID where the
 	// Knative component was built from in logs
 	GitHubCommitID = "commit"
+
+	// AdmissionReviewUID is the key used to represent the admission review
+	// request/response UID in logs
+	AdmissionReviewUID = "knative.dev/admissionreview/uid"
+
+	// AdmissionReviewAllowed is the key used to represent whether or not
+	// the admission request was permitted in logs
+	AdmissionReviewAllowed = "knative.dev/admissionreview/allowed"
+
+	// AdmissionReviewResult is the key used to represent extra details into
+	// why an admission request was denied in logs
+	AdmissionReviewResult = "knative.dev/admissionreview/result"
+
+	// AdmissionReviewPatchType is the key used to represent the type of Patch in logs
+	AdmissionReviewPatchType = "knative.dev/admissionreview/patchtype"
 )
--- a/logging/logkey/constants.go
+++ b/logging/logkey/constants.go
@@ -65,16 +65,16 @@ const (
 
 	// AdmissionReviewUID is the key used to represent the admission review
 	// request/response UID in logs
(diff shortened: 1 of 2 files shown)


webhook/admission.go
```diff
@@ -101,9 +101,12 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
 			patchType = string(*reviewResponse.PatchType)
 		}
 
-		logger.Infof("AdmissionReview for %#v: %s/%s response={ UID: %#v, Allowed: %t, Status: %#v, Patch: %s, PatchType: %s, AuditAnnotations: %#v}",
+		logger.Infof("AdmissionReview for %#v: %s/%s response={ UID: %#v, Allowed: %t, Status: %#v, PatchType: %s, AuditAnnotations: %#v}",
```
evankanderson, line 104: I wonder if we want to store a few of these fields (status, allowed, patchType) using `.With` for structured logging, rather than burying them in a string.

evankanderson (commented):
I'm wondering if we can load the logging context before the first `Infof` and then share the context between the two log calls.

webhook/admission.go
toVersus: Done. Also remove some admission request info from the message since they have already stored in logging context.

logging/logkey/constants.go
```diff
@@ -62,4 +62,19 @@ const (
 	// GitHubCommitID is the key used to represent the GitHub Commit ID where the
 	// Knative component was built from in logs
 	GitHubCommitID = "commit"
+
+	// AdmissionReviewUID is the key used to represent the admission review
+	// request/response UID in logs
+	AdmissionReviewUID = "knative.dev/admissionreview/uid"
```
evankanderson, line 68: I don't think it's necessary to namespace these logging keys, but it doesn't especially hurt, either. They should compress fairly well.

webhook/admission.go
```diff
@@ -101,15 +101,20 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
...
-		logger.Infof("AdmissionReview for %#v: %s/%s response={ UID: %#v, Allowed: %t, Status: %#v, Patch: %s, PatchType: %s, AuditAnnotations: %#v}",
-			review.Request.Kind, review.Request.Namespace, review.Request.Name,
-			reviewResponse.UID, reviewResponse.Allowed, reviewResponse.Result, string(reviewResponse.Patch), patchType, reviewResponse.AuditAnnotations)
-
 		if !reviewResponse.Allowed || reviewResponse.PatchType != nil || response.Response == nil {
 			response.Response = reviewResponse
 		}
 		response.Response.UID = review.Request.UID
 
+		logger = logger.With(
```
evankanderson, line 109: `SugaredLogger.With` takes pairs of untyped arguments:

https://godoc.org/go.uber.org/zap#SugaredLogger.With

(Personally, I find the non-sugared logger more intuitive because the types help you out, but we seem to prefer the sugared version.)

evankanderson (commented):
You can take or leave the logkey comment, but I think you got the wrong form of `.With`.

logging/logkey/constants.go
toVersus: Agreed. Saving logging data is important for us.

webhook/admission.go
toVersus: Thank you for pointing it out. Fixed to use sugared version for both places. Done.

logging/logkey/constants.go
```diff
@@ -62,4 +62,19 @@ const (
...
+	// AdmissionReviewAllowed is the key used to represent whether or not
+	// the admission request was permitted in logs
+	AdmissionReviewAllowed = "admissionreview/allowed"
+
+	// AdmissionReviewResult is the key used to represent extra details into
+	// why an admission request was denied in logs
+	AdmissionReviewResult = "admissionreview/result"
+
+	// AdmissionReviewPatchType is the key used to represent the type of Patch in logs
+	AdmissionReviewPatchType = "admissionreview/patchtype"
```
dprotaso, line 79: I would just pull these into `webhook/admission.go` since they're not use anywhere else in the project

webhook/admission.go
```diff
@@ -101,15 +101,20 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
...
+			logkey.AdmissionReviewUID, string(reviewResponse.UID),
+			logkey.AdmissionReviewAllowed, reviewResponse.Allowed,
+			logkey.AdmissionReviewResult, reviewResponse.Result.String(),
+			logkey.AdmissionReviewPatchType, patchType)
```
dprotaso, line 113: patch type would only be relevant when printing the `Patch` body

dprotaso (commented):
hey sorry for the delay - just got back from vacation and going through a backlog of stuff

logging/logkey/constants.go
toVersus: Thanks for looking into this. That makes sense. Done.

PR #1703, opened 2020-09-14
Make Do have consistent retry semantics to Poll.

mattmoor:
Today Poll delegates to Do and has additional logic to retry certain error conditions returned by Do focused around connection issues.  This means that we are inconsistent about retrying connection issues between Poll and Do.

This change has two main parts:
1. Inline the main logic previously within Do into Poll.
2. Invert the previous relationship (Poll -> Do) to have Do effectively become "Poll once".

With this, Do should have consistent connection retry semantics to Poll (because it IS just a call to Poll).

See also: https://knative.slack.com/archives/CA4DNJ9A4/p1600011374118700

/cc @tcnghia 
/assign @tcnghia 
/hold

I want to HOLD this while I validate it downstream, since Serving is the real test.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -62,7 +62,7 @@ func (r *Response) String() string {
 
 // Interface defines the actions that can be performed by the spoofing client.
 type Interface interface {
-	Do(*http.Request) (*Response, error)
+	Do(*http.Request, ...ErrorRetryChecker) (*Response, error)
 	Poll(*http.Request, ResponseChecker, ...ErrorRetryChecker) (*Response, error)
 }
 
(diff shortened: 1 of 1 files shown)


mattmoor:
Here's the downstream PR: https://github.com/knative/serving/pull/9393

tcnghia:
https://github.com/knative/serving/pull/9393 is green

mattmoor:
Yup, the e2e tests are, which is all that really matter.

WDYT?

test/spoof/spoof.go
```diff
@@ -166,71 +166,47 @@ func ResolveEndpoint(ctx context.Context, kubeClientset *kubernetes.Clientset, d
...
 		}
+		defer rawResp.Body.Close()
 
+		rawResp.Header.Add(zipkin.ZipkinTraceIDHeader, span.SpanContext().TraceID.String())
+		body, err := ioutil.ReadAll(rawResp.Body)
+		if err != nil {
+			return true, err
+		}
+
+		resp = &Response{
```
tcnghia, line 204:
```suggestion
		defer rawResp.Body.Close()
		body, err := ioutil.ReadAll(rawResp.Body)
		if err != nil {
			return true, err
		}

		rawResp.Header.Add(zipkin.ZipkinTraceIDHeader, span.SpanContext().TraceID.String())
		resp = &Response{
```

tcnghia (approved):

test/spoof/spoof.go
mattmoor: I think I got it, but LMK if I missed anything.

tcnghia (approved):
/lgtm

mattmoor:
@markusthoemmes if this makes sense to you, feel free to `/unhold` whenever, otherwise I'll try to catch you in the morning.

markusthoemmes (approved):
/unhold
/lgtm
/approve

PR #1705, opened 2020-09-14
Fix a nit in the spoof

vagababov:
Not all errors result in polling stop — it must be an error we don't retry upon

/assign mattmoor


mattmoor (commented):
/lgtm

PR #1710, opened 2020-09-16
Fix flakes in EnqueueAfter tests

julz:
Fixes flakes such as https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/1709/pull-knative-pkg-unit-tests/1305911274105737216.

the [docs for the Len() function in queue package say](https://github.com/kubernetes/client-go/blob/master/util/workqueue/queue.go#L129):

~~~~
// Len returns the current queue length, for informational purposes only. You
// shouldn't e.g. gate a call to Add() or Get() on Len() being a particular
// value, that can't be synchronized properly.
~~~~

Testing locally, when we see the flake we'll then see the correct length immediately after, so it appears to be caused by inconsistent reads. Directly sending the len to the channel when the poll finishes rather than effectively re-polling an extra time in the test seems to fix up the flakes. Since no-one should be using len() other than for information purposes, this solution seems fine. Ran with `-count 100` locally cleanly.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -756,8 +756,8 @@ const (
 
 func pollQ(q workqueue.RateLimitingInterface, sig chan int) func() (bool, error) {
 	return func() (bool, error) {
-		if q.Len() > 0 {
-			sig <- q.Len()
+		if ql := q.Len(); ql > 0 {
+			sig <- ql
 			return true, nil
 		}
 		return false, nil
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -754,10 +754,10 @@ const (
 	queueCheckTimeout = shortDelay + 500*time.Millisecond
 )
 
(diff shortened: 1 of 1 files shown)


controller/controller_test.go
```diff
@@ -754,18 +754,17 @@ const (
 	queueCheckTimeout = shortDelay + 500*time.Millisecond
 )
 
-func pollQ(q workqueue.RateLimitingInterface, sig chan struct{}) func() (bool, error) {
+func pollQ(q workqueue.RateLimitingInterface, sig chan int) func() (bool, error) {
 	return func() (bool, error) {
 		if q.Len() > 0 {
-			sig <- struct{}{}
+			sig <- q.Len()
```
markusthoemmes, line 760: Fetch q.Len() only once here for consistency?

julz: oops, I swear I meant to do that. yep one sec.

julz: done

PR #1716, opened 2020-09-18
Remove now unused test helpers

vagababov:
They were used in perf tests in serving, but those don't exit anymore either.
Let's free some code and let it go :)

/assign mattmoor @chaodaiG @chizhg


chizhg:
/lgtm
Thanks @vagababov !

coryrc:
For visibility, the `junithelper` is still being installed in prow-tests:

https://github.com/knative/test-infra/blob/master/images/prow-tests/Dockerfile#L97

vagababov:
That's actually quite since test-infra should not depend on pkg, at least that's my understanding.

chizhg:
> For visibility, the `junithelper` is still being installed in prow-tests:
> 
> https://github.com/knative/test-infra/blob/22269d552166def1d8a53775b974aa5258c97afa/images/prow-tests/Dockerfile#L97

Thanks @coryrc for noticing this. I'll submit a PR to pin the tool to one previous `knative/pkg` release before we can completely get rid of it. Let's keep the master of `knative/pkg` cleaner.

coryrc:
> That's actually quite since test-infra should not depend on pkg, at least that's my understanding.

That was done because we use `go mod` to install scripts instead of a correct tool and it broke things, otherwise we would have carried on as before.

PR #1717, opened 2020-09-18
Add codecov for pkg

vagababov:
/assign mattmoor


mattmoor:
/hold

We should pull in the actions first.

mattmoor:
/lgtm
/unhold

Looks like we may want to disable a couple more directories, but nice start 🤩

PR #1718, opened 2020-09-18
We don't really expect high test coverage for test helpers

vagababov:
There's some, but we don't want to be in the red for that.

/assign mattmoor


mattmoor (commented):
/lgtm

PR #1720, opened 2020-09-19
Deep copy views to avoid races

ian-mi:


metrics/resource_view.go
```diff
@@ -295,6 +287,21 @@ func resourceFromKey(s string) *resource.Resource {
 	return r
 }
 
+func copyViews(views []*view.View) []*view.View {
+	viewsCopy := make([]*view.View, 0, len(resourceViews.views))
+	for _, v := range views {
+		c := *v
+		c.TagKeys = make([]tag.Key, 0, len(v.TagKeys))
```
vagababov, line 294: use `copy` builtin.

metrics/resource_view.go
```diff
@@ -295,6 +287,21 @@ func resourceFromKey(s string) *resource.Resource {
...
+		c.TagKeys = append(c.TagKeys, v.TagKeys...)
+		c.Aggregation = new(view.Aggregation)
+		*c.Aggregation = *v.Aggregation
```
vagababov, line 297:
```suggestion
agg := *v.Aggregation
c.Aggregation = &agg
```

vagababov (commented):
If only we could codegen deepcopy on it :)

metrics/resource_view.go
ian-mi: Isn't this equivalent?

vagababov: Probably, but we're mostly doing this in the codebase. I.e. readability mostly.

metrics/resource_view.go
```diff
@@ -295,6 +287,21 @@ func resourceFromKey(s string) *resource.Resource {
...
 }
 
+func copyViews(views []*view.View) []*view.View {
+	viewsCopy := make([]*view.View, 0, len(resourceViews.views))
+	for _, v := range views {
+		c := *v
+		c.TagKeys = make([]tag.Key, len(v.TagKeys))
+		copy(c.TagKeys, v.TagKeys)
+agg := *v.Aggregation
+c.Aggregation = &agg
```
mattmoor, line 297: Format Go code:
```suggestion
		agg := *v.Aggregation
		c.Aggregation = &agg
```

mattmoor (commented):
Produced via:

yanweiguo:
Are the failed tests relative?

ian-mi:
The downstream test is failing because it does not always unregister views after each test and opencensus is no longer able to recognize that the aggregations are equal since it relied on a shallow comparison. https://github.com/knative/serving/pull/9497 should fix this.

PR #1725, opened 2020-09-21
Apply gosec checks to the codebase, except tests.

markusthoemmes:
Pretty much the same as https://github.com/knative/serving/pull/9489.
/hold

To review the TLS stuff.


configmap/hash-gen/main.go
```diff
@@ -49,7 +49,7 @@ func processFile(fileName string) error {
 		return err
 	}
 
-	if err := ioutil.WriteFile(fileName, out, 0644); err != nil {
+	if err := ioutil.WriteFile(fileName, out, 0600); err != nil {
```
mattmoor, line 52: should we get this from an `os.Stat`?

mattmoor:
TestMetrics (cc @evankanderson )
/retest

kmeta/names.go
```diff
@@ -58,6 +59,7 @@ func ChildName(parent, suffix string) string {
 			// remove it.
 			return strings.TrimRight(ret, "-")
 		}
+		// nolint:gosec // No strong cryptography needed.
```
evankanderson, line 62: Other than cross-version compatibility, this could be a crew, which is a lot cheaper to compute (for example, it doesn't need to be fixed-duration).

evankanderson:
Ugh, thanks!

configmap/hash-gen/main.go
evankanderson: I don't see a need to copy permissions from somewhere else here.

markusthoemmes: I think this is only relevant if the file doesn't already exist.

mattmoor: Probably worth checking / adding a comment?

markusthoemmes: Done. I rolled the change back, didn't seem important enough to fuzz about a lot tbh and this currently seems to work just fine.

mattmoor (commented):
/lgtm
/approve

PR #1729, opened 2020-09-22
Add a func to return a connection which has been established

yanweiguo:
`NewDurableSendingConnection` creates a managed connection which is connecting asynchronously. We have to wait the connection to be established before sending message.

Proposed to add a function `NewDurableSendingConnectionGuaranteed` to return a connection which can be used immediately by wrapping a wait.

Also remove `IsEstablished` as we can use `Status` instead. Its downstream usage has been removed.

/assign @vagababov

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -64,6 +64,9 @@ type ManagedConnection struct {
 	closeChan chan struct{}
 	closeOnce sync.Once
 
+	establishChan chan struct{}
+	establishOnce sync.Once
+
 	// Used to capture asynchronous processes to be waited
 	// on when shutting the connection down.
 	processingWg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


websocket/connection.go
```diff
@@ -92,6 +92,25 @@ func NewDurableSendingConnection(target string, logger *zap.SugaredLogger) *Mana
...
+// that can only send messages to the endpoint it connects to. It returns
+// the connection if the connection can be established within the given
+// `duration`. Otherwise it returns nil.
+//
+// The connection will continuously be kept alive and reconnected
+// in case of a loss of connectivity.
+func NewDurableSendingConnectionGuaranteed(target string, duration time.Duration, logger *zap.SugaredLogger) *ManagedConnection {
+	c := NewDurableConnection(target, nil, logger)
+
+	if err := wait.PollImmediate(10*time.Millisecond, duration, func() (bool, error) {
```
vagababov, line 105:
```suggestion
	if err := wait.PollImmediate(5*time.Millisecond, duration, func() (bool, error) {
```
Perhaps?

vagababov (commented):
Not a fan of the name, but 🤷 
Don't have a better suggestion.

/lgtm

I'll let Markus approve, but it seems straightforward (alternative is first `connect` call to be in the sync flow.

/assign @markusthoemmes

websocket/connection.go
```diff
@@ -92,6 +92,25 @@ func NewDurableSendingConnection(target string, logger *zap.SugaredLogger) *Mana
...
+// that can only send messages to the endpoint it connects to. It returns
+// the connection if the connection can be established within the given
+// `duration`. Otherwise it returns nil.
+//
+// The connection will continuously be kept alive and reconnected
+// in case of a loss of connectivity.
+func NewDurableSendingConnectionGuaranteed(target string, duration time.Duration, logger *zap.SugaredLogger) *ManagedConnection {
+	c := NewDurableConnection(target, nil, logger)
+
+	if err := wait.PollImmediate(5*time.Millisecond, duration, func() (bool, error) {
```
markusthoemmes, line 105: Can we achieve this without a poll potentially? :thinking:

In broad strokes I'm thinking about adding a channel that is closed when the first connect is successful.

websocket/connection.go
```diff
@@ -92,6 +92,25 @@ func NewDurableSendingConnection(target string, logger *zap.SugaredLogger) *Mana
...
 }
 
+// NewDurableSendingConnectionGuaranteed creates a new websocket connection
+// that can only send messages to the endpoint it connects to. It returns
+// the connection if the connection can be established within the given
+// `duration`. Otherwise it returns nil.
+//
+// The connection will continuously be kept alive and reconnected
+// in case of a loss of connectivity.
+func NewDurableSendingConnectionGuaranteed(target string, duration time.Duration, logger *zap.SugaredLogger) *ManagedConnection {
```
markusthoemmes, line 102: Should we return an error if we don't create the connection in the desired timeframe? Just returnin `nil` feels like abusing the type system a little imo. We could potentially use the error to surface the reason too.

websocket/connection.go
yanweiguo: done

websocket/connection.go
yanweiguo: done

PR #1735, opened 2020-09-23
use buoy for pkg

n3wscott:
This floats the current pkg (on master) to release `v0.19`.


n3wscott:
/hold

Moving buoy to knative-sandbox

PR #1738, opened 2020-09-23
Align linters with serving (enables stylecheck and asciicheck)

markusthoemmes:
As per title, to align this repo with the serving repo as it stands.


mattmoor (commented):
/lgtm

We might want to pick up @tcnghia's codecov change so the `codecov/patch` is less annoying

PR #1740, opened 2020-09-23
Make mako stub results reader cross-platform

tommyreddad:
The original mako stub `read_results.sh` script relies on `netstat` options which are not available on Mac, in order to find a free TCP port, which makes it impossible to launch some performance tests on Mac. This change instead uses the cross-platform utility `lsof` and portable bash environment for convenience.


tommyreddad:
Closing until the discussion on reworking the new test framework is resolved.

PR #1741, opened 2020-09-23
Add TTL expiry for items in allMeters.meters map

MontyCarter:
Currently, pkg/metrics keeps one meterExporter for each resource the system is sending metrics about.

As resources  fall off the system and no longer send metrics (deleting a knative revision, for example), a slow leak occurs as orphaned meterExporters are never cleaned up.

This PR adds a TTL expiry bit for each meterExporter, and cleans them up when they expire.

### DIFF
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -29,6 +29,7 @@ import (
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	"k8s.io/apimachinery/pkg/util/clock"
 )
 
 type storedViews struct {
@@ -40,6 +41,7 @@ type meterExporter struct {
 	m view.Meter    // NOTE: DO NOT RETURN THIS DIRECTLY; the view.Meter will not work for the empty Resource
 	o stats.Options // Cache the option to reduce allocations
 	e view.Exporter
+	t time.Time // Time when last access occurred
 }
 
 // ResourceExporterFactory provides a hook for producing separate view.Exporters
@@ -48,17 +50,70 @@ type meterExporter struct {
 // Tags are.
 type ResourceExporterFactory func(*resource.Resource) (view.Exporter, error)
 type meters struct {
-	meters  map[string]*meterExporter
-	factory ResourceExporterFactory
-	lock    sync.Mutex
+	meters         map[string]*meterExporter
+	factory        ResourceExporterFactory
+	lock           sync.Mutex
+	clock          clock.Clock
+	ticker         clock.Ticker
+	tickerStopChan chan struct{}
 }
 
 // Lock regime: lock allMeters before resourceViews. The critical path is in
 // optionForResource, which must lock allMeters, but only needs to lock
 // resourceViews if a new meter needs to be created.
-var resourceViews = storedViews{}
-var allMeters = meters{
-	meters: map[string]*meterExporter{"": &defaultMeter},
+var (
+	resourceViews = storedViews{}
+	allMeters     = meters{
+		meters: map[string]*meterExporter{"": &defaultMeter},
+		clock:  clock.Clock(clock.RealClock{}),
+	}
+
+	cleanupOnce                = new(sync.Once)
+	meterExporterScanFrequency = 1 * time.Minute
+	maxMeterExporterAge        = 10 * time.Minute
+)
+
+// cleanup looks through allMeter's meterExporter and prunes any that were last visited
+// more than maxMeterExporterAge ago. This will clean up any exporters for resources
+// that no longer exist on the system (a replaced revision, e.g.).  All meterExporter
+// lookups (and creations) run meterExporterForResource, which updates the timestamp
+// on each access.
+func cleanup() {
+	expiryCutoff := allMeters.clock.Now().Add(-1 * maxMeterExporterAge)
+	allMeters.lock.Lock()
+	defer allMeters.lock.Unlock()
+	for key, meter := range allMeters.meters {
+		if key != "" && meter.t.Before(expiryCutoff) {
+			flushGivenExporter(meter.e)
+			delete(allMeters.meters, key)
+		}
+	}
+}
+
+// startCleanup is configured to only run once in production.  For testing,
+// calling startCleanup again will terminate old cleanup threads,
+// and restart anew.
+func startCleanup() {
+	if allMeters.tickerStopChan != nil {
+		close(allMeters.tickerStopChan)
+	}
+
+	curClock := allMeters.clock
+	newTicker := curClock.NewTicker(meterExporterScanFrequency)
+	newStopChan := make(chan struct{})
+	allMeters.tickerStopChan = newStopChan
+	allMeters.ticker = newTicker
+	go func() {
+		defer newTicker.Stop()
+		for {
+			select {
+			case <-newTicker.C():
+				cleanup()
+			case <-newStopChan:
+				return
+			}
+		}
+	}()
 }
 
 // RegisterResourceView is similar to view.Register(), except that it will
@@ -174,7 +229,9 @@ func flushResourceExporters() {
 }
 
 // ClearMetersForTest clears the internal set of metrics being exported,
-// including cleaning up background threads.
+// including cleaning up background threads, and restarts cleanup thread.
+//
+// If a replacement clock is desired, it should be in allMeters.clock before calling.
 func ClearMetersForTest() {
 	allMeters.lock.Lock()
 	defer allMeters.lock.Unlock()
@@ -186,6 +243,7 @@ func ClearMetersForTest() {
 		meter.m.Stop()
 		delete(allMeters.meters, k)
 	}
+	startCleanup()
 }
 
 func meterExporterForResource(r *resource.Resource) *meterExporter {
@@ -195,6 +253,7 @@ func meterExporterForResource(r *resource.Resource) *meterExporter {
 		mE = &meterExporter{}
 		allMeters.meters[key] = mE
 	}
+	mE.t = allMeters.clock.Now()
 	if mE.o != nil {
 		return mE
 	}
@@ -219,6 +278,9 @@ func meterExporterForResource(r *resource.Resource) *meterExporter {
 // optionForResource finds or creates a stats exporter for the resource, and
 // returns a stats.Option indicating which meter to record to.
 func optionForResource(r *resource.Resource) (stats.Options, error) {
+	// Start the allMeters cleanup thread, if not already started
+	cleanupOnce.Do(startCleanup)
+
 	allMeters.lock.Lock()
 	defer allMeters.lock.Unlock()
 
@@ -313,6 +375,7 @@ var _ view.Meter = (*defaultMeterImpl)(nil)
 var defaultMeter = meterExporter{
 	m: &defaultMeterImpl{},
 	o: stats.WithRecorder(nil),
+	t: time.Now(), // time.Now() here is ok - defaultMeter never expires, so this won't be checked
 }
 
 func (*defaultMeterImpl) Record(*tag.Map, interface{}, map[string]interface{}) {
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -37,6 +37,7 @@ import (
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
+	"k8s.io/apimachinery/pkg/util/clock"
 
 	emptypb "github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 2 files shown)


metrics/resource_view.go
```diff
@@ -57,8 +58,23 @@ type meters struct {
...
+const cacheSize int = 4096
+
+func init() {
+	// lru.New() only throws on negative size
+	// The default meter is required - make sure it's not evicted
+	cache, _ := lru.NewWithEvict(cacheSize, func(key interface{}, value interface{}) {
+		if key == "" {
+			allMeters.lock.Lock()
+			defer allMeters.lock.Unlock()
+			allMeters.meters.Add(key, value)
```
vagababov, line 73: the `cache` is thread safe, so 1) don't need to lock 2) just use the `cache` var.

yanweiguo:
/assign @evankanderson 

I'm not sure the reason why we need to use a cache to replace map. Evan has more knowledge than me on these codes.

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
```diff
@@ -57,8 +58,22 @@ type meters struct {
...
+// Sizing is largely arbitrary - selected to easily cover a realistic number of revisions
+const cacheSize int = 4096
+
+func init() {
+	// lru.New() only throws on negative size
+	// The default meter is required - make sure it's not evicted
+	var cache *lru.Cache
+	cache, _ = lru.NewWithEvict(cacheSize, func(key interface{}, value interface{}) {
+		if key == "" {
+			cache.Add(key, value)
```
evankanderson, line 72: Codecov complains that this branch/code is not tested. I'd normally say "meh", but in this case, it seems fairly critical.

metrics/resource_view.go
```diff
@@ -94,7 +111,9 @@ func UnregisterResourceView(views ...*view.View) {
 	resourceViews.lock.Lock()
 	defer resourceViews.lock.Unlock()
 
-	for _, meter := range allMeters.meters {
+	for _, key := range allMeters.meters.Keys() {
```
evankanderson, line 114: It's unfortunate that we may need to generate and throw away a slice here... it would be nice if there was an iteration interface on the Cache.

Still, this is run fairly infrequently (only when registering views, which should be at startup time).

evankanderson (commented):
(From Slack)

Let me think about this PR — the problem is that we are trying to keep statistics on the resources that our component is currently handling (e.g. as the `activator` or `autoscaler`). Currently, we never expire metrics exports, which causes a slow leak of metrics over time.

Your PR changes that behavior to set a hard cap on the number of metrics kept over time, even if the component is handling more than N=4096 resources.

>> What other resources might have exporters besides each knative revision?
>
> This could also be used for Knative Brokers or Triggers in eventing.
> Or Channels.

Basically, there should be a custom object in Kubernetes that we can point to that justifies the memory usage in this cache, but I’m not sure we want to tie this to the informer cache. It should be bounded on the upper side by the informer cache

Maybe something with a simple access-age test would be sufficient?

Basically, I'm not sure that a fixed-size cache is the right choice here, particularly when other parts of the system (like the informer cache) have no such limit.

metrics/resource_view.go
```diff
@@ -59,6 +63,38 @@ type meters struct {
 var resourceViews = storedViews{}
 var allMeters = meters{
 	meters: map[string]*meterExporter{"": &defaultMeter},
+	clock:  cleanupClock,
+}
+
+var cleanupClock = clock.Clock(clock.RealClock{}) // Var needed to avoid allMeters->defaultMeter->allMeters.clock cycle
+var cleanupOnce sync.Once
+var meterExporterScanFrequency = 1 * time.Minute
+var maxMeterExporterAge = 10 * time.Minute
```
vagababov, line 72:
```suggestion
var (
resourceViews = storedViews{}
allMeters = meters{
	meters: map[string]*meterExporter{"": &defaultMeter},
	clock:  cleanupClock,
}

cleanupClock = clock.Clock(clock.RealClock{}) // Var needed to avoid allMeters->defaultMeter->allMeters.clock cycle
cleanupOnce sync.Once
meterExporterScanFrequency = 1 * time.Minute
maxMeterExporterAge = 10 * time.Minute
)
```

vagababov (commented):
Not a fan of bespoke things likes this, but 🤷

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
```diff
@@ -51,14 +54,51 @@ type meters struct {
...
+	}
+
+	cleanupClock               = clock.Clock(clock.RealClock{}) // Var needed to avoid allMeters->defaultMeter->allMeters.clock cycle
+	cleanupOnce                = new(sync.Once)
+	meterExporterScanFrequency = 1 * time.Minute
+	maxMeterExporterAge        = 10 * time.Minute
+)
+
+func cleanup() {
+	log.Printf("in cleanup")
```
vagababov, line 78:
```suggestion
```

metrics/resource_view.go
```diff
@@ -51,14 +54,51 @@ type meters struct {
...
+
+}
+
+func cleanupLoop() {
+	for range allMeters.ticker.C() {
+		cleanup()
+	}
+}
+
+func startCleanup() {
```
vagababov, line 97: should this just be in `init()` then locking is not necessary?

metrics/resource_view_test.go
```diff
@@ -144,6 +145,68 @@ func TestSetFactory(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Should succeed getting option, instead got error %v", err)
+	}
+	// (123=0, 456=1.5m)
+
+	// Expire the second entry
+	fakeClock.Step(9 * time.Minute) // t+12m
+	time.Sleep(1 * time.Second)     // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
+	if len(allMeters.meters) != 2 {
+		t.Errorf("2 meterExporters should exist, however %v found.", len(allMeters.meters))
```
vagababov, line 195: Got before want:
```suggestion
		t.Errorf("len(allMeters)=%d, want: 2", len(allMeters.meters))
```

evankanderson:
Two notes:

1. You probably want to update the title of this PR.
2. You probably want a description.

Taking a look now.

metrics/resource_view.go
```diff
@@ -19,6 +19,7 @@ package metrics
 import (
 	"errors"
 	"fmt"
+	"log"
```
evankanderson, line 22: The `log` packages produces a different (unstructured) log format -- I'd try to either wire a logger in somehow (if necessary) or drop the use of `log` here.

metrics/resource_view.go
```diff
@@ -51,14 +54,51 @@ type meters struct {
...
-var allMeters = meters{
-	meters: map[string]*meterExporter{"": &defaultMeter},
+var (
+	resourceViews = storedViews{}
+	allMeters     = meters{
+		meters: map[string]*meterExporter{"": &defaultMeter},
+		clock:  cleanupClock,
+	}
+
+	cleanupClock               = clock.Clock(clock.RealClock{}) // Var needed to avoid allMeters->defaultMeter->allMeters.clock cycle
```
evankanderson, line 71: Is this still true? Can't you just initialize the clock in `allMeters` with this now?

metrics/resource_view.go
```diff
@@ -51,14 +54,51 @@ type meters struct {
...
+	maxMeterExporterAge        = 10 * time.Minute
+)
+
+func cleanup() {
+	log.Printf("in cleanup")
+	allMeters.lock.Lock()
+	defer allMeters.lock.Unlock()
+	expiryCutoff := allMeters.clock.Now().Add(-1 * maxMeterExporterAge)
+	for key, meter := range allMeters.meters {
+		if key != "" && meter.t.Before(expiryCutoff) {
```
evankanderson, line 83: This is probably worth a comment about how it's avoiding a memory leak from Resources that are no longer monitored / emitting statistics.

metrics/resource_view.go
```diff
@@ -51,14 +54,51 @@ type meters struct {
...
+			flushGivenExporter(meter.e)
+			delete(allMeters.meters, key)
+		}
+	}
+
+}
+
+func cleanupLoop() {
```
evankanderson, line 91: I'd just make this an anonymous function in `startCleanup` unless there's something else that needs to call this.

metrics/resource_view.go
```diff
@@ -186,6 +226,7 @@ func ClearMetersForTest() {
 		meter.m.Stop()
 		delete(allMeters.meters, k)
 	}
+	cleanupOnce = new(sync.Once)
```
evankanderson, line 229: Hmm -- this will clear the the `Once`, but won't the old goroutine still hang around?

metrics/resource_view.go
```diff
@@ -313,6 +358,7 @@ var _ view.Meter = (*defaultMeterImpl)(nil)
 var defaultMeter = meterExporter{
 	m: &defaultMeterImpl{},
 	o: stats.WithRecorder(nil),
+	t: cleanupClock.Now(),
```
evankanderson, line 361: Why does this need a clock?

evankanderson (commented):
There are also some code suggestions by golint as GitHub Actions notes.

metrics/resource_view.go
MontyCarter: It doesn't need a clock - it needs a timestamp.  However, I just realized that we can avoid ever reading this for the defaultMeter, since it never expires.

metrics/resource_view.go
MontyCarter: See comment on defaultMeter initialization.  The meters need a timestamp, but for defaultMeter the value doesn't matter since it never expires (so I'll just avoid reading the timestamp in that case).

I'll remove this.

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
MontyCarter: Done.

evankanderson:
> Not a fan of bespoke things likes this, but 🤷

I think Monty has also looked at various lifetime-based caches. Ideally, the set of current `meterExporter`s would be a subset of the current set of items in the Informer cache, which is theoretically unbounded, but at least people expect it to grow as clusters grow, but not over time for the same set of live objects.

The current implementation in `metrics` suffers from a leak as resources are cycled, since it never expires an exporter from the map, though they may quiesce a bit if there are no further metrics recorded.

metrics/resource_view.go
MontyCarter: Yes, I realized that once I noticed that `go test --race` was failing.

I've still got a bit of ramping up to do on concurrency in go...  I'll

metrics/resource_view.go
MontyCarter: This section (this section's logical replacement, really) has coverage, now.

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view_test.go
MontyCarter: Done.

MontyCarter:
> Two notes:
> 
> 1. You probably want to update the title of this PR.
> 2. You probably want a description.
> 
> Taking a look now.

Done.

metrics/resource_view.go
```diff
@@ -48,17 +50,70 @@ type meterExporter struct {
...
+
+// cleanup looks through allMeter's meterExporter and prunes any that were last visited
+// more than maxMeterExporterAge ago.  This will clean up any exporters for resources
+// that no longer exist on the system (a replaced revision, e.g.).  All meterExporter
+// lookups (and creations) run meterExporterForResource, which updates the timestamp
+// on each access.
+func cleanup() {
+	allMeters.lock.Lock()
+	defer allMeters.lock.Unlock()
+	expiryCutoff := allMeters.clock.Now().Add(-1 * maxMeterExporterAge)
```
vagababov, line 84: nit: this can probably be outside the lock

metrics/resource_view.go
```diff
@@ -48,17 +50,70 @@ type meterExporter struct {
...
+// calling startCleanup again will terminate old cleanup threads,
+// and restart anew.
+func startCleanup() {
+	if allMeters.tickerStopChan != nil {
+		close(allMeters.tickerStopChan)
+	}
+
+	curClock := allMeters.clock
+	newTicker := curClock.NewTicker(meterExporterScanFrequency)
+	newStopChan := make(chan bool)
```
vagababov, line 103:
```suggestion
	newStopChan := make(chan struct{})
```

MontyCarter:
> > Not a fan of bespoke things likes this, but 🤷
> 
> I think Monty has also looked at various lifetime-based caches. Ideally, the set of current `meterExporter`s would be a subset of the current set of items in the Informer cache, which is theoretically unbounded, but at least people expect it to grow as clusters grow, but not over time for the same set of live objects.
> 
> The current implementation in `metrics` suffers from a leak as resources are cycled, since it never expires an exporter from the map, though they may quiesce a bit if there are no further metrics recorded.
> 
> Would you prefer to make the expiring-map construct a library import? I'm of two minds here, particularly with the lack of Go generics. On one hand, it's sort of silly to maintain our own implementation; on the other hand, relying on a third-party implementation which isn't actively maintained may not improve things much. (Guava has the benefit that there's a team continually improving it, which a library posted by an individual may not have.)

There are three candidates that I identified, if we decide to go this way:

- https://github.com/ReneKroon/ttlcache - Good interface, active project, but doesn't have ability to iterate over contents - would have to add
- https://github.com/patrickmn/go-cache - Acceptable interface, seems active enough, but "iterating" creates a new map
- https://github.com/muesli/cache2go - Not as active.  Less mature interface, less active project, but provides Foreach which takes a function pointer to run against each item.

metrics/resource_view.go
MontyCarter: In resource_view.go, there is a ClearMetersForTest.  I felt like this was close enough in its intent to being a reset function, so I thought I would keep this around so the whole library can be reinitialized.

The only test that currently needs the clock is TestAllMeterExpiration, which is new.  If we don't allow swapping the clock out after initialization (that is, allow swapping a running time.Ticker for a new one based on a fake clock), then this test function needs to run before anything else to set the clock.  This is error prone as any new test could break this ordering.

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
MontyCarter: Is this just a memory optimization?

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view.go
MontyCarter: Done, btw.

metrics/resource_view.go
```diff
@@ -313,6 +375,7 @@ var _ view.Meter = (*defaultMeterImpl)(nil)
 var defaultMeter = meterExporter{
 	m: &defaultMeterImpl{},
 	o: stats.WithRecorder(nil),
+	t: time.Now(), // time.Now() here is ok - defaultMeter never expires, so we wont read
```
vagababov, line 378:
```suggestion
	t: time.Now(), // time.Now() here is ok - defaultMeter never expires, so this will not be checked.
```

metrics/resource_view_test.go
```diff
@@ -144,6 +145,71 @@ func TestSetFactory(t *testing.T) {
 	}
 }
 
+func TestAllMetersExpiration(t *testing.T) {
+
```
vagababov, line 149:
```suggestion
```

vagababov:
/lgtm
In general

metrics/resource_view.go
MontyCarter: Done.

metrics/resource_view_test.go
MontyCarter: Done.

vagababov (commented):
/lgtm

I'll let Evan approve.

metrics/resource_view.go
```diff
@@ -313,6 +375,7 @@ var _ view.Meter = (*defaultMeterImpl)(nil)
 var defaultMeter = meterExporter{
 	m: &defaultMeterImpl{},
 	o: stats.WithRecorder(nil),
+	t: time.Now(), // time.Now() here is ok - defaultMeter never expires, so this won't be checked
```
evankanderson, line 378: Do you need to initalize this at all if it won't be checked?

evankanderson (commented):
I'm slightly worried about adding some flakiness in background operations here (particularly since `resource_view_test.go` already has some flakiness), but overall this seems like a necessary improvement.

I'm going to keep poking at the flakiness I've seen in the other tests.

/approve

PR #1747, opened 2020-09-25
Some more replacements to atomic.* types

vagababov:
More readable and declarative.

/assign @julz mattmoor

### DIFF
--- a/apis/duck/cached_test.go
+++ b/apis/duck/cached_test.go
@@ -31,7 +31,7 @@ import (
 
 type BlockingInformerFactory struct {
 	block  chan struct{}
-	nCalls *atomic.Int32
+	nCalls atomic.Int32
 }
 
 var _ InformerFactory = (*BlockingInformerFactory)(nil)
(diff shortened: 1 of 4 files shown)


apis/duck/cached_test.go
```diff
@@ -19,24 +19,25 @@ package duck
...
 	"golang.org/x/sync/errgroup"
+
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/cache"
 )
 
 type BlockingInformerFactory struct {
 	block  chan struct{}
-	nCalls int32
+	nCalls *atomic.Int32
```
julz, line 34: Pretty much a nit: think you can just use `atomic.Int32` here since the zero value is valid, and then you don't need the NewInt32 later (eg like in [throttler](https://github.com/knative/serving/blob/98f3589c694b7ed987b075c4fb4465edead6e4e9/pkg/activator/net/throttler.go#L80))

yanweiguo (commented):
/lgtm
/hold for julz's comment.

vagababov:
Awesome, didn't know that.
Done!

apis/duck/cached_test.go
vagababov: Done.

julz (commented):
/lgtm
/unhold

PR #1746, opened 2020-09-25
[Automated] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc n3wscott vagababov
/assign n3wscott vagababov

PR #1753, opened 2020-09-29
Fix the racy test.

vagababov:
The problem is that the original test presumed the 2 other backgound calls have
finished already.
Which is not really true always.
E.g. here: https://github.com/knative/pkg/pull/1746/checks?check_run_id=1179720602
So just wait for all 3 items that we expect to pass.

Also other nits around the file

/assign mattmoor


vagababov:
unrelated 💣 
/retest

vagababov:
cc @evankanderson  — exporter seems to flare up again
/retest

tcnghia (approved):
/lgtm

PR #1754, opened 2020-09-29
Format optimization cleanup

vagababov:
Running the script at test items first
The script is smart enough not to capture '\n%s' for proper diff printing even :-)

Note that `t.X()` methods always put a space between the output items, vs what `fmt` does (only if next param is not string) and zap (never).

/assign mattmoor


mattmoor (commented):
/lgtm

vagababov:
Unrelated downstream failure... flake.

tcnghia (approved):
/lgtm

PR #1755, opened 2020-09-29
Fix fmt and logger formatting

vagababov:
/assign mattmoor @tcnghia


vagababov:
I had to apply a lot of coercing to the pkg/apis/tests.

Should pass the tests now.

mattmoor (commented):
/lgtm

PR #1757, opened 2020-09-29
Delay informer start after controller.

### LINKED ISSUE #1756: PR #1658 is preventing resync period of informer to be overwritten
/kind bug

PR #1658 changes the order of informer setup and controller setup, however the resync period of informer can only be overwritten before the informer is started. As a result, all the event handler with customized resync period will have the default of 10 hours resync period after this PR.

## Expected Behavior

Event handlers in `AddEventHandlerWithResyncPeriod` with resync period set to 5 mins should get called every 5 mins.

## Actual Behavior

The event handlers above will only get called every 10 hours or whatever the k8s default has.

In any controller code using `Informer().AddEventHandlerWithResyncPeriod` add resync period of 5 mins, following error will be found when the controller is started.

`W0922 21:37:44.196866 1 shared_informer.go:461] resyncPeriod 300000000000 is smaller than resyncCheckPeriod 36000000000000 and the informer has already started. Changing it to 36000000000000`

zhongduo:
Starting informer before controller will cause resync period set
in the informer not respected, this commit makes EnableInjectionOrDie
to return informers so that the callers can delay the start of the
informer.

Fix #1756

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -127,9 +127,8 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
 	return leaderelection.NewConfigFromConfigMap(leaderElectionConfigMap)
 }
 
-// EnableInjectionOrDie enables Knative Injection and starts the informers.
-// Both Context and Config are optional.
-func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) context.Context {
+// InjectionSetup sets up the context and informers.
+func InjectionSetup(ctx context.Context, cfg *rest.Config) (context.Context, []controller.Informer) {
 	if ctx == nil {
 		ctx = signals.NewContext()
 	}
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -146,18 +146,7 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) context.Context
 	}
 	ctx = injection.WithConfig(ctx, cfg)
 
-	ctx, informers := injection.Default.SetupInformers(ctx, cfg)
-
-	// Start the injection clients and informers.
-	logging.FromContext(ctx).Info("Starting informers...")
-	go func(ctx context.Context) {
```
n3wscott, line 153: The whole point of this method is to start the informers here and not need to duplicate that code in the caller. 

Perhaps we can pass a delay config in the inbound context object instead? The goal is to reduce the amount of special code each injection enabled cmd needs to run. `MainWithConfig` is just one simple example.

Another example of integration: https://github.com/knative-sandbox/eventing-rabbitmq/blob/56208970df1d5db5526a12137d54ea16b37fe149/test/e2e/cmd/recorder/main.go#L32

I don't want to have to copy the informer startup code in all the places I have called enable injection.

injection/sharedmain/main.go
```diff
@@ -129,7 +129,7 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
 
 // EnableInjectionOrDie enables Knative Injection and starts the informers.
 // Both Context and Config are optional.
-func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) context.Context {
+func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Context, []controller.Informer) {
```
n3wscott, line 132: This is a breaking api change for all things that already use this and we can't do that for all the downstreams.

/hold

leaderelection/chaosduck/main.go
```diff
@@ -117,8 +118,10 @@ func quack(ctx context.Context, kc kubernetes.Interface, component string, leade
 
 func main() {
 	ctx := signals.NewContext()
-	ctx = sharedmain.EnableInjectionOrDie(ctx, nil)
-
+	ctx, informers := sharedmain.EnableInjectionOrDie(ctx, nil)
+	if err := controller.StartInformers(ctx.Done(), informers...); err != nil {
```
n3wscott, line 122: yeah ^ exactly, I don't want to need to copy this chunk of code in the dozens of places `EnableInjectionOrDie` is used.

injection/sharedmain/main.go
zhongduo: Delaying the start of goroutine might be a little bit tricky. If we use a magic time duration, then it might cause lots of unit test to fail. If we use a signal instead, then we still need to change all the downstream codes to signal start, which is same as calling `StartInformers` directly.

leaderelection/chaosduck/main.go
zhongduo: Same here, will it be better if the `EnableInjectionOrDie` function return a start function instead of `[]informers`, so downstream just need to call a function instead of this chunk of code? I do know this will still break the API.

injection/sharedmain/main.go
zhongduo: I guess this logic only works if `StartInformers` is called immediately after this function. At least `sharedmain` does not do that, so it needs to be reverted. Callers like `chaosduck/main.go` however seems less a problem, though I still think it should be fixed.

injection/sharedmain/main.go
n3wscott: Please check this PR again. I think I prefer a single method that returns a callback over two new methods.

zhongduo:
Closing this because the fix is done in #1772

PR #1761, opened 2020-09-30
Fix the resource naming for known terms

vagababov:
It's been a longterm pet peeve of mine that we don't name test functions TestAutoTLS etc, because this
function will mutilate the names in to t-l-s, etc.
So collect a list of well known names and catch them separately.

This is not superperformant, some trie would do this job better, but since
this is test name generation, I'll let it slide :-D

/assign @tcnghia mattmoor


tcnghia (approved):
/lgtm

PR #1767, opened 2020-10-02
Delay start of informers in sharedmain

zhongduo:
The EnableInjectionOrDie starts informers immedidately in a go
routine, which changes the order of informer start and controller
setup. As a result setting resync period to informer handler will
be ignored as the informer is already started. Though this will
likely affect more components, changing the EnableInjectionOrDie
function will break the API, so this is a temporary solution to
fix all the controllers depending on sharedmain.

This is a temporary solution to #1757 . We might change it to
use whatever solution #1757 will result in, but for now this can be
an easy fix for controllers depending on sharedmain.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -225,7 +225,16 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 		cfg.Burst = len(ctors) * rest.DefaultBurst
 	}
 
-	ctx = EnableInjectionOrDie(ctx, cfg)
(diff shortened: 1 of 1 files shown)


n3wscott:
/lgtm
/approve

We can follow up next week with a fix to enable injection.

zhongduo:
> We can follow up next week with a fix to enable injection.

Hi @n3wscott I am not sure about the process, but my original plan is to update my go.mod and regenerate the vendor directory. Are you saying that all the knative shared libraries, like this `pkg` repo, is normally updated from the upstream?

PR #1768, opened 2020-10-02
Fail louder when bad context given to injection

Cynocracy:
This uses the presence of the config key to determine whether
this is an injection context, and if it is not, the panic
message includes a notice about injection context behavior.

### DIFF
--- a/codegen/cmd/injection-gen/generators/client.go
+++ b/codegen/cmd/injection-gen/generators/client.go
@@ -98,8 +98,13 @@ func withClient(ctx {{.contextContext|raw}}, cfg *{{.restConfig|raw}}) context.C
 func Get(ctx {{.contextContext|raw}}) {{.clientSetInterface|raw}} {
 	untyped := ctx.Value(Key{})
 	if untyped == nil {
-		{{.loggingFromContext|raw}}(ctx).Panic(
-			"Unable to fetch {{.clientSetInterface}} from context.")
+		if injection.GetConfig(ctx) == nil {
+		    {{.loggingFromContext|raw}}(ctx).Panic(
+		    	    "Unable to fetch {{.clientSetInterface}} from context. This context is not the application context (which is typically given to constructors via sharedmain).")
+		} else {
+		    {{.loggingFromContext|raw}}(ctx).Panic(
+		    	    "Unable to fetch {{.clientSetInterface}} from context.")
+		}
 	}
 	return untyped.({{.clientSetInterface|raw}})
 }
--- a/client/injection/apiextensions/client/client.go
+++ b/client/injection/apiextensions/client/client.go
@@ -42,8 +42,13 @@ func withClient(ctx context.Context, cfg *rest.Config) context.Context {
 func Get(ctx context.Context) clientset.Interface {
 	untyped := ctx.Value(Key{})
 	if untyped == nil {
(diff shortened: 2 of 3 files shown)


codegen/cmd/injection-gen/generators/client.go
```diff
@@ -98,8 +98,13 @@ func withClient(ctx {{.contextContext|raw}}, cfg *{{.restConfig|raw}}) context.C
 func Get(ctx {{.contextContext|raw}}) {{.clientSetInterface|raw}} {
 	untyped := ctx.Value(Key{})
 	if untyped == nil {
-		{{.loggingFromContext|raw}}(ctx).Panic(
-			"Unable to fetch {{.clientSetInterface}} from context.")
+		if injection.GetConfig(ctx) == nil {
+		    {{.loggingFromContext|raw}}(ctx).Panic(
+		    	    "Unable to fetch {{.clientSetInterface}} from context. This context is not the injection context. Only controllers can inject clients.")
```
n3wscott, line 103: `"Only controllers can inject clients."` is not quite true, it is constructors called from the sharedmain injection enabled thread or something like that.

I make these calls too from just main threads.

Cynocracy: mm, ok. "Only the application context (given to controllers via sharedmain)"?

codegen/cmd/injection-gen/generators/client.go
```diff
@@ -98,8 +98,13 @@ func withClient(ctx {{.contextContext|raw}}, cfg *{{.restConfig|raw}}) context.C
 func Get(ctx {{.contextContext|raw}}) {{.clientSetInterface|raw}} {
 	untyped := ctx.Value(Key{})
 	if untyped == nil {
-		{{.loggingFromContext|raw}}(ctx).Panic(
-			"Unable to fetch {{.clientSetInterface}} from context.")
+		if injection.GetConfig(ctx) == nil {
```
n3wscott, line 101: did this method already exist?

Cynocracy: Yup, I'm just reusing the rest config key

codegen/cmd/injection-gen/generators/client.go
n3wscott: given to "constructors", they can be anything, also you can hack it and register a fake informer and grab a client)

codegen/cmd/injection-gen/generators/client.go
n3wscott: cool!

codegen/cmd/injection-gen/generators/client.go
```diff
@@ -98,8 +98,13 @@ func withClient(ctx {{.contextContext|raw}}, cfg *{{.restConfig|raw}}) context.C
...
+		    {{.loggingFromContext|raw}}(ctx).Panic(
+		    	    "Unable to fetch {{.clientSetInterface}} from context. This context is not the application context (which is typically given to controllers via sharedmain).")
```
n3wscott, line 103: controllers -> constructors and I am gtg

Cynocracy:
ah right. codegen.

n3wscott (commented):
/lgtm
/approve

markusthoemmes (approved):
/lgtm
/approve

Issue #1358, opened 2020-05-26
Add fuzzer to the tests of pkg.Status.Conversion


vagababov:
We added a field and then didn't add conversion and that broke downstream.

So we need to add fuzzers that will ensure this doesn't happen.

/cc @dprotaso @mattmoor

vaikas:
I think #1770 fixed this? @vagababov ok to close?

PR #1770, opened 2020-10-05
add fuzzers

vaikas:


dprotaso (changes requested):
I don't think you need to register a custom fuzzer if you're not doing anything special

So I think you can drop most FuzzerFuncs in this PR

vaikas:
Thanks @dprotaso and also for mentioning the status, I'll do that as well then, wasn't aware of the open issue.

vaikas:
Fixes: #1358

PR #1772, opened 2020-10-05
Put EnableInjectionOrDie back on the main path

n3wscott:
Supersedes https://github.com/knative/pkg/pull/1757
Reverts https://github.com/knative/pkg/pull/1767

### DIFF
--- a/injection/config.go
+++ b/injection/config.go
@@ -19,13 +19,14 @@ package injection
 import (
 	"errors"
 	"flag"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/klog"
 	"log"
 	"os"
 	"os/user"
 	"path/filepath"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
 )
 
 // ParseAndGetRestConfigOrDie parses the rest config flags and creates a client or
--- a/injection/config.go
+++ b/injection/config.go
@@ -43,7 +43,7 @@ func ParseAndGetRestConfigOrDie() *rest.Config {
 
 	cfg, err := GetRestConfig(*serverURL, *kubeconfig)
 	if err != nil {
-		log.Fatalf("Error building kubeconfig: %v", err)
+		log.Fatal("Error building kubeconfig: ", err)
 	}
 
 	return cfg
--- a/injection/config.go
+++ b/injection/config.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package injection
+
+import (
+	"errors"
+	"flag"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+	"log"
+	"os"
+	"os/user"
+	"path/filepath"
+)
+
+// ParseAndGetRestConfigOrDie parses the rest config flags and creates a client or
+// dies by calling log.Fatalf.
+func ParseAndGetRestConfigOrDie() *rest.Config {
+	var (
+		serverURL = flag.String("server", "",
+			"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+		kubeconfig = flag.String("kubeconfig", "",
+			"Path to a kubeconfig. Only required if out-of-cluster.")
+	)
+	klog.InitFlags(flag.CommandLine)
+	flag.Parse()
+
+	cfg, err := GetRestConfig(*serverURL, *kubeconfig)
+	if err != nil {
+		log.Fatalf("Error building kubeconfig: %v", err)
+	}
+
+	return cfg
+}
+
+// GetConfig returns a rest.Config to be used for kubernetes client creation.
+// It does so in the following order:
+//   1. Use the passed kubeconfig/serverURL.
+//   2. Fallback to the KUBECONFIG environment variable.
+//   3. Fallback to in-cluster config.
+//   4. Fallback to the ~/.kube/config.
+func GetRestConfig(serverURL, kubeconfig string) (*rest.Config, error) {
+	if kubeconfig == "" {
+		kubeconfig = os.Getenv("KUBECONFIG")
+	}
+
+	// We produce configs a bunch of ways, this gives us a single place
+	// to "decorate" them with common useful things (e.g. for debugging)
+	decorate := func(cfg *rest.Config) *rest.Config {
+		return cfg
+	}
+
+	// If we have an explicit indication of where the kubernetes config lives, read that.
+	if kubeconfig != "" {
+		c, err := clientcmd.BuildConfigFromFlags(serverURL, kubeconfig)
+		if err != nil {
+			return nil, err
+		}
+		return decorate(c), nil
+	}
+	// If not, try the in-cluster config.
+	if c, err := rest.InClusterConfig(); err == nil {
+		return decorate(c), nil
+	}
+	// If no in-cluster config, try the default location in the user's home directory.
+	if usr, err := user.Current(); err == nil {
+		if c, err := clientcmd.BuildConfigFromFlags("", filepath.Join(usr.HomeDir, ".kube", "config")); err == nil {
+			return decorate(c), nil
+		}
+	}
+
+	return nil, errors.New("could not create a valid kubeconfig")
+}
--- a/injection/config.go
+++ b/injection/config.go
@@ -29,9 +29,9 @@ import (
 	"k8s.io/klog"
 )
 
-// ParseAndGetKubeconfigOrDie parses the rest config flags and creates a client or
+// ParseAndGetRESTConfigOrDie parses the rest config flags and creates a client or
 // dies by calling log.Fatalf.
-func ParseAndGetKubeconfigOrDie() *rest.Config {
+func ParseAndGetRESTConfigOrDie() *rest.Config {
 	var (
 		serverURL = flag.String("server", "",
 			"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
@@ -41,7 +41,7 @@ func ParseAndGetKubeconfigOrDie() *rest.Config {
 	klog.InitFlags(flag.CommandLine)
 	flag.Parse()
 
-	cfg, err := GetKubeconfig(*serverURL, *kubeconfig)
+	cfg, err := GetRESTConfig(*serverURL, *kubeconfig)
 	if err != nil {
 		log.Fatal("Error building kubeconfig: ", err)
 	}
@@ -49,13 +49,13 @@ func ParseAndGetKubeconfigOrDie() *rest.Config {
 	return cfg
 }
 
-// GetKubeconfig returns a rest.Config to be used for kubernetes client creation.
+// GetRESTConfig returns a rest.Config to be used for kubernetes client creation.
 // It does so in the following order:
 //   1. Use the passed kubeconfig/serverURL.
 //   2. Fallback to the KUBECONFIG environment variable.
 //   3. Fallback to in-cluster config.
 //   4. Fallback to the ~/.kube/config.
-func GetKubeconfig(serverURL, kubeconfig string) (*rest.Config, error) {
+func GetRESTConfig(serverURL, kubeconfig string) (*rest.Config, error) {
 	if kubeconfig == "" {
 		kubeconfig = os.Getenv("KUBECONFIG")
 	}
--- a/injection/config.go
+++ b/injection/config.go
@@ -29,9 +29,9 @@ import (
 	"k8s.io/klog"
 )
 
-// ParseAndGetRestConfigOrDie parses the rest config flags and creates a client or
+// ParseAndGetKubeconfigOrDie parses the rest config flags and creates a client or
 // dies by calling log.Fatalf.
-func ParseAndGetRestConfigOrDie() *rest.Config {
+func ParseAndGetKubeconfigOrDie() *rest.Config {
 	var (
 		serverURL = flag.String("server", "",
 			"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
@@ -41,7 +41,7 @@ func ParseAndGetRestConfigOrDie() *rest.Config {
 	klog.InitFlags(flag.CommandLine)
 	flag.Parse()
 
-	cfg, err := GetRestConfig(*serverURL, *kubeconfig)
+	cfg, err := GetKubeconfig(*serverURL, *kubeconfig)
 	if err != nil {
 		log.Fatal("Error building kubeconfig: ", err)
 	}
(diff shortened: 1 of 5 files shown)


injection/sharedmain/main.go
```diff
@@ -127,6 +127,32 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
...
+
+func getInformerStartChanTimeout(ctx context.Context) (<-chan struct{}, time.Duration) {
+	timeout := defaultStartInformerDelayTime
+	tuntyped := ctx.Value(informerStartTimeoutKey{})
+	if tuntyped != nil {
+		timeout = tuntyped.(time.Duration)
+	}
+
+	untyped := ctx.Value(informerStartChanKey{})
+	if untyped == nil {
```
vagababov, line 150: given return is `interface{}` this might not always work.
https://medium.com/@mangatmodi/go-check-nil-interface-the-right-way-d142776edef1

n3wscott: thanks

zhongduo:
Hi @n3wscott,

I think this PR needs some discussion. I understand you are trying to avoid changing the function signature to maintain compatibility, but I believe API compatibility also includes underlying logic too and this PR changes the logic considerably IMO. If I understand correctly, with this PR:
1. For users other than shared_main, the informer start with be delayed for 5 seconds.
2. For shared_main, it chooses to start the informer after 5 seconds or after controller setup, whichever comes first.

For 1, though I know the timeout is configurable, I do believe most users won't notice this change until they encounter a problem. I am afraid that we are making the issue harder to reproduce to most users and the magic 5 seconds can cause some unexpected problems. Eg, in some case if the controller setup takes about 5 seconds for whatever reason, they will experience some flakiness depending on the actual runtime > 5s or < 5s. So to me it is not much different from the original logic, which starts the informer immediately.

For 2 or any other users that are aware of this problem and decide to start it manually, they can easily just use existing code without this PR. If we want to keep API and at the same time avoid duplicate code, here is what I would suggest:
a) Break the existing `EnableInjectionOrDie` to two parts: `BeforeStartInformer` and `StartInfomer`
b) Rewrite `EnableInjectionOrDie` to `BeforeStartInformer + go StartInformer`, which is same logic as the original `EnableInjectionOrDie`.
c) For shared_main or any users that want to start manually, we can do: `BeforeStartInformer` + do sth + `StartInformer`

Thanks,
Jimmy

zhongduo:
I have updated #1757 with this idea, which will keep API compatibility, could you please see if that can be an alternative solution too?

injection/config.go
```diff
@@ -0,0 +1,88 @@
...
+*/
+
+package injection
+
+import (
+	"errors"
+	"flag"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
```
mattmoor, line 24: Format Go code:
```suggestion

```

injection/config.go
```diff
@@ -0,0 +1,88 @@
...
+	"log"
+	"os"
+	"os/user"
+	"path/filepath"
```
mattmoor, line 28: Format Go code:
```suggestion
	"path/filepath"

	"k8s.io/client-go/rest"
	"k8s.io/client-go/tools/clientcmd"
	"k8s.io/klog"
```

injection/context.go
```diff
@@ -18,7 +18,6 @@ package injection
 
 import (
 	"context"
```
mattmoor, line 20: Format Go code:
```suggestion
	"context"

```

injection/injection.go
```diff
@@ -0,0 +1,57 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package injection
+
+import (
+	"context"
```
mattmoor, line 20: Format Go code:
```suggestion
	"context"

```

leaderelection/chaosduck/main.go
```diff
@@ -23,6 +23,7 @@ import (
 	"context"
 	"errors"
 	"flag"
+	"knative.dev/pkg/injection"
```
mattmoor, line 26: Format Go code:
```suggestion

```

leaderelection/chaosduck/main.go
```diff
@@ -23,6 +23,7 @@ import (
...
 	"log"
 	"strings"
 	"time"
```
mattmoor, line 29: Format Go code:
```suggestion
	"time"

	"knative.dev/pkg/injection"
```

mattmoor (commented):
Produced via:

injection/config.go
mattmoor, line 24: Format Go code:
```suggestion

```

injection/config.go
mattmoor, line 28: Format Go code:
```suggestion
	"path/filepath"

	"k8s.io/client-go/rest"
	"k8s.io/client-go/tools/clientcmd"
	"k8s.io/klog"
```

injection/context.go
mattmoor, line 20: Format Go code:
```suggestion
	"context"

```

injection/injection.go
mattmoor, line 20: Format Go code:
```suggestion
	"context"

```

leaderelection/chaosduck/main.go
mattmoor, line 26: Format Go code:
```suggestion

```

leaderelection/chaosduck/main.go
mattmoor, line 29: Format Go code:
```suggestion
	"time"

	"knative.dev/pkg/injection"
```

mattmoor (commented):
Produced via:

n3wscott:
> Hi @n3wscott,
> 
> I think this PR needs some discussion. I understand you are trying to avoid changing the function signature to maintain compatibility, but I believe API compatibility also includes underlying logic too and this PR changes the logic considerably IMO. If I understand correctly, with this PR:
> 
> 1. For users other than shared_main, the informer start with be delayed for 5 seconds.
> 2. For shared_main, it chooses to start the informer after 5 seconds or after controller setup, whichever comes first.
> 
> For 1, though I know the timeout is configurable, I do believe most users won't notice this change until they encounter a problem. I am afraid that we are making the issue harder to reproduce to most users and the magic 5 seconds can cause some unexpected problems. Eg, in some case if the controller setup takes about 5 seconds for whatever reason, they will experience some flakiness depending on the actual runtime > 5s or < 5s. Another possible problem is that some logic might be assuming informer start at that time, and this delay start will be quite difficult to debug. So to me it is not much different from the original logic, which starts the informer immediately.
> 
> For 2 or any other users that are aware of this problem and decide to start it manually, they can easily just use existing code without this PR. If we want to keep API and at the same time avoid duplicate code, here is what I would suggest:
> a) Break the existing `EnableInjectionOrDie` to two parts: `BeforeStartInformer` and `StartInfomer`
> b) Rewrite `EnableInjectionOrDie` to `BeforeStartInformer + go StartInformer`, which is same logic as the original `EnableInjectionOrDie`.
> c) For shared_main or any users that want to start manually, we can do: `BeforeStartInformer` + do sth + `StartInformer`
> 
> Thanks,
> Jimmy

Thanks for the thoughtful response. I agree that the timeout was strange. I have moved the `EnableInjectionOrDie`, and changed the signature, it now returns back a callback the caller should call if it wants to start the informers. It is basically `BeforeStartInformer` + do sth + `StartInformer` except wrapped up in one call with a callback. I think the naming of `BeforeStartInformer` and `StartInformer` is strange and calling `StartInformer` alone will cause issues.

Using a callback means the caller can not have access to start informers out of order.

injection/sharedmain/main.go
```diff
@@ -148,21 +150,19 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) context.Context
 
 	ctx, informers := injection.Default.SetupInformers(ctx, cfg)
 
-	// Start the injection clients and informers.
-	logging.FromContext(ctx).Info("Starting informers...")
-	go func(ctx context.Context) {
+	go func() {
```
zhongduo, line 153: Do we really need this go function? I think it is safer to let it finishes.

injection/injection.go
```diff
@@ -0,0 +1,58 @@
...
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/signals"
+
+	"k8s.io/client-go/rest"
+)
+
+// EnableInjectionOrDie enables Knative Injection and starts the informers.
+// Both Context and Config are optional. Returns context with rest config set
+// and a function to start the informers after watches have been set up.
+func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Context, func()) {
```
zhongduo, line 33: This makes sense to me, though the name is a little bit confusing as the original `EnableInjectionOrDie` is exported too.

n3wscott:
> I have updated #1757 with this idea, which will keep API compatibility, could you please see if that can be an alternative solution too?

How do you like this way?

leaderelection/chaosduck/main.go
```diff
@@ -116,8 +117,7 @@ func quack(ctx context.Context, kc kubernetes.Interface, component string, leade
 }
 
 func main() {
-	ctx := signals.NewContext()
-	ctx = sharedmain.EnableInjectionOrDie(ctx, nil)
+	ctx, _ := injection.EnableInjectionOrDie(signals.NewContext(), nil)
```
zhongduo, line 120: If we have to do this, that means we are really trying to avoid calling the original function. I would recommend to have a roadmap to deprecate the original function.

n3wscott: It is marked deprecated and documented redirected to the new one version.

injection/injection.go
n3wscott: yeah, that is intended, the original one is in a different package and marked deprecated.

injection/sharedmain/main.go
n3wscott: This will be deleted asap after knative rolls to the new one, shared main now calls down to the new version. this method is not changed and is legacy.

leaderelection/chaosduck/main.go
zhongduo: That's awesome. Is this on purpose not to call the start function here? Note that the original logic will have informers started here.

injection/injection.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"go.uber.org/zap"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/signals"
+
+	"k8s.io/client-go/rest"
+)
+
+// EnableInjectionOrDie enables Knative Injection and starts the informers.
```
zhongduo, line 30: Remove `starts the informers`.

injection/sharedmain/main.go
zhongduo: My original question is about using go func or not in the old function, but I guess the whole function will be removed anyway.

zhongduo:
> > Hi @n3wscott,
> > I think this PR needs some discussion. I understand you are trying to avoid changing the function signature to maintain compatibility, but I believe API compatibility also includes underlying logic too and this PR changes the logic considerably IMO. If I understand correctly, with this PR:
> > 
> > 1. For users other than shared_main, the informer start with be delayed for 5 seconds.
> > 2. For shared_main, it chooses to start the informer after 5 seconds or after controller setup, whichever comes first.
> > 
> > For 1, though I know the timeout is configurable, I do believe most users won't notice this change until they encounter a problem. I am afraid that we are making the issue harder to reproduce to most users and the magic 5 seconds can cause some unexpected problems. Eg, in some case if the controller setup takes about 5 seconds for whatever reason, they will experience some flakiness depending on the actual runtime > 5s or < 5s. Another possible problem is that some logic might be assuming informer start at that time, and this delay start will be quite difficult to debug. So to me it is not much different from the original logic, which starts the informer immediately.
> > For 2 or any other users that are aware of this problem and decide to start it manually, they can easily just use existing code without this PR. If we want to keep API and at the same time avoid duplicate code, here is what I would suggest:
> > a) Break the existing `EnableInjectionOrDie` to two parts: `BeforeStartInformer` and `StartInfomer`
> > b) Rewrite `EnableInjectionOrDie` to `BeforeStartInformer + go StartInformer`, which is same logic as the original `EnableInjectionOrDie`.
> > c) For shared_main or any users that want to start manually, we can do: `BeforeStartInformer` + do sth + `StartInformer`
> > Thanks,
> > Jimmy
> 
> Thanks for the thoughtful response. I agree that the timeout was strange. I have moved the `EnableInjectionOrDie`, and changed the signature, it now returns back a callback the caller should call if it wants to start the informers. It is basically `BeforeStartInformer` + do sth + `StartInformer` except wrapped up in one call with a callback. I think the naming of `BeforeStartInformer` and `StartInformer` is strange and calling `StartInformer` alone will cause issues.
> 
> Using a callback means the caller can not have access to start informers out of order.

The naming is just for illustration purpose only. :)

injection/injection.go
n3wscott: Updated.

leaderelection/chaosduck/main.go
zhongduo: I think this might skip your attention. Otherwise LGTM.

n3wscott: I checked and the duck does not use informers so it is not needed.

n3wscott:
flakes: https://github.com/knative/pkg/issues/1779

n3wscott:
g2g?

/cc @mattmoor

leaderelection/chaosduck/main.go
zhongduo: Thanks for checking.

injection/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+)
+
+// ParseAndGetRestConfigOrDie parses the rest config flags and creates a client or
+// dies by calling log.Fatalf.
+func ParseAndGetRestConfigOrDie() *rest.Config {
```
mattmoor, line 34: nit: REST

injection/config.go
```diff
@@ -0,0 +1,89 @@
...
+	return cfg
+}
+
+// GetConfig returns a rest.Config to be used for kubernetes client creation.
+// It does so in the following order:
+//   1. Use the passed kubeconfig/serverURL.
+//   2. Fallback to the KUBECONFIG environment variable.
+//   3. Fallback to in-cluster config.
+//   4. Fallback to the ~/.kube/config.
+func GetRestConfig(serverURL, kubeconfig string) (*rest.Config, error) {
```
mattmoor, line 58: nit: REST

injection/config.go
```diff
@@ -0,0 +1,89 @@
...
+	if kubeconfig == "" {
+		kubeconfig = os.Getenv("KUBECONFIG")
+	}
+
+	// We produce configs a bunch of ways, this gives us a single place
+	// to "decorate" them with common useful things (e.g. for debugging)
+	decorate := func(cfg *rest.Config) *rest.Config {
+		return cfg
+	}
```
mattmoor, line 67: I don't understand what this is for...

n3wscott: looks like OG debug. Deleted.

injection/config.go
n3wscott: ick, how about GetKubeconfig

mattmoor: it is getting a `rest.Config`...

injection/sharedmain/main.go
```diff
@@ -129,12 +130,13 @@ func GetLeaderElectionConfig(ctx context.Context) (*leaderelection.Config, error
 
 // EnableInjectionOrDie enables Knative Injection and starts the informers.
 // Both Context and Config are optional.
+// Deprecated: use injection.EnableInjectionOrDie
```
mattmoor, line 133: Should we change this method to:
```
   ctx, si := injection.EnableInjectionOrDie(ctx, cfg)
   go si()
   return ctx
```

injection/sharedmain/main.go
```diff
@@ -63,6 +63,7 @@ import (
 //   2. Fallback to the KUBECONFIG environment variable.
 //   3. Fallback to in-cluster config.
 //   4. Fallback to the ~/.kube/config.
+// Deprecated: use injection.GetRESTConfig
```
mattmoor, line 66: Should this call the other method?

n3wscott: good idea.

injection/sharedmain/main.go
n3wscott: yes, good idea!

mattmoor (commented):
/lgtm

PR #1774, opened 2020-10-05
move binding duck to v1

capri-xiyue:
-  Creating the v1 Binding duck (equal to v1beta1)


capri-xiyue:
The codecov/project fails because there is no UT for biding_types.go. But I checked the code, it only defines some struct and implement some interface. There is no need to write UT for it. And we don't have UT for biding_types.go for both v1beta1 and v1alpha1.

capri-xiyue:
assign @n3wscott

n3wscott (commented):
/lgtm
/approve

n3wscott:
/cc @mattmoor 

Matt to approve

apis/duck/v1/biding_types.go
```diff
@@ -0,0 +1,92 @@
+/*
```
mattmoor, line 1: The file name is missing an `n`

capri-xiyue: done

mattmoor (approved):
/lgtm

PR #1775, opened 2020-10-06
Enable e2e tests to be able to use e2e client injection.

n3wscott:
This is a breaking api change to the testing api. The old testing api defined the `--kubeconfig` flag. Disappoint. This does not mix with also using client injection.

- Semi-breaking API change: cleaned up all the usage of the `*kubernetes.ClientSet`, and converted them to `kubernetes.Interface` as they should have been.
- ~Breaking API change: `logstream.Start` now needs a client injection enabled context to pull the kubeclient out.~
- ~Breaking API change: Removed the global `pkg/test/Flags` var and replaced with an exported `Flags()` method that returns the same thing, it is guarded in a once. There are cases where this will need to be called to let `flags.Parse()` work as expected in test main methods. This is a better model: opt-in vs always on.~

Working on serving upgrade here: https://github.com/knative/serving/pull/9699


test/injection.go
```diff
@@ -0,0 +1,44 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"context"
```
mattmoor, line 20: Format Go code:
```suggestion
	"context"
	"sync"

```

test/injection.go
```diff
@@ -0,0 +1,44 @@
...
+	"knative.dev/pkg/injection"
+	"sync"
```
mattmoor, line 22: Format Go code:
```suggestion

```

mattmoor (commented):
Produced via:

n3wscott:
/hold until major things are staged.

Knative:

[] Serving - knative/serving#9699
[] Eventing
[] eventing-operator
[] serving-operator
[] networking

Knative-sandbox:
[] eventing-rabbitmq
[] discovery
[] net-istio
[] net-certmanager
[] net-kourier 
[] net-http01
[] net-contour

n3wscott:
/unhold

I am ready with downstream PRs of the complex migrations. The minor migrations (net-*, eventing-*) can done one at a time and are quick to update, save me time: merge this now and I can update the downstreams all at once in a single pr without the staged PR 🙏 .

test/clients.go
```diff
@@ -35,14 +35,14 @@ import (
 
 // KubeClient holds instances of interfaces for making requests to kubernetes client.
 type KubeClient struct {
-	Kube *kubernetes.Clientset
+	Kube kubernetes.Interface
```
vagababov, line 38: I wonder if it'll be better to anonymous aggregate it.

test/logstream/null.go
```diff
@@ -16,7 +16,9 @@ limitations under the License.
 
 package logstream
 
-import "knative.dev/pkg/test"
+import (
```
vagababov, line 19: why this change?

n3wscott: I think the linter did it :idk:

test/clients.go
n3wscott: tbh I think we should delete this resource 100%

vagababov: that works too.

n3wscott:
Ok I made some moves to make this a mostly non-breaking change. Now the only thing that breaks the api is the move away from k.ClientSet vs k.Interface.

PR #1777, opened 2020-10-06
configmap parse int

runzexia:
Signed-off-by: RyderXia <ryder.xia@sap.com>
to replace https://github.com/knative/eventing/blob/master/pkg/configmap/reader.go#L36

### DIFF
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -81,7 +81,7 @@ func AsInt64(key string, target *int64) ParseFunc {
 	}
 }
 
-// AsInt parses the value at key as an uint32 into the target. if it exists.
+// AsInt parses the value at key as an int into the target, if it exists.
 func AsInt(key string, target *int) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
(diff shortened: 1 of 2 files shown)


vagababov:
it doesn't look 1 to 1 with the function you say to replace?

runzexia:
> it doesn't look 1 to 1 with the function you say to replace?

`ReadInt` is used to read confingmap at https://github.com/knative/eventing/blob/master/pkg/channel/config.go#L54, we can use configmap parse instead, and remove `ReadInt`

vagababov:
I am not against the method itself, but the ReadInt has a list and stuff...

runzexia:
I would create PR in event to discuss readint. My idea is to use utils in 'pkg'  to parse configmap. 😊

configmap/parse.go
```diff
@@ -81,6 +81,20 @@ func AsInt64(key string, target *int64) ParseFunc {
 	}
 }
 
+// AsInt parses the value at key as an uint32 into the target. if it exists.
```
markusthoemmes, line 84:
```suggestion
// AsInt parses the value at key as an into into the target, if it exists.
```

markusthoemmes (approved):
/lgtm
/approve
/ok-to-test

PR #1780, opened 2020-10-06
Include REST and reorder items to be sorted

vagababov:
We're still doing a linear search... but one day may be.

/assign @tcnghia @chizhg mattmoor


n3wscott (commented):
/lgtm
/approve

PR #1790, opened 2020-10-09
BucketList cleanup

yanweiguo:
The `BucketList` func is only used in tests. No need to return a random order. Make the return sorted to simplify tests.

/assign @vagababov


vagababov (commented):
/lgtm
/approve

PR #1791, opened 2020-10-09
Drop the usage of knative.dev/pkg/test in the helpers package

dprotaso:
This has the side effects of pulling in a bunch of global variables


pierDipi (approved):
/lgtm

PR #1792, opened 2020-10-09
leverage the go_update_deps method from test-infra

n3wscott:
Steps to get to this point:

1. run `go get -d knative.dev/test-infra`
1. run `go mod vendor`
1. edit `hack/update-deps.sh`, drop all floating deps that are knative.dev, and see changelist.
1. run `./hack/update-deps.sh --upgrade --release v0.19`


coryrc:
I think you'll need to update test-infra first.

It would be nice to use the method client uses: https://github.com/knative/client/blob/4ebd7ed6042c1344e724e0c58b7460c54af91865/hack/update-deps.sh#L51 (instructions [here](https://github.com/knative/test-infra/blob/master/scripts/update-test-infra.sh#L28)

n3wscott:
> I think you'll need to update test-infra first.
> 
> It would be nice to use the method client uses: https://github.com/knative/client/blob/4ebd7ed6042c1344e724e0c58b7460c54af91865/hack/update-deps.sh#L51 (instructions [here](https://github.com/knative/test-infra/blob/master/scripts/update-test-infra.sh#L28)

ah, this line is why my PR failed. Thanks for pointing that out. I am going to drop that.

n3wscott:
Related to client: https://github.com/knative/test-infra/issues/2493

mattmoor (approved):
/lgtm

PR #1794, opened 2020-10-10
Final sync/atomic removals

vagababov:
There's one sync/atomic.Value, but value already is a wholesome
type, so no need to change that one.

I guess this one is test, so I ignored before :/

/assign @tcnghia mattmoor


n3wscott (commented):
/lgtm
/approve

PR #1796, opened 2020-10-10
Clean up leader election code for StatefulSet

yanweiguo:
We no longer plan to use StatefulSet.

/assign @mattmoor


mattmoor:
I still think that this is an interesting alternative way to run our core controllers if folks want even bucket distribution and no leaderelection.  I've been using this to do exactly that downstream in mink, and I still think it's a reasonable use case in general.

mattmoor:
/hold

for discussion

yanweiguo:
> I still think that this is an interesting alternative way to run our core controllers if folks want even bucket distribution and no leaderelection. I've been using this to do exactly that downstream in mink, and I still think it's a reasonable use case in general.

Sure. If it is or could be used somewhere (downstream), let's keep it.

PR #1800, opened 2020-10-12
Shush autogomaxproc.

vagababov:
I am not as sure about this change, but those logs are not following the standard formt
and given we need to inject the logger in init() makes this next to impossible.

Corollary, is that we set cpu limits on all our binaries, except QP, so it only mattered there
since otherwise the package logs nothing.

I am 50/50 on this.

/assign @julz mattmoor

For https://github.com/knative/serving/issues/9787


injection/sharedmain/main.go
```diff
@@ -50,6 +51,10 @@ import (
 	"knative.dev/pkg/webhook"
 )
 
+func init() {
+	maxprocs.Set(maxprocs.Logger(func(string, ...interface{}) {}))
+}
```
mattmoor, line 56: Couldn't we just have done this in QP?

vagababov:
The QP was pulling in more stuff, so the solution I actually did is preferable (we shaved of probably some of the binary size, etc). But, yes, we could've done this as well.

I am not bent on this change, though.

mattmoor:
/lgtm

🤷

PR #1804, opened 2020-10-13
drop the default log url template

dprotaso:
this had references to the monitoring bundle which is being removed here: https://github.com/knative/serving/pull/9807

### DIFF
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -27,7 +27,8 @@ import (
 
 const (
 	// The following is used to set the default log url template
-	DefaultLogURLTemplate = "http://localhost:8001/api/v1/namespaces/knative-monitoring/services/kibana-logging/proxy/app/kibana#/discover?_a=(query:(match:(kubernetes.labels.knative-dev%2FrevisionUID:(query:'${REVISION_UID}',type:phrase))))"
+	DefaultLogURLTemplate = ""
+
 	// DefaultRequestLogTemplate is the default format for emitting request logs.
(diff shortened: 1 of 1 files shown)


dprotaso:
interesting logstream flaked 

```go
--- FAIL: TestNamespaceStream (1.00s)
    stream_test.go:87: timed out: log message wasn't received
```

vagababov:
This will break stuff.
We use it in the serving.

dprotaso:
https://github.com/knative/serving/pull/9807

vagababov:
I think the default _is not_ an empty string anymore. Those were the changes we did recently.
Since we added a new boolean toggle so default _always_ has a value now.

dprotaso:
> I think the default _is not_ an empty string anymore. Those were the changes we did recently.
> Since we added a new boolean toggle so default _always_ has a value now.

I'm confused - the logging template is a string value

vagababov:
Yes it always has a value now.

Initial implementation was:
```
if template == "" { no logging }
else {use the template}
```

And the default value was empty.

Current is
```
if !enable_logging {no logging}
else {use the template}
```

And template is *always* non empty now (since the toggle is false by default).
Test here: https://github.com/knative/serving/blob/master/pkg/metrics/config_test.go verify that.
Also it has to match the example set in config/core/configmaps/observability.yaml.

dprotaso:
it looks like the change your made was for the request log template

I'm changing the `DefaultLogURLTemplate`

vagababov:
Too many templates! :-)

I think it still has to be non-empty so that we can have a non-empty example.

dprotaso:
> I think it still has to be non-empty so that we can have a non-empty example.

I'm going to drop that bit of logic in serving

vagababov:
Which part? Checking in UT that the example value is not empty?

vagababov:
Not really. We actually had an issue about that. 
Why do you want to drop the default?

dprotaso:
> Why do you want to drop the default?

With the monitoring bundle gone an empty default make sense - but we _would_ want an example that highlights the revision UID substitution

dprotaso:
downstream serving fix is here: https://github.com/knative/serving/pull/9807

vagababov:
The problem with example not matching the default, is that just copy-pasting example as actual config is going to break stuff. Which might or might not be expected :)

Anyway, I am not going to fight this a lot :)

/cc @mattmoor 
For more insights :)

mattmoor (commented):
/lgtm

PR #1807, opened 2020-10-14
Port tracing setup code to pkg

slinkydeveloper:
Related to https://github.com/knative/eventing/issues/4315

### DIFF
--- a/tracing/config/tracing.go
+++ b/tracing/config/tracing.go
@@ -40,6 +40,28 @@ const (
 	stackdriverProjectIDKey = "stackdriver-project-id"
 )
 
+var (
+	// OnePercentSampling is a configuration that samples 1% of the requests.
+	// TODO(#1712): Remove this and pull "static" configuration from the
+	// environment instead.
+	OnePercentSampling = &Config{
+		Backend:        Zipkin,
+		Debug:          false,
+		SampleRate:     0.01,
+		ZipkinEndpoint: "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans",
+	}
+
+	// AlwaysSample is a configuration that samples 100% of the requests and sends them to Zipkin.
+	// It is expected to be used only for testing purposes (e.g. in e2e tests).
+	// TODO(#1712): Remove this and pull "static" configuration from the environment instead.
+	AlwaysSample = &Config{
+		Backend:        Zipkin,
+		Debug:          true,
+		SampleRate:     1.0,
+		ZipkinEndpoint: "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans",
+	}
+)
+
 // BackendType specifies the backend to use for tracing
 type BackendType string
 
--- a/tracing/config/tracing.go
+++ b/tracing/config/tracing.go
@@ -40,28 +40,6 @@ const (
 	stackdriverProjectIDKey = "stackdriver-project-id"
 )
 
-var (
-	// OnePercentSampling is a configuration that samples 1% of the requests.
-	// TODO(#1712): Remove this and pull "static" configuration from the
-	// environment instead.
-	OnePercentSampling = &Config{
-		Backend:        Zipkin,
-		Debug:          false,
-		SampleRate:     0.01,
-		ZipkinEndpoint: "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans",
-	}
-
-	// AlwaysSample is a configuration that samples 100% of the requests and sends them to Zipkin.
-	// It is expected to be used only for testing purposes (e.g. in e2e tests).
-	// TODO(#1712): Remove this and pull "static" configuration from the environment instead.
-	AlwaysSample = &Config{
-		Backend:        Zipkin,
-		Debug:          true,
-		SampleRate:     1.0,
-		ZipkinEndpoint: "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans",
-	}
-)
-
 // BackendType specifies the backend to use for tracing
 type BackendType string
 
@@ -89,7 +67,8 @@ func (cfg *Config) Equals(other *Config) bool {
 	return reflect.DeepEqual(other, cfg)
 }
 
-func defaultConfig() *Config {
+// NoopConfig returns a new noop default config
+func NoopConfig() *Config {
 	return &Config{
 		Backend:    None,
 		Debug:      false,
@@ -99,7 +78,7 @@ func defaultConfig() *Config {
 
 // NewTracingConfigFromMap returns a Config given a map corresponding to a ConfigMap
 func NewTracingConfigFromMap(cfgMap map[string]string) (*Config, error) {
-	tc := defaultConfig()
+	tc := NoopConfig()
 
 	if backend, ok := cfgMap[backendKey]; ok {
 		switch bt := BackendType(backend); bt {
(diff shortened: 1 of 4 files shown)


slinkydeveloper:
/hold 
Preparing a pr in eventing to test it

slinkydeveloper:
If the integration tests pass here https://github.com/knative/eventing/pull/4316 then we can merge this one

tracing/setup.go
```diff
@@ -0,0 +1,97 @@
...
+			err := oct.ApplyConfig(cfg)
+			if err != nil {
+				logger.Errorw("Unable to apply open census tracer config", zap.Error(err))
+				return
+			}
+		}
+	}
+
+	// Set up our config store.
+	configStore := configmap.NewDefaultUntypedStore(
```
markusthoemmes, line 72: Is this required? Can't you use the "usual" store and wrap the constructor function like so:

```golang
func enableZeroSamplingCM(ns string, tracingConfigName string) func(*corev1.ConfigMap) (*Config, error) {
  return func(*corev1.ConfigMap) (*Config, error) {
    cm := &corev1.ConfigMap{
		ObjectMeta: metav1.ObjectMeta{
			Name:      tracingConfigName,
			Namespace: ns,
		},
		Data: map[string]string{
			"backend":         "zipkin",
			"debug":           "False",
			"sample-rate":     "0",
			"zipkin-endpoint": "http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans",
		},
	}
  return config.NewTracingConfigFromConfigMap(cm)
  }
}

Comment (slinkydeveloper) [reply tracing/setup.go:72]: This is just code porting, should we modify the actual code in a separate pr?

Comment (markusthoemmes) [reply tracing/setup.go:72]: well I'm questioning if you need to add the helpers at all

Comment (slinkydeveloper) [reply tracing/setup.go:72]: I think we don't even need this zero sampling stuff

Comment (slinkydeveloper) [reply tracing/setup.go:72]: let me try

Comment (slinkydeveloper) [reply tracing/setup.go]: Ok I tried to remove it, now let me sync in eventing and let's see if e2e pass

File: tracing/config/tracing.go [hunk 2]
```diff
@@ -40,6 +40,28 @@ const (
 	stackdriverProjectIDKey = "stackdriver-project-id"
 )
 
+var (
+	// OnePercentSampling is a configuration that samples 1% of the requests.
+	// TODO(#1712): Remove this and pull "static" configuration from the
```
markusthoemmes, line 45: Where does this TODO point to? :thinking:. The configs certainly look like they're not "generic", should they really live here?

slinkydeveloper: These are mostly useless now, lemme remove these

tracing/config/tracing.go
```diff
@@ -67,7 +67,8 @@ func (cfg *Config) Equals(other *Config) bool {
 	return reflect.DeepEqual(other, cfg)
 }
 
-func defaultConfig() *Config {
+// NoopConfig returns a new noop default config
```
vagababov, line 70:
```suggestion
// NoopConfig returns a new noop config
```
Or `default config`, but I think both are a bit too much :)

tracing/config/tracing.go
```diff
@@ -127,6 +128,9 @@ func NewTracingConfigFromMap(cfgMap map[string]string) (*Config, error) {
 
 // NewTracingConfigFromConfigMap returns a Config for the given configmap
 func NewTracingConfigFromConfigMap(config *corev1.ConfigMap) (*Config, error) {
+	if config == nil {
+		return NewTracingConfigFromMap(map[string]string{})
```
vagababov, line 132:
```suggestion
		return NewTracingConfigFromMap(nil)
```

tracing/setup.go
```diff
@@ -0,0 +1,79 @@
...
+	"knative.dev/pkg/tracing/config"
+)
+
+// setupPublishing sets up trace publishing for the process. Note that other pieces
+// still need to generate the traces, this just ensures that if generated, they are collected
+// appropriately. This is normally done by using tracing.HTTPSpanMiddleware as a middleware HTTP
+// handler.
+func setupPublishing(serviceName string, logger *zap.SugaredLogger) *OpenCensusTracer {
+	return NewOpenCensusTracer(WithExporter(serviceName, logger))
+}
```
vagababov, line 34: Given this is an internal method consisting of a single line — why do we need it?

tracing/setup.go
```diff
@@ -0,0 +1,79 @@
...
+
+// SetupStaticPublishing sets up trace publishing for the process. Note that other
+// pieces still need to generate the traces, this just ensures that if generated, they are collected
+// appropriately. This is normally done by using tracing.HTTPSpanMiddleware as a middleware HTTP
+// handler. The configuration will not be dynamically updated.
+func SetupStaticPublishing(logger *zap.SugaredLogger, serviceName string, cfg *config.Config) error {
+	oct := setupPublishing(serviceName, logger)
+	err := oct.ApplyConfig(cfg)
+	if err != nil {
+		return fmt.Errorf("unable to set OpenCensusTracing config: %v", err)
```
vagababov, line 44:
```suggestion
		return fmt.Errorf("unable to set OpenCensusTracing config: %w", err)
```

tracing/setup.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// SetupStaticPublishing sets up trace publishing for the process. Note that other
+// pieces still need to generate the traces, this just ensures that if generated, they are collected
+// appropriately. This is normally done by using tracing.HTTPSpanMiddleware as a middleware HTTP
+// handler. The configuration will not be dynamically updated.
+func SetupStaticPublishing(logger *zap.SugaredLogger, serviceName string, cfg *config.Config) error {
+	oct := setupPublishing(serviceName, logger)
+	err := oct.ApplyConfig(cfg)
+	if err != nil {
```
vagababov, line 43:
```suggestion
	if err := oct.ApplyConfig(cfg); err != nil {
```

tracing/setup.go
```diff
@@ -0,0 +1,79 @@
...
+// updated when the ConfigMap is updated.
+func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher *configmap.InformedWatcher, serviceName, tracingConfigName string) error {
+	oct := setupPublishing(serviceName, logger)
+
+	tracerUpdater := func(name string, value interface{}) {
+		if name == tracingConfigName {
+			cfg := value.(*config.Config)
+			logger.Debugw("Updating tracing config", zap.Any("cfg", cfg))
+			err := oct.ApplyConfig(cfg)
+			if err != nil {
```
vagababov, line 62:
```suggestion
			if err := oct.ApplyConfig(cfg); err != nil {
```

tracing/setup.go
```diff
@@ -0,0 +1,79 @@
...
+				logger.Errorw("Unable to apply open census tracer config", zap.Error(err))
+				return
```
vagababov, line 64:
```suggestion
```

tracing/config/tracing.go
slinkydeveloper: Not sure this works

markusthoemmes: It should since there are only read operations on the map, which work for `nil` maps.

slinkydeveloper: I think that function needs a default value in there

slinkydeveloper: I wasn't aware of that, thanks for the tip!

vagababov: This 100% works :)

tracing/setup.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return nil
+}
+
+// SetupDynamicPublishing sets up trace publishing for the process, by watching a
+// ConfigMap for the configuration. Note that other pieces still need to generate the traces, this
+// just ensures that if generated, they are collected appropriately. This is normally done by using
+// tracing.HTTPSpanMiddleware as a middleware HTTP handler. The configuration will be dynamically
+// updated when the ConfigMap is updated.
+func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher *configmap.InformedWatcher, serviceName, tracingConfigName string) error {
```
vagababov, line 45: Final nit: this function never returns an error inside.Any reason it's in the interface?

slinkydeveloper: TBH I don't know :man_shrugging: , maybe it was returning an error in the past... Should I change the signature or we keep as is (maybe useful for forward compatibility, given at some point we'll switch to the new tracing lib)

vagababov: :shrug:
we can always change that.
I'd rather not have to check that we return an error when we shouldn't.
But it's fine. I can clean this up later, since it probably requires downstream changes.

vagababov (commented):
/lgtm
/approve

PR #1809, opened 2020-10-14
test all downstreams.

n3wscott:
Drops eventing-contrib, adds everything else. 😎

### DIFF
--- a/.github/workflows/knative-downstream.yaml
+++ b/.github/workflows/knative-downstream.yaml
@@ -25,12 +25,53 @@ jobs:
     strategy:
       fail-fast: false # Keep running if one leg fails.
       matrix:
-        downstream: # only for github.com/knative
-          - serving
-          - networking
-          - eventing
-          - operator
-          - client
+        include:
+          - repo: serving
+            org: knative
+          - repo: networking
+            org: knative
+          - repo: eventing
+            org: knative
+          - repo: operator
+            org: knative
+          - repo: client
+            org: knative
+          - repo: discovery
+            org: knative-sandbox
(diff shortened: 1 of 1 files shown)


.github/workflows/knative-downstream.yaml
```diff
@@ -20,211 +20,90 @@ on:
...
-        platform: [ubuntu-latest]
-
-    runs-on: ${{ matrix.platform }}
-
+        downstream: # only for github.com/knative
+          - serving
+          - networking
+          - eventing
+          - operator
+          - client
```
mattmoor, line 33:
```suggestion
        include:
          - downstream: serving
             org: knative
          - downstream: networking
             org: knative
          - downstream: eventing
             org: knative
          - downstream: operator
             org: knative
          - downstream: client
             org: knative
```

n3wscott: rad.

mattmoor:
Can we wait until they're passing...? 😉

n3wscott:
> Can we wait until they're passing...? 😉

they might not... that is kinda the point

mattmoor:
The set of things that run presubmit here should be expected to stay green the vast majority of the time.

If things aren't green, we should remove them here.  I don't think 11 failing legs is where we expect the steady state to be, so let's get what passes passing and comment out the rest.

n3wscott:
@mattmoor It looks like all the current failures are related to those deps not being in sync with networking or eventing, which will get resolved at 6pm

mattmoor:
@n3wscott should I run it now?  it's a click away 😇

PR #1811, opened 2020-10-15
Update vegeta to a version supporting go modules

markusthoemmes:
This allows us to drop the explicit overrides in downstream repositories and it also looks so much nicer :).

I'd hope the change is non-breaking to downstream repositories, given the interface is strictly widened here.

/assign @julz @vagababov


markusthoemmes:
Okay, I guess I'll have to stage downstream PRs :cry:

PR #1812, opened 2020-10-15
Add suggested fix when checksums dont match

julz:
This confused someone in https://github.com/knative/serving/pull/9778#issuecomment-709285563, and I can't think of any obvious downside to including a suggested fix since the person hitting this may often not know why.

/assign @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #1815, opened 2020-10-19
Simplify test helpers for prometheus API

markusthoemmes:
As per title, it's not necessary to specify all the functions that are never called anyway.

/assign @vagababov @julz


vagababov (commented):
/lgtm

PR #1816, opened 2020-10-19
Remove unnecessary override for dependencies

markusthoemmes:
It's not clear why we'd have to override these, so let's try not to override them.

/assign @vagababov


vagababov (commented):
/лгтм

markusthoemmes:
/retest

Unrelated.

PR #1819, opened 2020-10-19
Wait for prometheus server to start serving before fetching metrics

markusthoemmes:
This should avoid failures like in https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/1818/pull-knative-pkg-unit-tests/1318114477454921728, where the test failed because of:

```
    resource_view_test.go:371: failed to fetch prometheus metrics: Get "http://localhost:19090/metrics": dial tcp 127.0.0.1:19090: connect: connection refused
```

/assign @julz @vagababov

### DIFF
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -38,6 +38,7 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
 	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	emptypb "github.com/golang/protobuf/ptypes/empty"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


metrics/resource_view_test.go
```diff
@@ -362,7 +363,14 @@ func TestMetricsExport(t *testing.T) {
 	}{{
 		name: "Prometheus",
 		init: func() error {
-			return UpdateExporter(context.Background(), configForBackend(prometheus), logtesting.TestLogger(t))
+			if err := UpdateExporter(context.Background(), configForBackend(prometheus), logtesting.TestLogger(t)); err != nil {
```
julz, line 366: just to ask the maybe-obvious question.. why is this function returning before the server is listening? Seems just about possible that this is not only a test bug if this can return before it's safe to ask for metrics so double-checking? (I guess it's not a big deal if metrics aren't immediately available since stuff generally polls?)

markusthoemmes: Aside from stuff generally not really caring (as you said), I don't even know if there's a good way to make confidently sure it is serving without running health checks during the setup which seems excessive.

julz: I think generally you can split Listen and Serve if you want to be sure something is listening before returning. Saying that I totally buy that it probably isn't needed here, just instinctively feels a bit like the test may be be telling us the function is a bit fishy (clients need to poll before they can actually fetch metrics). I guess we can always fix the fishiness independently if we decide we care, and this solves the flake so lgtm.

julz (approved):
/lgtm

PR #1820, opened 2020-10-19
Remove re-definition of kubeconfig from pkg/test

n3wscott:


vagababov (commented):
/cc @coryrc 
if we can do something better here
but in general
/lgtm

coryrc:
`... Define your own interface.`

why?

n3wscott:
Drat, missed one in the null class

n3wscott:
> `... Define your own interface.`
> 
> why?

There is no need to define a central interface for the testing class, and doing so causes a circular dependency.

vagababov (commented):
/lgtm

Client side interface are standard in go (in fact they are recommended pattern)

PR #1823, opened 2020-10-20
Fix nolint warnings and adhere to best practices

markusthoemmes:
As per title.

This gets rid of

```
WARN [runner/nolint] Found unknown linters in //nolint directives: stylecheck for backcompat., the newer package has different interface. 
```

and also adheres to the best practice of not having a space in front of `nolint`, see https://golangci-lint.run/usage/false-positives/#nolint


n3wscott (commented):
/lgtm
/approve

PR #1824, opened 2020-10-20
Harden logstream test, make it quicker and repeatable

markusthoemmes:
This hardens the logstream test by not relying for the cancel signal to arrive before the pod creation signal. Golang's runtime doesn't guarantee such ordering.

It also makes the test significantly quicker by cutting down on the log timeouts and makes it repeatable by deep copying the global pod into it's visibility.

/assign @vagababov

### DIFF
--- a/test/logstream/v2/stream_test.go
+++ b/test/logstream/v2/stream_test.go
@@ -29,6 +29,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


test/logstream/v2/stream_test.go
```diff
@@ -145,15 +149,25 @@ func TestNamespaceStream(t *testing.T) {
...
+		case <-logFuncInvoked:
+			t.Log("Log was still produced, trying again...")
+			if err := podClient.Delete(context.Background(), pod.Name, metav1.DeleteOptions{}); err != nil {
+				return false, err
+			}
+			return false, nil
+		}
+	}); err != nil {
+		t.Fatal("No watching should have happened", err)
 	}
```
dprotaso, line 171: Dunno if this is a factor but the logFunc in the test is now blocking

```go
logFunc := func(format string, args ...interface{}) {
	logFuncInvoked <- struct{}{}
}
```

While it's blocked it blocks cancellation
https://github.com/knative/pkg/blob/eafc8c6f72a7aad7d1110e19f9d3d6cfbb7c0332/test/logstream/v2/stream.go#L198-L199

markusthoemmes: I don't think that'd be a factor for the test honestly. The test's race is a race of us cancelling the context vs. creating the pod. If both signals arrive at the same time (from the PoV of the `select` statement), one will be picked "randomly".

dprotaso: then there's a bug somewhere

`cancel()` is synchronous so `logFuncInvoked` should _never_ be invoked

markusthoemmes: Cancel is, but is producing the signal onto the channel synchronous too?

dprotaso: realized cancel was for the context and not the specific stream

PR #1834, opened 2020-10-22
Some cleanups arounds the metrics code

vagababov:
- use switch
- add new method that won't trigger lint, with a deprecation notice when deps are updated
- other minor things
- string(nil) actually works, so no need for an extra if

/assign @yanweiguo @evankanderson


tcnghia (approved):
/lgtm

evankanderson (commented):
/lgtm
/approve

That set of if statements bugged me, too.

vagababov:
Sorry, watching debates, pressing random buttons.

PR #1836, opened 2020-10-23
Remove NamespacedName logging helper

markusthoemmes:
As per title. This doesn't behave the way we expected it to. See https://github.com/knative/serving/pull/9923 for more info. This is only used in Serving now, so we can safely drop this after https://github.com/knative/serving/pull/9923 merges.

/assign @vagababov


markusthoemmes:
/hold

Until the serving PR merges.

PR #1838, opened 2020-10-23
Cherry-pick informer fix

n3wscott:
The EnableInjectionOrDie starts informers immedidately in a go
routine, which changes the order of informer start and controller
setup. As a result setting resync period to informer handler will
be ignored as the informer is already started. Though this will
likely affect more components, changing the EnableInjectionOrDie
function will break the API, so this is a temporary solution to
fix all the controllers depending on sharedmain.

Cherry-Pick of https://github.com/knative/pkg/pull/1767


n3wscott:
Note: the downstream tests are not smart enough to test on the release branches yet. Something I could look into later.

antoineco (approved):

PR #1843, opened 2020-10-26
Fix produced events if patching a finalizer fails

markusthoemmes:
The returned `resource` is broken (as nothing on it) if the `Patch` operation fails, which causes the produced event to be incomplete and in the wrong namespace (`default`).

Actually using a known-to-be-correct version of the object fixes the issue.

/assign @n3wscott


n3wscott (commented):
/lgtm
/approve

PR #1847, opened 2020-10-26
Remove deprecated methods now that downstream has been updated

vagababov:
/assign @yanweiguo


vagababov:
Downstream will work fine since it's the eventing dep that doesn't build, but eventing is already updated.

PR #1851, opened 2020-10-27
Splitting configmap package to avoid heavy dependencies

mattmoor:
/hold

Abusing the downstream test to see what breaks 🤞

### DIFF
--- a/configmap/nodep_test.go
+++ b/configmap/nodep_test.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package configmap_test
+
+import (
+	"testing"
+
+	"knative.dev/pkg/depcheck"
+)
+
+func TestNoDeps(t *testing.T) {
+	depcheck.AssertNoDependency(t, map[string][]string{
+		"knative.dev/pkg/configmap": depcheck.KnownHeavyDependencies,
+	})
+}
--- a/configmap/informer/informed_watcher.go
+++ b/configmap/informer/informed_watcher.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package configmap
+package informer
 
 import (
 	"errors"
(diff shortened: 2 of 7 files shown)


configmap/nodep_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"k8s.io/apimachinery/pkg/util/sets"
+	"knative.dev/pkg/nodep"
+)
+
+func TestNoDeps(t *testing.T) {
+	nodep.AssertNoDependency(t, map[string]sets.String{
+		"knative.dev/pkg/configmap": sets.NewString(
+			"k8s.io/client-go/informers",
+		),
+	})
```
dprotaso, line 31: - how does this affect test duration?
- can we avoid the package name `knative.dev/pkg/configmap` 
- do we have instances where binaries are depending on the configmap water interface but don't have informers?

vaikas: Just curious, is your feeling that exclusion list is going to be smaller, or the allow list? Just thinking about which way we think it's easier to create these tests. If we use exclusion list, then any new dependency is allowed by default. I wonder if just like we have test coverage, we could also in our actions do things like:
 * show the resulting binary size changes
 * show the resulting binary dep deltas
 etc.

Neat stuff indeed.

mattmoor: I renamed this package `depcheck` here: https://github.com/knative/pkg/pull/1853

We can add other assertions about the dependency graph (e.g. `AssertOnlyDeps`), which I could see being very useful for our most widely used libraries.  However, generally I think this will be prohibitive to maintain for non-leaf libraries.

mattmoor:
Related: https://github.com/knative/serving/issues/9957

dprotaso:
> Test took 17s

```
$ time go list -f '{{ join .Deps "\n" }}'
...
real	0m0.255s
user	0m0.760s
sys	0m0.416s
```

mattmoor:
```
go test ./configmap/
ok      knative.dev/pkg/configmap       0.922s
```
🙃   I was going off of the test timing on Prow.

dprotaso:
lol that's not 17s

dprotaso:
why is prow so slow?

vagababov:
Has it ever been fast?

dprotaso:
also @mattmoor run `go clean -testcache -cache` ;)

mattmoor:
I've never run `go clean` in my life, why start now?  What problem are we trying to solve...?

The test wasn't cached, or it wouldn't print the time. 🤷

vagababov:
`-count=1` ensures it is not cached

mattmoor:
I'm going to wait on this until after the release cut because it contains breaking changes, and we won't unlock the bigger win until we remove `k8s.io/client-go/kubernetes` from `knative.dev/metrics`

mattmoor:

prometheus export flake 🙃

mattmoor:
I staged PRs for the three broken downstream repositories.

Tekton has a couple places to fixup, but they are just in test code.

Issue #1854, opened 2020-10-27
Return typed (or wrapped context) error in pkg/signal.go


embano1:
A pattern I see used a lot when reacting to program termination is checking for the returned error to distinct an expected, e.g. context error, from an abnormal error.

E.g.:

```go
func main() {
	ctx := signals.NewContext()
	err := run(ctx, os.Args, os.Stdout)
	if err != nil && err != context.Canceled {
		panic(err)
	}
}
```

However, this does not work as `signals` slightly deviates from the `context` behavior and returns a plain error string:

```go
// Err implements context.Context
func (scc *signalContext) Err() error {
	select {
	case _, ok := <-scc.Done():
		if !ok {
			return errors.New("received a termination signal")
		}
	default:
	}
	return nil
}
```

Proposal: either return `context.Canceled` or a wrapped error including the error message so one can assert the error, e.g. using `errors.Is(err, context.Canceled)`.

mattmoor:
cc @dprotaso @n3wscott @markusthoemmes for thoughts

markusthoemmes:
Absolutely, let's do it. `context.Canceled` seems fine to me.

embano1:
Will file PR 👍

PR #1857, opened 2020-10-27
Return context.Canceled error in pkg/signals

embano1:
Return a typed `context.Canceled` error following the context package
semantic.

Closes: #1854

Might break existing users if they rely on the current error string
being returned with string matching, which is unlikely though.

Signed-off-by: Michael Gasch <mgasch@vmware.com>

### DIFF
--- a/signals/signal.go
+++ b/signals/signal.go
@@ -18,7 +18,6 @@ package signals
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):
/lgtm
/approve

PR #1859, opened 2020-10-27
Add AssertOnlyDeps variant

mattmoor:


depcheck/depcheck.go
```diff
@@ -109,3 +109,32 @@ func AssertNoDependency(t *testing.T, banned map[string][]string) {
...
+			allowed[x] = struct{}{}
+		}
+		t.Run(ip, func(t *testing.T) {
+			g, err := buildGraph(ip)
+			if err != nil {
+				t.Fatal("buildGraph(queue) =", err)
+			}
+			for name := range g {
+				if _, ok := allowed[name]; !ok {
+					t.Errorf("dependency of %s not explicitly allowed %s\n%s", ip, name,
```
vagababov, line 134:
```suggestion
					t.Errorf("Dependency %s of %s is not explicitly allowed %s\n%s", name, ip,
```
I think is more intuitive.

vagababov (commented):
In general, I think the core functionality should be extracted into a bool returning function so that the unit tests can test both positive and negative coverage. Right now we don't know if the failures work, besides manual checking.

mattmoor:
We can add `Check` variants that return an `error`, but I think a `bool` is far too lossy.

Probably worth doing separately.

vagababov:
OK I can do that :)

/lgtm

PR #1864, opened 2020-10-28
race was fixed in k8s 1.18

dprotaso:
See discussion: https://github.com/kubernetes/kubernetes/issues/90952

### DIFF
--- a/leaderelection/context_test.go
+++ b/leaderelection/context_test.go
@@ -1,6 +1,3 @@
-// +build !race
-// TODO(https://github.com/kubernetes/kubernetes/issues/90952): Remove the above.
-
 /*
 Copyright 2020 The Knative Authors
 


leaderelection/context_test.go
```diff
@@ -1,6 +1,3 @@
-// +build !race
-// TODO(https://github.com/kubernetes/kubernetes/issues/90952): Remove the above.
```
mattmoor, line 2: We should delete this too: https://github.com/knative/pkg/blob/ed076316cbe1107590822f930f4ed20b109290f1/test/presubmit-tests.sh#L48

vagababov (commented):
/lgtm
/approve

leaderelection/context_test.go
mattmoor: @vagababov 👀

vagababov: Sorry :| 
Somehow I missed this.
😢

PR #1867, opened 2020-10-28
Run update-deps.sh to gain correct licenses

markusthoemmes:
Somehow, this wasn't caught by our presubmit checks. Here we go then.

/assign @julz @vagababov


julz (approved):
weird

/lgtm

PR #1868, opened 2020-10-28
Use the correct module import for 'gopkg.in/evanphx/json-patch.v4'

markusthoemmes:
This changes the import name of the respective dependency to actually point to a valid gomodule.

/assign @julz @vaikas


julz (approved):
/lgtm

PR #1875, opened 2020-10-29
Remove TLogger and relative interfaces

markusthoemmes:
Mostly want to see if this has any impact currently. It's an open discussion whether or not we want to move forward with this.


markusthoemmes:
@coryrc any specific reason to hold this one?

PR #1882, opened 2020-11-02
Add host config option to the prometheus exporter

skonto:
Right now controller, autoscaler pods etc emit metrics using "0.0.0.0" which allows an arbitrary pod to view sensitive service information eg list all svcs etc in a multi-tenancy setting. We can disable metrics by setting no metrics backend, but ideally we would like a fine-grained configuration. My use case is to allow metrics to be exposed safely via a side-car like [rbac-proxy](https://github.com/brancz/kube-rbac-proxy). In this scenario first step is to constraint metrics on localhost by using an env var to pick up the right addr.

### DIFF
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"knative.dev/pkg/ptr"
 	"net"
 	"net/url"
 	"os"
@@ -33,6 +32,7 @@ import (
 	"go.opencensus.io/stats"
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/metrics/metricskey"
+	"knative.dev/pkg/ptr"
 )
 
 // metricsBackend specifies the backend to use for metrics
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net"
 	"net/url"
 	"os"
 	"path"
@@ -32,7 +31,6 @@ import (
 	"go.opencensus.io/stats"
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/metrics/metricskey"
-	"knative.dev/pkg/ptr"
 )
 
 // metricsBackend specifies the backend to use for metrics
@@ -61,7 +59,6 @@ const (
 
 	defaultBackendEnvName = "DEFAULT_METRICS_BACKEND"
 	defaultPrometheusPort = 9090
-	defaultPrometheusHost = ""
 	maxPrometheusPort     = 65535
 	minPrometheusPort     = 1024
 	prometheusPortEnvName = "METRICS_PROMETHEUS_PORT"
@@ -361,20 +358,18 @@ func prometheusPort() (int, error) {
 // Validation is performed via parsing the host value as part of a url.
 func prometheusHost() (*string, error) {
 	phStr := os.Getenv(prometheusHostEnvName)
+	// empty string either explicitly set via the env var or by default
+	// if the env var is not set translates to "0.0.0.0"
 	if phStr == "" {
-		return ptr.String(defaultPrometheusHost), nil
+		return &phStr, nil
 	}
 	u, err := url.Parse(phStr)
 	if err != nil {
 		return nil, fmt.Errorf("the environment variable %q could not be parsed as a host: %w",
 			prometheusHostEnvName, err)
 	}
-	host, _, err := net.SplitHostPort(u.Host)
-	if err != nil {
-		return nil, fmt.Errorf("could not extract host from %q: %w",
-			u.Host, err)
-	}
-	return &host, nil
+	hostname := u.Hostname()
+	return &hostname, nil
 }
 
 // JSONToOptions converts a json string to ExporterOptions.
--- a/metrics/config.go
+++ b/metrics/config.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net/url"
 	"os"
 	"path"
 	"strconv"
@@ -61,6 +60,7 @@ const (
 	defaultPrometheusPort = 9090
 	maxPrometheusPort     = 65535
 	minPrometheusPort     = 1024
+	defaultPrometheusHost = "0.0.0.0"
 	prometheusPortEnvName = "METRICS_PROMETHEUS_PORT"
 	prometheusHostEnvName = "METRICS_PROMETHEUS_HOST"
 )
@@ -247,11 +247,8 @@ func createMetricsConfig(ctx context.Context, ops ExporterOptions) (*metricsConf
 
 		mc.prometheusPort = pp
 
-		phost, err := prometheusHost()
-		if err != nil {
-			return nil, fmt.Errorf("failed to determine Prometheus host: %w", err)
-		}
-		mc.prometheusHost = *phost
+		phost := prometheusHost()
+		mc.prometheusHost = phost
 	case stackdriver:
 		// If stackdriverClientConfig is not provided for stackdriver backend destination, OpenCensus will try to
 		// use the application default credentials. If that is not available, Opencensus would fail to create the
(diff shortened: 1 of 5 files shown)


metrics/config.go
```diff
@@ -58,9 +61,11 @@ const (
 
 	defaultBackendEnvName = "DEFAULT_METRICS_BACKEND"
 	defaultPrometheusPort = 9090
+	defaultPrometheusHost = ""
```
vagababov, line 64: do we need a const for ""? Or it has to be 0.0.0.0

metrics/config.go
```diff
@@ -341,6 +356,27 @@ func prometheusPort() (int, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("the environment variable %q could not be parsed as a host: %w",
+			prometheusHostEnvName, err)
+	}
+	host, _, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		return nil, fmt.Errorf("could not extract host from %q: %w",
+			u.Host, err)
+	}
+	return &host, nil
```
vagababov, line 377: https://golang.org/pkg/net/url/#URL.Hostname perhaps?

metrics/config.go
skonto: If we keep the convention of "" http will use `0.0.0.0` anyway, I guess I can remove the `const`.

metrics/config.go
```diff
@@ -341,6 +353,25 @@ func prometheusPort() (int, error) {
...
 
+// prometheusHost returns the host configured via the environment
+// for the Prometheus metrics exporter if it's set, a default value otherwise.
+// Validation is performed via parsing the host value as part of a url.
+func prometheusHost() (*string, error) {
+	phStr := os.Getenv(prometheusHostEnvName)
+	// empty string either explicitly set via the env var or by default
+	// if the env var is not set translates to "0.0.0.0"
+	if phStr == "" {
+		return &phStr, nil
```
zhongduo, line 364: Shouldn't this return "0.0.0.0" instead of empty?

metrics/config.go
```diff
@@ -341,6 +353,25 @@ func prometheusPort() (int, error) {
 	return int(pp), nil
 }
 
+// prometheusHost returns the host configured via the environment
+// for the Prometheus metrics exporter if it's set, a default value otherwise.
+// Validation is performed via parsing the host value as part of a url.
+func prometheusHost() (*string, error) {
```
zhongduo, line 359: I don't think returning string pointer is a common practice in Golang or eventing. For this particular case, it seems to me that you don't need a nil host value at all. For performance, RVO will optimize it out anyway.

metrics/config.go
```diff
@@ -105,6 +107,10 @@ type metricsConfig struct {
 	// format. It defaults to 9090.
 	prometheusPort int
 
+	// prometheusHost is the host where the metrics are exposed in Prometheus
+	// format. It defaults to "0.0.0.0"
```
zhongduo, line 111: If I understand the logic correctly, you actually mean "an empty host will default to 0.0.0.0 by prometheus"

metrics/prometheus_exporter_test.go
```diff
@@ -40,8 +41,9 @@ func TestNewPrometheusExporter(t *testing.T) {
 			component:          testComponent,
 			backendDestination: prometheus,
 			prometheusPort:     9091,
+			prometheusHost:     "127.0.0.1",
```
zhongduo, line 44: Maybe add a test for configuration from env?

metrics/config.go
skonto: ok.

metrics/config.go
skonto: This is what http uses as a convention and it is commonly used, returning `0.0.0.0` will add more changes to this PR for the sake of maybe clarity.

metrics/config.go
skonto: Yes that is the http convention re-used.

metrics/prometheus_exporter_test.go
skonto: ok I can do that.

metrics/config.go
zhongduo: Yes, that's the common case of using *string, but in your case you already have error return value, so when error is not nil, string should be ignored. That's why I said "in this particular case".

metrics/config.go
zhongduo: Yes, I realized that after reading the whole code.

metrics/config.go
zhongduo: Thanks for clarifying. I was confused by the comment, thinking that you want to force it to 0.0.0.0 if not set in env.

metrics/prometheus_exporter.go
```diff
@@ -82,7 +82,7 @@ func startNewPromSrv(e *prom.Exporter, port int) *http.Server {
 		curPromSrv.Close()
 	}
 	curPromSrv = &http.Server{
-		Addr:    fmt.Sprint(":", port),
+		Addr:    fmt.Sprintf("%s:%d", host, port),
```
vagababov, line 85: `host + ":" + strconv.Itoa(port)`
But up to you, this is going to be called once probably in the lifetime of the service.

vagababov (commented):
/lgtm
In general

metrics/config.go
vagababov: Returning *string is useful if we need to discern between empty and nil and is used a lot in APIs around K8s.
Here it is probably a tossup, so should not matter.

metrics/config.go
skonto: This will not work if you pass an ip btw, it returns "".

skonto:
@zhongduo @vagababov thanks for reviewing. Is simplified so no validation, because validating a domain is too much and not that easy unless there is a util to use out of the box (url is not correct to use). I am now returning "0.0.0.0" to make the default hostname explicit.  Added a couple of testcases as suggested,

metrics/config.go
```diff
@@ -240,6 +246,9 @@ func createMetricsConfig(ctx context.Context, ops ExporterOptions) (*metricsConf
 		}
 
 		mc.prometheusPort = pp
+
+		phost := prometheusHost()
+		mc.prometheusHost = phost
```
yanweiguo, line 251:
```suggestion
		mc.prometheusHost = prometheusHost()
```

metrics/prometheus_exporter_test.go
```diff
@@ -57,6 +61,68 @@ func TestNewPrometheusExporter(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.prometheusPortVarName != "" {
+				os.Setenv(tc.prometheusPortVarName, tc.prometheusPortVarValue)
+				defer os.Unsetenv(tc.prometheusPortVarName)
+			}
+			if tc.prometheusHostVarName != "" {
+				os.Setenv(tc.prometheusHostVarName, tc.prometheusHostVarValue)
+				defer os.Unsetenv(tc.prometheusHostVarName)
+			}
+			mc, err := createMetricsConfig(context.Background(), tc.ops)
```
yanweiguo, line 113: The error here is not checked.

metrics/exporter.go
```diff
@@ -70,6 +70,11 @@ type ExporterOptions struct {
 	// Prometheus.
 	PrometheusPort int
 
+	// PrometheusHost is the host to expose metrics on if metrics backend is Prometheus.
+	// The dfault value is "0.0.0.0" If is ignored if metrics backend is not
```
zhongduo, line 74: default

zhongduo: "If is ignored"?

metrics/prometheus_exporter_test.go
skonto: fixed

skonto:
Not sure if I can fix this:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.32.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: newPrometheusExporter - result 1 (knative.dev/pkg/metrics.ResourceExporterFactory) is never used (unparam)
  
  Error: issues found
```
This is a limitation of the linter it seems.

metrics/prometheus_exporter_test.go
```diff
@@ -57,6 +61,71 @@ func TestNewPrometheusExporter(t *testing.T) {
...
+			if err != nil {
+				t.Error(err)
```
yanweiguo, line 115: Should this be `t.Fatal`? I think if there's error, it's meaningless to use the returned `mc`.

metrics/prometheus_exporter_test.go
```diff
@@ -57,6 +61,71 @@ func TestNewPrometheusExporter(t *testing.T) {
...
+			}
+			e, _, err := newPrometheusExporter(mc, TestLogger(t))
+			if err != nil {
+				t.Error(err)
```
yanweiguo, line 119: Same here.

yanweiguo (commented):
some nits

PR #1884, opened 2020-11-02
Add tests for pre-release versions and add a better error message

### LINKED ISSUE #1883: KubernetesMinVersionKey check doesn't work with EKS?
## Expected Behavior

When the minimum version of kubernetes required is 1.15.11,  EKS 1.15 with KUBERNETES_MIN_VERSION=1.15.11 should pass the minimum version check.

## Actual Behavior

Using EKS 1.15 with KUBERNETES_MIN_VERSION=1.15.11 and Tekton Pipelines v0.16.3 raises this error:

```
kubernetes version \"1.15.11-eks-065dce\" is not compatible, need at least \"1.15.11\" (this can be overridden with the env var \"KUBERNETES_MIN_VERSION\")
```

Strangely, the Pipelines 0.16.3 docs indicate we actually need 1.16 https://github.com/tektoncd/pipeline/blob/v0.16.3/docs/install.md#before-you-begin but the error seems to indicate that the check is looking for 1.15.11 - I think the question here is whether or not the check is able to handle the string `1.15.11-eks-065dce`. If it is, then maybe we're misunderstanding the real root problem.

## Additional Info

Bug opened in Tekton: https://github.com/tektoncd/pipeline/issues/3335 

/kind bug

Not sure what area to pick!

vagababov:
When we have pre-release version it is very unintuitive with the default error message, since just looking at the string
`1.15.11-abs-1234` is larger than `1.15.11`, but semver treats `-abs-1234` as a prerelease version and hence smaller than
`1.15.11`.
So expand the error in this case

Fixes #1883

/assign @dprotaso mattmoor

### DIFF
--- a/version/version_test.go
+++ b/version/version_test.go
@@ -43,6 +43,21 @@ func TestVersionCheck(t *testing.T) {
 		name:          "greater version (patch)",
 		actualVersion: &testVersioner{version: "v1.17.2"},
 	}, {
+		name:          "greater version (patch), no v",
+		actualVersion: &testVersioner{version: "1.17.2"},
+	}, {
+		name:          "greater version (patch), pre-release",
+		actualVersion: &testVersioner{version: "1.17.2-kpn-065dce"},
+	}, {
+		name:            "greater version (patch), pre-release, envvar override",
+		actualVersion:   &testVersioner{version: "1.15.11-kpn-065dce"},
+		versionOverride: "1.15.11",
+		wantError:       true,
+	}, {
+		name:            "greater version (patch), pre-release, envvar override, -0 hack",
+		actualVersion:   &testVersioner{version: "1.15.11-kpn-065dce"},
+		versionOverride: "1.15.11-0",
+	}, {
 		name:          "greater version (minor)",
 		actualVersion: &testVersioner{version: "v1.18.0"},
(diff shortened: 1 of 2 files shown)


version/version_test.go
```diff
@@ -42,6 +42,21 @@ func TestVersionCheck(t *testing.T) {
...
+	}, {
+		name:          "greater version (patch), pre-release",
+		actualVersion: &testVersioner{version: "1.17.2-kpn-065dce"},
+	}, {
+		name:            "greater version (patch), pre-release, envvar override",
+		actualVersion:   &testVersioner{version: "1.15.11-kpn-065dce"},
+		versionOverride: "1.15.11",
+		wantError:       true,
+	}, {
+		name:            "greater version (patch), pre-release, envvar override, -0 hack",
```
dprotaso, line 57: whats the `-0` hack for?

vagababov: It's _smallest_ reasonable pre-release version

vagababov:
Also 100% test coverage 🎉

version/version_test.go
dprotaso: ok - so we're comparing two prerelease versions

thanks

PR #1892, opened 2020-11-03
include a message when the logging config is bad

dprotaso:


logging/config.go
```diff
@@ -65,7 +65,10 @@ func NewLogger(configJSON string, levelOverride string, opts ...zap.Option) (*za
 	if err2 != nil {
 		panic(err2)
 	}
-	return enrichLoggerWithCommitID(logger.Named(fallbackLoggerName)), loggingCfg.Level
+
+	slogger := enrichLoggerWithCommitID(logger.Named(fallbackLoggerName))
+	slogger.Warnw("Failed to parse logging config", zap.Error(err))
```
yanweiguo, line 70: Shall we also mention that production config is used instead? I read this logs as something fails and causes panic but actually the error is ignored.

dprotaso: yeah good point

PR #1896, opened 2020-11-05
Add GetObservers to tracker interface

markusthoemmes:
This allows an arbitrary client to fetch all the observers of a given object. Ultimately, this enables an informer setup that can use this knowledge as a filter to decide whether or not the current object is actually watched by anything.

/assign @mattmoor @dprotaso

### DIFF
--- a/reconciler/testing/tracker.go
+++ b/reconciler/testing/tracker.go
@@ -56,11 +56,10 @@ func (n *FakeTracker) GetObservers(obj interface{}) []types.NamespacedName {
 		Name:       or.Name,
 	}
 
-	keys := make([]types.NamespacedName, 0, len(n.references[ref]))
-
 	n.Lock()
 	defer n.Unlock()
 
+	keys := make([]types.NamespacedName, 0, len(n.references[ref]))
 	for key := range n.references[ref] {
 		keys = append(keys, key)
 	}
--- a/reconciler/testing/tracker.go
+++ b/reconciler/testing/tracker.go
@@ -41,6 +41,32 @@ var _ tracker.Interface = (*FakeTracker)(nil)
 // OnChanged implements OnChanged.
 func (*FakeTracker) OnChanged(interface{}) {}
 
+// GetObservers implements GetObservers.
+func (n *FakeTracker) GetObservers(obj interface{}) []types.NamespacedName {
+	item, err := kmeta.DeletionHandlingAccessor(obj)
+	if err != nil {
+		return nil
+	}
+
+	or := kmeta.ObjectReference(item)
+	ref := tracker.Reference{
+		APIVersion: or.APIVersion,
+		Kind:       or.Kind,
+		Namespace:  or.Namespace,
+		Name:       or.Name,
+	}
+
+	keys := make([]types.NamespacedName, 0, len(n.references[ref]))
+
+	n.Lock()
+	defer n.Unlock()
+
+	for key := range n.references[ref] {
+		keys = append(keys, key)
+	}
+	return keys
+}
+
 // OnDeletedObserver implements OnDeletedObserver.
 func (n *FakeTracker) OnDeletedObserver(obj interface{}) {
 	item, err := kmeta.DeletionHandlingAccessor(obj)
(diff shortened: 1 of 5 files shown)


dprotaso:
> This allows an arbitrary client to fetch all the observers of a given object. Ultimately, this enables an informer setup that can use this knowledge as a filter to decide whether or not the current object is actually watched by anything.

The tracker does this for you so I don't understand the use case here

markusthoemmes:
@dprotaso it does, yes, but I cannot add more filters to the informer that would be applied **after** the tracker filter.

My use-case in Kourier is, that I want to filter endpoint updates. I only want to do that for updates to endpoints we're actually watching though. Computing the filter is relatively expensive, as I compare the set of all IPs from both endpoints and determine if they changed at all.

I have a broader change in mind that I'd work on after this lands, which would pull the `impl.EnqueueKey` callback out of the tracker's constructor and into a `MapObservers` (bikeshed) function to more naturally flow with how informers are "usually" setup.

dprotaso:
Why not just add a separate constructor `NewWithFilters`

https://github.com/knative/pkg/blob/f98befda69eb2f154bc0bc147622ae35d004cde0/tracker/enqueue.go#L43-L48

markusthoemmes:
Because I think we're actually kinda holding it wrong currently. Additionally, the callback we currently have assumes just the `new` object to be passed through. To compute my filter, I need the new and the old object though (i.e. in the update handler).

That's why I'd like to flip the current machinery on its head a bit and allow the tracker to integrate better with informer setup.

tracker/interface.go
```diff
@@ -71,6 +72,10 @@ type Interface interface {
 	// so that we are notified for appropriate object changes.
 	OnChanged(obj interface{})
 
+	// GetObservers returns the names of all observers for the given
+	// object.
+	GetObservers(obj interface{}) []types.NamespacedName
+
```
dprotaso, line 78: I'd drop this from the interface since it breaks anyone using this package

tracker/enqueue.go
```diff
@@ -229,14 +238,9 @@ func (i *impl) OnChanged(obj interface{}) {
...
 	}
 
-	i.m.Lock()
-	// Call the callbacks without the lock held.
 	var keys []types.NamespacedName
-	defer func(cb func(types.NamespacedName)) {
-		for _, key := range keys {
-			cb(key)
-		}
-	}(i.cb) // read i.cb with the lock held
```
dprotaso, line 239: was `cb` every mutated?

reconciler/testing/tracker.go
```diff
@@ -41,6 +41,32 @@ var _ tracker.Interface = (*FakeTracker)(nil)
 // OnChanged implements OnChanged.
 func (*FakeTracker) OnChanged(interface{}) {}
 
+// GetObservers implements GetObservers.
+func (n *FakeTracker) GetObservers(obj interface{}) []types.NamespacedName {
```
dprotaso, line 45: Can optionally drop this if it's not part of the interface

tracker/interface.go
markusthoemmes: How am I supposed to use the added functionality then though? Return the struct type from `New`?

tracker/enqueue.go
markusthoemmes: I don't think so, but might be worth doublechecking.

reconciler/testing/tracker.go
```diff
@@ -41,6 +41,32 @@ var _ tracker.Interface = (*FakeTracker)(nil)
...
+		APIVersion: or.APIVersion,
+		Kind:       or.Kind,
+		Namespace:  or.Namespace,
+		Name:       or.Name,
+	}
+
+	keys := make([]types.NamespacedName, 0, len(n.references[ref]))
+
+	n.Lock()
+	defer n.Unlock()
```
dprotaso, line 62: this should guard line 59

tracker/interface.go
dprotaso: > Return the struct type from New?

blah I thought it was already doing that :(

dprotaso:
Given https://github.com/knative/pkg/pull/1896#discussion_r520015346

> > Return the struct type from New?
> 
> blah I thought it was already doing that :(

/approve
/lgtm
/hold for the misplaced lock line

reconciler/testing/tracker.go
markusthoemmes: ugh, right, cause I take `len`

PR #1898, opened 2020-11-06
Add a flag to disable QP profiling.

vagababov:
Turning profiling on or off toggles full rolling restart
of knative pods. And if profiling QP is not of interest, this
might be quite undesireable. To permit that.

For https://github.com/knative/serving/issues/10051

### DIFF
--- a/metrics/config_observability.go
+++ b/metrics/config_observability.go
@@ -73,6 +73,12 @@ type ObservabilityConfig struct {
 	// the pods via an HTTP server in the format expected by the pprof visualization tool.
 	EnableProfiling bool
 
+	// DisableQPProfiling indicates whether EnableProfiling flag applies to the QueueProxy processes.
+	// If false (default) — enabling or disabling profiling causes a global rolling restart of the
+	// Knative service pods, which might not be desireable (especially if there is no need to profile
+	// QP processes).
+	DisableQPProfiling bool
+
 	// EnableRequestLog enables activator/queue-proxy to write request logs.
 	EnableRequestLog bool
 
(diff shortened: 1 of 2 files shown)


metrics/config_observability.go
```diff
@@ -73,6 +73,12 @@ type ObservabilityConfig struct {
 	// the pods via an HTTP server in the format expected by the pprof visualization tool.
 	EnableProfiling bool
 
+	// DisableQPProfiling indicates whether EnableProfiling flag applies to the QueueProxy processes.
```
yanweiguo, line 76: nit: Renamed to `EnableQPProfiling` to be consistent with other flags?

vagababov: I had that initially. But it means default true, which is non-intuitive.

metrics/config_observability.go
```diff
@@ -73,6 +73,12 @@ type ObservabilityConfig struct {
...
+	// If false (default) — enabling or disabling profiling causes a global rolling restart of the
+	// Knative service pods, which might not be desireable (especially if there is no need to profile
```
yanweiguo, line 78:
```suggestion
	// Knative service pods, which might not be desirable (especially if there is no need to profile
```

vagababov: yeah, I saw it :)

markusthoemmes:
It feels quite a bit odd to have a queue-proxy specific flag in pkg to be honest.

vagababov:
It is a bit... I can replay this as a `deployment.yaml` setting.

PR #1902, opened 2020-11-12
Fix a few more occurrences of divisive language

markusthoemmes:
As per title. Dummy and "sanity check" are considered divisive language.

/assign @mattmoor


mattmoor (approved):
/lgtm

PR #1906, opened 2020-11-14
drainer support kprobe

runzexia:
Signed-off-by: runzexia <root@junwuhui.cn>
ref https://github.com/knative/eventing/issues/3911
https://github.com/knative/networking/blob/bf808e4cd3f37ea514d365ffb2bb5ffa908e141c/pkg/probe/handler.go#L47-L48

### DIFF
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -17,7 +17,6 @@ limitations under the License.
 package handlers
 
 import (
-	"fmt"
 	"net/http"
 	"sync"
 	"time"
@@ -88,14 +87,9 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		if d.draining() {
 			http.Error(w, "shutting down", http.StatusServiceUnavailable)
 		} else {
-			hh := r.Header.Get(network.HashHeaderName)
-			if hh == "" {
-				http.Error(w, fmt.Sprintf("a probe request must contain a non-empty %q header", network.HashHeaderName), http.StatusBadRequest)
-				return
-			}
-			w.Header().Set(network.HashHeaderName, hh)
-			w.WriteHeader(http.StatusOK)
+			network.ServeKProbe(w, r)
 		}
+		return
 	}
 
 	d.reset()
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -17,6 +17,7 @@ limitations under the License.
 package handlers
 
 import (
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


network/handlers/drain.go
```diff
@@ -83,6 +84,19 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if d.draining() {
+			http.Error(w, "shutting down", http.StatusServiceUnavailable)
+		} else {
+			hh := r.Header.Get(network.HashHeaderName)
+			if hh == "" {
+				http.Error(w, fmt.Sprintf("a probe request must contain a non-empty %q header", network.HashHeaderName), http.StatusBadRequest)
+				return
+			}
+			w.Header().Set(network.HashHeaderName, hh)
+			w.WriteHeader(http.StatusOK)
```
mattmoor, line 97: Can we factor a function for this?  We'll want to deduplicate this logic (incl. constants) when it lands in knative/networking, so we should make this reusable.

mattmoor:
cc @ZhiminXiang @tcnghia @nak3

network/handlers/drain.go
runzexia: done😁

network/handlers/drain_test.go
```diff
@@ -236,6 +236,58 @@ func TestDrainMechanics(t *testing.T) {
...
 }
 
+func TestDrainerKProbe(t *testing.T) {
+	var (
+		w          http.ResponseWriter
+		req        = &http.Request{}
+		kprobehash = "hash"
+		kprobe     = &http.Request{
+			Header: http.Header{
+				"K-Network-Probe": []string{"probe"},
```
nak3, line 246: Could you use the const value `K-Network-Probe`, `K-Network-Hash` and `probe` defined in network/network.go?

network/network_test.go
```diff
@@ -42,3 +43,58 @@ func TestIsKubeletProbe(t *testing.T) {
...
+
+func TestIsKnativeProbe(t *testing.T) {
+	req, err := http.NewRequest(http.MethodGet, "http://example.com/", nil)
+	if err != nil {
+		t.Fatal("Error building request:", err)
+	}
+	if IsKProbe(req) {
+		t.Error("Not a knative probe but counted as such")
+	}
+	req.Header.Set("K-Network-Probe", "probe")
```
nak3, line 55: Same with above. Could you use the const values in this test file?

network/network_test.go
```diff
@@ -42,3 +43,58 @@ func TestIsKubeletProbe(t *testing.T) {
...
+	if !IsKProbe(req) {
+		t.Error("knative probe but not counted as such")
+	}
+	req.Header.Del("K-Network-Probe")
+	if IsKProbe(req) {
+		t.Error("Not a knative probe but counted as such")
+	}
+	req.Header.Set("K-Network-Probe", "no matter")
+	if IsKProbe(req) {
+		t.Error("Not a knative probe but not counted as such")
```
nak3, line 65: This test would be `"Not a knative probe but counted as such"`?

nak3 (changes requested):

network/handlers/drain_test.go
runzexia: ok, thanks for review😋

nak3 (approved):

mattmoor (commented):
/lgtm

tcnghia (approved):

Issue #1764, opened 2020-10-01
FilterFunc in the informer is not working on the first run of reconciler


danyinggu:
## Expected Behavior
The informer should always successfully filter objects based on filterFunc defined the informer

## Actual Behavior
In the first run of the reconcile, it lists all objects and objects are not filtered by filterFunc in the informer 

## Steps to Reproduce the Problem

1. Given a controller with an informer that has a filterFunc
2. Deploy the controller 
3. In the first run of the reconciler, it reconciles all the objects but not only the objects should pass the filterFunc

## Additional Info
We did some investigation, in the genreconciler: https://github.com/google/knative-gcp/blob/3707bd3ca3d010532b334201dc7da6c9bc382959/pkg/client/injection/reconciler/broker/v1beta1/broker/controller.go#L73-L81
When a controller starts, a new leader is promoted with a listAll operation and there is no informer resync.

danyinggu:
@mattmoor @n3wscott @grantr

danyinggu:
kind/bug

mattmoor:
Yup, there's no filter passed to genreconciler currently, but you could plumb one through in the controller options.

grantr:
@mattmoor can you point to an example of plumbing through? Looks like the code directly enqueues every object in the lister, skipping any informer FilterFunc.

mattmoor:
By "plumbing through" I mean adding a field and calling it from that logic in genreconciler itself, there is no field today.

PR #1909, opened 2020-11-16
Allow global resync & reconcilerImpl informers to be filtered via filterfunc

lberk:
related to #1764

There were two missing places where the informers from the shared cache we're not being filtered.
1. in the globalresync functions (where the filterFunc being passed through was ~= `return true`
1. the informer in the `LeaderAwareFuncs.PromoteFunc` which was marked with a TODO to allow filtering

The former has been changed to allow for a `controllerImpl.GlobalResyncFilterFunc` to be set (and if not, the former default of `return true` is passed to allow objects.

The latter was complicated to pass as a function, if done as a `controller.Option`, it would be too late for the closure in the initial `reconcilerImpl`.  I've opted instead for a `pkg/controller/controller.go` variable that will store via the shared context and again -- if not set -- pass a default filter of `return true` filtering over the objects. 

/cc @vaikas @n3wscott @matzew

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_controller.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller.go
@@ -138,6 +138,10 @@ func (g *reconcilerControllerGenerator) GenerateType(c *generator.Context, t *ty
 			Package: "knative.dev/pkg/controller",
 			Name:    "OptionsFn",
 		}),
+		"controllerGetFilterFunc": c.Universe.Type(types.Name{
+			Package: "knative.dev/pkg/controller",
+			Name:    "GetFilterFunc",
+		}),
 		"contextContext": c.Universe.Type(types.Name{
 			Package: "context",
 			Name:    "Context",
@@ -203,6 +207,7 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 
 	lister := {{.type|lowercaseSingular}}Informer.Lister()
 
+	filterFunc := {{.controllerGetFilterFunc|raw}}(ctx)
 	rec := &reconcilerImpl{
 		LeaderAwareFuncs: {{.reconcilerLeaderAwareFuncs|raw}}{
 			PromoteFunc: func(bkt {{.reconcilerBucket|raw}}, enq func({{.reconcilerBucket|raw}}, {{.typesNamespacedName|raw}})) error {
@@ -211,11 +216,12 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 					return err
 				}
 				for _, elt := range all {
-					// TODO: Consider letting users specify a filter in options.
-					enq(bkt, {{.typesNamespacedName|raw}}{
-						Namespace: elt.GetNamespace(),
-						Name: elt.GetName(),
-					})
+					if ok := filterFunc(elt); ok {
+						enq(bkt, {{.typesNamespacedName|raw}}{
+							Namespace: elt.GetNamespace(),
+							Name: elt.GetName(),
+						})
+					}
 				}
 				return nil
 			},
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -174,6 +174,8 @@ func FilterWithNameAndNamespace(namespace, name string) func(obj interface{}) bo
 	}
 }
 
+type filterFunc func(obj interface{}) bool
+
 // Impl is our core controller implementation.  It handles queuing and feeding work
 // from the queue to an implementation of Reconciler.
 type Impl struct {
@@ -203,15 +205,22 @@ type Impl struct {
 
 	// StatsReporter is used to send common controller metrics.
 	statsReporter StatsReporter
+
+	// GlobalResyncFilterFunc is used to filter our objects from
+	// the shared cache on a global resync, be default and if not
+	// set by the controller implemenation, it will default to
+	// allowing every object in the cache
+	GlobalResyncFilterFunc filterFunc
 }
 
 // ControllerOptions encapsulates options for creating a new controller,
 // including throttling and stats behavior.
 type ControllerOptions struct { //nolint // for backcompat.
-	WorkQueueName string
-	Logger        *zap.SugaredLogger
-	Reporter      StatsReporter
-	RateLimiter   workqueue.RateLimiter
+	WorkQueueName          string
+	Logger                 *zap.SugaredLogger
+	Reporter               StatsReporter
+	RateLimiter            workqueue.RateLimiter
+	GlobalResyncFilterFunc filterFunc
 }
 
 // NewImpl instantiates an instance of our controller that will feed work to the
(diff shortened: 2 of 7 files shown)


controller/controller.go
```diff
@@ -236,12 +245,16 @@ func NewImplFull(r Reconciler, options ControllerOptions) *Impl {
 	if options.Reporter == nil {
 		options.Reporter = MustNewStatsReporter(options.WorkQueueName, options.Logger)
 	}
+	if options.GlobalResyncFilterFunc == nil {
+		options.GlobalResyncFilterFunc = func(obj interface{}) bool { return true }
```
vagababov, line 249: use a fixed object rather than lambda?

controller/controller.go
```diff
@@ -203,15 +205,22 @@ type Impl struct {
 
 	// StatsReporter is used to send common controller metrics.
 	statsReporter StatsReporter
+
+	// GlobalResyncFilterFunc is used to filter our objects from
+	// the shared cache on a global resync, be default and if not
+	// set by the controller implemenation, it will default to
+	// allowing every object in the cache
+	GlobalResyncFilterFunc filterFunc
```
julz, line 213: does this need to be public, since it's set by an option?

controller/controller.go
```diff
@@ -542,19 +555,18 @@ func (c *Impl) handleErr(err error, key types.NamespacedName) {
...
 func (c *Impl) GlobalResync(si cache.SharedInformer) {
-	alwaysTrue := func(interface{}) bool { return true }
-	c.FilteredGlobalResync(alwaysTrue, si)
+	c.FilteredGlobalResync(c.GlobalResyncFilterFunc, si)
 }
 
 // FilteredGlobalResync enqueues all objects from the
 // SharedInformer that pass the filter function in to the slow queue.
-func (c *Impl) FilteredGlobalResync(f func(interface{}) bool, si cache.SharedInformer) {
+func (c *Impl) FilteredGlobalResync(_ func(interface{}) bool, si cache.SharedInformer) {
```
julz, line 563: Do we really want to ignore this parameter rather than just changing the default in GlobalResync from alwaysTrue to GlobalResyncFilterFunc (as we already do above)? Seems quite confusing to have this parameter entirely ignored, and [the readme specifically documents being able to pass a custom filter here](https://github.com/knative/pkg/blob/6f19c52bf18091ef7f344e57c72db8cbf9d95be3/injection/README.md#configstore).

controller/controller.go
```diff
@@ -542,19 +555,18 @@ func (c *Impl) handleErr(err error, key types.NamespacedName) {
 
 // GlobalResync enqueues into the slow lane all objects from the passed SharedInformer
 func (c *Impl) GlobalResync(si cache.SharedInformer) {
-	alwaysTrue := func(interface{}) bool { return true }
-	c.FilteredGlobalResync(alwaysTrue, si)
+	c.FilteredGlobalResync(c.GlobalResyncFilterFunc, si)
```
julz, line 558: I think this parameter is actually ignored since `FilteredGlobalResync` hard-codes using the GlobalResyncFilterFunc below and ignores its argument.. seems potentially confusing/wrong?

controller/controller.go
julz: (kind of surprised that no tests failed on this change, but it looks like we currently only have tests via `GlobalResync` which is probably why :D wonder if it's worth adding some tests specifically for `FilteredGlobalResync`)

markusthoemmes: Ack. Looks like you actually want `c.GlobalResyncFilterFunc(obj) && f(obj)` in this case.

markusthoemmes, line 563:
```suggestion
func (c *Impl) FilteredGlobalResync(_ filterFunc, si cache.SharedInformer) {
```

lberk: Nope, thanks for catching that (and the above).  I've fixed the immediate issue -- working on tests in a different commit

controller/controller.go
```diff
@@ -56,6 +56,11 @@ var (
 	// may adjust this process-wide default.  For finer control, invoke
 	// Run on the controller directly.
 	DefaultThreadsPerController = 2
+
+	// alwaysTrue is the default FilterFunc, which allows all objects
+	// passed through this is used in the default case for passing all
+	// objects to the slowlane from the passed sharedInformer
```
julz, line 62: super nit: I think there's a bit of missing punctuation here -
```suggestion
	// alwaysTrue is the default FilterFunc, which allows all objects
	// passed through. This is used in the default case for passing all
	// objects to the slowlane from the passed sharedInformer.
```

controller/controller.go
```diff
@@ -203,15 +210,22 @@ type Impl struct {
 
 	// StatsReporter is used to send common controller metrics.
 	statsReporter StatsReporter
+
+	// GlobalResyncFilterFunc is used to filter our objects from
+	// the shared cache on a global resync, be default and if not
+	// set by the controller implemenation, it will default to
+	// allowing every object in the cache
```
julz, line 217:
```suggestion
	// GlobalResyncFilterFunc is used to filter out objects from
	// the shared cache on a global resync. By default and if not
	// set by the controller implemenation, it will default to
	// allowing every object in the cache.
```

julz (commented):
lgtm apart from a couple comment typos

lberk:
thanks for the reviews @julz, should be fixed at this point.

julz (commented):
/lgtm

codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -203,6 +207,7 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 
 	lister := {{.type|lowercaseSingular}}Informer.Lister()
 
+	filterFunc := {{.controllerGetFilterFunc|raw}}(ctx)
```
mattmoor, line 210: I'd rather see this come through the `controller.Options`, but the ordering and capture make this awkward.

I think that if you have `filterFunc := noFilter` up here, and optionally overwrite it below the capture should work out correctly though?

mattmoor: in fact the rest of the things already seem to be doing this, so you are pretty close  🤔

markusthoemmes:
Which state is this one in? Seems like it fell between the cracks @lberk ?

lberk:
@markusthoemmes this is still a problem in eventing.  I was asked for further justification that this solves the root cause of the issue, I haven't had time to return to it since.

codegen/cmd/injection-gen/generators/reconciler_controller.go
markusthoemmes: That's what @benmoss did in https://github.com/knative/pkg/pull/1909 too.

benmoss: #2170 rather 😄 . Just clicked your link and momentarily was very confused

dprotaso:
I'm going to close this out since it's been a number of months - feel free reopen

mattmoor:
I’m pretty sure Options now has PromoteFilterFunc…

Issue #923, opened 2019-12-04
Verify that the K8s service actually exists in K8s in func URIFromObjectReference


capri-xiyue:
## Expected Behavior
We should be able to know whether K8s service actually exists in K8s in func URIFromObjectReference()
## Actual Behavior
Currently, we can still resolve URI from K8s service even though it doesn't exists.

vagababov:
@capri-xiyue 
is this still a desired improvement?

capri-xiyue:
> @capri-xiyue
> is this still a desired improvement?

Yep. This is still a desired improvement.

vagababov:
Are you going to work on it?

capri-xiyue:
> Are you going to work on it?

I may not have cycles in a short time. If someone else have free cycles,  please feel free to pick up this tasks. 
If not, I will revisit it when I have cycles.

PR #1910, opened 2020-11-16
Some small changes for the URL test

vagababov:
I was looking at the #923 (it'll be quite some effort to get the
informer there, or create dynamic one for k8s services),
but nonetheless clean the test up.

- Rename the getters to exclude word get (as go style suggests).
- The errors test contains only errors, so stream line it
- Ensure errors are also idempotent, which wasn't.

/assign @n3wscott mattmoor


n3wscott (commented):
/lgtm
/approve

PR #1915, opened 2020-11-17
Use the same dial with backoff for H2C as we do for HTTP/1.1

vagababov:
Also remove the constant that is not used anywhere.

/assign @tcnghia mattmoor


tcnghia (approved):
/lgtm

Issue #1894, opened 2020-11-04
Add `knative_revision` tags to knative provided reconcile metrics


rusde:
/area monitoring

/kind feature

## Expected Behavior
Lot of controller metrics knative provides does not contains all the labels `knative_revision` [schema](https://github.com/knative/pkg/blob/master/metrics/metricskey/constants_serving.go#L43-L51) expects. Targeting following [3 metrics](https://github.com/knative/pkg/blob/master/controller/stats_reporter.go#L35-L37) in this bug

## Actual Behavior
These metrics should contain following tags which are missing when exported 
* namespace
* config
* revision
* service

For these metrics config, revision and service will be unknown. Namespace can be assigned to the namespace last object belongs to. 

In addition we should add following metric keys

* pod_name -> to avoid metric collision when HA is enabled
* resource_name -> resource that was reconciled

## Steps to Reproduce the Problem

1. Export metrics to prometheus and will not see these tags populated.

## Additional Info

None

markusthoemmes:
Why do all metrics need to contain all these labels? Have you considered the increase in cardinality (effectively unbound) and the effects that has on memory allocation for accounting and for prometheus as a monitoring system?

rusde:
How `knative_revision` schema is setup, it expects these tags to be present. We cannot leverage these metrics until these tags are populated. 

We already handle such cardinality data on dataplane side when request metrics has revision name tagged to it. 

I understand this might break prometheus, do you think this proposal can go ahead if this is behind a configmap flag?

rusde:
From the PR

Is it fine if this is behind a config map setting to add these tags?

CM: https://github.com/knative/serving/blob/master/config/core/configmaps/observability.yaml#L25

Saying metrics.controller-add-knative-revision-tags

It will be turned off by default

rusde:
Closing issue as related PR is merged

PR #1916, opened 2020-11-18
Add  tags to knative provided reconcile metrics

rusde:
This is related to Issue https://github.com/knative/pkg/issues/1894

### DIFF
--- a/controller/stats_reporter.go
+++ b/controller/stats_reporter.go
@@ -50,9 +50,6 @@ var (
 	reconcilerTagKey = tag.MustNewKey("reconciler")
 	successTagKey    = tag.MustNewKey("success")
 	NamespaceTagKey  = tag.MustNewKey(metricskey.LabelNamespaceName)
-	ServiceTagKey    = tag.MustNewKey(metricskey.LabelServiceName)
-	ConfigTagKey     = tag.MustNewKey(metricskey.LabelConfigurationName)
-	RevisionTagKey   = tag.MustNewKey(metricskey.LabelRevisionName)
 )
 
 func init() {
@@ -171,17 +168,17 @@ func init() {
 		Description: "Depth of the work queue",
 		Measure:     workQueueDepthStat,
 		Aggregation: view.LastValue(),
-		TagKeys:     []tag.Key{reconcilerTagKey, NamespaceTagKey, ServiceTagKey, ConfigTagKey, RevisionTagKey},
+		TagKeys:     []tag.Key{reconcilerTagKey},
 	}, {
 		Description: "Number of reconcile operations",
 		Measure:     reconcileCountStat,
 		Aggregation: view.Count(),
-		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey, ServiceTagKey, ConfigTagKey, RevisionTagKey},
+		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey},
 	}, {
 		Description: "Latency of reconcile operations",
 		Measure:     reconcileLatencyStat,
 		Aggregation: reconcileDistribution,
-		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey, ServiceTagKey, ConfigTagKey, RevisionTagKey},
+		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey},
 	}}
 	views = append(views, wp.DefaultViews()...)
 	views = append(views, rp.DefaultViews()...)
@@ -215,11 +212,7 @@ func NewStatsReporter(reconciler string) (StatsReporter, error) {
 	// Reconciler tag is static. Create a context containing that and cache it.
 	ctx, err := tag.New(
 		context.Background(),
-		tag.Insert(reconcilerTagKey, reconciler),
-		tag.Insert(NamespaceTagKey, metricskey.ValueUnknown),
-		tag.Insert(ServiceTagKey, metricskey.ValueUnknown),
-		tag.Insert(ConfigTagKey, metricskey.ValueUnknown),
-		tag.Insert(RevisionTagKey, metricskey.ValueUnknown))
+		tag.Insert(reconcilerTagKey, reconciler))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


markusthoemmes:
This explodes the cardinality of these metrics doesn't it? We had huge issues with memory pressure due to this and I think in general prometheus wants you to keep the labels on the metrics relatively bounded.

rusde:
Hmm without these tags these metrics does not have enough information to be in sync with `knative_revision` metric schema. https://github.com/knative/pkg/blob/master/metrics/metricskey/constants_serving.go#L43-L51

Without that we are not able to ingest these metrics into our system.

vagababov (commented):
Resource name — is the most expensive, but it kind of doubles on Service/Config/Revision (it is one of those probably). Also Route is missing.

I wonder if we can export those not as tags, basically keys, but just added metadata (probably not, though).

rusde:
Is it fine if this is behind a config map setting to add these tags?

CM: https://github.com/knative/serving/blob/master/config/core/configmaps/observability.yaml#L25

Saying `metrics.controller-add-knative-revision-tags`

vagababov:
This is `pkg` so it is used outside of Knative Serving proper, so the flag should be more generic `enable-detailed-flags`, or along those lines (I am notoriously bad with naming).

Now, question is also how this is going to be consumed downstream, e.g. do you have in mind how activator or autoscaler will receive this flag, and attach the additional tags. A POC would be super useful.

Generally speaking, I like more dimensions than fewer, but we have to balance against system capabilities as well.

evankanderson:
I chatted a bit with Rushabh; my $0.02 is that it probably makes sense to include a namespace in the controller metrics (so that you can determine if a particular customer is stuck), but it may not make sense to add resource tags to every controller metric, since a lot of monitoring systems do struggle with high cardinality, and these metrics are likely to be infrequently updated.

evankanderson:
@vagababov and @markusthoemmes 

If we only add namespace, do you think we can avoid needing to make this a toggle? I worry about adding configuration knobs that need to be set one way or another to be able to get useful information, but I also don't want to introduce large amounts of memory growth for no reason.

vagababov:
Namespace should be alright I think.
I think @julz might be more interested, since they run multitenant systems, so might have many more namespaces (so both more useful and more resources :))

julz:
I _think_ we'd be ok with namespaces since they're 1 per customer (more granular than that would indeed be bad I think, see https://github.com/knative/serving/issues/8609 for previous context). Reserve the right to plead to add a toggle after all if I turn out to be wrong, though 😅. 

@cdlliuy does adding just namespace sound reasonable to you, too?

rusde:
Updated the PR by remove resource_name tag. PTAL

cdlliuy:
@julz , thank you for bringing me into the discussion.  
I do agree that the detail metrics are valuable to knative users, and I do agree as well the huge amount of metrics will bring more pressure to metric systems.   It depends on the different user case of knative.  so there should be some trade-off, . 

So,  personally,  I support the idea to make the key configurable by the knative operator  to define the level of metric that we would like to retrieve, i.e. 
```
metrics.controller-add-tags: "" (as default)
metrics.controller-add-tags: "namespace" 
metrics.controller-add-tags: "namespace, services"
metrics.controller-add-tags: "namespace, service, configuration"
metrics.controller-add-tags: "namespace, service, configuration, revision"
```   
That means the pkg level need to open a func to initialize the key dynamically, then in serving side , we can call the metric init according to the config map values. 

And, on the other side,  the metric "reconcile_latency" is a histogram, then we can think of the metric type as well to see whether it really need to be a histogram. 
```
# HELP controller_reconcile_latency Latency of reconcile operations
# TYPE controller_reconcile_latency histogram
```
If we only expose `"reconcile_latency"` metric without any namespace, services and etc tags, definitely, it need to be a histogram , so that we can understand the latency behaves in the large scope. 

When adding namespace as the tag,  maybe histogram is still making sense. 

But if it is service, configuration, revision level,  whether it is still worth to be as a histogram,  considering the count of reconcile on the detail level is not big enough .. maybe providing an avg value as gauge is good enough ..

rusde:
Thanks alot for all these insights. 

As per I read, adding just namespace as additional field should be a good addition to these metrics. Other tags are added to the metrics to be in par with `knative_revision` metric schema expectation and they are set as always `unknown`. 

Do we feel configuration proposed by @cdlliuy is something we want for such changes to metrics? If yes I will go with proposal above and try to execute on this. Thanks

vagababov:
So we set unknown in order to clump them together by default but perhaps initialize with precise values downstream at the call site?

rusde:
removed all unknown tags. PTAL

controller/stats_reporter.go
```diff
@@ -171,17 +168,17 @@ func init() {
...
 		Measure:     workQueueDepthStat,
 		Aggregation: view.LastValue(),
-		TagKeys:     []tag.Key{reconcilerTagKey, NamespaceTagKey, ServiceTagKey, ConfigTagKey, RevisionTagKey},
+		TagKeys:     []tag.Key{reconcilerTagKey},
 	}, {
 		Description: "Number of reconcile operations",
 		Measure:     reconcileCountStat,
 		Aggregation: view.Count(),
-		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey, ServiceTagKey, ConfigTagKey, RevisionTagKey},
+		TagKeys:     []tag.Key{reconcilerTagKey, successTagKey, NamespaceTagKey},
```
yanweiguo, line 176: If we don't define other keys (service, config, revision) here, what's your plan to insert them when reporting? IIUC, this PR is to support these metrics for stackdriver. If we use `knative_revision` as resource type, the metrics must have those resource labels.

rusde: After some testing discovered stackdriver is adding this tags by default if not present

https://github.com/knative/pkg/blob/master/metrics/stackdriver_exporter.go#L263

controller/testing/fake_stats_reporter.go
```diff
@@ -43,7 +45,7 @@ func (r *FakeStatsReporter) ReportQueueDepth(v int64) error {
 }
 
 // ReportReconcile records the call and returns success.
-func (r *FakeStatsReporter) ReportReconcile(duration time.Duration, success string) error {
+func (r *FakeStatsReporter) ReportReconcile(duration time.Duration, success string, key types.NamespacedName) error {
```
vagababov, line 48:
```suggestion
func (r *FakeStatsReporter) ReportReconcile(duration time.Duration, success string, _ types.NamespacedName) error {
```

vagababov (commented):
/lgtm

yanweiguo:
ah, i though it was in `pkg/metrics`. Actually I don't have permission to approve.

/hold cancel. will let @evankanderson to approve

controller/stats_reporter.go
evankanderson: Absolutely, it should be up to the various importing engines which have specific schema needs to adapt the schema to their needs, rather than requiring that Knative contort the metrics export code to match the intersection of schema needs as more vendors integrate with different metrics backends.

evankanderson (commented):
/lgtm
/approve

PR #1918, opened 2020-11-19
Add webhookname from env

arghya88:
Signed-off-by: Arghya Sadhu <arghya88@gmail.com>

Ref: https://github.com/knative/eventing/issues/4530

This is currently used here
https://github.com/knative/eventing/blob/52597d174b52a680756b52b377c81bfb96fb636b/pkg/logconfig/config.go#L37

Moving it to pkg to be used across all knative components

### DIFF
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -24,6 +24,9 @@ import (
 
 const portEnvKey = "WEBHOOK_PORT"
 
+// Webhook is the name of the override key used inside of the logging config for Webhook Controller.
+const webhookNameEnv = "WEBHOOK_NAME"
+
 // PortFromEnv returns the webhook port set by portEnvKey, or default port if env var is not set.
 func PortFromEnv(defaultPort int) int {
 	if os.Getenv(portEnvKey) == "" {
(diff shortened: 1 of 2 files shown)


webhook/env.go
```diff
@@ -37,3 +40,17 @@ func PortFromEnv(defaultPort int) int {
 	}
 	return port
 }
+
+func NameFromEnv() string {
+	if webhook := os.Getenv(webhookNameEnv); webhook != "" {
+		return webhook
+	}
+
+	panic(fmt.Sprintf(`The environment variable %q is not set
```
vagababov, line 49: Why would this be required? Can't it be sensibly defaulted?

arghya88: As it's being done for the `PortFromEnv` ? or you mean just set a hardcoded default value if anything is not set in env?

mattmoor: @vagababov I think requiring it is better than defaulting it.  If we default it for two components in a namespace, then they will end up with the same lease names and have a bad time.  I helped the Tekton folks debug this with pipelines and triggers collocated recently.

We should probably change the panic message to indicate that this should be unique for the webhooks in a namespace.

webhook/env.go
```diff
@@ -37,3 +40,17 @@ func PortFromEnv(defaultPort int) int {
...
+If this is a process running on Kubernetes, then it should be using the downward
+API to initialize this variable via:
```
mattmoor, line 51: This isn't the downward API 😂 

I think this is carryover from `system.Namespace()` 🤦

arghya88: Done some changes.PTAL

mattmoor (approved):
/lgtm

PR #1923, opened 2020-11-23
Fix logging message when creating a Prometheus exporter

lionelvillard:
The logging message is quite confusing since there is an opencensus exporter.


vagababov (commented):
/lgtm
/approve

PR #1924, opened 2020-11-24
Provide the information about why did the connection fail

vagababov:
All the juicy details are in that http response,
the error is basically meaningless

/assign mattmoor @markusthoemmes


tcnghia (approved):
/lgtm

Issue #1786, opened 2020-10-08
Gracefully apply multiple PodSpec Bindings to a single subject resource


scothis:
## Expected Behavior

When multiple PodSpec-able Bindings target the same subject resource, each binding should be applied by the webhook when the subject is created/updated.

## Actual Behavior

Only one of the Bindings is applied to the subject via the webhook. The other Bindings are applied via the controller. While the subject will reach a level state, there is unnecessary churn. For example, extra Revisions are created for a subject Configuration. Depending on what other controllers are in the system, there could be a reconciliation cycle thrashing the API Server.

## Additional Info

Refs https://github.com/vmware-labs/service-bindings/issues/78

n3wscott:
Note the `Undo` method will also be destructive.

scothis:
> Note the `Undo` method will also be destructive.

`Undo` isn't inherently destructive, it depends on what's being undone.

glyn:
Clarification of intended behaviour:

1. For a given subject, if some Bindings match exactly and others inexactly, should they all be applied to the subject? Or should the presence of an exact match cause inexact matches to be ignored?

2. For a given subject, if no Bindings match exactly and more than one Binding matches inexactly, should all the inexactly matching Bindings be applied to the subject? (The answer seems obvious, but I want to be certain since it's a change from the current behaviour.)

3. What should happen when more than one Binding targets the same subject, but the Bindings' `Do` methods make incompatible changes (such as setting an env var to distinct values)?

Clarification of current behaviour:

1. Is there an implicit precondition on a Binding's `Do` method that the changes it wants to make must not already be present in the subject? If not, `Undo` could corrupt the subject instead of undoing the effects of `Do` (i.e. not changing the subject).

scothis:
Good questions @glyn 

> 1. For a given subject, if some Bindings match exactly and others inexactly, should they all be applied to the subject? Or should the presence of an exact match cause inexact matches to be ignored?
>
> 2. For a given subject, if no Bindings match exactly and more than one Binding matches inexactly, should all the inexactly matching Bindings be applied to the subject? (The answer seems obvious, but I want to be certain since it's a change from the current behaviour.)

Every binding referencing the subject should be applied. So all exact matches **and** all inexact matches. This is the current behavior of the controller, the webhook's behavior should match.

> 3. What should happen when more than one Binding targets the same subject, but the Bindings' Do methods make incompatible changes (such as setting an env var to distinct values)?
>
[...]

PR #1925, opened 2020-11-24
Apply multiple bindings to the same subject

glyn:
Fixes https://github.com/knative/pkg/issues/1786.

cc @mattmoor @n3wscott

### DIFF
--- a/webhook/psbinding/index.go
+++ b/webhook/psbinding/index.go
@@ -7,7 +7,7 @@ You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless requ ired by applicable law or agreed to in writing, software
+Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
@@ -32,17 +32,16 @@ type exactKey struct {
 
 // exactMatcher is our reverse index from subjects to the Bindings that apply to
 // them.
-type exactMatcher map[exactKey]Bindable
+type exactMatcher map[exactKey][]Bindable
 
-// add writes a key into the reverse index.
+// add associates a Binding with a key in the reverse index.
 func (em exactMatcher) add(key exactKey, b Bindable) {
-	em[key] = b
+	em[key] = append(em[key], b)
 }
 
 // get fetches the key from the reverse index, if present.
-func (em exactMatcher) get(key exactKey) (bindable Bindable, present bool) {
-	b, ok := em[key]
-	return b, ok
+func (em exactMatcher) get(key exactKey) []Bindable {
+	return em[key]
 }
 
 // inexactKey is the type for keys that match inexactly (via selector)
@@ -73,15 +72,16 @@ func (im inexactMatcher) add(key inexactKey, selector labels.Selector, b Bindabl
 }
 
 // get fetches the key from the reverse index, if present.
-func (im inexactMatcher) get(key inexactKey, ls labels.Set) (bindable Bindable, present bool) {
-	// Iterate over the list of pairs matched for this GK + namespace and return the first
-	// Bindable that matches our selector.
+func (im inexactMatcher) get(key inexactKey, ls labels.Set) []Bindable {
+	found := []Bindable{}
+	// Iterate over the list of pairs matched for this GK + namespace and collect the
+	// Bindables that matches our selector.
 	for _, p := range im[key] {
 		if p.selector.Matches(ls) {
-			return p.sb, true
+			found = append(found, p.sb)
 		}
 	}
-	return nil, false
+	return found
 }
 
 // index is a collection of Bindables indexed by their subject resources.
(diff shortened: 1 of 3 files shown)


mattmoor:
This generally seems reasonable to me.

webhook/psbinding/index.go
```diff
@@ -79,3 +83,76 @@ func (im inexactMatcher) Get(key inexactKey, ls labels.Set) (bindable Bindable,
...
+func (idx *index) setIndex(exact exactMatcher, inexact inexactMatcher) {
+	idx.lock.Lock()
+	defer idx.lock.Unlock()
+	idx.exact = exact
+	idx.inexact = inexact
+}
+
+// LookUp returns the Bindables associated with a resource with the given group, kind, namespace, name, and labels.
+// TODO: find all the matching Bindables instead of just one.
+func (idx *index) LookUp(group string, kind string, namespace string, name string, labels labels.Set) []Bindable {
```
vagababov, line 134:
```suggestion
func (idx *index) LookUp(group, kind, namespace, name string, labels labels.Set) []Bindable {
```

webhook/psbinding/index.go
```diff
@@ -79,3 +83,76 @@ func (im inexactMatcher) Get(key inexactKey, ls labels.Set) (bindable Bindable,
...
+// index is a collection of Bindables indexed by their subject resources.
+type index struct {
+	// lock protects access to exact and inexact
+	lock    sync.RWMutex
+	exact   exactMatcher
+	inexact inexactMatcher
+}
+
+// IndexBuilder allows an index to be built atomically
+type IndexBuilder struct {
```
vagababov, line 96: this and the next function seem to be local, so no need to export.

webhook/psbinding/index.go
glyn: @vagababov Agreed. What I actually did was group the first four parameters into an `exactKey` which makes the call site more readable.

webhook/psbinding/index.go
glyn: @vagababov Yes. My reasoning was that the matcher methods in `index.go` were exported and so I tried to adopt the same convention. (A lint complains about unexported `indexBuilder` being returned from `NewIndexBuilder`, so exporting the latter leads to exporting the former).

Maybe the best convention is to just unexport everything in `index.go`.

glyn: @vagababov Fixed. Please see the second commit.

vagababov: yeah, my point was that `NewIndexBuilder` should not be exported either 😄

webhook/psbinding/index.go
```diff
@@ -79,3 +83,71 @@ func (im inexactMatcher) Get(key inexactKey, ls labels.Set) (bindable Bindable,
...
+	// Next look for inexact matches.
+	if sb, ok := idx.inexact.get(inexactKey{
+		Group:     key.Group,
+		Kind:      key.Kind,
+		Namespace: key.Namespace,
+	}, labels); ok {
+		return []Bindable{sb}
+	}
+
+	return []Bindable{}
```
vagababov, line 152: just nil here?

vagababov (commented):
Looks good.
Thanks.

webhook/psbinding/index.go
glyn: Sure. That's addressed now.

webhook/psbinding/index.go
glyn: Fixed. Clearly I've been writing too much Rust code...

glyn:
@vagababov Hi. The PR is functionally complete. I'm going to beef up the exact and inexact reverse index tests. Meanwhile, PTAL.

glyn:
Note: the coverage of `psbinding.go` has decreased only because code was factored out and the uncovered lines are now a larger proportion of the total.

glyn:
@scothis PTAL and see if this fixes the downstream problem.

glyn:
>  I'm going to beef up the exact and inexact reverse index tests.

Done.

webhook/psbinding/psbinding.go
```diff
@@ -197,56 +193,48 @@ func (ac *Reconciler) Admit(ctx context.Context, request *admissionv1.AdmissionR
...
-	}()
-	if fb == nil {
+	// Look up the Bindables for this resource.
+	fbs := ac.index.lookUp(exactKey{
+		Group:     request.Kind.Group,
+		Kind:      request.Kind.Kind,
+		Namespace: request.Namespace,
+		Name:      orig.Name,
+	},
+		labels.Set(orig.Labels))
```
scothis, line 203: this formatting is a bit odd

scothis (commented):
Tested this PR with the [vmware-labs/service-bindings](https://github.com/vmware-labs/service-bindings) repo, and now multiple bindings apply to a Job on creation as expected. Thanks @glyn

/lgtm

webhook/psbinding/psbinding.go
glyn: Agreed. Note that the same oddity occurred in the deleted code just above. Thanks `go fmt`.

vagababov: I'd join with the previous line to avoid this.

glyn: Fixed two occurrence. Looks a little better now.

glyn: Fixed.

glyn: Fixed.

Issue #1927, opened 2020-11-24
Add controller metrics as allowed stackdriver metrics


rusde:
/area monitoring

/kind feature

## Expected Behavior
Export below metrics as known metrics via stackdriver.

```
        workQueueDepthStat   = stats.Int64("work_queue_depth", "Depth of the work queue", stats.UnitNone)
	reconcileCountStat   = stats.Int64("reconcile_count", "Number of reconcile operations", stats.UnitNone)
	reconcileLatencyStat = stats.Int64("reconcile_latency", "Latency of reconcile operations", stats.UnitMilliseconds)
```

## Actual Behavior

They are exported as custom metrics today.

PR #1928, opened 2020-11-24
Adding controller metrics as supported stackdriver serving metrics

rusde:
Related to https://github.com/knative/pkg/issues/1927


yanweiguo:
/hold

Stackdriver doesn't support these metrics yet.

PR #1930, opened 2020-11-27
feat/perf-test: continue test when fail to get commit id

zhongduo:
changeset.Get does not work well when the test image is in the
vendor directory because there is no kodata setup. After this
commit, failure to get the commit id will result in a
log entry instead of failing the whole test.

### DIFF
--- a/test/mako/sidecar.go
+++ b/test/mako/sidecar.go
@@ -93,7 +93,7 @@ func SetupHelper(ctx context.Context, benchmarkKey *string, benchmarkName *strin
 	// Get the commit of the benchmarks
 	commitID, err := changeset.Get()
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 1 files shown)


test/mako/sidecar.go
```diff
@@ -93,7 +93,7 @@ func SetupHelper(ctx context.Context, benchmarkKey *string, benchmarkName *strin
 	// Get the commit of the benchmarks
 	commitID, err := changeset.Get()
 	if err != nil {
-		return nil, err
+		log.Print("Cannot find commit ID")
```
pierDipi, line 96:
```suggestion
		log.Println("Cannot find commit ID")
```

pierDipi (approved):

pierDipi (approved):
/lgtm

PR #1931, opened 2020-11-27
Add a parallel test to upgrade test framework to assert no race conditions are present

cardil:
This is just to showcase that parallel steps in new upgrade test framework do work.

### DIFF
--- a/test/upgrade/steps_in_parallel_test.go
+++ b/test/upgrade/steps_in_parallel_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


cardil:
Of course, it produces a race... I didn't bother to write it with channels, as this is just to showcase that operations in single steps can by running in parallel. That should be enough, as other types of parallelism seams unwanted.

/cc @mgencur

markusthoemmes:
Still required?

cardil:
Yeah. Maybe, I'll add this test to the framework, so the support would be explicit, and fixing the race in the same time.

/remove-lifecycle stale

markusthoemmes:
What's the goal of this PR? Is this supposed to land at some point? What's missing?

PR #1939, opened 2020-12-04
Log if the deserialized line is semi-normal.

vagababov:
Right now they panic and we lose all the context of what logged semi-correct line

/assign @tcnghia @mattmoor


tcnghia (approved):
/lgtm

PR #1940, opened 2020-12-04
add LabelSelector to InformerFactory so that the informers only list and watch resources with the given labels

qibobo:
In knative/serving issue https://github.com/knative/serving/issues/10098#issuecomment-738553604, when the knative autoscaler watches too many non-knative resources, it causes huge memory usage. This PR is to fix the issue.
After this PR is finished,  I will create another PR in knative/serving repo to pick up this change in autoscaler. The code change will be like: https://github.com/qibobo/serving/commit/05c2523f4d303261942795c50f7fdadb08340f6b


julz (commented):
/ok-to-test

.. but these changes are to generated files, looks like the actual code that generates them is in cmd/injection-gen/generators/factory.go

qibobo:
@julz updated the code gen part

client/injection/apiextensions/informers/factory/factory.go
```diff
@@ -35,12 +36,24 @@ func init() {
...
+	opts := make([]externalversions.SharedInformerOption, 0, 2)
 	if injection.HasNamespaceScope(ctx) {
 		opts = append(opts, externalversions.WithNamespace(injection.GetNamespaceScope(ctx)))
 	}
+	untyped := ctx.Value(LabelKey{})
+	if untyped != nil {
+		label := untyped.(string)
+		opts = append(opts, externalversions.WithTweakListOptions(func(l *v1.ListOptions) {
+			if l == nil {
+				l = &v1.ListOptions{}
```
mattmoor, line 52: Can this get nil?  Why not just return?

mattmoor:
I'm a little concerned that this alters the behavior of the entire shared informer factory 🤔 

In the context of the autoscaler, this is probably fine, but we watch Pods in the serving controller as well (where this would apply to Deployments, Services, and other resources), and Tekton is watching this and would be similarly affected.

I also don't like that the same filter is applies to the entire shared informer.  If we could distinguish between: unfiltered / filtered, and distinct filters, then I think I'd be happy. 

Perhaps we should introduce a new `/filtered` subpackage (like we have `/fake`) that enables folks to spin up informer factories for different filters?

What I'm thinking (out loud) is something like this in the binary entrypoint:

```go
ctx = WithFilters(ctx, l1, l2, l3)
```
Each of `l1`, `l2`, and `l3` would gets its own shared informer factory.

Then in the controllers folks would use:

```go
podInformer := filteredpodinformer.Get(ctx, l1) // Get our pod informer from the shared informer factory with the "l1" filter.
```

mattmoor:
Also, this is subtle but useful: There's a clear error path when folks try to `Get` a filter that wasn't set up in shared main.

cdlliuy:
@mattmoor ,  as explained in https://github.com/knative/serving/issues/10098#issuecomment-738553604,  we tried to give the option to use / not use tweakOption by the label selector assigned in context.    
```
In this PR, the label selector can be defined in context.
If no specific label selector, then the same shardInformerFactory is returned as before.
If a label selector is defined, then the informer generated by this factory will be tweak per the designed label.
```

As shown in https://github.com/knative/serving/pull/10266,  we only add label in autoscaler main.go, so only autosclaer is affected (with pod informer, metric informer) , but the other informer won't be changed. 

But I agree with you that we can create a filterdSharedInformerFactory as well, then we can claim to use the filter factory in podInformer.go as a fine-gained control.  

So,  a separated file for filterdSharedInformerFactory is recommended, right ?

mattmoor:
Yeah, I get what's going on and what you are going for.  What I worry about is changing the existing stuff to include a filter expression in a way that doesn't force consumers of that existing stuff to Ack the filter.

I think it should be relatively straightforward to code-generate filtered versions that include the filter expressions in the create/get contract.  This should be only slightly more complex outside of the code-gen, and allow us to colocate informers on the same resource with different filters.

I get that the autoscaler is the main place hitting these problems right now, but I can also see this hitting the main serving controller and I don't want to have to rework this because that has different needs.

Generally I'm pretty excited to see us land this!

qibobo:
@mattmoor 
There are 2 commits, the first is to generated filter-informer-factory and the second is to generate the filtered-informer based on the first one.
The changes has been tested in https://github.com/knative/serving/pull/10266

client/injection/apiextensions/informers/factory/filtered/filteredFactory.go
```diff
@@ -0,0 +1,69 @@
...
+		label := untyped.(string)
+		opts = append(opts, externalversions.WithTweakListOptions(func(l *v1.ListOptions) {
+			if l == nil {
+				l = &v1.ListOptions{}
+			}
+			l.LabelSelector = label
+		}))
+	}
+	return context.WithValue(ctx, Key{},
+		externalversions.NewSharedInformerFactoryWithOptions(c, controller.GetResyncPeriod(ctx), opts...))
```
mattmoor, line 58: LabelKey should give you a list of label selectors, and we should attach a shared informer factory for each.

client/injection/apiextensions/informers/factory/filtered/filteredFactory.go
```diff
@@ -0,0 +1,69 @@
...
+	injection "knative.dev/pkg/injection"
+	logging "knative.dev/pkg/logging"
+)
+
+func init() {
+	injection.Default.RegisterInformerFactory(withInformerFactory)
+}
+
+// Key is used as the key for associating information with a context.Context.
+type Key struct{}
```
mattmoor, line 37: Key should include the selector string

client/injection/apiextensions/informers/factory/filtered/filteredFactory.go
```diff
@@ -0,0 +1,69 @@
...
+			l.LabelSelector = label
+		}))
+	}
+	return context.WithValue(ctx, Key{},
+		externalversions.NewSharedInformerFactoryWithOptions(c, controller.GetResyncPeriod(ctx), opts...))
+}
+
+// Get extracts the InformerFactory from the context.
+func Get(ctx context.Context) externalversions.SharedInformerFactory {
+	untyped := ctx.Value(Key{})
```
mattmoor, line 63: Get should take the selector string we're getting for, and pass it with Key.

client/injection/apiextensions/informers/factory/filtered/filteredFactory.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func init() {
+	injection.Default.RegisterInformerFactory(withInformerFactory)
+}
+
+// Key is used as the key for associating information with a context.Context.
+type Key struct{}
+
+type LabelKey struct{}
```
mattmoor, line 39: This package should provide an accessor for setting this up.

client/injection/apiextensions/informers/apiextensions/v1/customresourcedefinition/filtered/customresourcedefinition.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Key is used for associating the Informer inside the context.Context.
+type Key struct{}
+
+func withInformer(ctx context.Context) (context.Context, controller.Informer) {
+	f := filtered.Get(ctx)
+	inf := f.Apiextensions().V1().CustomResourceDefinitions()
+	return context.WithValue(ctx, Key{}, inf), inf.Informer()
+}
```
mattmoor, line 42: We should be associating a filtered informer for each of the filters applied to the shared informer factory.

client/injection/apiextensions/informers/apiextensions/v1/customresourcedefinition/filtered/customresourcedefinition.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Get extracts the typed informer from the context.
+func Get(ctx context.Context) v1.CustomResourceDefinitionInformer {
+	untyped := ctx.Value(Key{})
```
mattmoor, line 46: This should take the selector and access the appropriate informer.

mattmoor (commented):
Left some comments on the generated code, as that seems a bit easier to reason about.

qibobo:
@mattmoor 
Thank you for your comments and help. I have updated the PR according to our discussion.

qibobo:
@julz @markusthoemmes @vagababov 
Any additional comments?

injection/interface.go
```diff
@@ -53,9 +53,16 @@ type Interface interface {
 	// a new informer with a context.
 	RegisterInformer(InformerInjector)
 
+	// RegisterInformers registers a new filtered informer injector callback for associating
```
julz, line 56:
```suggestion
	// RegisterFilteredInformer registers a new filtered informer injector callback for associating
```

injection/interface.go
```diff
@@ -53,9 +53,16 @@ type Interface interface {
...
+	// a new set of informers with a context.
+	RegisterFilteredInformer(FilteredInformersInjector)
```
julz, line 58: should this method be plural to match the argument it takes?

```suggestion
	RegisterFilteredInformers(FilteredInformersInjector)
```

injection/interface.go
```diff
@@ -53,9 +53,16 @@ type Interface interface {
...
+
 	// GetInformers fetches all of the registered informer injectors.
 	GetInformers() []InformerInjector
 
+	// GetInformers fetches all of the registered filtered informer injectors.
```
julz, line 63:
```suggestion
	// GetFilteredInformers fetches all of the registered filtered informer injectors.
```

injection/interface.go
```diff
@@ -84,8 +91,9 @@ var (
...
 	m sync.RWMutex
 
-	clients   []ClientInjector
-	factories []InformerFactoryInjector
-	informers []InformerInjector
-	ducks     []DuckFactoryInjector
+	clients           []ClientInjector
+	factories         []InformerFactoryInjector
+	informers         []InformerInjector
+	filteredinformers []FilteredInformersInjector
```
julz, line 97: bit of a nit, but I think this should probably be camelCase:
```suggestion
	filteredInformers []FilteredInformersInjector
```

injection/interface.go
julz: ..though having said that, I'm wondering if we actually want FilteredInformersInjector to be plural anyway. Could it match RegisterInformer(InformerInjector) and be singular (and inject a single filteredinformer at a time), or is there a reason it has to differ?

qibobo: @julz Thank you for your comments. The plural informers comes from this comment: https://github.com/knative/pkg/pull/1940#discussion_r541150349
And pair work with @mattmoor to have this design.

qibobo:
@julz fixed.

qibobo:
@mattmoor @julz 
Could you help review the latest change?

client/injection/apiextensions/informers/factory/filtered/fake/fakeFilteredFactory.go
```diff
@@ -0,0 +1,62 @@
...
+	if untyped == nil {
+		logging.FromContext(ctx).Panic(
+			"Unable to fetch labelkey from context.")
+	}
+	labelSelectors := untyped.([]string)
+	for _, selector := range labelSelectors {
+		thisOpts := append(opts, externalversions.WithTweakListOptions(func(l *v1.ListOptions) {
+			if l == nil {
+				l = &v1.ListOptions{}
+			}
```
mattmoor, line 55: Why do we need this?

codegen/cmd/injection-gen/generators/fakefilteredfactory.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2019 The Knative Authors
```
mattmoor, line 2: 2021

injection/informers_test.go
```diff
@@ -61,9 +65,11 @@ func TestRegisterInformersAndSetup(t *testing.T) {
 	i.RegisterInformer(injectFooInformer)
 	i.RegisterInformer(injectBarInformer)
 
+	i.RegisterFilteredInformers(injectFooFilteredInformer)
+
 	_, infs := i.SetupInformers(context.Background(), &rest.Config{})
 
-	if want, got := 2, len(infs); got != want {
+	if want, got := 4, len(infs); got != want {
```
mattmoor, line 72: Can you explain the math here in a comment?

mattmoor:
Can we use `foo_bar.go` style naming for consistency?

client/injection/apiextensions/informers/factory/filtered/fake/fakeFilteredFactory.go
qibobo: @mattmoor Fixed.
But one question is that given l is a pointer, why do not we need to check whether it is nil?

codegen/cmd/injection-gen/generators/fakefilteredfactory.go
qibobo: Fixed.

injection/informers_test.go
qibobo: Fixed.

qibobo:
> Can we use `foo_bar.go` style naming for consistency?

Also renamed some existed files to foo_bar.go style.

mattmoor:
I still see the old files here 🤔

client/injection/apiextensions/informers/factory/filtered/fake/fakeFilteredFactory.go
mattmoor: Depends on the api contract.  Technically yes, but it is likely just passed as a pointer for mutability.

ZhuangYuZY:
@qibobo could you please check the latest comment ?  We will need make this PR merged. Thank you.

qibobo:
@mattmoor For the informers, all existed files are not foo_bar.go, so leave them.

mattmoor (commented):
/ok-to-test
/lgtm
/approve

PR #1942, opened 2020-12-07
Add dependabot configuration

pierDipi:
# Changes

- Add dependabot configuration

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,6 @@ updates:
     target-branch: "master"
     schedule:
       interval: "daily"
-      
     ignore:
       # Ignore updates to packages that start 'k8s'
       # Wildcards match zero or more arbitrary characters
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,4 +9,4 @@ updates:
     ignore:
       # Ignore updates to packages that start 'k8s'
       # Wildcards match zero or more arbitrary characters
-      - dependency-name: "k8s*"
+      - dependency-name: "k8s.io*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+
+version: 2
+updates:
+
+  - package-ecosystem: "gomod"
+    vendor: true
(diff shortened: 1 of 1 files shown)


markusthoemmes:
Thanks for trying this out! We'll have to make sure this doesn't butcher our K8s deps though. Your experiments showed that it currently does (see https://github.com/pierDipi/pkg/pull/4). Can we ignore k8s dep bumps?

pierDipi:
https://github.com/pierDipi/pkg/pull/4#issuecomment-739856702

markusthoemmes (approved):
I like, don't see a reason not to try this out!

I'll hold off on a LGTM to get input from more peeps. Thanks for playing with this!

pierDipi:
Probably this https://github.com/pierDipi/pkg/blob/db193b096f34d415da6e3e3fefb6ac231ee0bc4b/hack/update-deps.sh#L52-L53, won't work on verify because it includes OWNERS files: https://github.com/pierDipi/pkg/pull/5/files#diff-70721b9094f3050454588463842e68eef3e12218fa7bd22455c5948052ad4614R1.

pierDipi:
Does having OWNERS in a vendor dir break Prow?

markusthoemmes:
Oh yeah, yes it does :(. We'll have to run a script after updating go.mod.

markusthoemmes:
/hold

/assign @mattmoor @n3wscott 

See above, we currently can't have nice things like dependabot because of our additional massaging of `go mod vendor`. I wonder if we should consider leaving it alone instead :thinking:.

mattmoor:
There are enough gotchas that I'm reluctant to enable it wholesale.  Top of mind:
1. License compliance
2. Carrying patches
3. Removing symlinks

Most of the rest I've seen feel cosmetic.  Could we instead run it as an action with some of our extra cleanup logic as a post-step?

pierDipi:
@mattmoor we can run an action on a PR created by Dependabot and commit the diff.
In that case, the PR would have 1 commit from Dependabot and 1 commit from Knobots.
Unclear how that plays with the CLA bot.

PR #1944, opened 2020-12-07
Default to the lowest pre-release version to be the minimum

markusthoemmes:
As per title.

Several users have hit this inconvience already (pre-release versions are by definition smaller than "normal" versions) and we've added a more verbose log message in the past to accomodate.

I can't think of a situation where you would not want to actually allow a pre-release version of a specific Major.Minor.Patch level, if you do not specify a very specific pre-build number, so this always assumes "-0" if no other pre-release version is specified.

/assign @vagababov

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -61,6 +61,12 @@ func CheckMinimumVersion(versioner discovery.ServerVersionInterface) error {
 		return err
 	}
 
+	// If no specific pre-release requirement is set, we default to "-0" to always allow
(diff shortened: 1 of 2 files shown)


mattmoor (commented):
/lgtm

thanks, based on the tests I'm surprised we ever wanted to fail.  🤦

vagababov:
Well this is debatable, since pre-release version _is_ smaller than the release version. But I guess the possible failure space there is very small.

PR #1949, opened 2020-12-08
Add smart handling of selectors in webhooks

### LINKED ISSUE #1943: Trouble due to "arbitrary" control-plane label exclusion in Webhooks
Through #1590 we added an arbitrary control-plane label exclusion to fix support with AKS. This is causing headaches for at least one applicatione downstream in our case, which happens to rely on the same label. Bummer.

I think we should fix this "properly" as the current workaround can break anytime somebody is mucking with webhook COs anyway.

markusthoemmes:
# Changes

This is an alternative fix for #1590. Instead of arbitrarily adding a label from a different project to avoid the reconcilers racing, this adds "smart" handling of the selectors in that labels not inside the knative.dev domain are plainly ignored and our own selectors are added additively.

Fixes #1943

**Release Note**

```release-note
Fixed an issue where our webhooks arbitrarily ignored namespaces labeled with `control-plane`.
```

/assign @mattmoor @n3wscott @dprotaso

### DIFF
--- a/webhook/psbinding/psbinding.go
+++ b/webhook/psbinding/psbinding.go
@@ -349,7 +349,7 @@ func (ac *Reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
 		if cur.ClientConfig.Service == nil {
 			return fmt.Errorf("missing service reference for webhook: %s", wh.Name)
 		}
-		current.Webhooks[i].ClientConfig.Service.Path = ptr.String(ac.Path())
+		cur.ClientConfig.Service.Path = ptr.String(ac.Path())
 	}
 
 	if ok := equality.Semantic.DeepEqual(configuredWebhook, current); !ok {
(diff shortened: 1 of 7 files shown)


webhook/psbinding/psbinding.go
```diff
@@ -338,31 +328,34 @@ func (ac *Reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
...
+		cur.MatchPolicy = &matchPolicy
+		cur.Rules = rules
+		cur.NamespaceSelector = selector
+		cur.ObjectSelector = selector // 1.15+ only
+		cur.ClientConfig.CABundle = caCert
+		if cur.ClientConfig.Service == nil {
 			return fmt.Errorf("missing service reference for webhook: %s", wh.Name)
 		}
-		webhook.Webhooks[i].ClientConfig.Service.Path = ptr.String(ac.Path())
+		current.Webhooks[i].ClientConfig.Service.Path = ptr.String(ac.Path())
```
mattmoor, line 352: why not use `cur` here?

mattmoor (commented):
/lgtm

mattmoor:
/hold

For nit

webhook/psbinding/psbinding.go
markusthoemmes: prolly forgot, fixing.

markusthoemmes:
/unhold

Done!

Issue #1948, opened 2020-12-08
Support customizing the profiling port


mattmoor:
/area monitoring
/kind good-first-issue

## Expected Behavior

Users should be able to customize the profiling port.

## Actual Behavior

The profiling port is hardcoded as 8008.

PR #1950, opened 2020-12-08
Make the profiling port configurable via envvar.

mattmoor:
With this folks can override the default 8008 with PROFILING_PORT.

Fixes: https://github.com/knative/pkg/issues/1948


mattmoor:
Looks like serving is using the exported constant 🤔

mattmoor:
/unhold

Ok I reverted the breaking part that constant folded the strconv.

vagababov (commented):
/lgtm

PR #1951, opened 2020-12-08
Add Release Note for Logging Format Change

ian-mi:
**Release Note**

```release-note
The default logging format has changed: https://github.com/knative/pkg/pull/1773. Knative structured logs will now use the keys timestamp, message, and severity instead of ts, msg, and level. Action required: to keep the old behavior configure these keys in the logging configmap.
```


matzew:
/close

there is no diff on this PR

PR #1953, opened 2020-12-09
Expose the statefulset constructor

mattmoor:
I want to be able to use this directly (in place of `WithDynamic`) from the autoscaler, since it wants the bucket set, which `WithDynamic` builds but doesn't return 😞


vagababov (commented):
/lgtm

PR #1957, opened 2020-12-11
Refactor and fix metrics export tests.

### LINKED ISSUE #1672: TestMetricsExport is super flaky

evankanderson:
# Changes

- :bug: Fix long-standing in e2e tests for OpenCensus. Ran 200 times without error; previously would see at least 1 error in 60-80 runs.
  - It turns out the "that could work" part of "do the simplest thing that could work" is important... there was a race collecting results off the channel, and we didn't read all channels to completion. /sigh
- :broom: Move e2e tests to a separate file from more unit-testy `resource_view_test` files.
- :bug: Moving tests between files exposed a bug in `TestFlushExporter`. Fixed that, too.
- 🎁 Made the "none" exporter a real exporter to allow disabling all metric exports.

/kind bug
/kind cleanup

Fixes #1672 

/assign @vagababov 

It took embarassingly long to find this... I had to walk away twice for at least a week after bashing my head on the export code to figure out what was really going on. (The bug was inside the test all along!)


evankanderson:
If splitting `resource_view_test` is too annoying in this PR, I can merge things back into a single file. Looking at https://github.com/knative/pkg/pull/1957/commits/3e094f8e3b2f531159852229ede93d8f519319a9 should show a reasonable diff.

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+				return err
+			}
+			t.Log("Created exporter at", ocFake.address)
+			return UpdateExporter(context.Background(), configForBackend(openCensus), logtesting.TestLogger(t))
+		},
+		validate: func(t *testing.T) {
+			// We unregister the views because this is one of two ways to flush
+			// the internal aggregation buffers; the other is to have the
+			// internal reporting period duration tick, which is at least
+			// [new duration] in the future.
```
yanweiguo, line 243: Are the comments here still correct?

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+							if record.Resource != nil {
+								labels = record.Resource.Labels
+							}
+							metric := metricExtract{
+								Name:   m.MetricDescriptor.Name,
+								Labels: labels,
+								Value:  m.Timeseries[0].Points[0].GetInt64Value(),
+							}
+							records = append(records, metric)
+							keys[metric.Key()] = struct{}{}
```
yanweiguo, line 266: Could you add a comment that why using a set here fixes the problem?

yanweiguo (commented):
Is the failing downstream test relative?

metrics/e2e_test.go
evankanderson: Whoops, no. We have a function for that now.

metrics/e2e_test.go
evankanderson: Done. I was convinced for a long time that the RPCs weren't actually going to the right place, but I finally figured out that we simply weren't reading enough off the channel to find them.

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+
+func (m metricExtract) Key() string {
+	return fmt.Sprintf("%s<%s>", m.Name, resource.EncodeLabels(m.Labels))
+}
+
+func (m metricExtract) String() string {
+	return fmt.Sprintf("%s:%d", m.Key(), m.Value)
+}
+
+func initSdFake(sdFake *stackDriverFake) error {
```
vagababov, line 77: I guess?
```suggestion
func initSDFake(sdFake *stackDriverFake) error {
```

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+			},
+		},
+		{
+			Type: "revision",
+			Labels: map[string]string{
+				"project":  "p1",
+				"revision": "r2",
+			},
+		},
+	}
```
vagababov, line 159:
```suggestion
	resources := []*resource.Resource{{
			Type: "revision",
			Labels: map[string]string{
				"project":  "p1",
				"revision": "r1",
			},
		},{
			Type: "revision",
			Labels: map[string]string{
				"project":  "p1",
				"revision": "r2",
			},
		}}
```

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+			resp, err := http.Get(fmt.Sprintf("http://localhost:%d/metrics", prometheusPort))
+			if err != nil {
+				t.Fatalf("failed to fetch prometheus metrics: %+v", err)
+			}
+			defer resp.Body.Close()
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				t.Fatalf("failed to read prometheus response: %+v", err)
+			}
+			want := `# HELP testComponent_global_export_counts Count of exports via standard OpenCensus view.
```
vagababov, line 215:
```suggestion
			const want = `# HELP testComponent_global_export_counts Count of exports via standard OpenCensus view.
```

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+				label2,
+				2,
+			},
+			{
+				"custom.googleapis.com/knative.dev/autoscaler/not_ready_pods",
+				batchLabels,
+				3,
+			},
+		},
+	}, {
```
vagababov, line 406:
```suggestion
		expected: []metricExtract{{
				"knative.dev/serving/autoscaler/actual_pods",
				label1,
				1,
			},{
				"knative.dev/serving/autoscaler/desired_pods",
				label2,
				2,
			},{
				"custom.googleapis.com/knative.dev/autoscaler/not_ready_pods",
				batchLabels,
				3,
			}},
	}, {
```

metrics/e2e_test.go
```diff
@@ -0,0 +1,629 @@
...
+		name:               "Don't allow custom metrics",
+		allowCustomMetrics: "false",
+		expected: []metricExtract{
```
vagababov, line 409: ditto

vagababov (commented):
LG in general
Left some stylistic comments.

metrics/e2e_test.go
evankanderson: Done. I have no idea, the product name is "Stackdriver", so I just spelled it out the 3 places it is used.

metrics/e2e_test.go
evankanderson: Done.

metrics/e2e_test.go
evankanderson: Done.

metrics/e2e_test.go
evankanderson: Done

metrics/e2e_test.go
evankanderson: Done

yanweiguo:
LG but the failed unit test indicates there may be another problem.

evankanderson:
> LG but the failed unit test indicates there may be another problem.

I've seen prometheus fail to be able to listen on the port; I can choose a random port to see if that helps. Probably the other fix would be to add a SO_REUSEADDR to the prometheus server setup; let me see if that helps.

evankanderson:
I'm not sure that `SO_REUSEADDR` helps; I added that code but was still able to get a timeout in the prometheus poll-for-ready loop.

evankanderson:
Regardless, the Prometheus code has not been changed, so I don't think that should be a barrier on this PR. (Though I'd love to figure out why the server sometimes won't respond for > 10s).

metrics/exporter.go
```diff
@@ -161,13 +162,13 @@ func UpdateExporter(ctx context.Context, ops ExporterOptions, logger *zap.Sugare
 		flushGivenExporter(curMetricsExporter)
 		e, f, err := newMetricsExporter(newConfig, logger)
 		if err != nil {
-			logger.Errorw("Failed to update a new metrics exporter based on metric config", newConfig, zap.Error(err))
+			logger.Errorw("Failed to update a new metrics exporter based on metric config", "config", newConfig, "error", err)
```
vagababov, line 165: zap.String("config", newConfig), zap.Error(err)? Same below?

evankanderson: Ugh... I want to replace this with a non-sugared logger. It looks like [`With`](https://godoc.org/go.uber.org/zap#SugaredLogger.With) takes *both* Field objects and key-value pairs. Unfortunately, this ended up with the key-value format where the value was a `zap.Field`.

WDYT?

vagababov: Well you do use sugared.
You don't need `with`, just `logger.Error("...", zap.Error(err))` on the desugared one

evankanderson: I switched this to put all the `Field` objects first so it's not possible to log things as "string" => Field, which was what the old code was doing.

Using `logger.Error()` means taking the error out of a separate JSON field, which makes it harder (for example) to filter all the logs for just the ones with errors in them.

vagababov: Yeah, I know. That's why when I joined I was surprised we;re using this crap (zap that is) :)
But that's the zap philosophy — use jq to parse :)

vagababov (commented):
/lgtm

PR #1964, opened 2020-12-15
Fix excessive allocation in the resource metrics

markusthoemmes:
# Changes

Through a bit of cleanup work in Knative Serving, it got kinda apparent that our metric recording calls are kinda expensive. I triaged it down to the `resourceToKey` being quite expensive.

Discussing with @evankanderson, we came up with this quick fix to make the allocations of the functions much more reasonable without coming up with a different way of handling the keys here altogether.

## Benchmark results

```
benchmark                               old ns/op     new ns/op     delta
BenchmarkResourceToKey/0-labels-16      139           2.38          -98.29%
BenchmarkResourceToKey/1-labels-16      677           168           -75.18%
BenchmarkResourceToKey/5-labels-16      2344          802           -65.78%
BenchmarkResourceToKey/10-labels-16     4625          1645          -64.43%

benchmark                               old allocs     new allocs     delta
BenchmarkResourceToKey/0-labels-16      2              0              -100.00%
BenchmarkResourceToKey/1-labels-16      6              1              -83.33%
BenchmarkResourceToKey/5-labels-16      18             3              -83.33%
BenchmarkResourceToKey/10-labels-16     33             3              -90.91%

benchmark                               old bytes     new bytes     delta
BenchmarkResourceToKey/0-labels-16      40            0             -100.00%
BenchmarkResourceToKey/1-labels-16      128           32            -75.00%
BenchmarkResourceToKey/5-labels-16      432           192           -55.56%
BenchmarkResourceToKey/10-labels-16     800           320           -60.00%
```

/assign @evankanderson

### DIFF
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -322,7 +322,7 @@ func resourceToKey(r *resource.Resource) string {
 	}
 
 	var s strings.Builder
-	writeKV := func(key, value string) {
+	writeKV := func(key, value string) { // This lambda doesn't force an allocation.
 		// We use byte values 1 and 2 to avoid colliding with valid resource labels
 		// and to make unpacking easy
 		s.WriteByte('\x01')
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -333,16 +333,12 @@ func resourceToKey(r *resource.Resource) string {
 
 	// If there's only one label, we can skip building and sorting a slice of keys.
 	if len(r.Labels) == 1 {
-		l := len(r.Type)
-		var key string
 		for k, v := range r.Labels {
-			key = k
-			l += len(k) + len(v) + 2
+			// We know this only executes once.
+			s.Grow(len(r.Type) + len(k) + len(v) + 2)
+			s.WriteString(r.Type)
+			writeKV(k, v)
 		}
-		s.Grow(l)
-
-		s.WriteString(r.Type)
-		writeKV(key, r.Labels[key])
 		return s.String()
 	}
 
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -316,25 +316,48 @@ func resourceToKey(r *resource.Resource) string {
 		return ""
 	}
 
+	// If there are no labels, we have just the Type.
+	if len(r.Labels) == 0 {
+		return r.Type
+	}
+
+	var s strings.Builder
+	writeKV := func(key, value string) {
+		// We use byte values 1 and 2 to avoid colliding with valid resource labels
+		// and to make unpacking easy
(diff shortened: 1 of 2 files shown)


metrics/resource_view.go
```diff
@@ -315,21 +315,49 @@ func resourceToKey(r *resource.Resource) string {
...
+
 	var s strings.Builder
+	writeKV := func(key, value string) {
+		// We use byte values 1 and 2 to avoid colliding with valid resource labels
+		// and to make unpacking easy
+		s.WriteByte('\x01')
+		s.WriteString(key)
+		s.WriteByte('\x02')
+		s.WriteString(value)
+	}
```
vagababov, line 332: not sure if at this point `func writeKV(Builder, string, string)` isn't better, or definitely more readable.
Shouldn't it save an allocation, since otherwise it'll dynamically create the lambda to bind it to  `s`?

metrics/resource_view.go
```diff
@@ -315,21 +315,49 @@ func resourceToKey(r *resource.Resource) string {
...
+		l := len(r.Type)
+		var key string
+		for k, v := range r.Labels {
+			key = k
+			l += len(k) + len(v) + 2
+		}
+		s.Grow(l)
+
+		s.WriteString(r.Type)
+		writeKV(key, r.Labels[key])
```
vagababov, line 345: bind `v` to `val` to avoid lookup?
Or, which will even avoid creating both stack variables just put everything in the loop.
We know it will execute only once. Even `l` will be redundant (s.Grow(len(type)+len(k)+len(v)+2))

metrics/resource_view.go
markusthoemmes: The lambda doesn't escape apparently. Moving it out works too though.

metrics/resource_view.go
markusthoemmes: Good one, will do.

metrics/resource_view.go
```diff
@@ -315,21 +315,45 @@ func resourceToKey(r *resource.Resource) string {
...
+
+	// If there's only one label, we can skip building and sorting a slice of keys.
+	if len(r.Labels) == 1 {
+		for k, v := range r.Labels {
+			// We know this only executes once.
+			s.Grow(len(r.Type) + len(k) + len(v) + 2)
+			s.WriteString(r.Type)
+			writeKV(k, v)
+		}
+		return s.String()
```
vagababov, line 342: or even just `return` after `writeKV`, but up to you.

vagababov (commented):
/lgtm

vagababov (commented):
/lgtm

PR #1965, opened 2020-12-15
Reproduce inconsistent Prometheus metric

antoineco:
This is an attempt to reproduce knative/eventing#4645 using a test, mostly for discussion, therefore marking as draft+hold.

So far I haven't been successful, although the issue happens 100% of the time in production after a variable amount of time.


antoineco:
Closing in favor of knative/eventing#4662

PR #1966, opened 2020-12-15
Remove useless logger creation

markusthoemmes:
This call is redundant as the method is already put onto the logger a few lines above.

/assign @mattmoor @vagababov


mattmoor (approved):
/lgtm

Issue #1968, opened 2020-12-16
Review usage of logger to avoid unnecessary allocations


markusthoemmes:
I've done a bit of a cleanup the other day and noticed that our logger allocates way more at times then you would expect. The biggest offender for us is `logger.With(...).Info` etc. where the `With` call unnecessarily clones and creates a new logger with the respective fields added. Note, that the `With` call comes before even asking for a loglevel, so this would cause allocations even if the respective loglevel is not enabled.

I've whipped up a quick benchmark to measure the exact overhead:

```golang
func BenchmarkZap(b *testing.B) {
	l, err := zap.NewProduction()
	if err != nil {
		b.Fatal(err)
	}
	logger := l.Sugar()

	b.Run("debug-with", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			logger.With(zap.String("foo", "bar")).
				Debugf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
					"foo", 10, 20)
		}
	})

	b.Run("debug-check-level", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			desugared := logger.Desugar()
			if desugared.Core().Enabled(zapcore.DebugLevel) {
				desugared.Debug(fmt.Sprintf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
					"foo", 10, 20), zap.String("foo", "bar"))
			}
		}
	})

	b.Run("info-with", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			logger.With(zap.String("foo", "bar")).
				Infof("Adding to the slow queue %s (depth(total/slow): %d/%d)",
					"foo", 10, 20)
		}
	})

	b.Run("info-sprintf", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			logger.Infow(fmt.Sprintf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
				"foo", 10, 20), zap.String("foo", "bar"))
		}
	})
}
```

Results:

```
goos: linux
goarch: amd64
pkg: knative.dev/pkg/websocket
BenchmarkZap/debug-with-16         	  803221	      1455 ns/op	    1450 B/op	       9 allocs/op
BenchmarkZap/debug-check-level-16  102735457	      11.4 ns/op	       0 B/op	       0 allocs/op
BenchmarkZap/info-with-16          	  509606	      2277 ns/op	    1519 B/op	      10 allocs/op
[...]

PR #1969, opened 2020-12-16
Remove unnecessary intermediate loggers

markusthoemmes:
Ref #1968 

This removes all occurrences where we unnecessarily allocate an intermediate logger rather than using the respective logger directly. Debug statements are additionally guarded as that level is disabled in production most likely.

/assign @mattmoor @vagababov

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -267,9 +267,8 @@ func (c *Impl) EnqueueAfter(obj interface{}, after time.Duration) {
 func (c *Impl) EnqueueSlowKey(key types.NamespacedName) {
 	c.workQueue.SlowLane().Add(key)
 
-	desugared := c.logger.Desugar()
-	if desugared.Core().Enabled(zapcore.DebugLevel) {
-		desugared.Debug(fmt.Sprintf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
+	if logger := c.logger.Desugar(); logger.Core().Enabled(zapcore.DebugLevel) {
+		logger.Debug(fmt.Sprintf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
 			safeKey(key), c.workQueue.Len(), c.workQueue.SlowLane().Len()),
 			zap.String(logkey.Key, key.String()))
 	}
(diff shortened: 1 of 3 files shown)


websocket/connection.go
```diff
@@ -165,7 +166,7 @@ func NewDurableConnection(target string, messageChan chan []byte, logger *zap.Su
 				}
 				logger.Debug("Connected to ", target)
 				if err := c.keepalive(); err != nil {
-					logger.With(zap.Error(err)).Errorf("Connection to %s broke down, reconnecting...", target)
+					logger.Errorw(fmt.Sprintf("Connection to %s broke down, reconnecting...", target), zap.Error(err))
```
antoineco, line 169: nit, saves one string parsing if I'm not mistaken (although to be fair I've never any difference comparing both in benchmarks).
```suggestion
					logger.Errorw(fmt.Sprint("Connection to ", target, " broke down, reconnecting..."), zap.Error(err))
```

controller/controller.go
```diff
@@ -409,8 +418,12 @@ func (c *Impl) MaybeEnqueueBucketKey(bkt reconciler.Bucket, key types.Namespaced
...
 func (c *Impl) EnqueueKeyAfter(key types.NamespacedName, delay time.Duration) {
 	c.workQueue.AddAfter(key, delay)
-	c.logger.With(zap.String(logkey.Key, key.String())).
-		Debugf("Adding to queue %s (delay: %v, depth: %d)", safeKey(key), delay, c.workQueue.Len())
+
+	desugared := c.logger.Desugar()
+	if desugared.Core().Enabled(zapcore.DebugLevel) {
+		desugared.Debug(fmt.Sprintf("Adding to queue %s (delay: %v, depth: %d)", safeKey(key), delay, c.workQueue.Len()),
+			zap.String(logkey.Key, key.String()))
+	}
```
antoineco, line 426: Do we need to assign this to a new variable when we could just use `c.logger` (sugared) inside the block?

websocket/connection.go
markusthoemmes: Yeaaaaah, well, I'll stick the current version then, if there's no significant diff anyway.

antoineco: The reason I commented is because @vagababov consistently asked me to not use `Sprintf` whenever possible in the past, and he's assigned to this PR.

controller/controller.go
antoineco: Makes sense. You might want to assign the var in the `if` context if you don't use it anywhere else.
```suggestion
	if desugared := c.logger.Desugar(); desugared.Core().Enabled(zapcore.DebugLevel) {
		desugared.Debug(fmt.Sprintf("Adding to queue %s (delay: %v, depth: %d)", safeKey(key), delay, c.workQueue.Len()),
			zap.String(logkey.Key, key.String()))
	}
```

websocket/connection.go
markusthoemmes: That is indeed a different convo and our linter currently only catches trailing transforms.

I do think we should disable that linter if it does not result in a measurable performance benefit (which AFAIK it doesn't, at least not in the cases I checked) as it just makes code-writing (especially logging) feel more foreign then it should.

I'm all for good code-style etc. but IMO this is just overly verbose (and I have to remember which functions add spaces between the arguments and which don't).

antoineco (approved):
/lgtm

PR #1971, opened 2020-12-17
Unify usage of YAML libraries

markusthoemmes:
As per title. `sigs.k8s.io/yaml` is a permanent fork of `ghodss/yaml` and maintained by the K8s community, so it seems to make sense for us to use that one too.

/assign @mattmoor @vagababov


mattmoor (commented):
But I was getting so good at typing `ghodss` without typos... 😂 

/lgtm

🎉

PR #1974, opened 2020-12-23
Introduce a nostackdriver build tag

mattmoor:
This teaches `depcheck` to support passing build flags through, and uses it to ensure that with `-tags=nostackdriver` that the expensive stack driver libraries are not linked.

I'm starting in `./tracing` since it is considerably simpler than the mess in metrics, but my intention is to do something similar in `./metrics` next.


vagababov (commented):
/lgtm

PR #1977, opened 2020-12-28
Drainer supports customized kubelet probe check

grac3gao:
# Changes

- Besides of directly returning "200 OK", drainer accepts customized kubelet probe check to perform when the handler is not in Drain status.

### DIFF
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -38,7 +38,8 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
-type CustomizeChecker func(w http.ResponseWriter, req *http.Request)
+// CheckOption enables configuration of an optional checker.
+type CheckOption func(w http.ResponseWriter, req *http.Request)
 
 // This constructor is overridden in tests to control the progress
 // of time in the test.
@@ -49,9 +50,8 @@ var newTimer = func(d time.Duration) timer {
 }
 
 // Drainer wraps an inner http.Handler to support responding to kubelet
-// probes and KProbes until the handler is told to Drain.
-// If the handler is not in Drain status, kubelet probe will get "200 OK" or
-// status from CustomizeCheck. KProbes will get "200 OK".
+// probes and KProbes with a "200 OK" until the handler is told to Drain,
+// or Drainer will optionally run the callback checker if CheckOption is defined.
 // When the Drainer is told to Drain, it will immediately start to fail
 // probes with a "500 shutting down", and the call will block until no
 // requests have been received for QuietPeriod (defaults to
@@ -60,9 +60,9 @@ type Drainer struct {
 	// Mutex guards the initialization and resets of the timer
 	sync.RWMutex
 
-	// CustomizeCheck is the optional checker
+	// CheckOption is the optional checker
 	// for kubelet probe to perform when the handler is not in Drain status.
-	CustomizeCheck CustomizeChecker
+	CheckOption CheckOption
 
 	// Inner is the http.Handler to which we delegate actual requests.
 	Inner http.Handler
@@ -86,11 +86,9 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if network.IsKubeletProbe(r) { // Respond to probes regardless of path.
 		if d.draining() {
 			http.Error(w, "shutting down", http.StatusServiceUnavailable)
+		} else if d.CheckOption != nil {
+			d.CheckOption(w, r)
 		} else {
-			if d.CustomizeCheck != nil {
-				d.CustomizeCheck(w, r)
-				return
-			}
 			w.WriteHeader(http.StatusOK)
 		}
 		return
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -38,6 +38,8 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
+type CustomizeChecker func(w http.ResponseWriter, req *http.Request)
+
 // This constructor is overridden in tests to control the progress
 // of time in the test.
 var newTimer = func(d time.Duration) timer {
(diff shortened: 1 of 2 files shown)


network/handlers/drain.go
```diff
@@ -38,6 +38,8 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
+type CustomizeChecker func(w http.ResponseWriter, req *http.Request)
```
vagababov, line 41: `CheckOption` is a more classical name in our codebase.

network/handlers/drain.go
```diff
@@ -47,7 +49,9 @@ var newTimer = func(d time.Duration) timer {
 }
 
 // Drainer wraps an inner http.Handler to support responding to kubelet
-// probes and KProbes with a "200 OK" until the handler is told to Drain.
+// probes and KProbes until the handler is told to Drain.
+// If the handler is not in Drain status, kubelet probe will get "200 OK" or
```
vagababov, line 53: It also seems to be basically a generic additional callback check, i.e. you don't have to restrict it to the kubelet check or anything else, fwiw.

vagababov (commented):
I've just got nits
/assign mattmoor

network/handlers/drain.go
grac3gao: done

network/handlers/drain.go
grac3gao: done

network/handlers/drain.go
```diff
@@ -38,6 +38,8 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
+type CheckOption func(w http.ResponseWriter, req *http.Request)
```
vagababov, line 41: Add a comment to the public type.

network/handlers/drain.go
```diff
@@ -79,6 +86,10 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		if d.draining() {
 			http.Error(w, "shutting down", http.StatusServiceUnavailable)
 		} else {
+			if d.CheckOption != nil {
+				d.CheckOption(w, r)
+				return
+			}
```
vagababov, line 92:
```suggestion
		} else  if d.CheckOption != nil {
				d.CheckOption(w, r)
	         } else {
	                         w.WriteHeader(http.StatusOK)
	         }         
```

vagababov (commented):
A few nits, but I'll leave to Matt to approve.

/lgtm

network/handlers/drain.go
grac3gao: done

network/handlers/drain.go
grac3gao: done

grac3gao:
TestMetricsExport/OpenCensus failed, which seems has nothing to do with this change.

network/handlers/drain.go
grac3gao: It seems after changing to this, Unit test TestMetricsExport/OpenCensus starts to fail and downstream unit test in serving fails as well.

vagababov:
may be you still need the return after you report failure.

vagababov:
Nah, that doesn't seem right....

network/handlers/drain.go
grac3gao: After retesting, all tests passed. Not sure if the failure is just a coincidence. I'll rerun the test for more times to verify.

network/handlers/drain.go
```diff
@@ -38,6 +38,9 @@ func (s *sysTimer) tickChan() <-chan time.Time {
 	return s.Timer.C
 }
 
+// CheckOption enables configuration of an optional checker.
+type CheckOption func(w http.ResponseWriter, req *http.Request)
```
mattmoor, line 42: This is `http.HandlerFunc`, so not strictly necessary.

network/handlers/drain.go
```diff
@@ -56,6 +60,10 @@ type Drainer struct {
 	// Mutex guards the initialization and resets of the timer
 	sync.RWMutex
 
+	// CheckOption is the optional checker
+	// for kubelet probe to perform when the handler is not in Drain status.
+	CheckOption CheckOption
```
mattmoor, line 65:
```suggestion
	// HealthCheck is an optional health check that is performed until the drain signal is received.
	// When unspecified, a "200 OK" is returned, otherwise this function is invoked.
	HealthCheck http.HandlerFunc
```

network/handlers/drain.go
mattmoor: https://godoc.org/net/http#HandlerFunc

grac3gao: Thanks for pointing out, removed.

network/handlers/drain.go
grac3gao: done

vagababov:
I kind of liked custom name, but 🤷

network/handlers/drain.go
grac3gao: Personally, I am in favor of custom name too. Do you mind if I change back to `CheckOption` and keep the comment you suggested?

mattmoor: I don't understand the value we're getting from the type alias?  inlined, I know it's a well known handler interface I need to implement.  outlined I need to look it up before doing the same.

We're not defining our own methods on the type alias, or "constants" we expect to be passed as values.

mattmoor (commented):
/lgtm
/approve
/hold

Holding for discussion around the type alias

vagababov:
I am not going to fight for it. For me the alias was preferable, since tomorrow if we decide to change the signature (e.g. to add context) it's just signature and call-sites, now it's more work. 
But, all in all, it's soft subjective preference that I am not fighting for, so don't hold on me :)

network/handlers/drain.go
grac3gao: Thanks, that sounds reasonable to me. I'll keep the current change you suggested, and remove the hold.

PR #1984, opened 2021-01-12
Harden logstream against non-object events

markusthoemmes:
# Changes

The `Watch` can return events that do not deserialize into a pod object, for example if errors occur. Currently this [fails tests](https://prow.knative.dev/view/gcs/knative-prow/logs/ci-knative-serving-continuous/1348721427838144512).

This makes that part of the code not crash the test but rather log that an unexpected event happened, so if somebody uses the logs for debugging, they know that lines might be missing.

/kind bug

/assign @vagababov @mattmoor

### DIFF
--- a/test/logstream/v2/stream.go
+++ b/test/logstream/v2/stream.go
@@ -102,7 +102,13 @@ func (s *namespaceSource) watchPods() error {
 					if ev.Object == nil || reflect.ValueOf(ev.Object).IsNil() {
 						continue
 					}
-					p := ev.Object.(*corev1.Pod)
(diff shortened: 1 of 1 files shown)


dprotaso:
Looks like when it's a status it's an error with the watch - I wonder it means we should restart the watch

https://github.com/kubernetes/apimachinery/blob/6c16abd7175829b60c6be7384d8a2dd05e6f34a8/pkg/watch/watch.go#L60-L69

mattmoor (commented):
/lgtm

PR #1985, opened 2021-01-12
Make Deployment OwnerRefable

lionelvillard:
# Changes

- :gift: Small utility to make deployment OwnerRefable
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/kmeta/owner_references.go
+++ b/kmeta/owner_references.go
@@ -38,15 +38,13 @@ func NewControllerRef(obj OwnerRefable) *metav1.OwnerReference {
 	return metav1.NewControllerRef(obj.GetObjectMeta(), obj.GetGroupVersionKind())
 }
 
-type ownerRefableDeployment struct {
-	*appsv1.Deployment
-}
+type ownerRefableDeployment appsv1.Deployment
 
-func (o *ownerRefableDeployment) GetGroupVersionKind() schema.GroupVersionKind {
+func (o ownerRefableDeployment) GetGroupVersionKind() schema.GroupVersionKind {
 	return appsv1.SchemeGroupVersion.WithKind("Deployment")
 }
 
 // DeploymentAsOwnerRefable returns a owner refable deployment
 func DeploymentAsOwnerRefable(d *appsv1.Deployment) OwnerRefable {
-	return &ownerRefableDeployment{d}
+	return (*ownerRefableDeployment)(d)
 }
--- a/kmeta/owner_references.go
+++ b/kmeta/owner_references.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kmeta
 
 import (
+	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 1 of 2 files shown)


kmeta/owner_references_test.go
```diff
@@ -62,3 +63,28 @@ func TestNewControllerRef(t *testing.T) {
...
+func TestOwnerRefableDeployment(t *testing.T) {
+	d := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "foo",
+			UID:  "42",
+		},
+	}
+	blockOwnerDeletion := true
+	isController := true
```
vagababov, line 76: `ptr.Bool()` for both at point of invocation.

kmeta/owner_references.go
```diff
@@ -36,3 +37,16 @@ type OwnerRefable interface {
...
 	return metav1.NewControllerRef(obj.GetObjectMeta(), obj.GetGroupVersionKind())
 }
+
+type ownerRefableDeployment struct {
+	*appsv1.Deployment
+}
+
+func (o *ownerRefableDeployment) GetGroupVersionKind() schema.GroupVersionKind {
+	return appsv1.SchemeGroupVersion.WithKind("Deployment")
+}
```
mattmoor, line 47: This could just be a type alias instead of a field, but I would like to understand the motivation for this.  Do you have a method that takes `kmeta.OwnerRefable`?  How are you planning to use this?

lionelvillard: +1 for alias. Yes I'm planning to use it in eventing for mt adapters owned by controllers. 

I can leave it in eventing for now if you prefer and once the PR (yet to be open) gets accepted and merged I can move it here. WDYT?

mattmoor (commented):
Can't reply to the other thing (b/c github), but I'm wondering how much going through OwnerRefable here actually saves us vs. just constructing the owner ref directly or exposing a similar method that takes Deployment itself (probably in eventing).

lionelvillard:
> Can't reply to the other thing (b/c github), but I'm wondering how much going through OwnerRefable here actually saves us vs. just constructing the owner ref directly or exposing a similar method that takes Deployment itself (probably in eventing).

It helps pushing the construction of the owner ref down to the utility function. It's also nice to treat all resources a OwnerRefable, not just the Knative ones.

kmeta/owner_references_test.go
```diff
@@ -46,19 +49,40 @@ func TestNewControllerRef(t *testing.T) {
...
+			UID:  "42",
+		},
+	}
+	want := &metav1.OwnerReference{
+		APIVersion:         "apps/v1",
+		Kind:               "Deployment",
+		Name:               "foo",
+		UID:                "42",
+		BlockOwnerDeletion: ptr.Bool(true),
+		Controller:         ptr.Bool(true),
```
n3wscott, line 81: the deployment is the controller of whatever you are making? This seems strange to me at first read, likely you are not reconciling deployments and I will bet you want more than one owner on the child object? 

What is the usecase?

lionelvillard: https://github.com/knative/eventing/pull/4731/files#diff-3d78ac5bf206c11ebf4dae96f827bdb6a30b4d50a03c12a1bd65ee1af73782c2R167

markusthoemmes:
Any news on this one?

PR #1987, opened 2021-01-14
Close websocket connection in parallel to read/write calls

markusthoemmes:
# Changes

This speeds up the closure of websocket connections by not waiting for (potentially long) i/o operations to finish and rather closes the connection immediately, potentially in parallel to these operations.

/kind enhancement

/assign @vagababov @julz

### DIFF
--- a/websocket/connection.go
+++ b/websocket/connection.go
@@ -63,8 +63,9 @@ type ManagedConnection struct {
 	connection        rawConnection
 	connectionFactory func() (rawConnection, error)
 
-	closeChan chan struct{}
-	closeOnce sync.Once
+	closeChan   chan struct{}
+	closeOnce   sync.Once
+	shutdownErr error
 
 	establishChan chan struct{}
 	establishOnce sync.Once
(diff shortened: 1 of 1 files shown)


websocket/connection.go
```diff
@@ -268,15 +268,19 @@ func (c *ManagedConnection) keepalive() error {
...
-
-	if c.connection != nil {
-		err := c.connection.Close()
-		c.connection = nil
-		return err
+	// Close the connection under a read-lock to abort in-flight i/o operations and not
+	// having to wait for a read timeout to happen.
+	c.connectionLock.RLock()
+	if c.connection == nil {
+		return nil
```
julz, line 275: Is it safe not to call RUnlock before this?

markusthoemmes: ugh, of course it isn't. Need :coffee:

websocket/connection.go
```diff
@@ -268,15 +268,19 @@ func (c *ManagedConnection) keepalive() error {
...
 	}
-	return nil
+	err := c.connection.Close()
+	c.connectionLock.RUnlock()
+
+	c.connectionLock.Lock()
```
julz, line 281: Trying to think whether there's a potential race here: is it possible between the RUnlock and the Lock someone could reconnect? if so I think we'd null out connection without closing it

julz: actually maybe it's slightly worse.. right after the RUnlock the connection is closed but not nil, so we could try to use it in other methods which check if c.connection == nil under lock

markusthoemmes: Yeah I've been thinking along similar lines. Will recheck

markusthoemmes:
Discussed with @julz, while this is prolly fine now, it changes the semantics of Shutdown in maybe unexpected ways.

PR #1988, opened 2021-01-14
Use special error type to designate skips.

mattmoor:
This change introduces a new `controller.NewSkipKey` method to designate certain reconciliations as "skipped".

The primary motivation for this is to squelch useless logging on non-leader replicas, which currently report success with trivial latency.

I have plumbed this through existing reconcilers and the code-gen so most things downstream should get this for free.  In places where a key is observed, I do not mark the reconcile as skipped as the reconciler did some processing for which the awareness of side-effects and reported latency may be interesting.

### DIFF
--- a/webhook/resourcesemantics/conversion/reconciler.go
+++ b/webhook/resourcesemantics/conversion/reconciler.go
@@ -65,8 +65,7 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 	logger := logging.FromContext(ctx)
 
 	if !r.IsLeaderFor(types.NamespacedName{Name: key}) {
-		logger.Debugf("Skipping key %q, not the leader.", key)
-		return nil
+		return controller.NewSkipKey(key)
 	}
 
 	// Look up the webhook secret, and fetch the CA cert bundle.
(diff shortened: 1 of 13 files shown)


webhook/resourcesemantics/conversion/reconciler.go
```diff
@@ -65,8 +65,7 @@ func (r *reconciler) Reconcile(ctx context.Context, key string) error {
 	logger := logging.FromContext(ctx)
 
 	if !r.IsLeaderFor(types.NamespacedName{Name: key}) {
-		logger.Debugf("Skipping key %q, not the leader.", key)
-		return nil
+		return controller.NewSkipKey(key)
```
vagababov, line 68: Keep the log?

vagababov (commented):
/lgtm
/hold
for comment

webhook/resourcesemantics/conversion/reconciler.go
mattmoor: I intentionally dropped these since we use debug logging in a lot of our CI jobs and it is unnecessarily noisy.

mattmoor:
Looks like things should "just work"(tm) downstream. 💪 

/unhold

PR #1989, opened 2021-01-15
New constructor functions for defaulting and validation webhook controllers to customize the work queue name

### LINKED ISSUE #1979: More than one defaulting webhook controller won't work in the same binary with HA enabled
## Expected Behavior
I can set more than one [defaulting webhook controller](https://github.com/knative/pkg/blob/master/webhook/resourcesemantics/defaulting/controller.go#L39) in the same binary with leader election (HA) enabled. Same for configmaps and validation webhook.

## Actual Behavior
If more than one defaulting webhook controllers are set in the same binary with leader election (HA) enabled, only one of the controllers would reconcile as expected.

1. In knative serving webhook [main.go](https://github.com/knative/serving/blob/master/cmd/webhook/main.go#L78), set up another `defaulting.NewAdmissionController` for Istio [MutatingWebhookConfiguration](https://github.com/knative-sandbox/net-istio/blob/master/config/500-mutating-webhook.yaml)
2. Deploy knative serving and the Istio MutatingWebhookConfiguration.
3. Check Istio MutatingWebhookConfiguration and the serving [MutatingWebhookConfiguration](https://github.com/knative/serving/blob/master/config/core/webhooks/defaulting.yaml), you will fine that only one of them gets reconciled.

## Additional Info
This is because for defaulting webhook controllers, they share the same lease. The lease name is `defaultingwebhook`: https://github.com/knative/pkg/blob/master/webhook/resourcesemantics/defaulting/controller.go#L77. However when setting up leader election, each controller uses different resource lock ID: https://github.com/knative/pkg/blob/master/leaderelection/context.go#L114. So only one of them will actually got set up successfully.

In Knative serving autoscaler, KPA controller and metrics controller share the same lease as well. But they have the same resource lock ID: https://github.com/knative/serving/blob/master/cmd/autoscaler/main.go#L269. The callbacks of those two controllers will be triggered when starting leading the lease.

yanweiguo:
# Changes

- Add new constructor functions for defaulting and validation webhook controllers to customize the work queue name, which is used as lease name as well for leader election. Downstream can use these functions to pass a different work queue name as a workaround to solve issue #1979.

Fixes #1979. IMO, setting some labels is also setting default values. I'm not sure why we need a different controller for https://github.com/knative-sandbox/net-istio/blob/master/config/500-mutating-webhook.yaml.

/assign @mattmoor 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/webhook/resourcesemantics/defaulting/controller.go
+++ b/webhook/resourcesemantics/defaulting/controller.go
@@ -47,7 +47,7 @@ func NewAdmissionController(
 	return NewAdmissionControllerWithWorkQueueName(ctx, name, "DefaultingWebhook", path, handlers, wc, disallowUnknownFields)
 }
 
-// NewAdmissionController constructs a reconciler.
+// NewAdmissionControllerWithWorkQueueName constructs a reconciler.
 func NewAdmissionControllerWithWorkQueueName(
 	ctx context.Context,
 	name, wqName, path string,
--- a/webhook/resourcesemantics/defaulting/controller.go
+++ b/webhook/resourcesemantics/defaulting/controller.go
@@ -35,10 +35,22 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics"
 )
 
-// NewAdmissionController constructs a reconciler
+// NewAdmissionController constructs a reconciler with "DefaultingWebhook"
+// as the work queue name.
 func NewAdmissionController(
 	ctx context.Context,
 	name, path string,
+	handlers map[schema.GroupVersionKind]resourcesemantics.GenericCRD,
+	wc func(context.Context) context.Context,
+	disallowUnknownFields bool,
+) *controller.Impl {
+	return NewAdmissionControllerWithWorkQueueName(ctx, name, "DefaultingWebhook", path, handlers, wc, disallowUnknownFields)
+}
+
+// NewAdmissionController constructs a reconciler.
+func NewAdmissionControllerWithWorkQueueName(
+	ctx context.Context,
+	name, wqName, path string,
 	handlers map[schema.GroupVersionKind]resourcesemantics.GenericCRD,
 	wc func(context.Context) context.Context,
 	disallowUnknownFields bool,
@@ -74,7 +86,7 @@ func NewAdmissionController(
 	}
 
 	logger := logging.FromContext(ctx)
-	c := controller.NewImpl(wh, logger, "DefaultingWebhook")
+	c := controller.NewImpl(wh, logger, wqName)
 
 	// Reconcile when the named MutatingWebhookConfiguration changes.
 	mwhInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
--- a/webhook/resourcesemantics/validation/controller.go
+++ b/webhook/resourcesemantics/validation/controller.go
@@ -48,7 +48,7 @@ func NewAdmissionController(
 	return NewAdmissionControllerWithWorkQueueName(ctx, name, "ValidationWebhook", path, handlers, wc, disallowUnknownFields, callbacks...)
 }
 
-// NewAdmissionController constructs a reconciler.
+// NewAdmissionControllerWithWorkQueueName constructs a reconciler.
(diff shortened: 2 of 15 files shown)


webhook/resourcesemantics/validation/controller.go
```diff
@@ -44,6 +45,18 @@ func NewAdmissionController(
 	disallowUnknownFields bool,
 	callbacks ...map[schema.GroupVersionKind]Callback,
 ) *controller.Impl {
+	return NewAdmissionControllerWithWorkQueueName(ctx, name, "ValidationWebhook", path, handlers, wc, disallowUnknownFields, callbacks...)
+}
+
+// NewAdmissionController constructs a reconciler.
```
julz, line 51:
```suggestion
// NewAdmissionControllerWithWorkQueueName constructs a reconciler.
```

julz (commented):
we lint these in serving but looks like we dont in pkg

mattmoor:
/hold

🤔   This isn't what we discussed in slack...  The Istio webhook is an abuse of the defaulting webhook, and with the infrastructure we have these days it really isn't that hard to write your own webhook that would trivially live alongside these.

yanweiguo:
> /hold
> 
>  This isn't what we discussed in slack... The Istio webhook is an abuse of the defaulting webhook, and with the infrastructure we have these days it really isn't that hard to write your own webhook that would trivially live alongside these.

I'm not sure why it is an abuse of the defaulting webhook. If we have a new webhook, its target is to set two labels. Doesn't it belong to setting defaults?

mattmoor:
It's a stretch to call it defaulting when it isn't our resource.  This is absolutely a mutating webhook, which you can write with the same infrastructure without abusing defaulting in this way.

yanweiguo:
> It's a stretch to call it defaulting when it isn't our resource. This is absolutely a mutating webhook, which you can write with the same infrastructure without abusing defaulting in this way.

Does defaulting webhook do anything specific for knative resource that can't be used for other resource? I don't understand what the differences would be to add a new mutating webhook.

mattmoor:
The contract is that is calls `SetDefaults` on all resources of our given type.

I think the most fundamental conceptual departure from "defaulting" is that net-istio is using an `objectSelector` to avoid most resources.  Why?  Deployments are extremely common and we presumable don't want to be on the critical path for all of them, but perhaps most importantly: we typically run the webhook as a Deployment, which creates a dependency cycle 😬 

Generally my rule for hacks is that they are fine if you can keep them relatively contained (as this is today downstream in net-istio), but we're starting to let the hack grow tentacles that creep out of its box.  The fact that the hack is starting to want upstream changes is a strong signal that maybe we should ditch the hack and implement something better suited for this.

webhook/resourcesemantics/defaulting/controller.go
```diff
@@ -74,7 +86,7 @@ func NewAdmissionController(
 	}
 
 	logger := logging.FromContext(ctx)
-	c := controller.NewImpl(wh, logger, "DefaultingWebhook")
```
n3wscott, line 77: why can't the work queue name be `"DefaultingWebhook-"+name` ?

dprotaso: That would work and make webhook match the behaviour of our other controllers

yanweiguo: It would work. FWIW, this is not only used for the queue name but also the lease name for lead election. By changing this, a lease object will be leaked after upgrading an existing cluster. Are we fine with this?

This is for https://github.com/knative/pkg/issues/1979. @n3wscott @dprotaso are you fine with the solution to modify upstream code? Matt prefers solution 3 mentioned in https://github.com/knative/pkg/issues/1979#issuecomment-755814906.

dprotaso: See my comments here:

https://github.com/knative/pkg/issues/1979#issuecomment-797855624

dprotaso:
I'm going to close out this PR since it's been a few months

PR #1993, opened 2021-01-15
drop excess and duplicate logger properties

dprotaso:
Slack thread: https://knative.slack.com/archives/C964EFQAW/p1610738747008200

`zapdriver` package isn't maintained and has some weird quirks causing dupe info in our log output.

thus the following remains the same:
- all keys remain stable as before
- encoding levels are mapped to stack driver levels
- encoding time is RFC3339 nano

following changed (since we're not wrapping the core):
- I changed the duration output to a string that includes units (cc @mattmoor @vagababov)
- de-duped `caller` and sourceLocation
- no more empty `googleapis.com/labels` fields
- [this isn't invoked anymore](https://github.com/blendle/zapdriver/blob/9200777f8a3d4173247db5db22f6463772bb6f20/core.go#L112-L141) 
  - so no labels extraction to the above labels field (we weren't really use it)
  - no service context field attachment

/assign @yanweiguo @ian-mi

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -55,14 +55,15 @@ func NewLogger(configJSON string, levelOverride string, opts ...zap.Option) (*za
 		return enrichLoggerWithCommitID(logger), atomicLevel
 	}
 
-	loggingCfg := zapdriver.NewProductionConfig()
+	loggingCfg := stackdriverConfig()
+
 	if levelOverride != "" {
 		if level, err := levelFromString(levelOverride); err == nil {
 			loggingCfg.Level = zap.NewAtomicLevelAt(*level)
 		}
 	}
 
-	logger, err = loggingCfg.Build(append(opts, zapdriver.WrapCore())...)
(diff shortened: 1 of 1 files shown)


dprotaso:
cc @slinkydeveloper given https://github.com/knative/pkg/pull/1773#issuecomment-745350926

slinkydeveloper:
Sounds like a good change, thanks @dprotaso !

logging/config.go
```diff
@@ -249,3 +251,9 @@ func ConfigToJSON(cfg *Config) (string, error) {
 	})
 	return string(jsonCfg), err
 }
+
+func stackdriverConfig() zap.Config {
```
yanweiguo, line 255: Why is this called `stackdriverConfig`?

dprotaso: `zapdriver` is specific to stack driver - I figured that'd be more explicit than `zapdriverConfig`

PR #2005, opened 2021-02-01
Unregister views to avoid slow oom issue during meter cleanup

skonto:
# Changes

- Unregisters views that cause part of https://github.com/knative/serving/issues/10516, details here: https://github.com/knative/serving/issues/10516#issuecomment-770771649
/kind bug

/cc @evankanderson

### DIFF
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -87,7 +87,7 @@ func cleanup() {
 	for key, meter := range allMeters.meters {
 		if key != "" && meter.t.Before(expiryCutoff) {
 			flushGivenExporter(meter.e)
-			// make a copy of views to avoid data races
+			// Make a copy of views to avoid data races
 			viewsCopy := copyViews(resourceViews.views)
 			meter.m.Unregister(viewsCopy...)
 			delete(allMeters.meters, key)
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -82,9 +82,14 @@ func cleanup() {
 	expiryCutoff := allMeters.clock.Now().Add(-1 * maxMeterExporterAge)
 	allMeters.lock.Lock()
 	defer allMeters.lock.Unlock()
+	resourceViews.lock.Lock()
+	defer resourceViews.lock.Unlock()
 	for key, meter := range allMeters.meters {
 		if key != "" && meter.t.Before(expiryCutoff) {
 			flushGivenExporter(meter.e)
+			// make a copy of views to avoid data races
+			viewsCopy := copyViews(resourceViews.views)
+			meter.m.Unregister(viewsCopy...)
 			delete(allMeters.meters, key)
 		}
 	}
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -91,6 +91,7 @@ func cleanup() {
 			viewsCopy := copyViews(resourceViews.views)
 			meter.m.Unregister(viewsCopy...)
 			delete(allMeters.meters, key)
+			meter.m.Stop()
 		}
 	}
 }
--- a/metrics/resource_view.go
+++ b/metrics/resource_view.go
@@ -145,7 +145,7 @@ func RegisterResourceView(views ...*view.View) error {
 	return nil
 }
 
-// UnregisterResourceView is similar to view.Unregiste(), except that it will
+// UnregisterResourceView is similar to view.Unregister(), except that it will
 // unregister the view across all Resources tracked byt he system, rather than
 // simply the default view.
 func UnregisterResourceView(views ...*view.View) {
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
 	// (123=9m, 456=evicted, 789=0m)
 }
 
+func TestIfAllMeterResourcesAreRemoved(t *testing.T) {
+	allMeters.clock = clock.Clock(clock.NewFakeClock(time.Now()))
+	var fakeClock *clock.FakeClock = allMeters.clock.(*clock.FakeClock)
+	ClearMetersForTest() // t+0m
+	// Register many resources at once
+	for i := 1; i <= 1000; i++ {
+		res := resource.Resource{Labels: map[string]string{"foo": "bar"}}
+		res.Labels["id"] = string(i)
+		_, err := optionForResource(&res)
+		if err != nil {
+			t.Error("Should succeed getting option, instead got error ", err)
+		}
+
+	}
+	allMeters.lock.Lock()
+	// make a copy to test against as allMeters should be cleaned up
+	copyAllMeters := make(map[string]*meterExporter)
+	for k, v := range allMeters.meters {
+		copyAllMeters[k] = v
+	}
+	for _, meter := range copyAllMeters {
+		for _, mView := range resourceViews.views {
+			v := meter.m.Find(mView.Name)
+			if v == nil {
+				t.Errorf("Got a view that should be registered")
+			}
+		}
+	}
+	allMeters.lock.Unlock()
+	// expire all views
+	fakeClock.Step(12 * time.Minute) // t+12m
+	time.Sleep(time.Second)          // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
+	allMeters.lock.Lock()
+	// non-expiring defaultMeter should be available
+	if len(allMeters.meters) != 1 {
+		t.Errorf("len(allMeters)=%d, want: 1", len(allMeters.meters))
+	}
+	allMeters.lock.Unlock()
+	resourceViews.lock.Lock()
+	for _, meter := range copyAllMeters {
+		for _, mView := range resourceViews.views {
+			v := meter.m.Find(mView.Name)
+			if v != nil {
+				t.Errorf("Got a view that should be unregistered")
+			}
+			if meter.e != nil {
+				t.Errorf("Got a meter exporter set")
+			}
+		}
+	}
+	resourceViews.lock.Unlock()
+}
+
 func TestResourceAsString(t *testing.T) {
 	r1 := &resource.Resource{Type: "foobar", Labels: map[string]string{"k1": "v1", "k3": "v3", "k2": "v2"}}
 	r2 := &resource.Resource{Type: "foobar", Labels: map[string]string{"k2": "v2", "k3": "v3", "k1": "v1"}}
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -75,6 +75,8 @@ type testExporter struct {
 	id string
 }
 
+func (testExporter) ExportView(viewData *view.Data) {}
+
 func TestSetFactory(t *testing.T) {
 	var oldFactory ResourceExporterFactory
 	func() {
@@ -171,7 +173,7 @@ func TestAllMetersExpiration(t *testing.T) {
 
 	// Expire the second entry
 	fakeClock.Step(9 * time.Minute) // t+12m
-	_ = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+	_ = wait.PollImmediate(100*time.Millisecond, 5*time.Second, func() (bool, error) {
 		// Non-expiring defaultMeter should be available
 		allMeters.lock.Lock()
 		defer allMeters.lock.Unlock()
@@ -235,8 +237,7 @@ func TestIfAllMeterResourcesAreRemoved(t *testing.T) {
 
 	// Expire all meters and views
 	fakeClock.Step(12 * time.Minute) // t+12m
-
-	_ = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+	_ = wait.PollImmediate(100*time.Millisecond, 5*time.Second, func() (bool, error) {
 		// Non-expiring defaultMeter should be available
 		allMeters.lock.Lock()
 		defer allMeters.lock.Unlock()
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -247,16 +247,6 @@ func TestIfAllMeterResourcesAreRemoved(t *testing.T) {
 	if len(allMeters.meters) != 1 {
 		t.Errorf("len(allMeters)=%d, want: 1", len(allMeters.meters))
 	}
-	resourceViews.lock.Lock()
-	for k, meter := range copyAllMeters {
-		for _, mView := range resourceViews.views {
-			// Default meter should be skipped as views are not removed from it during cleanup
-			if k != "" && meter.m.Find(mView.Name) != nil {
-				t.Errorf("Got a view that should be unregistered: %s", mView.Name)
-			}
-		}
-	}
-	resourceViews.lock.Unlock()
 }
 
 func TestResourceAsString(t *testing.T) {
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -196,7 +196,7 @@ func TestIfAllMeterResourcesAreRemoved(t *testing.T) {
 	// Register many resources at once
 	for i := 1; i <= 1000; i++ {
 		res := resource.Resource{Labels: map[string]string{"foo": "bar"}}
-		res.Labels["id"] = fmt.Sprintf("id:%d", i)
+		res.Labels["id"] = fmt.Sprintf("%d", i)
 		_, err := optionForResource(&res)
 		if err != nil {
 			t.Error("Should succeed getting option, instead got error ", err)
--- a/metrics/resource_view_test.go
+++ b/metrics/resource_view_test.go
@@ -25,6 +25,7 @@ import (
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"k8s.io/apimachinery/pkg/util/clock"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var (
@@ -170,12 +171,14 @@ func TestAllMetersExpiration(t *testing.T) {
 
 	// Expire the second entry
 	fakeClock.Step(9 * time.Minute) // t+12m
-	time.Sleep(time.Second)         // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
-	allMeters.lock.Lock()
+	_ = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+		// Non-expiring defaultMeter should be available
+		return len(allMeters.meters) == 2, nil
+	})
+
 	if len(allMeters.meters) != 2 {
 		t.Errorf("len(allMeters)=%d, want: 2", len(allMeters.meters))
 	}
-	allMeters.lock.Unlock()
 	// (123=9m, 456=10.5m)
 	// non-expiring defaultMeter was just tested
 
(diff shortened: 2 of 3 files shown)


skonto:
Unit tests pass locally, need to retest.

markusthoemmes:
Do we have a reproducer that allows us to more confidently test that we are no longer leaking memory? Can we potentially write a unit(ish) test that makes sure we don't leak memory?

skonto:
@markusthoemmes I added a unit test hope that helps. Not sure if there is way to check this better eg. compare actual mem allocations.

skonto:
@markusthoemmes could you add `/ok-to-test`?

skonto:
Not sure what iw rong here with the unit tests but they pass locally.

skonto:
`TestWorkqueueMetrics` misbehaves.

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
 }
 
+func TestIfAllMeterResourcesAreRemoved(t *testing.T) {
+	allMeters.clock = clock.Clock(clock.NewFakeClock(time.Now()))
+	var fakeClock *clock.FakeClock = allMeters.clock.(*clock.FakeClock)
+	ClearMetersForTest() // t+0m
+	// Register many resources at once
+	for i := 1; i <= 1000; i++ {
+		res := resource.Resource{Labels: map[string]string{"foo": "bar"}}
+		res.Labels["id"] = fmt.Sprintf("%d", i)
```
vagababov, line 199:
```suggestion
		res.Labels["id"] = fmt.Sprint(i)
```

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
+		_, err := optionForResource(&res)
+		if err != nil {
```
vagababov, line 201:
```suggestion
		if _, err := optionForResource(&res);  err != nil {
```

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
+			t.Error("Should succeed getting option, instead got error ", err)
+		}
+
+	}
+	allMeters.lock.Lock()
+	// make a copy to test against as allMeters should be cleaned up
+	copyAllMeters := make(map[string]*meterExporter)
```
vagababov, line 208:
```suggestion
	// Make a copy to test against as allMeters should be cleaned up.
	copyAllMeters := make(map[string]*meterExporter, len(allMeters.meters))
```

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
+	for k, v := range allMeters.meters {
+		copyAllMeters[k] = v
+	}
+	for _, meter := range copyAllMeters {
+		for _, mView := range resourceViews.views {
+			v := meter.m.Find(mView.Name)
+			if v == nil {
+				t.Errorf("Got a view that should be registered")
+			}
```
vagababov, line 217:
```suggestion
			if meter.m.Find(mView.Name) == nil {
				t.Error("Got a view that should be registered")
			}
```

vagababov: Also the error text and check don't match. Do we want to log the view that was unexpected?
Could you just use `sets.String` to get the names and then do `Equal`?

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
+		}
+	}
+	allMeters.lock.Unlock()
+	// expire all views
+	fakeClock.Step(12 * time.Minute) // t+12m
+	time.Sleep(time.Second)          // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
```
vagababov, line 223: this is going to flake on prow. Is there a chan we can use, or a poll?

metrics/resource_view_test.go
```diff
@@ -189,6 +189,59 @@ func TestAllMetersExpiration(t *testing.T) {
...
+	resourceViews.lock.Lock()
+	for _, meter := range copyAllMeters {
+		for _, mView := range resourceViews.views {
+			v := meter.m.Find(mView.Name)
+			if v != nil {
+				t.Errorf("Got a view that should be unregistered")
+			}
+			if meter.e != nil {
+				t.Errorf("Got a meter exporter set")
+			}
```
vagababov, line 239:
```suggestion
			if meter.m.Find(mView.Name) != nil {
				t.Error("Got a view that should be unregistered")
			}
			if meter.e != nil {
				t.Error("Got a meter exporter set")
			}
```

metrics/resource_view_test.go
skonto: Maybe a poll, I copied what was there from the other test.

metrics/resource_view_test.go
```diff
@@ -170,12 +171,14 @@ func TestAllMetersExpiration(t *testing.T) {
 
 	// Expire the second entry
 	fakeClock.Step(9 * time.Minute) // t+12m
-	time.Sleep(time.Second)         // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
-	allMeters.lock.Lock()
+	_ = wait.PollImmediate(time.Second, time.Minute, func() (bool, error) {
+		// Non-expiring defaultMeter should be available
+		return len(allMeters.meters) == 2, nil
```
vagababov, line 176: well that's just one — what's the second one?
Also I'd suggest 100ms/5s params for poll

skonto: Second one?

skonto:
@vagababov thanks for the review, I addressed the comments.

metrics/resource_view.go
```diff
@@ -82,9 +82,14 @@ func cleanup() {
 	expiryCutoff := allMeters.clock.Now().Add(-1 * maxMeterExporterAge)
 	allMeters.lock.Lock()
 	defer allMeters.lock.Unlock()
+	resourceViews.lock.Lock()
+	defer resourceViews.lock.Unlock()
 	for key, meter := range allMeters.meters {
 		if key != "" && meter.t.Before(expiryCutoff) {
 			flushGivenExporter(meter.e)
+			// make a copy of views to avoid data races
```
julz, line 90:
```suggestion
			// Make a copy of views to avoid data races.
```

metrics/resource_view_test.go
```diff
@@ -74,6 +75,8 @@ type testExporter struct {
 	id string
 }
 
+func (testExporter) ExportView(viewData *view.Data) {}
```
julz, line 78: Out of interest: what is this needed for? I quickly checked the branch out and the tests still seem to pass with this line deleted

metrics/resource_view_test.go
```diff
@@ -189,6 +196,69 @@ func TestAllMetersExpiration(t *testing.T) {
...
+		}
+	}
+	allMeters.lock.Lock()
+	// Make a copy to test against as allMeters should be cleaned up
+	copyAllMeters := make(map[string]*meterExporter)
+	for k, v := range allMeters.meters {
+		copyAllMeters[k] = v
+	}
+	allMeters.lock.Unlock()
+	resourceViews.lock.Lock()
```
julz, line 225: [the comment on `resourceViews`](https://github.com/knative/pkg/blob/ba2137c16f4f4eb562b523e72f7267a6f6d4e3d6/metrics/resource_view.go#L61) says:

> // Lock regime: lock allMeters before resourceViews. The critical path is in
 // optionForResource, which must lock allMeters, but only needs to lock
 // resourceViews if a new meter needs to be created.

.. which makes me wonder if this locking is strictly correct. Maybe the allMeters lock should be held until after we unlock resourceViews.lock?

metrics/resource_view_test.go
```diff
@@ -189,6 +196,69 @@ func TestAllMetersExpiration(t *testing.T) {
...
+		for _, mView := range resourceViews.views {
+			want, got := mView, meter.m.Find(mView.Name)
+			if got == nil {
+				t.Errorf("View %s is not available", want.Name)
+			} else if want.Name != got.Name {
+				t.Errorf("Want %v, got %v", want.Name, got.Name)
+			}
+		}
+	}
+	resourceViews.lock.Unlock()
```
julz, line 236: wonder if it's worth reworking this a bit so that we can defer the Unlocks

metrics/resource_view_test.go
```diff
@@ -170,12 +173,16 @@ func TestAllMetersExpiration(t *testing.T) {
...
-	time.Sleep(time.Second)         // Wait a second on the wallclock, so that the cleanup thread has time to finish a loop
-	allMeters.lock.Lock()
+	_ = wait.PollImmediate(100*time.Millisecond, 5*time.Second, func() (bool, error) {
+		// Non-expiring defaultMeter should be available
+		allMeters.lock.Lock()
+		defer allMeters.lock.Unlock()
+		return len(allMeters.meters) == 2, nil
+	})
+
 	if len(allMeters.meters) != 2 {
```
julz, line 183: Wonder why we need a lock for the other accesses to `len(allMeters.meters)` (e.g. L179 and L157) but we dont here 🤔 (race detector isn't complaining, though, so maybe it's fine 🤷)

metrics/resource_view_test.go
skonto: Not all tests pass when using -tagnostackdriver. When a view is delete it exports the data before it is unregistered and if there is no method it will fail with a nill pointer exception.

skonto:
I found another place where we keep mem. Every meter is also a producer and it is added in [this](https://github.com/knative/pkg/blob/887806985c099edb5fc24db1ab4bfa4d5b59c0e5/vendor/go.opencensus.io/metric/metricproducer/manager.go#L27) global structure. We need to stop each meter in our cleanup func otherwise for every meter removed from the global allmeters list there is mem leaking. [Meter.Stop](https://github.com/knative/pkg/blob/887806985c099edb5fc24db1ab4bfa4d5b59c0e5/vendor/go.opencensus.io/stats/view/worker.go#L292) does that. This leads to a slow oom. I will double test and provide an update.

skonto:
Seems TestMetricsExport is [flacky](https://github.com/knative/pkg/issues/1672), locally it runs fine and cleanup function should no have run.

skonto:
[TestMetricsExport](https://github.com/knative/pkg/issues/1672) is failing while it passes locally (`./test/presubmit-tests.sh --unit-tests`): 
```
DONE 2012 tests, 3 skipped in 125.165s
Finished run, return code is 0
XML report written to /tmp/tmp.lWI0YXUruT/junit_MZNievKY.xml
JUnit file /tmp/tmp.lWI0YXUruT/junit_gsbXufbL.xml is created for reporting the test result
--- FAIL: kntest not installed, please clone knative test-infra repo and run `go install ./kntest/cmd/kntest` to install it
Test log written to /tmp/tmp.lWI0YXUruT/go_test_MZNievKY.log
===========================
==== UNIT TESTS PASSED ====
===========================
==== Sun 07 Feb 2021 11:17:52 AM PST
===========================
```
Wondering why the prometheus endpoint never came up or if 10 secs is not enough time on ci.

skonto:
I run this for over 16h and the soak test passes. Here are the screenshots and [data](https://drive.google.com/file/d/1EfQQLX4HTeNiz7xovobXSSgMbIRubs3S/view?usp=sharing):

Here `container_memory_working_set_bytes` it is measured as the indication for OOM (related discussion [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/227) and [here](https://www.bwplotka.dev/2019/golang-memory-monitoring/)).
image

```
$ go tool pprof --base knative-serving.autoscaler-6d86d96786-jpdkr.heap.2021-02-09_09:25:55+02:00.pb.gz knative-serving.autoscaler-6d86d96786-jpdkr.heap.2021-02-09_10:49:58+02:00.pb.gz
File: autoscaler
Type: inuse_space
Time: Feb 9, 2021 at 9:25am (EET)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for -425.31kB, 2.49% of 17097.11kB total
Dropped 11 nodes (cum <= 85.49kB)
Showing top 10 nodes out of 83
      flat  flat%   sum%        cum   cum%
-3640.29kB 21.29% 21.29% -3640.29kB 21.29%  go.opencensus.io/stats/view.NewMeter
-1024.28kB  5.99% 27.28% -1024.28kB  5.99%  go.opencensus.io/stats/view.(*collector).addSample
 1024.16kB  5.99% 21.29%  1536.18kB  8.99%  knative.dev/serving/pkg/metrics.AugmentWithRevision
 1024.05kB  5.99% 15.30%  1024.05kB  5.99%  container/list.(*List).insertValue
  650.62kB  3.81% 11.50%   650.62kB  3.81%  golang.org/x/net/http2.(*ClientConn).frameScratchBuffer
  514.63kB  3.01%  8.49%   514.63kB  3.01%  k8s.io/apimachinery/pkg/apis/meta/v1.(*FieldsV1).UnmarshalJSON
  513.50kB  3.00%  5.48%   513.50kB  3.00%  github.com/json-iterator/go.(*Iterator).stopCapture
  512.22kB  3.00%  2.49%   512.22kB  3.00%  github.com/json-iterator/go.(*Iterator).readStringSlowPath
  512.12kB  3.00%  0.51%   512.12kB  3.00%  fmt.Errorf
 -512.04kB  2.99%  2.49%  -511.62kB  2.99%  go.opencensus.io/stats/view.(*worker).tryRegisterView
```
image

```
$ kubectl top pods -n knative-serving
NAME                                     CPU(cores)   MEMORY(bytes)   
activator-688c899b4-5bxrb                5m           42Mi            
activator-688c899b4-kdknr                5m           40Mi            
autoscaler-6d86d96786-jpdkr              42m          91Mi            
autoscaler-hpa-69b5787778-28jzv          3m           25Mi            
autoscaler-hpa-69b5787778-m7brw          2m           28Mi            
controller-7cbd786664-6gzt8              23m          119Mi           
controller-7cbd786664-snx8b              17m          42Mi            
domain-mapping-7989f9b9fb-lxmq6          2m           24Mi            
domainmapping-webhook-6ddcdf6c98-vbdt6   8m           26Mi            
webhook-7dd8c645d-9rwdk                  30m          29Mi            
webhook-7dd8c645d-hxcdw                  7m           34Mi      
```

```
Every 10.0s: oc get pods -n knative-serving                                                         easy.box: Tue Feb  9 11:03:54 2021

NAME                                     READY   STATUS    RESTARTS   AGE
activator-688c899b4-5bxrb                1/1     Running   0          16h
activator-688c899b4-kdknr                1/1     Running   0          16h
autoscaler-6d86d96786-jpdkr              1/1     Running   0          16h
autoscaler-hpa-69b5787778-28jzv          1/1     Running   0          16h
autoscaler-hpa-69b5787778-m7brw          1/1     Running   0          16h
controller-7cbd786664-6gzt8              1/1     Running   0          16h
controller-7cbd786664-snx8b              1/1     Running   0          16h
domain-mapping-7989f9b9fb-lxmq6          1/1     Running   0          16h
domainmapping-webhook-6ddcdf6c98-vbdt6   1/1     Running   0          16h
webhook-7dd8c645d-9rwdk                  1/1     Running   0          16h
webhook-7dd8c645d-hxcdw                  1/1     Running   0          16h
``` 
Without stopping the meters I get an oom at some point:
image

In general there is an increase in memory over time with metrics enabled (this is obvious from the diagram above), while without metrics after 12h the mem is around ~65MB.

metrics/resource_view_test.go
skonto: Will change to be on the safe side. I have it this way here because I dont expect the views to be modified by some other entity.

metrics/resource_view_test.go
skonto: Will take a look.

metrics/resource_view_test.go
skonto: Because of the nature of the test there should not be a race we progress the clock manually and when poll ends we should read the right value. No other goroutine is expected to write anything in allMeters at that point. For the sake of readability and having less confusion I can change it and do a lock/unlock there.

metrics/resource_view_test.go
skonto: If copying is done in a separate func then yes but at after copying we need to progress the clock which means we have to unlock to allow the cleanup goroutine to run, we can simply defer the Unlocks for end of the test.

metrics/resource_view_test.go
```diff
@@ -189,6 +198,64 @@ func TestAllMetersExpiration(t *testing.T) {
...
+		v := view.View{Name: fmt.Sprintf("testview-%d", i), Measure: m, Aggregation: view.Sum()}
+
+		err := RegisterResourceView(&v)
+		if err != nil {
+			t.Fatal("RegisterResourceView =", err)
+		}
+	}
+	allMeters.lock.Lock()
+	// Make a copy to test against as allMeters should be cleaned up
+	copyAllMeters := make(map[string]*meterExporter)
```
vagababov, line 222:
```suggestion
	copyAllMeters := make(map[string]*meterExporter, len(allMeters.meters))
```

metrics/resource_view_test.go
vagababov: you can just put it into an anonymous
```
func() {
lock
defer unlock
do stuf
}()
```

skonto:
I managed to reproduce locally the failure of `TestMetricsExport` and it seems that a new exporter is not always created because the condition for when to create one, only checks if the backend destination is different, however we do have a test that sets prometheus port to 9091 and that is causing issues as we later on we wait for the server to be up on port 9090.

skonto:
Now I see 
```--- FAIL: TestDefaultObserved (0.00s)
    informed_watcher_test.go:392: foo1.count = 1, want 2```

**Comment by skonto** [comment]
```
--- FAIL: TestMetricsExport (6.07s)
    --- FAIL: TestMetricsExport/OpenCensus (5.03s)
        e2e_test.go:235: Created exporter at localhost:12345
        logger.go:130: 2021-02-09T19:31:22.391Z	INFO	metrics/exporter.go:161	Flushing the existing exporter before setting up the new exporter.
        logger.go:130: 2021-02-09T19:31:22.398Z	INFO	metrics/opencensus_exporter.go:56	Created OpenCensus exporter with config:	{"config": {}}
        logger.go:130: 2021-02-09T19:31:22.398Z	INFO	metrics/exporter.go:174	Successfully updated the metrics exporter; old config: &{knative.dev/serving testComponent prometheus 1000000000 <nil> <nil>  false 19090 0.0.0.0 false   {   false}}; new config &{knative.dev/serving testComponent opencensus 1000000000 <nil> <nil> localhost:12345 false 0  false   {   false}}
        e2e_test.go:273: Timeout reading input
        e2e_test.go:279: Unexpected OpenCensus exports (-want +got):
              []metrics.metricExtract(Inverse(Sort, []string{
              	"knative.dev/serving/testComponent/global_export_counts<>:2",
              	"knative.dev/serving/testComponent/resource_global_export_count<>:2",
              	`knative.dev/serving/testComponent/testing/value<project="p1",rev`...,
            - 	`knative.dev/serving/testComponent/testing/value<project="p1",revision="r2">:1`,
              }))
{"level":"info","ts":1612899093.698988,"logger":"fallback","caller":"metrics/prometheus_exporter.go:51","msg":"Created Prometheus exporter with config: &{test test prometheus 0 <nil> <nil>  false 0  false   {   false}}. Start the server for Prometheus exporter."}
FAIL 

```

skonto:
@julz @vagababov @markusthoemmes should this be merged now? I think the failing test should be handled on its own separately.

metrics/resource_view_test.go
skonto: There is the default one and one that comes from the test.

vagababov (commented):
/lgtm

/assign @evankanderson 
For Evan to stamp approval.

skonto:
@evankanderson gentle ping.

skonto:
@evankanderson @vagababov  should we merge? wdyth?

PR #2007, opened 2021-02-03
Allow setting DisableCompression in NewAutoTransport

### LINKED ISSUE #2006: Transports should allow setting DisableCompression
/kind bug

## Expected Behavior

The transports returned by NewAutoTransport should support disabling go's helpful functionality where it silently adds an Accept-Encoding: gzip header and uncompresses the result for you. While this feature is often useful, it's not what we want for Activator and QP's ReverseProxy transports since it (a) injects an extra header not in the client side in to the request the user container receives, and (b) causes the QP/Activator to end up uncompressing things uselessly and wastefully if the user container happens to co-operate with that request

## Actual Behavior

The transports do not set DisableCompression=true, so Go helpfully injects an `Accept-Encoding: gzip` header. This ends up causing the user container to compress the response (if it supports gzip encoding), which the QP immediately then immediately uncompresses :).

1. ko apply ksvc.yaml from https://github.com/julz/gziphello
2. curl the container, don't pass any extra headers
2. see in logs that gzip header was nevertheless added
3. notice (by logging resp.Uncompressed) that queue proxy uncompressed the response and the container compressed it entirely pointlessly

julz:
Fixes #2006.

Go's http client helpfully adds an "Accept-Encoding: gzip" header if not
already present, and unzips the response if it did so. While this is
often the right thing to do, it is not the right thing to do if we're
reverse proxying since (a) it means we add a header to the request the
client didn't actually ask for (b) it means we end up uncompressing
things in the proxy, which e.g. for activator/QP is not what we want.

/kind enhancement

# Changes

- :gift: Add `NewProxyAutoTransport` which is an auto-transport with DisableCompression=true set on the underlying transports.

**Release Note**

```release-note
Adds NewProxyAutoTransport(), this method returns a transport which automatically selects between HTTP and H2c, and which is suitable for use by reverse proxies. Specifically it sets DisableCompression=true on the underlying transports.
```

/assign @vagababov @markusthoemmes

### DIFF
--- a/network/h2c.go
+++ b/network/h2c.go
@@ -41,8 +41,13 @@ func NewServer(addr string, h http.Handler) *http.Server {
 // to explicitly allow h2c (http2 without TLS) transport.
 // See https://github.com/golang/go/issues/14141 for more details.
 func NewH2CTransport() http.RoundTripper {
+	return newH2CTransport(false)
+}
+
+func newH2CTransport(disableCompression bool) http.RoundTripper {
 	return &http2.Transport{
-		AllowHTTP: true,
+		AllowHTTP:          true,
+		DisableCompression: disableCompression,
 		DialTLS: func(netw, addr string, _ *tls.Config) (net.Conn, error) {
 			return DialWithBackOff(context.Background(),
 				netw, addr)
--- a/network/transports.go
+++ b/network/transports.go
@@ -94,13 +94,14 @@ func dialBackOffHelper(ctx context.Context, network, address string, bo wait.Bac
 	return nil, fmt.Errorf("timed out dialing after %.2fs", elapsed.Seconds())
 }
 
-func newHTTPTransport(disableKeepAlives bool, maxIdle, maxIdlePerHost int) http.RoundTripper {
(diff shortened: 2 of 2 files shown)


markusthoemmes (approved):
Seems useful but also kind of seems like a bug for the reverse proxy :thinking:. I'd think it should take care of that, should we open a bug against Golang here as well?

vagababov:
My history as chrome developer taught me that most people out there don't set gzip and you can reap up to 20-30% bandwidth savings just doing that. Now, question is, would it _stop_ gzipping if user asked to?

julz:
@vagababov no it won't, it does the right thing if the header was actually passed by the client (i.e keeps it uncompressed all the way along the chain) -- it only does the silently adding the header and uncompressing if the header _isnt_ supplied:

>  // DisableCompression, if true, prevents the Transport from
    // requesting compression with an "Accept-Encoding: gzip"
    // request header when the Request contains no existing
    // Accept-Encoding value. If the Transport requests gzip on
    // its own and gets a gzipped response, it's transparently
    // decoded in the Response.Body. However, if the user
    // explicitly requested gzip it is not automatically
    // uncompressed.

PR #2016, opened 2021-02-09
adding retry event count metric for source

kthakar1990:
# Changes

- adding a code change that will be used by knative/eventing to report retry event count for knative source.

### DIFF
--- a/.github/workflows/knative-releasability.yaml
+++ b/.github/workflows/knative-releasability.yaml
@@ -36,8 +36,8 @@ jobs:
     env:
       #########################################
       #   Update this section each release.   #
-      RELEASE: 'v0.20'
-      SLACK_CHANNEL: 'release-20'
+      RELEASE: 'v0.21'
+      SLACK_CHANNEL: 'release-21'
       #########################################
 
     steps:
--- a/.github/workflows/knative-releasability.yaml
+++ b/.github/workflows/knative-releasability.yaml
@@ -36,8 +36,8 @@ jobs:
     env:
       #########################################
       #   Update this section each release.   #
-      RELEASE: 'v0.21'
-      SLACK_CHANNEL: 'release-21'
+      RELEASE: 'v0.20'
+      SLACK_CHANNEL: 'release-20'
       #########################################
 
     steps:
--- a/source/source_stats_reporter.go
+++ b/source/source_stats_reporter.go
@@ -10,7 +10,7 @@ You may obtain a copy of the License at
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific langusage governing permissions and
+See the License for the specific language governing permissions and
 limitations under the License.
 */
 
--- a/source/source_stats_reporter.go
+++ b/source/source_stats_reporter.go
@@ -57,7 +57,7 @@ var (
 	responseTimeout        = tag.MustNewKey(metricskey.LabelResponseTimeout)
 )
 
-//ReportArgs defines the arguments for reporting metrics
+// ReportArgs defines the arguments for reporting metrics.
 type ReportArgs struct {
 	Namespace     string
 	EventType     string
--- a/source/source_stats_reporter.go
+++ b/source/source_stats_reporter.go
@@ -35,8 +35,8 @@ var (
 		stats.UnitDimensionless,
 	)
 
-	// RetryEventCountM is a counter which records the number of events sent by the source in retries.
-	RetryEventCountM = stats.Int64(
+	// retryEventCountM is a counter which records the number of events sent by the source in retries.
+	retryEventCountM = stats.Int64(
 		"retry_event_count",
 		"Number of retry events sent",
 		stats.UnitDimensionless,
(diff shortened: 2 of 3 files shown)


source/source_stats_reporter.go
```diff
@@ -51,6 +57,7 @@ var (
 	responseTimeout        = tag.MustNewKey(metricskey.LabelResponseTimeout)
 )
 
+//ReportArgs interface
```
skonto, line 60: This is a struct, not sure if we want to add this here.

source/source_stats_reporter_test.go
```diff
@@ -51,14 +51,31 @@ func TestStatsReporter(t *testing.T) {
...
 	expectSuccess(t, func() error {
 		return r.ReportEventCount(args, http.StatusAccepted)
 	})
 	expectSuccess(t, func() error {
 		return r.ReportEventCount(args, http.StatusAccepted)
 	})
+	expectSuccess(t, func() error {
+		return r.ReportRetryEventCount(args, http.StatusServiceUnavailable)
+	})
+	expectSuccess(t, func() error {
```
skonto, line 74: You repeat the same function twice?

source/source_stats_reporter.go
kthakar1990: I can remove it.. it was just giving an suggestion to be exported or commented.. so I added it there

kthakar1990:
United test failures would be resolved by this [PR](https://github.com/knative/eventing/pull/4861)

.github/workflows/knative-releasability.yaml
```diff
@@ -36,8 +36,8 @@ jobs:
     env:
       #########################################
       #   Update this section each release.   #
-      RELEASE: 'v0.20'
-      SLACK_CHANNEL: 'release-20'
+      RELEASE: 'v0.21'
+      SLACK_CHANNEL: 'release-21'
```
skonto, line 40: I think this file is not related to this PR.

source/source_stats_reporter.go
```diff
@@ -51,6 +57,7 @@ var (
 	responseTimeout        = tag.MustNewKey(metricskey.LabelResponseTimeout)
 )
 
+//ReportArgs defines the arguments for reporting metrics
```
julz, line 60:
```suggestion
// ReportArgs defines the arguments for reporting metrics.
```

source/source_stats_reporter.go
```diff
@@ -35,6 +35,12 @@ var (
 		stats.UnitDimensionless,
 	)
 
+	// RetryEventCountM is a counter which records the number of events sent by the source in retries.
+	RetryEventCountM = stats.Int64(
```
julz, line 39: just to check: does this need to be exported? (I notice `eventCountM` isn't and I'm wondering if this can be private too)

kthakar1990: you're right it doesn't have to be!

julz (commented):
/lgtm

evankanderson:
Just wondering if you want to use a distribution of retry counts to distinguish between "everything retries once" and "20% retry 5 times".

evankanderson:
Also, `pkg` seems like an unfortunate place for us to need to do development for sources. I don't have a pat solution, though.

vagababov:
> Also, `pkg` seems like an unfortunate place for us to need to do development for sources. I don't have a pat solution, though.

Yeah, I think I mentioned that in one other PR.
Would just eventing be a good place? DOn't sources depend on it anyway?

kthakar1990:
> Just wondering if you want to use a distribution of retry counts to distinguish between "everything retries once" and "20% retry 5 times".

What we actually want is to see various count by http status codes.. so retry count is enough for us.. because we can segment by http status codes..

kthakar1990:
> > Also, `pkg` seems like an unfortunate place for us to need to do development for sources. I don't have a pat solution, though.
> 
> Yeah, I think I mentioned that in one other PR.
> Would just eventing be a good place? DOn't sources depend on it anyway?

I am fine with that idea.. However, there is a plan to make standardize metrics API and we can include the refactoring with that work..

vaikas:
I don't have a super strong feeling about this landing in pkg, just wanted to comment on the fact that not all sources necessarily have to depend on eventing so it might make sense to have a way to get the stats in from here.

kthakar1990:
@evankanderson is this ready to merge?

source/source_stats_reporter.go
```diff
@@ -98,6 +106,15 @@ func (r *reporter) ReportEventCount(args *ReportArgs, responseCode int) error {
 	return nil
 }
 
+func (r *reporter) ReportRetryEventCount(args *ReportArgs, responseCode int) error {
+	ctx, err := r.generateTag(args, responseCode)
+	if err != nil {
+		return err
+	}
+	metrics.Record(ctx, retryEventCountM.M(1))
```
n3wscott, line 114: retry can happen more than once per attempt. I have not seen how this is intended to be used

kthakar1990: you're right.. it happens more than once per attempt.. so the related PR: https://github.com/knative/eventing/pull/4861 is submitted in eventing repo.. that will call this reporter for every retry and that's why I am adding count here.

kthakar1990: @n3wscott does that help?

vaikas: Yeah, that makes sense to me. Otherwise you'd have to pass in something like an array of structs since if you collate the retry count with all the responseCodes that caused the failure, that seems wrong? Or, maybe I misunderstand what the question here was.

n3wscott (approved):
/lgtm
/approve

PR #2020, opened 2021-02-15
Remove reflector metrics

### LINKED ISSUE #2019: Reflector metrics are not available
## Expected Behavior
See reflector metrics exported
## Actual Behavior
No reflector metrics are exported
1. Installed v0.19.0 for Knative Serving on OCP, K8s version v1.18.2.
2. Port-forward and listed metrics for the Controller pod, for a dump check [here](https://gist.github.com/skonto/e3dcc5c28c24644003895fef943a3cf4)

It seems this is due to https://github.com/kubernetes/kubernetes/pull/74636 check relevant discussion here https://github.com/kubernetes-sigs/controller-runtime/issues/817. Not sure if this has been discussed before, I see a related comment [here](https://github.com/knative/pkg/blob/3154117dcfaf7a691017d22cd4c5461e09a27fb3/controller/stats_reporter.go#L97). Should we remove these metrics?

/cc @mattmoor @evankanderson @vaikas

skonto:
# Changes

- Removes the reflector metrics which are not being exposed any more from the client-go lib, see my comment [here](https://github.com/knative/pkg/issues/2019#issuecomment-779118506). I prefer to remove the dead code than simply deactivating with a noop metric provider (we save some allocations as well). Reflector metrics has been deactivated for a long time in client-go, in case they are re-enabled we can discuss the value they bring and if we see fit we can re-enable them.

Fixes #2019 

/cc @evankanderson @mattmoor @markusthoemmes


antoineco (approved):
/lgtm

evankanderson (commented):
/lgtm
/approve

PR #2028, opened 2021-02-18
Use the constant for UserAgent

vagababov:
rather than inline string.

/assign @tcnghia


vagababov:
metrics

tcnghia (approved):
/lgtm

PR #2031, opened 2021-02-19
Fix potential deadlock when k8s client is used

### LINKED ISSUE #807: Potential deadlock in knative.dev/pkg/metrics when using k8s.io/client-go/kubernetes

skonto:
# Changes

- Removes the global metrics config mutex to avoid a potential deadlock when k8s client is used due to the 
registered metrics hooks in the client.
- This PR preserves the previous semantics when managing the metrics config etc so it should be transparent from a user point of view.
- It follows the worker approach in the opencensus lib.

/kind bug

Fixes #807 

/cc @vagababov @evankanderson

### DIFF
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -35,15 +35,8 @@ var (
 )
 
 func init() {
-	setupMetricsWorker()
-}
-
-func setupMetricsWorker() {
-	// No lock is needed because we only need this for the testing initialization
-	if mWorker == nil {
-		mWorker = newMetricsWorker()
-		go mWorker.start()
-	}
+	mWorker = newMetricsWorker()
+	go mWorker.start()
 }
 
 // SecretFetcher is a function (extracted from SecretNamespaceLister) for fetching
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"sync"
 
 	"go.opencensus.io/resource"
 	"go.opencensus.io/stats/view"
@@ -32,8 +31,20 @@ import (
 var (
 	curMetricsExporter view.Exporter
 	curMetricsConfig   *metricsConfig
-	metricsMux         sync.RWMutex
+	mWorker            *metricsWorker
 )
+
+func init() {
+	setupMetricsWorker()
+}
+
+func setupMetricsWorker() {
+	// No lock is needed because we only need this for the testing initialization
+	if mWorker == nil {
+		mWorker = newMetricsWorker()
+		go mWorker.start()
+	}
+}
 
 // SecretFetcher is a function (extracted from SecretNamespaceLister) for fetching
 // a specific Secret. This avoids requiring global or namespace list in controllers.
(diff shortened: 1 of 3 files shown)


skonto:
`TestMetricsExport` fails again.

metrics/testing.go
```diff
@@ -22,6 +22,7 @@ const (
 
 // InitForTesting initialize the necessary global variables for unit tests.
 func InitForTesting() {
+	setupMetricsWorker()
```
vagababov, line 25: I think this is already called due to `init`?

skonto: Nope that was really weird because I thought init from the metrics packed (since it is imported) it would be call first but it does not. It fails with a nil pointer exception when it tries to access the mWorker.

skonto: Cant reproduce it will update.

skonto:
@evankanderson gentle ping.

metrics/exporter.go
```diff
@@ -154,28 +158,14 @@ func UpdateExporter(ctx context.Context, ops ExporterOptions, logger *zap.Sugare
...
+		ctx:       ctx,
+		newConfig: newConfig,
+		done:      make(chan error),
 	}
-
-	setCurMetricsConfigUnlocked(newConfig)
-	return nil
+	mWorker.c <- updateCmd
+	err = <-updateCmd.done
+	return err
```
evankanderson, line 168: I think this can be `return <-updateCmd.done`

evankanderson (commented):
Other than the unit test failure, this looks reasonable to me.

evankanderson:
Running main and your branch 84 times, I see about an equivalent number of failures (8 vs 9(, but there seems to be a lot more output on failure with this PR, and I see a timeout/missing data failure in the OpenCensus export in addition to the `e2e_test.go:320` timeout on Prometheus errors. I'm willing to believe that the OpenCensus failure was a fluke and that this doesn't make the Prometheus failures any worse (though they seem much higher than I'd expect).

/retest
/lgtm
/approve

skonto:
@evankanderson I have seen that test failing multiple times in my other pr #2005, left a comment [there](https://github.com/knative/pkg/pull/2005#issuecomment-776174080), so I am not sure if this is any different.
I will try to debug that and see if it can be fixed.

PR #2032, opened 2021-02-19
Add symmetric filter helped based on OwnerRefable.

mattmoor:
Most resources stamped out by knative controllers have OwnerReferences synthesized via `kmeta.NewOwnerRef`, which requires the parent resource to implement `kmeta.OwnerRefable` for accessing the `GroupVersionKind`.

However, where we setup informer watches on those same children resources we have essentially relied on direct synthesis of the `Group[Version]Kind` where we could instead provide an empty instance of the controller resource and leverage `GetGroupVersionKind` to provide the GVK used for filtration.

So where before folks would write:

```golang
FilterFunc: controller.FilterControllerGK(v1alpha1.WithKind("MyType"))
```

They may now write:

```golang
FilterFunc: controller.FilterController(&v1alpha1.MyType{})
```

The latter is preferable in part because it is more strongly typed, but also shorter.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -152,6 +152,13 @@ func FilterControllerGK(gk schema.GroupKind) func(obj interface{}) bool {
 	}
 }
 
+// FilterController makes it simple to create FilterFunc's for use with
+// cache.FilteringResourceEventHandler that filter based on the
+// controlling resource.
+func FilterController(r kmeta.OwnerRefable) func(obj interface{}) bool {
+	return FilterControllerGK(r.GetGroupVersionKind().GroupKind())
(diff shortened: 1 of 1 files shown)


n3wscott (commented):
/lgtm
/approve

TAKE MY MONEY

mattmoor:
This is the code I was looking at when this struck me: https://github.com/knative-sandbox/net-contour/blob/ac41764ba80f3a1d9716bfe8c27b8cd74642de00/pkg/reconciler/contour/controller.go#L91-L100

I think we have bits like that ~everywhere we should clean up after the release cut.

PR #1929, opened 2020-11-26
fixes TestInformedWatcher flake

dprotaso:
Fixes: https://github.com/knative/pkg/issues/1907

Issue #1907, opened 2020-11-14
TestInformedWatcher flaky 


runzexia:
/area test-and-release
/kind bug

## Expected Behavior
test always success
## Actual Behavior
test failed in https://github.com/knative/pkg/runs/1399755328,
after rerun test success
## Steps to Reproduce the Problem
not easy to reproduce, but i guess it because `informer.HasSynced` not means all EventHandler has been called

runzexia:
failed again https://github.com/knative/pkg/runs/1401922270?check_suite_focus=true

bobcatfish:
Hi @runzexia !! I think this might be an actual bug and we are running into this in Tekton as well, causing https://github.com/tektoncd/pipeline/issues/2815 (longer explanation in https://github.com/tektoncd/pipeline/issues/2815#issuecomment-733207368)

Looking at the assertion that's failing:

https://github.com/knative/pkg/blob/c5296cd61e9f34e183f8010a40dc866f895252b9/configmap/informed_watcher_test.go#L86-L92

It looks like the intention is that after calling `cmw.Start`, all of the callbacks have been called, but these flakes are showing that that isn't always the case (which I think makes sense b/c the callbacks are called in separate go routines).

I hacked the failing test to run in a loop and added a script to run it repeatedly in https://github.com/bobcatfish/pkg/commit/32da8032bfb04caa34dcbaea1f50b1d024bc3f44 and every time I run `doit.sh` I eventually see a failure such as:

```
+ for i in '`seq 1 1000`'
+ go test ./configmap/... -run TestInformedWatcher -count=1
--- FAIL: TestInformedWatcher (0.00s)
    --- FAIL: TestInformedWatcher/AGAIN-140 (0.00s)
        informed_watcher_test.go:94: foo1.count = 0, want 1
FAIL
FAIL	knative.dev/pkg/configmap	2.319s
```

In Tekton this means that there is a chance that when a controller starts, the values returned by the config store using this watcher might not return the right values (if the go routines that update the config store have not completed).

So I think that `InformedWatcher.Start` isn't functioning as intended and this is a bug. The fix would be to make it so that `InformedWatcher.Start` actually somehow guarantees that those callbacks have been invoked before returning, what do you think?

( @mattmoor might be relevant to your interests! )

mattmoor:
That'd be a good find.  cc @dprotaso 

> which I think makes sense b/c the callbacks are called in separate go routines

Where do you mean?  Within the informer itself?

[...]

PR #2036, opened 2021-02-23
Fix race: Make informed watcher start wait for Add event 🏎️

### LINKED ISSUE #1960: Race condition when using config store backed by config maps
## Expected Behavior

When I try to use a config store backed by a config map immediately after the process using it starts up, when the config map exists before the process starts, I'd like to be able to guarantee that I am reading the configured value.

I think we could do this if one of the following was true:

1. When the reconciler starts, it could wait for the config store to be initialized, i.e. wait for the goroutines that handle events to fire at least once (see https://github.com/knative/pkg/issues/1907#issuecomment-733254238 for a description of this code) before returning
2. Provide a way for code using this reconciler to wait for this state (e.g. maybe don't do (1) by default, but at least provide a way for a caller to do this, e.g. [via the Reconciler object](https://github.com/tektoncd/pipeline/blob/8eaaeaac6957b7c7eff6694b00e792bc00f2c587/pkg/reconciler/taskrun/controller.go#L67-L79))
3. [Export the config store](https://github.com/knative/pkg/blob/a2394e88959df2ca1900855474ae43d7da0a9046/codegen/cmd/injection-gen/generators/reconciler_reconciler.go#L278-L280) so that I can look at the state

/kind bug

## Actual Behavior

When I use a generated reconciler, which has an unexported [configstore](https://github.com/knative/pkg/blob/a2394e88959df2ca1900855474ae43d7da0a9046/codegen/cmd/injection-gen/generators/reconciler_reconciler.go#L278-L280), if I try to back that config store with configmaps, there is a race condition where if I try to use that config store immediately after start up, the informed watcher's goroutines which read from the backing config maps and update the config store may not have fired yet, which means I might read something from the config store other than what I've configured.

* I believe https://github.com/knative/pkg/issues/1907 was actually catching this bug (explanation at https://github.com/knative/pkg/issues/1907#issuecomment-733211228)
* I repro-ed this on the Tekton side with these horrible hacks: https://github.com/bobcatfish/pipeline/commit/1fbc3889b8ab1fbdcb2f0510a8c0674b436c3bf7 (trying to more reliably reproduce https://github.com/tektoncd/pipeline/issues/2815)

## Additional Info

* More detail from a Tekton perspective in https://github.com/tektoncd/pipeline/issues/2815#issuecomment-733207368

bobcatfish:
# Changes

When using the informed watcher to watch a config map, previously add
events were being processed in a goroutine with no syncrhonization
making it so that code may try to access the values backed by the
configmaps before they are initialized.

This commit makes it so that the Start method of the informer will wait
for the add event to occur at least once for all config maps it is
watching.

This commit also undoes the workaround added in #1929 which was working
around the race condition identified in #1907 (and in
https://github.com/tektoncd/pipeline/issues/3720). This means that if
the synchronization was removed, the impacted test would start flaking
again. If we wanted it to reliably fail in that case, we could introduce
a sleep in the callback but that doesn't seem worth it.

I also tested this change by manually patching the changes into my
clone of tektoncd/pipeline and following the repro steps at
https://github.com/tektoncd/pipeline/issues/2815#issuecomment-733207368
Before the change I can reproduce the issue, and after the change, I
can't! :D

Fixes #1960 :bug:

@dprotaso @mattmoor 

/kind bug

**Release Note**

```release-note
Controllers built using knative/pkg and using the config store watcher for hot reloading of config maps can trust that once the controller has started, the values have been loaded from the backing config maps.
```

**Docs**

```docs
n/a
```


configmap/informer/informed_watcher.go
```diff
@@ -207,3 +241,91 @@ func (i *InformedWatcher) deleteConfigMapEvent(obj interface{}) {
 	}
 	// If there is no default value, then don't do anything.
 }
+
```
bobcatfish, line 244: * I'm not sure if there is a better package to put these kind of synchronization primitives in?
* Instead of making these wrappers around a WaitGroup, I could use polling (e.g. like `cache.WaitForCacheSync(stopCh, i.informer.Informer().HasSynced)`) but using go synchronization primitives to do synchronization between goroutines seemed like a better solution to me (less consistent with the surrounding code tho)

bobcatfish:
I can't tell if the pull-knative-pkg-unit-tests failure is legit 🤔

bobcatfish:
image

image

configmap/informer/informed_watcher.go
```diff
@@ -126,31 +130,56 @@ func (i *InformedWatcher) WatchWithDefault(cm corev1.ConfigMap, o ...configmap.O
...
 		return errors.New("error waiting for ConfigMap informer to sync")
 	}
 
-	return i.checkObservedResourcesExist()
+	if err := i.checkObservedResourcesExist(); err != nil {
+		return err
+	}
+
+	// Wait until all config maps have been at least initially processed
+	return s.WaitForAllKeys(stopCh)
```
dprotaso, line 179: This changes the contract a bit since prior you could run this watcher with supplied defaults and/or missing config maps.

I think being stricter is good but we should document this in the godoc

dprotaso (changes requested):

dprotaso:
Thanks for the PR! Yeah metrics tests are flakey

configmap/informer/informed_watcher.go
```diff
@@ -126,31 +130,56 @@ func (i *InformedWatcher) WatchWithDefault(cm corev1.ConfigMap, o ...configmap.O
...
+	// to wait for the add event to be processed for all configmaps
+	s := NewSyncedCallback(i.getConfigMapNames(), i.addConfigMapEvent)
+	addConfigMapEvent := func(obj interface{}) {
+		configMap := obj.(*corev1.ConfigMap)
+		s.Call(obj, configMap.Name)
+	}
+	// Pretend that all the defaulted ConfigMaps were just created. This is done before we start
+	// the informer to ensure that if a defaulted ConfigMap does exist, then the real value is
+	// processed after the default one.
+	i.triggerAddEventForDefaultedConfigMaps(addConfigMapEvent)
```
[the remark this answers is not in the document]
bobcatfish: right, that's the case but only when using defaulted config maps - i wasnt totally clear on the expectations when using defaulted them, does this match what you think it should do? (i can add some comments at least explaining this if you can confirm this is what we want)

it DOES wait when you're not using defaulted config maps (e.g. in tekton pipelines afaik we're not using defaulted configmaps: https://github.com/tektoncd/pipeline/blob/93e368dbcb014e8650a6ba8ad01ca3c5f6eb6554/vendor/knative.dev/pkg/configmap/store.go#L114)

configmap/informer/informed_watcher.go
bobcatfish: i'm not sure the contract has changed:

* if you supply defaults, it won't wait (as you commented above) (verified via [TestWatchMissingOKWithDefaultOnStart](https://github.com/knative/pkg/blob/8a8de5fc8c72f358ad3e7b5e29f4dfb8a264892d/configmap/informer/informed_watcher_test.go#L300)
* if you don't supply the config maps it expects, it'll fail before this change (checkObservedResourcesExist will return an error i think?), i think [TestWatchMissingFailsOnStart](https://github.com/knative/pkg/blob/8a8de5fc8c72f358ad3e7b5e29f4dfb8a264892d/configmap/informer/informed_watcher_test.go#L229) verifies that

configmap/informer/informed_watcher.go
bobcatfish: writing tests for unexported code gives me a no feeling but i can understand why you dont want ppl to start importing this and relying on it :D

bobcatfish:
should be ready for another look @dprotaso !

configmap/informer/informed_watcher.go
dprotaso: I missed the `checkObservedResourcesExist` bit - thanks for pointing that out

configmap/informer/informed_watcher.go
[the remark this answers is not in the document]
dprotaso: I'm good with what you described - ie. if you include defaulted config maps then we don't wait for a sync from the API

Can you add the comment to both `Start` & `WatchWithDefault` ?

configmap/informer/informed_watcher_test.go
```diff
@@ -18,18 +18,15 @@ package informer
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
-	"k8s.io/apimachinery/pkg/util/wait"
 	fakekubeclientset "k8s.io/client-go/kubernetes/fake"
+	"sync"
+	"testing"
```
dprotaso, line 29: run `goimports` here

bobcatfish: booo whoops

bobcatfish:
@dprotaso I've updated the comments, lemme know what you think!

dprotaso (approved):
thanks for the changes!

PR #2038, opened 2021-02-25
Move IsKProbe to drainer, since it's the only place it's used

vagababov:
And hide it, since no need to export it, if nobody's going to use it.

🧹 

 /assign @tcnghia


vagababov:

metrics :/

vagababov:
Added one more method which is only used in drain.

tcnghia (approved):
/lgtm

PR #2039, opened 2021-02-26
adding a prototype schema dump tool to help keep CRD openAPI up to date.

n3wscott:
# Changes

- :gift: New schema generator based on reflection and go comments. 

/kind  enhancement

I found an interesting tool that @Harwayne built inside of Knative-GCP, I have cherrypicked that tool and created the start of a CLI from that code. Work in progress.

In use https://github.com/knative-sandbox/discovery/pull/153

**Release Note**

```release-note
Adding the start of a base cli for generating schema examples in place for downstream projects.
```


schema/schema/generate_schema.go
```diff
@@ -0,0 +1,267 @@
...
+
+func GenerateForType(t reflect.Type) JSONSchemaProps {
+	s := generateStructSchema(t, true)
+	return s
+}
+
+func generateSchema(t reflect.Type, history ...reflect.Type) JSONSchemaProps {
+	if visited(t, history) {
+		return JSONSchemaProps{
+			Type: "object",
```
antoineco, line 48: We could have constants for all those JSON types and formats that are currently repeated across the file.

schema/schema/generate_schema.go
```diff
@@ -0,0 +1,267 @@
...
+	case reflect.Array:
+		fallthrough
+	case reflect.Chan:
+		fallthrough
+	case reflect.Func:
+		fallthrough
+	case reflect.UnsafePointer:
+		panic(fmt.Errorf("can't handle kind %+v", t))
+	default:
+		panic(fmt.Errorf("unknown kind: %+v", t))
```
antoineco, line 158: Did you mean `%T` here, and in the previous case?

antoineco (approved):
First shallow review, looking really nice, and it's not even that much code 👍 

Leaving someone else an opportunity to look and give the LGTM.

schema/schema/generate_schema.go
n3wscott: t is already a reflected Type so `%T`  I think would spit out reflection.Type

n3wscott:
/hold

I am going to iterate one more time with this in use downstream.

n3wscott:
/unhold

I think this will be a good first pass.

n3wscott:
hilarious, the spellchecker bot does not like `Lorem Ipsum` based text...

schema/schema/generate_schema.go
antoineco: Ah right, fair point.

dprotaso:
Have you seen the prior exploration in https://github.com/knative/serving/issues/912 ? There are upstream tools (https://github.com/kubernetes-sigs/controller-tools) that help with schema generation that let you add markers/annotation that influence the output schema (ie. constraints). Although it has issues with serving, because we're an anomaly that embeds PodSpec, other WG groups may benefit from it.

Plus the upstream tooling is actively maintained while Harwayne wrote these generators as a one-off to get an initial schema. It still required manual edits after generation.

dprotaso:
I guess to state explicitly if the upstream tooling works (or requires changes that they're receptive to fixing or accepting contributions) I'd prefer to not maintain a crd generator ourselves.

n3wscott:
> There are upstream tools 

I am well aware of the upstream tools, but they do not work if you are not in that ecosystem. And for the most part, kubebuilder and operator sdk have abandoned things like the openapi schema generator tool in place of their own very opinionated ways of creating and maintaining CRDs. But it tends to be a heavy handed choice that you have to use their annotations and then mix it together with kustomize, or let their cli generate the entire CRD. <-- @dprotaso 

> re: https://github.com/knative/serving/issues/912

That is how I found the code that @Harwayne developed that I wrapped up in a base cli to allow for downstream vendoring. There is no answer in the https://github.com/knative/serving/issues/912 thread, we would have to take on their entire toolchain.

> why in pkg?

Happy to revisit that choice, I had it in test-infra but it needs to link with the project so it would create a dep back on test-infra. I could move it to hack. <-- @ImJasonH

n3wscott:
/hold

Looking at moving this to knative/hack

dprotaso:
> kubebuilder and operator sdk have abandoned things like the openapi schema generator tool in place of their own very opinionated ways of creating and maintaining CRDs

K8s structural schema has restrictions on the OpenAPI schema that's allowed. 

> their own very opinionated ways of creating and maintaining CRDs

The tooling seems modular enough that you can run the crd generator without all the other bits.

n3wscott:
that is the problem: `The tooling seems modular enough that you can run the crd generator without all the other bits.`

you have to pick up the entire CRD generator. you can't just use the schema generator.

n3wscott:
Moved to https://github.com/knative/hack/pull/51

PR #2041, opened 2021-03-02
Fix comment for skipKeyError

markusthoemmes:
As per title. Small fixeroo.

/assign @julz @mattmoor

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -585,7 +585,7 @@ func NewSkipKey(key string) error {
 	return skipKeyError{key: key}
 }
 
-// skipKeyError is an error that is indicating that a key was skipped.
+// skipKeyError is an error that indicates a key was skipped.
 // We should not re-queue keys when it returns with thus error in reconcile.
 type skipKeyError struct {
 	key string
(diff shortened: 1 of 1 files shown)


controller/controller.go
```diff
@@ -585,7 +585,7 @@ func NewSkipKey(key string) error {
 	return skipKeyError{key: key}
 }
 
-// permanentError is an error that is considered not transient.
+// skipKeyError is an error that is indicating that a key was skipped.
```
julz, line 588: super-nit, but since we're editing this line anyway :)

```suggestion
// skipKeyError is an error that indicates a key was skipped.
```

julz (approved):

mattmoor (approved):
/ok-to-test
/lgtm
/approve

controller/controller.go
```diff
@@ -585,7 +585,7 @@ func NewSkipKey(key string) error {
 	return skipKeyError{key: key}
 }
 
-// permanentError is an error that is considered not transient.
+// skipKeyError is an error that indicates a key was skipped.
 // We should not re-queue keys when it returns with thus error in reconcile.
```
vagababov, line 589:
```suggestion
// We should not re-queue keys when it returns with this error in reconcile.
```

PR #2045, opened 2021-03-05
Add WaitForCacheSyncQuick and use it in tests

markusthoemmes:
As per title really.

`cache.WaitForCacheSync` has a default poll interval of `100ms` which quickly adds up as each informer will take 100ms to setup in each test that's using this.

I'd love to use this for every occurrence of `cache.WaitForCacheSync` (the duck factories for instance remain slow for now), I've poked sig-api-machinery to see if they think there's any harm in reducing this, maybe even in K8s, globally. Having gone through the code, it seems like the `HasSynced` functions just go check the internal state of the queue, all-in-mem and kinda cheap. I guess people could've started to rely on `cache.WaitForCacheSync` to have some long'ish interval but at least everywhere where we control the callback to be an informer's default `HasSynced` function, we **should** be safe to switch this over everywhere.

/assign @julz @mattmoor @n3wscott

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


mattmoor (approved):
/lgtm

PR #2046, opened 2021-03-05
Slash off 5s from TestAdmissionValidResponseForResource

markusthoemmes:
# Changes

As per title. Waiting for 5s is **really** long in unit-test time imo, so let's shorten this significantly.

/assign @mattmoor @julz


vagababov (approved):
/lgtm

PR #2054, opened 2021-03-10
Implement getter machinery to generically fetch all injected clients

markusthoemmes:
# Changes

This is a stepping stone towards moer unit test stability around informer usage. See https://github.com/knative/serving/pull/10930 for an end to end example.

In a nutshell, we'd like to be able to fetch all injected clients generically to be able to muck with them and make sure everything is synced when we think it is.

/assign @n3wscott


markusthoemmes:
/retest

Prometheus stuff.

n3wscott (approved):
/lgtm

PR #2057, opened 2021-03-14
add filter label selector for serving fake pod informer factory setup

cdlliuy:
/kind bug

Update the faked testing context to facilitate the faked `filterinformaterfactory` setup. 

Without the correct faked context,  the PR https://github.com/knative/serving/pull/10266 always failed with UT by:
```
ok  	knative.dev/serving/pkg/reconciler/autoscaling/hpa/resources	0.080s	coverage: 100.0% of statements
--- FAIL: TestGlobalResyncOnUpdateAutoscalerConfigMap (0.00s)
    logger.go:130: 2021-03-13T01:25:20.733Z	PANIC	fake/fake_filtered_factory.go:47	Unable to fetch labelkey from context.
panic: Unable to fetch labelkey from context. [recovered]
	panic: Unable to fetch labelkey from context.
```

### DIFF
--- a/reconciler/testing/context.go
+++ b/reconciler/testing/context.go
@@ -39,16 +39,21 @@ import (
 )
 
 // SetupFakeContext sets up the the Context and the fake informers for the tests.
-func SetupFakeContext(t testing.TB) (context.Context, []controller.Informer) {
-	c, _, is := SetupFakeContextWithCancel(t)
+// The optional fs() can be used to edit ctx before the SetupInformer steps
+func SetupFakeContext(t testing.TB, fs ...func(context.Context) context.Context) (context.Context, []controller.Informer) {
+	c, _, is := SetupFakeContextWithCancel(t, fs...)
 	return c, is
 }
 
 // SetupFakeContextWithCancel sets up the the Context and the fake informers for the tests
 // The provided context can be canceled using provided callback.
-func SetupFakeContextWithCancel(t testing.TB) (context.Context, context.CancelFunc, []controller.Informer) {
+// The optional fs() can be used to edit ctx before the SetupInformer steps
+func SetupFakeContextWithCancel(t testing.TB, fs ...func(context.Context) context.Context) (context.Context, context.CancelFunc, []controller.Informer) {
 	ctx, c := context.WithCancel(logtesting.TestContextWithLogger(t))
 	ctx = controller.WithEventRecorder(ctx, record.NewFakeRecorder(1000))
+	for _, f := range fs {
+		ctx = f(ctx)
+	}
 	ctx, is := injection.Fake.SetupInformers(ctx, &rest.Config{})
 	return ctx, c, is
 }


reconciler/testing/context.go
```diff
@@ -53,6 +53,17 @@ func SetupFakeContextWithCancel(t zaptest.TestingT) (context.Context, context.Ca
 	return ctx, c, is
 }
 
+// SetupFakeCustomizedContextWithCancel sets up the the Context and the fake informers for the tests
+// The provided context can be canceled using provided callback.
+// The provided customizeCtxFunc() can be used to edit ctx before the SetupInformer steps
+func SetupFakeCustomizedContextWithCancel(t zaptest.TestingT, customizeCtxFunc func(context.Context) context.Context) (context.Context, context.CancelFunc, []controller.Informer) {
+	ctx, c := context.WithCancel(logtesting.TestContextWithLogger(t))
+	ctx = controller.WithEventRecorder(ctx, record.NewFakeRecorder(1000))
+	ctx = customizeCtxFunc(ctx)
```
markusthoemmes, line 62: Can you describe how this is used? Could we instead pass a context to `SetupFakeContext` (or rather add a new function like `EnableFakeContext(ctx...)`?

cdlliuy: @markusthoemmes , the usage is demostrated in PR #10026 
https://github.com/knative/serving/pull/10266/files#diff-66f7c5c9446e2a41cfa409d097a7bd20ea73a45b684d93e32aed0fbbbb39f12bR1373-R1375

cdlliuy: I think we can add a new function named `EnableFakeContext(ctx...)` , but that required all the autoscaler related testcases need to execute the `ctx` initialization before calling the `SetupFakContextWithCancel`, i.e.
```
ctx, cancel := context.WithCancel(logtesting.TestContextWithLogger(t))
ctx = filteredinformerfactory.WithSelectors(ctx, serving.RevisionUID)
ctx, infs := EnableFakeContext(t, ctx)
```

Then, inside EnableFakeContext func , only 2 lines left. 
```
func EnableFakeContext(t zaptest.TestingT,context.Context) (context.Context, []controller.Informer) {
	ctx = controller.WithEventRecorder(ctx, record.NewFakeRecorder(1000))
	ctx, is := injection.Fake.SetupInformers(ctx, &rest.Config{})
	return ctx,  is
}
```

markusthoemmes: Is the logger required before calling `filteredinformerfactory.WithSelectors`?

cdlliuy: I don't think so, it is just to setup a ctx.WithValue() inside `filteredinformerfactory.WithSelectors`

cdlliuy: @markusthoemmes  any further comment based on above discussion?  
Your preference of how the further serving/scaling related UT testcase looks like?    I still tend to use the sample showed in https://github.com/knative/serving/pull/10266/files#diff-66f7c5c9446e2a41cfa409d097a7bd20ea73a45b684d93e32aed0fbbbb39f12bR1450-R1452

cdlliuy: @markusthoemmes  ^^

cdlliuy: @markusthoemmes   can you help to check the above comment , and let me your idea on this?

markusthoemmes: So, I've whipped up an example commit with what I had in mind here: https://github.com/knative/serving/commit/a95df7c12008235eec869d820d625b549a23ec64

I believe that'd lead to more reusability of those helpers without having to keep adding more and more `WithXXX` handlers.

One alternative that comes to mind is to add the mutation functions as variadic args to the existing functions, like so

```golang
// SetupFakeContext sets up the the Context and the fake informers for the tests.
func SetupFakeContext(t zaptest.TestingT, fs ...func(context.Context)) (context.Context, []controller.Informer) {
	c, _, is := SetupFakeContextWithCancel(t, fs...)
	return c, is
}

// SetupFakeContextWithCancel sets up the the Context and the fake informers for the tests
// The provided context can be canceled using provided callback.
func SetupFakeContextWithCancel(t zaptest.TestingT, fs ...func(context.Context)) (context.Context, context.CancelFunc, []controller.Informer) {
	ctx, c := context.WithCancel(logtesting.TestContextWithLogger(t))
	ctx = controller.WithEventRecorder(ctx, record.NewFakeRecorder(1000))
	for _, f := range fs {
		f(ctx)
	}
	ctx, is := injection.Fake.SetupInformers(ctx, &rest.Config{})
	return ctx, c, is
}
```

Those should not break existing usages (since the argument is variadic) but allow for "optional" customization of the context.

cdlliuy: yes,  I tends to use the 2nd approach with the optional fs passed in a parameter.  I will refine the PR with the suggestion

cdlliuy: @markusthoemmes ,     I updated the PR with the 2nd approach.   Also,  https://github.com/knative/serving/pull/10266/files#diff-66f7c5c9446e2a41cfa409d097a7bd20ea73a45b684d93e32aed0fbbbb39f12bR1210-R1212  shows the usage of the current proposal.  

Can you help to review again/

markusthoemmes (approved):
/lgtm
/approve

PR #2072, opened 2021-03-29
Enhance controller to pass context

jinchihe:
# Changes

I defined the key/value in context, but cannot pass to reconcile, the root cause is the context is not passed to `processNextWorkItem` below:
https://github.com/knative/pkg/blob/9d92ea16c0d3b2b03ac456782563321030726280/controller/controller.go#L468

The PR is going to fix the issue to pass context. Thanks.

/kind bug

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -458,7 +458,7 @@ func (c *Impl) RunContext(ctx context.Context, threadiness int) error {
 		sg.Add(1)
 		go func() {
 			defer sg.Done()
-			for c.processNextWorkItem() {
+			for c.processNextWorkItem(ctx) {
 			}
 		}()
 	}
@@ -483,7 +483,7 @@ func (c *Impl) Run(threadiness int, stopCh <-chan struct{}) error {
 
 // processNextWorkItem will read a single work item off the workqueue and
 // attempt to process it, by calling Reconcile on our Reconciler.
-func (c *Impl) processNextWorkItem() bool {
+func (c *Impl) processNextWorkItem(ctx context.Context) bool {
 	obj, shutdown := c.workQueue.Get()
 	if shutdown {
 		return false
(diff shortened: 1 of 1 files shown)


markusthoemmes (commented):
Could you elaborate on why you need this? I think we're not passing the "big" context to each reconcile by design currently.

jinchihe:
@markusthoemmes Thanks. I made a new controller basic on knative framework, and I would like to pass some args from controller main to reconcile by `context` so that the `reconcile` can handle some logic basic on the args, but I found the org context is not passed, and create new context in `processNextWorkItem`. Personally I think the context should be coherent, and not sure how big the context for general cases.

by the way, here shows the  `controller/controller.go` is conflicting, that's weird, I fetch and rebase, nothing new, what's problem here? Thanks.
```
➜  pkg git:(enhance_controller_to_pass_ctx) git fetch upstream
➜  pkg git:(enhance_controller_to_pass_ctx) git rebase upstream/master
Current branch enhance_controller_to_pass_ctx is up to date.
➜  pkg git:(enhance_controller_to_pass_ctx) git push origin enhance_controller_to_pass_ctx -f
Everything up-to-date
➜  pkg git:(enhance_controller_to_pass_ctx)
```

markusthoemmes:
The current way of sharing information from `NewController` into `Reconcile` is by adding them to the reconciler type. Is that not possible here?

jinchihe:
@markusthoemmes Thank you. yeah, in fact I can get this by others such as define a ENV or read from config map etc.. my point is if we need ensure the context to be coherent. Thanks.

markusthoemmes:
At least currently, I think the context's not inheriting from each other is by-design. Summoning @n3wscott for input.

jinchihe:
OK So I will close the PR, thanks @markusthoemmes

PR #2061, opened 2021-03-18
Add skonto to metrics reviewers

skonto:
Interested in reviewing PRs in this area.

PR #2078, opened 2021-03-31
Update OWNERS_ALIASES to match autogen in community

evankanderson:
See https://github.com/knative/community/pull/552 for the addition of groups/full OWNERS_ALIASES that will be synced.

This is a preparatory PR for landing OWNERS_ALIASES automation across the Knative org (alternative to the CODEOWNERS changes in knative-sandbox).

If this seems to be expanding permissions or otherwise likely to cause a problem, `/hold` this PR and comment here or on https://github.com/knative/community/pull/552.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,12 @@
 # The OWNERS file is used by prow to automatically merge approved PRs.
 
 approvers:
-- pkg-approvers
+- technical-oversight-committee
+- dprotaso
+- grantr
+- markusthoemmes
+- mattmoor
+- tcnghia
+- vagababov
+- vaikas
+- n3wscott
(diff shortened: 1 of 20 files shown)


evankanderson:
/assign @vagababov @markusthoemmes @dprotaso 

Picking a few people who may have opinions here.

/hold

In case one of those three wants to raise objections.

dprotaso:
Sorry - I gave precedence to this https://github.com/knative/pkg/pull/2061 cause it was first

evankanderson (commented):
Updated, added `skonto` directly since there was only one use of `metrics-reviewers` in the OWNERS files. If we need more directories with the same OWNERS, we can add them via peribolos to the autogen in the future.

dprotaso:
/lgtm
/approve

Going to let someone else unhold so we get a second set of eyes

markusthoemmes (approved):
LGTM

I'd say let's pull this in and fix any aftermath if we have any. The contents of this are reasonable from my PoV, but it's a bit hard to review.

PR #2079, opened 2021-04-01
Simplify loading of local kubeconfig in sharedmain

markusthoemmes:
# Changes

`RecommendedHomeFile` seems to be precisely what we're after, dealing with some windows shenanigans probably better than we do currently, so why not? :shrug: 

/assign @n3wscott @vagababov


dprotaso:
/retest

unf metrics

PR #2052, opened 2021-03-09
Add benchmark for metrics recording

skonto:
* Adds a benchmark to be used as a baseline for future changes or migrations.
This is helpful especially because metrics reporting happens in many places and after some user request eg. [eventing](https://github.com/knative/eventing/blob/149ebe660b60f43975fba414034a7c90682e8804/pkg/adapter/v2/cloudevents.go#L120).
/cc @evankanderson 
I used this benchmark to compare with branch release-0.20. Results are bellow.
Possibly due to our change in #2031 moving to a channel implementation we add some overhead ~1ms but for overloaded
pods with many goroutines things get better (I will compare in the future with using atomic values) probably due to [this](https://stackoverflow.com/questions/57562606/why-does-sync-mutex-largely-drop-performance-when-goroutine-contention-is-more-t).
It would be nice numbers to be validated independently (or percentage diffs).

**main**
```
$go test -run=^$ -test.v -test.benchmem=true -test.bench=^BenchmarkMetricsRecording$ ./metrics/ 
goos: linux
goarch: amd64
pkg: knative.dev/pkg/metrics
BenchmarkMetricsRecording
BenchmarkMetricsRecording/sequential
BenchmarkMetricsRecording/sequential-12         	  716590	      2110 ns/op	     368 B/op	       9 allocs/op
BenchmarkMetricsRecording/parallel
BenchmarkMetricsRecording/parallel-12           	  356906	      3212 ns/op	     368 B/op	       9 allocs/op
BenchmarkMetricsRecording/sequential-batch
BenchmarkMetricsRecording/sequential-batch-12   	  568777	      2252 ns/op	     400 B/op	       9 allocs/op
BenchmarkMetricsRecording/parallel-batch
BenchmarkMetricsRecording/parallel-batch-12     	  295448	      3850 ns/op	     400 B/op	       9 allocs/op
PASS
ok  	knative.dev/pkg/metrics	6.185s


$ go test -run=^$ -test.v -test.benchmem=true -cpu=3000 -test.bench=^BenchmarkMetricsRecording$ ./metrics/ 
goos: linux
goarch: amd64
pkg: knative.dev/pkg/metrics
BenchmarkMetricsRecording
BenchmarkMetricsRecording/sequential
BenchmarkMetricsRecording/sequential-3000         	  540606	      1984 ns/op	     368 B/op	       9 allocs/op
BenchmarkMetricsRecording/parallel
[...]

PR #2080, opened 2021-04-01
Add benchmarking for the http tracing handler

skonto:
* Adds a benchmark to be used as a baseline for future changes or migrations. This is helpful especially because tracing happens in several places like activator, queue-proxy. Note that eventing uses [direct calls](https://github.com/knative/eventing/blob/8f35d42544e064783f344538d2957c753e31c820/pkg/broker/ingress/ingress_handler.go#L141-L152) to opencesus lib eg. `trace.StartSpan()` in its event handlers and sets span attributes explicitly in several occasions (this is lighter compared to the handler we benchmark here). I am wondering if we can do the same for Serving as trace.StartSpan() is run anyway as part of the HTTPSpanMiddleware (worth investigating next and use this benchmark to compare changes). Maybe create an optimized handler to be used with activator, queue-proxy? 
Allocations btw seem to be pretty high check bellow. /cc @evankanderson 

```
$ go test   -benchmem -memprofile memprofile.out -cpuprofile profile.out -test.v -test.bench BenchmarkSpanMiddleware -test.run ^$ ./tracing/
goos: linux
goarch: amd64
pkg: knative.dev/pkg/tracing
BenchmarkSpanMiddleware
BenchmarkSpanMiddleware/sequential
BenchmarkSpanMiddleware/sequential-12         	  133786	      9358 ns/op	    6718 B/op	      97 allocs/op
BenchmarkSpanMiddleware/parallel
BenchmarkSpanMiddleware/parallel-12           	  300289	      4129 ns/op	    6586 B/op	      97 allocs/op
PASS
ok  	knative.dev/pkg/tracing	3.025s

```

From sequential execution:
```
Dropped 76 nodes (cum <= 12.05ms)
Showing top 10 nodes out of 142
      flat  flat%   sum%        cum   cum%
     480ms 19.92% 19.92%      910ms 37.76%  runtime.scanobject
     190ms  7.88% 27.80%      190ms  7.88%  runtime.procyield
     160ms  6.64% 34.44%      240ms  9.96%  runtime.findObject
     130ms  5.39% 39.83%      400ms 16.60%  runtime.mallocgc
     110ms  4.56% 44.40%      110ms  4.56%  runtime.pageIndexOf (inline)
      70ms  2.90% 47.30%       70ms  2.90%  runtime.memclrNoHeapPointers
      60ms  2.49% 49.79%       60ms  2.49%  runtime.memmove
      60ms  2.49% 52.28%       70ms  2.90%  runtime.spanOf
      50ms  2.07% 54.36%      170ms  7.05%  runtime.greyobject
      50ms  2.07% 56.43%       80ms  3.32%  runtime.heapBitsSetType
```
It means that for the most part there is gc work going on here. Let's check gc allocation:
```
Dropped 11 nodes (cum <= 63402)
Showing top 10 nodes out of 75
      flat  flat%   sum%        cum   cum%
   1327148 10.47% 10.47%    1327148 10.47%  go.opencensus.io/tag.Upsert
    819212  6.46% 16.93%    2306801 18.19%  go.opencensus.io/trace.(*span).copyToCappedAttributes
    699082  5.51% 22.44%     699082  5.51%  container/list.(*List).insertValue (inline)
    654138  5.16% 27.60%    1353220 10.67%  github.com/golang/groupcache/lru.(*Cache).Add
    546149  4.31% 31.90%    1043150  8.23%  go.opencensus.io/trace.newLruMap
    524333  4.13% 36.04%     622638  4.91%  contrib.go.opencensus.io/exporter/zipkin.zipkinSpan
    466960  3.68% 39.72%    1618794 12.77%  go.opencensus.io/plugin/ochttp.(*trackingResponseWriter).end.func1
    447849  3.53% 43.25%     447849  3.53%  context.WithValue
    400255  3.16% 46.41%    2726550 21.50%  go.opencensus.io/plugin/ochttp.(*Handler).startStats
    382302  3.01% 49.43%     382302  3.01%  go.opencensus.io/tag.newMap (inline)
```
By digging into [startStats](https://github.com/census-instrumentation/opencensus-go/blob/de37041b23a4ac17ac695473ec3876c40a4b65f8/plugin/ochttp/server.go#L149) I see that similar to the findings in [metrics benchmarking](https://github.com/knative/pkg/pull/2052#issuecomment-795296071), mutators and other tag/attributes manipulation cause significant overhead.

### DIFF
--- a/tracing/http_test.go
+++ b/tracing/http_test.go
@@ -166,3 +166,49 @@ func TestHTTPSpanIgnoringPaths(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkSpanMiddleware(b *testing.B) {
+	cfg := config.Config{
+		Backend: config.Zipkin,
+		Debug:   true,
+	}
+
+	// Create tracer with reporter recorder
+	reporter, co := FakeZipkinExporter()
+	oct := NewOpenCensusTracer(co)
+	b.Cleanup(func() {
+		reporter.Close()
+		oct.Finish()
+	})
+
+	if err := oct.ApplyConfig(&cfg); err != nil {
+		b.Error("Failed to apply tracer config:", err)
+	}
+
+	next := testHandler{}
+	middleware := HTTPSpanMiddleware(&next)
+
+	fw := &fakeWriter{}
+
+	req, err := http.NewRequest("GET", "http://test.example.com", nil)
+	if err != nil {
+		b.Fatal("Failed to make fake request:", err)
+	}
+	const traceID = "821e0d50d931235a5ba3fa42eddddd8f"
+	req.Header["X-B3-Traceid"] = []string{traceID}
+	req.Header["X-B3-Spanid"] = []string{"b3bd5e1c4318c78a"}
+
+	b.Run("sequential", func(b *testing.B) {
+		for j := 0; j < b.N; j++ {
+			middleware.ServeHTTP(fw, req)
+		}
+	})
+
+	b.Run("parallel", func(b *testing.B) {
+		b.RunParallel(func(pb *testing.PB) {
+			for pb.Next() {
+				middleware.ServeHTTP(fw, req)
+			}
+		})
+	})
+}
--- a/tracing/http_test.go
+++ b/tracing/http_test.go
@@ -41,17 +41,17 @@ func (fw *fakeWriter) Write(data []byte) (int, error) {
 
 func (fw *fakeWriter) WriteHeader(statusCode int) {}
 
-type dummyWriter struct {}
+type benchWriter struct{}
 
-func (dw *dummyWriter) Header() http.Header {
+func (dw *benchWriter) Header() http.Header {
 	return http.Header{}
 }
 
-func (dw *dummyWriter) Write(data []byte) (int, error) {
+func (dw *benchWriter) Write(data []byte) (int, error) {
 	return 0, nil
 }
 
-func (dw *dummyWriter) WriteHeader(statusCode int) {}
+func (dw *benchWriter) WriteHeader(statusCode int) {}
 
 type testHandler struct{}
 
@@ -200,7 +200,7 @@ func BenchmarkSpanMiddleware(b *testing.B) {
 	next := testHandler{}
 	middleware := HTTPSpanMiddleware(&next)
 
-	dw := &dummyWriter{}
+	bw := &benchWriter{}
 
 	req, err := http.NewRequest(http.MethodGet, "http://test.example.com", nil)
 	if err != nil {
@@ -212,14 +212,14 @@ func BenchmarkSpanMiddleware(b *testing.B) {
 
 	b.Run("sequential", func(b *testing.B) {
 		for j := 0; j < b.N; j++ {
-			middleware.ServeHTTP(dw, req)
+			middleware.ServeHTTP(bw, req)
 		}
 	})
 
 	b.Run("parallel", func(b *testing.B) {
 		b.RunParallel(func(pb *testing.PB) {
 			for pb.Next() {
-				middleware.ServeHTTP(dw, req)
+				middleware.ServeHTTP(bw, req)
 			}
 		})
 	})
--- a/tracing/http_test.go
+++ b/tracing/http_test.go
@@ -41,6 +41,18 @@ func (fw *fakeWriter) Write(data []byte) (int, error) {
 
 func (fw *fakeWriter) WriteHeader(statusCode int) {}
 
+type dummyWriter struct {}
+
+func (dw *dummyWriter) Header() http.Header {
+	return http.Header{}
+}
+
+func (dw *dummyWriter) Write(data []byte) (int, error) {
+	return 0, nil
+}
+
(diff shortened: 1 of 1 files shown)


markusthoemmes:
> I am wondering if we can do the same for Serving as trace.StartSpan() is run anyway as part of the HTTPSpanMiddleware (worth investigating next and use this benchmark to compare changes). Maybe create an optimized handler to be used with activator, queue-proxy?

Both the activator and the queue-proxy already do this, so I'd be curious what we're even getting from this handler.

tracing/http_test.go
```diff
@@ -166,3 +166,49 @@ func TestHTTPSpanIgnoringPaths(t *testing.T) {
...
+	if err := oct.ApplyConfig(&cfg); err != nil {
+		b.Error("Failed to apply tracer config:", err)
+	}
+
+	next := testHandler{}
+	middleware := HTTPSpanMiddleware(&next)
+
+	fw := &fakeWriter{}
+
+	req, err := http.NewRequest("GET", "http://test.example.com", nil)
```
markusthoemmes, line 193: nit

```suggestion
	req, err := http.NewRequest(http.MethodGet, "http://test.example.com", nil)
```

tracing/http_test.go
```diff
@@ -166,3 +166,49 @@ func TestHTTPSpanIgnoringPaths(t *testing.T) {
...
+	})
+
+	if err := oct.ApplyConfig(&cfg); err != nil {
+		b.Error("Failed to apply tracer config:", err)
+	}
+
+	next := testHandler{}
+	middleware := HTTPSpanMiddleware(&next)
+
+	fw := &fakeWriter{}
```
markusthoemmes, line 191: Is anything ever calling `Write` on this? That function would cause additional allocations I believe if called.

markusthoemmes (commented):
I'd love to understand what this handler even does for us. Even if disabled, it causes quite a bit of overhead, see https://github.com/knative/serving/pull/11016.

We're already manually adding traces in both the activator and the queue-proxy so... does this even add any value?

tracing/http_test.go
skonto: It is the baseline because the opencensus handler will not live alone but part of an http handler chain anyway. It's overhead seems minimal from the call graph:
image

skonto: I can create a dummy doing nothing.

markusthoemmes: Yeah if the tracing handler doesn't call Write (which I guess it shouldn't anyway), we should be fine.

skonto: Actually its one allocation less (pretty predictible):
```
$ go test   -benchmem -memprofile memprofile.out -cpuprofile profile.out -test.v -test.bench BenchmarkSpanMiddleware -test.run ^$ ./tracing/
goos: linux
goarch: amd64
pkg: knative.dev/pkg/tracing
BenchmarkSpanMiddleware
BenchmarkSpanMiddleware/sequential
BenchmarkSpanMiddleware/sequential-12         	  141504	     10679 ns/op	    6911 B/op	      97 allocs/op
BenchmarkSpanMiddleware/parallel
BenchmarkSpanMiddleware/parallel-12           	  661308	      4737 ns/op	    6742 B/op	      97 allocs/op
PASS
ok  	knative.dev/pkg/tracing	5.204s
```

markusthoemmes: Okay, no biggy then. Fine to keep as-is.

skonto: It seems it calls the handler [here](https://github.com/census-instrumentation/opencensus-go/blob/de37041b23a4ac17ac695473ec3876c40a4b65f8/plugin/ochttp/server.go#L92).

skonto:
```

skonto:
```TestMetricsExport``` again. 
/retest

skonto:
@markusthoemmes I see that the [broker ingress handler](https://github.com/knative/eventing/blob/main/pkg/broker/ingress/ingress_handler.go) and other handlers do not mix with the special [opencensus http handler](https://github.com/knative/pkg/blob/main/tracing/http.go#L43). This means that they cannot extract the request headers and add them to the local context (In Serving it is the opposite, for activator and queue-proxy). `Trace.StartSpan()` used by broker ingress handler only picks up the existing tracing info if already exists in current ctx. The special handler uses a [separate call](https://github.com/census-instrumentation/opencensus-go/blob/de37041b23a4ac17ac695473ec3876c40a4b65f8/plugin/ochttp/server.go#L113) to extract tracing info from the request context eg. B3 headers. If you check the example [here](https://knative.slack.com/archives/C017X0PFC0P/p1616513118038700?thread_ts=1615981746.000200&cid=C017X0PFC0P) with istio enabled (istio propagates headers) ping source span is not shown as the parent span of the broker ingress span. However the chain looks ok bellow starting with the broker ingress span. Btw since tracing is enabled ping source should send tracing headers since it uses the observed client from cloudevents sdk. It seems like a bug to me at the Eventing side.

markusthoemmes (approved):
/lgtm
/approve

PR #2081, opened 2021-04-02
Fix default domain name

wzshiming:
Signed-off-by: Shiming Zhang <wzshiming@foxmail.com>

# Changes

-
-
-

/kind bug

Fixes #

**Release Note**

```release-note
NONE
```

**Docs**

```docs
NONE
```

### DIFF
--- a/network/domain.go
+++ b/network/domain.go
@@ -46,6 +46,7 @@ func GetClusterDomainName() string {
 	once.Do(func() {
 		f, err := os.Open(resolverFileName)
 		if err != nil {
+			domainName = defaultDomainName
 			return
 		}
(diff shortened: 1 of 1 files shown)


network/domain.go
```diff
@@ -46,6 +46,7 @@ func GetClusterDomainName() string {
 	once.Do(func() {
 		f, err := os.Open(resolverFileName)
 		if err != nil {
+			domainName = defaultDomainName
```
markusthoemmes, line 49: This seems like not doing anything since this is set by default above, or am I missing something? In which case would this be necessary?

wzshiming:
Sorry for misreading

PR #2088, opened 2021-04-08
Try fix TestMetricsExport 

skonto:
TestMetricsExport creates a lot of failures on CI lately. Here is a case where [exporter never got updated](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_pkg/2082/pull-knative-pkg-unit-tests/1380101339379404800/build-log.txt). This means most likely that a previous test already had the backend set as expected but the port was different than the [one expected later on](https://github.com/knative/pkg/pull/2005#issuecomment-776174080).
I guess this happens only in tests and https://github.com/knative/pkg/issues/1672 is not fixed. 
This fix makes logic about prometheus reconfiguration correct for testing purposes.

/cc @evankanderson

### DIFF
--- a/metrics/exporter.go
+++ b/metrics/exporter.go
@@ -183,6 +183,10 @@ func isNewExporterRequired(newConfig *metricsConfig) bool {
 		return newConfig.collectorAddress != cc.collectorAddress || newConfig.requireSecure != cc.requireSecure
 	}
 
+	if newConfig.backendDestination == prometheus {
+		return newConfig.prometheusHost != cc.prometheusHost || newConfig.prometheusPort != cc.prometheusPort
+	}
+
 	return newConfig.backendDestination == stackdriver && newConfig.stackdriverClientConfig != cc.stackdriverClientConfig
 }
 


metrics/exporter.go
```diff
@@ -183,6 +183,10 @@ func isNewExporterRequired(newConfig *metricsConfig) bool {
 		return newConfig.collectorAddress != cc.collectorAddress || newConfig.requireSecure != cc.requireSecure
 	}
 
+	if newConfig.backendDestination == prometheus {
+		return newConfig.prometheusHost != cc.prometheusHost || newConfig.prometheusPort != cc.prometheusPort
+	}
+
```
skonto, line 189: I am wondering why we didn't have this here before?

skonto:
@markusthoemmes pls review.

skonto:
@evankanderson pls review.

n3wscott (approved):
/lgtm
/approve

metrics/exporter.go
evankanderson: I wonder if all the effort to avoid reloading the config is too clever by half, but this seems like a good fix for now.

evankanderson (commented):
/lgtm
/approve

markusthoemmes:
failing test seems related. It's the test that's supposed to be fixed, I  guess?

evankanderson:
The failing test is in the OpenCensus path, so it's not directly related to this PR. It looks like maybe we missed an export RPC in that test run.

skonto:
@evankanderson The failing path is due to the timeout where we dont get any records (assuming there is no race there or some bug at the opencensus side). Prometheus has a 10sec delay. Maybe if we are more generous to opencensus this will not happen.

evankanderson:
The last time time I fixed this, I discovered that we were using the first four reports, rather than waiting for four different series. I'm not sure what's happened in this case.

PR #2089, opened 2021-04-08
Add possibility to implement a deletion handler in a reconciler.

### LINKED ISSUE #2021: Rethink "ObserveFinalizeKind"
See https://github.com/knative-sandbox/net-kourier/issues/448 on why the current behavior is probably buggy and should not be encouraged.

We should however come up with a way to safely observe deletion events, even in scaled out controllers. Copying a thought of mine from the original issue:

> Maybe we can also offer similar callbacks for the "get-but-not-found-so-i-guess-deleted" flow? I.e. an Deleted/ObserveDeleted set of functions that determine deletion via the mechanism we're after here? That'd remove the potential races I think.

/cc @mattmoor

markusthoemmes:
Fixes #2021

# Changes

ObserveFinalizeKind is not guaranteed to be called for every resource
that gets deleted, even if they do have a finalizer. The leader can race
the observers in removing the finalizer so the observers would not even
see the deletion.

Using the OnDelete handler on the informer is equally racy in that it
can trigger the deletion handler while a potential "normal" reconcile is
still in-flight.

This adds an ObserveDeletedKind handler that can be implemented by
reconcilers. The events still go via the workQueue, so proper order and
deduplication is guaranteed. Observation is guaranteed as well. In most
if not all cases, this handler should replace the ObserveFinalizeKind
handler.

Will for example be used in https://github.com/knative-sandbox/net-kourier/issues/448

/assign @n3wscott

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -111,6 +111,7 @@ func (g *reconcilerReconcilerGenerator) GenerateType(c *generator.Context, t *ty
 		"reconcilerReconcilerEvent":      c.Universe.Type(types.Name{Package: "knative.dev/pkg/reconciler", Name: "ReconcilerEvent"}),
 		"reconcilerRetryUpdateConflicts": c.Universe.Function(types.Name{Package: "knative.dev/pkg/reconciler", Name: "RetryUpdateConflicts"}),
 		"reconcilerConfigStore":          c.Universe.Type(types.Name{Name: "ConfigStore", Package: "knative.dev/pkg/reconciler"}),
+		"reconcilerOnDeletionInterface":  c.Universe.Type(types.Name{Package: "knative.dev/pkg/reconciler", Name: "OnDeletionInterface"}),
 		// Deps
 		"clientsetInterface": c.Universe.Type(types.Name{Name: "Interface", Package: g.clientsetPkg}),
 		"resourceLister":     c.Universe.Type(types.Name{Name: g.listerName, Package: g.listerPkg}),
@@ -410,8 +411,15 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 	original, err := getter.Get(s.name)
 
 	if {{.apierrsIsNotFound|raw}}(err) {
-		// The resource may no longer exist, in which case we stop processing.
+		// The resource may no longer exist, in which case we stop processing and call
+		// the ObserveDeletion handler if appropriate.
 		logger.Debugf("Resource %q no longer exists", key)
+		if del, ok := r.reconciler.({{.reconcilerOnDeletionInterface|raw}}); ok {
+			return del.ObserveDeletion(ctx, {{.typesNamespacedName|raw}}{
+				Namespace: s.namespace,
+				Name: s.name,
+			})
+		}
 		return nil
 	} else if err != nil {
 		return err
--- a/reconciler/configstore.go
+++ b/reconciler/configstore.go
@@ -1,7 +1,7 @@
 /*
 Copyright 2020 The Knative Authors
 
-Licensed under the Apache License, Veroute.on 2.0 (the "License");
+Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
@@ -16,9 +16,7 @@ limitations under the License.
 
 package reconciler
 
-import (
-	"context"
-)
+import "context"
 
 // ConfigStore is used to attach the frozen configuration to the context.
 type ConfigStore interface {
--- a/reconciler/configstore.go
+++ b/reconciler/configstore.go
@@ -16,10 +16,26 @@ limitations under the License.
 
 package reconciler
 
-import "context"
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/types"
+)
 
 // ConfigStore is used to attach the frozen configuration to the context.
 type ConfigStore interface {
 	// ConfigStore is used to attach the frozen configuration to the context.
 	ToContext(ctx context.Context) context.Context
+}
+
+// OnDeletionInterface defines the strongly typed interface to be implemented by a
+// controller observing a deletion of {{.type|raw}}. Every controller that was active
+// during the deletion of the respective resource is guaranteed to observe this event,
+// leader or not. It's usually used to clear up in-memory state regarding the respective
+// resource. Finalizers should be used to ensure external resources are properly cleaned
+// up.
+type OnDeletionInterface interface {
+	// ObserveDeletion implements custom logic to observe deletion of the respective resource
+	// with the given key.
+	ObserveDeletion(ctx context.Context, key types.NamespacedName) error
 }
--- a/reconciler/configstore.go
+++ b/reconciler/configstore.go
@@ -18,24 +18,10 @@ package reconciler
 
 import (
 	"context"
-
-	"k8s.io/apimachinery/pkg/types"
 )
 
 // ConfigStore is used to attach the frozen configuration to the context.
 type ConfigStore interface {
 	// ConfigStore is used to attach the frozen configuration to the context.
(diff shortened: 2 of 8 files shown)


reconciler/configstore.go
```diff
@@ -16,10 +16,26 @@ limitations under the License.
...
+// OnDeletionInterface defines the strongly typed interface to be implemented by a
+// controller observing a deletion of {{.type|raw}}. Every controller that was active
+// during the deletion of the respective resource is guaranteed to observe this event,
+// leader or not. It's usually used to clear up in-memory state regarding the respective
+// resource. Finalizers should be used to ensure external resources are properly cleaned
+// up.
+type OnDeletionInterface interface {
+	// ObserveDeletion implements custom logic to observe deletion of the respective resource
+	// with the given key.
+	ObserveDeletion(ctx context.Context, key types.NamespacedName) error
```
markusthoemmes, line 40: opted for returning `error` vs. `reconciler.Event` due to the unavailablity of the entire `runtime.Object` on this path. Open to suggestions.

vagababov (commented):
/lgtm

reconciler/configstore.go
n3wscott: Observers should not return anything seems like?

n3wscott: why is this in config store?

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -410,8 +411,15 @@ func (r *reconcilerImpl) Reconcile(ctx {{.contextContext|raw}}, key string) erro
 	original, err := getter.Get(s.name)
 
 	if {{.apierrsIsNotFound|raw}}(err) {
-		// The resource may no longer exist, in which case we stop processing.
+		// The resource may no longer exist, in which case we stop processing and call
+		// the ObserveDeletion handler if appropriate.
 		logger.Debugf("Resource %q no longer exists", key)
+		if del, ok := r.reconciler.({{.reconcilerOnDeletionInterface|raw}}); ok {
+			return del.ObserveDeletion(ctx, {{.typesNamespacedName|raw}}{
```
n3wscott, line 418: I would not return anything here, no reason to requeue on error, the resource is gone

markusthoemmes: Doesn't that make retrying errors a bit easier? It'd go through the same codepath again and error handling would be similar to the usual reconciler functions.

n3wscott: the observer does not need to retry, I see this as an FYI and it should have no fail path, and should not be able to requeue the key.

Do you have a usecase for an observer that is able to re-enqueue?

markusthoemmes: Yep, https://github.com/knative-sandbox/net-kourier/blob/7387060469bcbeccb8e99e8abe141a2647c507c3/pkg/reconciler/ingress/ingress.go#L115-L150 for instance. Updating the envoy config there **can** error out and I'm not quite clear how I'd resolve such error if the reconciler framework doesn't do it for me.

I do see your point that this can be handled as more of an FYI, but I do believe it's valuable to offer the error return + retry facilities that reconcilers usually have as well.

n3wscott: my pushback is because it does not follow the other readonly methods we define: 

```
type ReadOnlyInterface interface {
	// ObserveKind implements logic to observe {{.type|raw}}.
	// This method should not write to the API.
	ObserveKind(ctx {{.contextContext|raw}}, o *{{.type|raw}}) {{.reconcilerEvent|raw}}
}
// ReadOnlyFinalizer defines the strongly typed interfaces to be implemented by a
// controller finalizing {{.type|raw}} if they want to process tombstoned resources
// even when they are not the leader.  Due to the nature of how finalizers are handled
// there are no guarantees that this will be called.
type ReadOnlyFinalizer interface {
	// ObserveFinalizeKind implements custom logic to observe the final state of {{.type|raw}}.
	// This method should not write to the API.
	ObserveFinalizeKind(ctx {{.contextContext|raw}}, o *{{.type|raw}}) {{.reconcilerEvent|raw}}
}
```

n3wscott: interesting, ok so the diff here is we are talking about Kind based reconcilers, but yours is for queue key based... until this point we have not been leaking the queue key. But I suppose you no longer have the deleted object to make a:

```
type ReadOnlyDeletion interface {
	ObserveDeleteKind(ctx {{.contextContext|raw}}, o *{{.type|raw}}) {{.reconcilerEvent|raw}}
}
```

markusthoemmes: Right, the deleted object is already gone.

reconciler/deletion.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Veroute.on 2.0 (the "License");
+you may not use this file except in compliance with the License.
```
n3wscott, line 5: lol on `Veroute.on 2.0 (the "License");`

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
n3wscott: In my opinion, we should not allow an observation method to return an error that will cause the key to be requeued.
Additionally, I do not think we should pass the key as a string to a consumer because it leaks more information that I would like for future changes that are "planned".

How about this spin on this idea:

```
ObserveDeletion(ctx context.Context, ref duckv1.KReference)
```

changes would be:

1. Drop the return error.
2. Drop the queue key string in favor of a object ref. (reason being: the downstream reconciler should need to know how we parse keys, leaving us free to change the queue key scheme in the future)

reconciler/configstore.go
```diff
@@ -1,7 +1,7 @@
 /*
 Copyright 2020 The Knative Authors
 
-Licensed under the Apache License, Veroute.on 2.0 (the "License");
+Licensed under the Apache License, Version 2.0 (the "License");
```
n3wscott, line 4: oops. changed here too, revert the search and replace typo

codegen/cmd/injection-gen/generators/reconciler_reconciler.go
markusthoemmes: FWIW, we're not passing strings, we pass a NamespacedName, which is already a parsed version of the string. If we don't return errors (and thus disallow requeuing) how are observers supposed to handle errors? To take the example from above: If we fail to update the envoy config, I'd actually like to retry that. I can't retry in the handler though, as that'd block a reconciler thread.

n3wscott: I need a second opinion. @vaikas

n3wscott: I am just an idiot... sorry.
/lgtm
/approve

vaikas: Just to make sure I understand. Is it correct to say that the problem we're trying to solve here is:
 1. We have a resource (call it "original" resource) that creates another resource. I reckon this resource is maybe not a k8s resource, because for managing ownership, we'd use the OwnerRef?
 2. That resource then needs to be cleaned up when the "original" resource gets deleted. Or Something needs to be updated that's not a k8s resource?
 3. If something goes wrong during that delete phase, things are left in an inconsistent state?
 
Is that the problem here?
IFF that's the problem my probably dump question is why not use finalizers here then to ensure that the object will get reconciled properly.

markusthoemmes: @vaikas see the linked issue for the motivation of this. This tries to solve the problem of cleaning up in-memory state consistent with the deletion of K8s objects. Finalizers are not guaranteed to be observed by all replicas and directly reacting to the informer events isn't guaranteed to not race an update event.

PR #2095, opened 2021-04-13
Allow URIs to be resolved without reconciler context

SwEngin:
/kind enhancement

Fixes #

As we noticed from test [here](https://github.com/knative/serving/pull/10587) that using the reconcilers context with timeout makes resolver fails (for DomainMapping) as the canceled context gets cached

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -21,26 +21,27 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/client/injection/ducks/duck/v1/addressable"
+	"knative.dev/pkg/controller"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/types"
-
 	"knative.dev/pkg/apis"
 	pkgapisduck "knative.dev/pkg/apis/duck"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
-	"knative.dev/pkg/controller"
 	"knative.dev/pkg/network"
 	"knative.dev/pkg/tracker"
-
-	"knative.dev/pkg/client/injection/ducks/duck/v1/addressable"
 )
 
 // URIResolver resolves Destinations and ObjectReferences into a URI.
 type URIResolver struct {
-	tracker         tracker.Interface
-	informerFactory pkgapisduck.InformerFactory
+	tracker       tracker.Interface
+	listerFactory func(schema.GroupVersionResource) (cache.GenericLister, error)
 }
 
 // NewURIResolver constructs a new URIResolver with context and a callback
@@ -49,11 +50,17 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
 	ret := &URIResolver{}
 
 	ret.tracker = tracker.New(callback, controller.GetTrackerLease(ctx))
-	ret.informerFactory = &pkgapisduck.CachedInformerFactory{
+
+	informerFactory := &pkgapisduck.CachedInformerFactory{
 		Delegate: &pkgapisduck.EnqueueInformerFactory{
 			Delegate:     addressable.Get(ctx),
 			EventHandler: controller.HandleAll(ret.tracker.OnChanged),
 		},
+	}
+
+	ret.listerFactory = func(gvr schema.GroupVersionResource) (cache.GenericLister, error) {
+		_, l, err := informerFactory.Get(ctx, gvr)
+		return l, err
 	}
 
 	return ret
@@ -163,7 +170,7 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		return url, nil
 	}
 
-	_, lister, err := r.informerFactory.Get(ctx, gvr)
+	lister, err := r.listerFactory(gvr)
 	if err != nil {
 		return nil, apierrs.NewNotFound(gvr.GroupResource(), "Lister")
 	}


SwEngin:
@julz

resolver/addressable_resolver.go
```diff
@@ -56,6 +58,28 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
+	ret := &URIResolver{}
+
+	ret.tracker = dmTracker
+
+	ret.listerFactory = func(gvr schema.GroupVersionResource) (cache.GenericLister, error) {
+		_, l, err := dmInformerFactory.Get(ctx, gvr)
+		return l, err
+	}
+
 	return ret
```
julz, line 83: we can collapse this:

```suggestion
	return &URIResolver{
	  tracker: dmTracker,
	  listerFactory = func(gvr schema.GroupVersionResource) (cache.GenericLister, error) {
		_, l, err := dmInformerFactory.Get(ctx, gvr)
		return l, err
	  },
	}
```

resolver/addressable_resolver.go
```diff
@@ -21,26 +21,28 @@ import (
...
 	"knative.dev/pkg/network"
 	"knative.dev/pkg/tracker"
-
-	"knative.dev/pkg/client/injection/ducks/duck/v1/addressable"
 )
 
 // URIResolver resolves Destinations and ObjectReferences into a URI.
 type URIResolver struct {
 	tracker         tracker.Interface
 	informerFactory pkgapisduck.InformerFactory
```
julz, line 44: looks like `informerFactory` wont be used outside of the constructor after this, so maybe we should drop the field?

resolver/addressable_resolver.go
```diff
@@ -56,6 +58,28 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
+		return l, err
+	}
+
+	return ret
+}
+
+// NewURIResolver constructs a new URIResolver with context and a callback
+// for a given listableType (Listable) passed to the URIResolver's tracker.
+// extra when controller context is used
+func NewURIResolverWithControllerContext(ctx context.Context, dmInformerFactory pkgapisduck.InformerFactory,
```
julz, line 72: I'd really like to see some test coverage for this, especially in order to be able to see how the API would look for consumers (a WIP version of the serving PR might be good for this too). It might also be worth adding a test for the new behaviour here (that cancelling the context passed to `URIFromObjectReference` etc doesn't cause problems for future calls to the resolver), though looks like I didn't do that either in https://github.com/knative/serving/pull/10673 🤭.

resolver/addressable_resolver.go
```diff
@@ -140,7 +164,6 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 	if ref == nil {
 		return nil, apierrs.NewBadRequest("ref is nil")
 	}
-
```
julz, line 143: I think the newline is worth keeping here, since it separates the guard clause from the rest of the method

resolver/addressable_resolver.go
```diff
@@ -56,6 +58,28 @@ func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *U
...
+		_, l, err := ret.informerFactory.Get(ctx, gvr)
+		return l, err
+	}
+
+	return ret
+}
+
+// NewURIResolver constructs a new URIResolver with context and a callback
+// for a given listableType (Listable) passed to the URIResolver's tracker.
+// extra when controller context is used
```
julz, line 71: I'd like to see this fleshed out into a full sentence.. WDYT of something like:

~~~~
// This method returns a URIResolver which uses the passed context to retrieve
// listerFactories, and should be used when the URIResolver will be reused for multiple reconciles.
~~~~

.. although writing this out makes me wonder if there's any use case where you _shouldnt_ use the new version, and if so whether there's a good reason not to update the behaviour of the original `NewURIResolver` method instead 🤔 

ping @n3wscott for thoughts on that cos I dont have that much context on URIResolver use outside of domainmapping (which is where we use it in serving).

julz (commented):
nice work and thanks @SwEngin! left a few comments, mainly Id really like to see a test case and/or a draft PR to see how the new constructor would be consumed, and Id like to understand if there's ever a case where people should still use the old version or if it's always better to use the new one.

resolver/addressable_resolver.go
SwEngin: Yeah, the uri resolver is used in trigger, pingsource, subscribtion and apiserversource reconcilers in eventing. Also discussed that in serving-api wg and we thought we'll have the context timeout in genreconciler as an opt in. That's why I created a new constructor.

resolver/addressable_resolver.go
SwEngin: I think we can

resolver/addressable_resolver.go
julz: I guess Im wondering if there'd be a downside to having NewURIResolver _always_ set things up to use the constructor context in the resolve calls when looking up the factory. That wouldn't be needed if you don't set a timeout on the context you pass in the reconcile, but it wouldn't (I think?) hurt in that case, and it seems maybe more correct anyway not to leak the per-reconcile context outside of the reconcile (plus avoids having two different code paths/constructors iyswim). 

It's hard to know when writing a reconciler if someone will later decide to add a timeout, so seems nice if things would work "just work" if they do without needing to remember to use the right constructor.

To be clear Im not saying there's definitely not a reason to keep the existing behaviour and add a new constructor for the new behaviour... I just can't think of one and if there _isn't_ it might be nice to be able to avoid adding surface area to the api. (paging @n3wscott to the courtesy telephone again 😊)

resolver/addressable_resolver.go
markusthoemmes, line 72: Hmm, the name of this functions seems kinda misleading. Why `WithControllerContext`? `NewURIResolver` seems to have changed its meaning of the context as well so maybe this is more like `NewURIResolverWithFactory` or something like that?

julz: Oh! Yes I think you're right @markusthoemmes, which answers my question in https://github.com/knative/pkg/pull/2095#discussion_r613030091 -- we already did fix the existing constructor. But now I'm not sure if/why we need the second ctor at all? 😊

resolver/addressable_resolver.go
```diff
@@ -163,7 +169,8 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		return url, nil
 	}
 
-	_, lister, err := r.informerFactory.Get(ctx, gvr)
+	lister, err := r.listerFactory(gvr)
+
```
julz, line 173: super-nit: I'd drop this newline (to keep the error handling together with the function that returns the err), and add back the one after the `ref == nil` guard clause (so after L149) that got dropped here

```suggestion
```

julz (commented):
lgtm apart from a bit of remaining pedantry about newlines :) - thanks @SwEngin!

resolver/addressable_resolver.go
julz: @SwEngin just waiting on adding this newline back to lgtm this!

SwEngin: Done!

julz (approved):
/lgtm
/retest

(btw it's a little easier to review things if you avoid force pushing because then it's obvious what changed, and prow will squash stuff when it merges anyway so there's generally no need unless you have to rebase)

markusthoemmes (approved):
/lgtm
/approve

PR #2098, opened 2021-04-20
set webhooks ownerReferences to namespace

### LINKED ISSUE #2044: Webhooks should get OwnerRef to namespace
/area API
/kind cleanup
/kind good-first-issue

## Expected Behavior

It's a common misconception that folks can delete our system namespace (e.g. `knative-serving`) to "clean up" their cluster for a fresh installation.

## Actual Behavior

Unfortunately, we install a lot of things that aren't deleted by deleting our system namespace, and some of those things block reinstallation.  In particular the configmap webhook has no pods, and so it blocks configmap creations.  This creates a chicken/egg problem because the webhook pods need those configmaps to start up properly.

## Proposed Solution

I think the simplest solution would be to have our webhooks set up an OwnerReference to the namespace resource within which its target service lives.  We used to do this for the webhook deployment, but removed that due to it being Cluster-scoped -> Namespaced (which isn't supported).  However, Namespace is a cluster-scoped resource, so I think this would work.

novahe:
# Changes

 - set webhooks ownerReferences to namespace 

/kind cleanup

Fixes #2044

**Release Note**

```release-note
NONE
```

**Docs**

```docs
```

### DIFF
--- a/webhook/configmaps/configmaps.go
+++ b/webhook/configmaps/configmaps.go
@@ -137,9 +137,13 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 
 	webhook := configuredWebhook.DeepCopy()
 
-	// Clear out any previous (bad) OwnerReferences.
-	// See: https://github.com/knative/serving/issues/5845
-	webhook.OwnerReferences = nil
+	// Set the owner to namespace.
+	ns, err := ac.client.CoreV1().Namespaces().Get(ctx, system.Namespace(), metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch namespace: %w", err)
+	}
+	nsRef := *metav1.NewControllerRef(ns, corev1.SchemeGroupVersion.WithKind("Namespace"))
+	webhook.OwnerReferences = []metav1.OwnerReference{nsRef}
 
 	for i, wh := range webhook.Webhooks {
 		if wh.Name != webhook.Name {
--- a/webhook/resourcesemantics/defaulting/defaulting.go
+++ b/webhook/resourcesemantics/defaulting/defaulting.go
@@ -30,6 +30,7 @@ import (
 	jsonpatch "gomodules.xyz/jsonpatch/v2"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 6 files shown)


markusthoemmes:
/ok-to-test

The code looks spot on, thanks! Have we tested this is actually fixing the referenced bug? Let's also make sure the OwnerReference doesn't cause the webhook config to be garbage collected. I guess manual tests for both of those are fine.

NovaHe:
Thank you for reviewing @markusthoemmes

In my test environment
**I tested to delete the ns, `webhook config` can be deleted:**

image

**I deleted webhook pod twice, The `webhook config` has not changed**
image
only updated once

Am I missing something? What else do I need to do?

webhook/configmaps/configmaps.go
```diff
@@ -137,9 +137,15 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 
 	webhook := configuredWebhook.DeepCopy()
 
-	// Clear out any previous (bad) OwnerReferences.
-	// See: https://github.com/knative/serving/issues/5845
-	webhook.OwnerReferences = nil
+	// Set the owner to namespace.
+	if len(webhook.OwnerReferences) == 0 {
```
markusthoemmes, line 141: Should we fully reconcile this? I.e. make sure it's always the correct ownerRef, even if this is non-zero (i.e. after the first reconcilation?)

webhook/resourcesemantics/validation/reconcile_config.go
```diff
@@ -135,9 +136,15 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 
 	current := configuredWebhook.DeepCopy()
 
-	// Clear out any previous (bad) OwnerReferences.
-	// See: https://github.com/knative/serving/issues/5845
-	current.OwnerReferences = nil
+	// Set the owner to namespace.
+	if len(current.OwnerReferences) == 0 {
```
markusthoemmes, line 140: Same q as above.

webhook/resourcesemantics/defaulting/defaulting.go
```diff
@@ -172,9 +173,14 @@ func (ac *reconciler) reconcileMutatingWebhook(ctx context.Context, caCert []byt
 
 	current := configuredWebhook.DeepCopy()
 
-	// Clear out any previous (bad) OwnerReferences.
-	// See: https://github.com/knative/serving/issues/5845
-	current.OwnerReferences = nil
+	if len(current.OwnerReferences) == 0 {
```
markusthoemmes, line 176: Same q as above.

markusthoemmes (commented):
/assign @dprotaso 

Sorry for having this dangling for a while. Adding Dave to assert if this isn't breaking assumptions in K8s API.

novahe:
I fixed it, PTAL thanks

dprotaso:
🎉 tested it out and it works

/lgtm
/approve

julz:
This is going to avoid a _lot_ of user confusion, thanks @novahe 🎉🎉 

cc @omerbensaadon @csantanapr I think you will like!

PR #2100, opened 2021-04-22
Allow errors.As to work with reconciler.Events

benmoss:
# Changes

- :gift: `reconciler.Event`s that wrap errors can be used with the [apimachinery errors package](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors), which uses [errors.As](https://github.com/kubernetes/apimachinery/blob/v0.21.0/pkg/api/errors/errors.go#L649-L654) to compare error types.

/kind enhancement

https://github.com/knative/eventing/issues/5146 involves a bug where we are [trying](https://github.com/knative/eventing/blob/7775e372bdc033da163cfb01194785d11a4f7d93/pkg/reconciler/subscription/subscription.go#L120) to ignore `IsNotFound`-type errors in our finalizer, but right now Events can never be used with these apimachinery predicates.

**Release Note**

```release-note
- reconciler.Events can be cast to their underlying error types
```

### DIFF
--- a/reconciler/events_test.go
+++ b/reconciler/events_test.go
@@ -19,6 +19,7 @@ package reconciler
 import (
 	"errors"
 	"io"
+	"net"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


vaikas (approved):
/lgtm

reconciler/events_test.go
```diff
@@ -19,6 +19,7 @@ package reconciler
 import (
 	"errors"
 	"io"
+	"io/fs"
```
n3wscott, line 22: I think this is a new package to go 1.16 and can not be used yet in 1.14

n3wscott (approved):
/lgtm
/approve

good catch!

PR #2103, opened 2021-04-23
Use semantic equality for status update retries

markusthoemmes:
# Changes

As per title. We use semantic equality to determine if we want to update the status in the first place. Using `reflect.DeepEqual` here means, that we're going to update status even if only the timestamp has "changed". We want to explicitly exclude that (that's why we use `VolatileTime` in the first place) so this should spare us a few unnecessary API calls.

/assign @n3wscott


vagababov (approved):
/lgtm

PR #2106, opened 2021-04-26
Better support for error wrapping with ReconcilerEvents

benmoss:
# Changes

- :broom: Use `ReconcilerEvent.Error()` when translating ReconcilerEvents to event message strings.
- :broom: Update codegen to use GOPATH hacks, modeled on https://github.com/knative/eventing/pull/4621

This is a follow-on to #2100, I didn't fully understand how ReconcilerEvents were translated to EventRecorder events. The goal is to more fully support wrapping errors in ReconcilerEvents, but using `Eventf` with `%w` format strings doesn't work. By generating the error message via `fmt.Errorf(...).Error()` we can get the proper formatting while still preserving the wrapping chain.


hack/update-codegen.sh
```diff
@@ -18,23 +18,20 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-export GO111MODULE=on
+source $(dirname $0)/../vendor/knative.dev/hack/codegen-library.sh
```
n3wscott, line 21: Thanks for migrating this. I missed it on my update pass.

n3wscott (approved):
/lgtm
/approve

dprotaso:
I think the hack/codegen lib doesn't set REPO_ROOT_DIR correctly

https://github.com/knative-sandbox/knobots/runs/2448216473?check_suite_focus=true

PR #2117, opened 2021-05-10
Allow disable via regex in chaosduck

markusthoemmes:
# Changes

It's been kinda annoying trying to come up with the exact list of webhooks to ignore and this has caused intermediate flakiness. This adds the possibility to exclude components via a regex, so all webhooks can be excluded more easily and scalably.

/assign @nak3 @dprotaso

### DIFF
--- a/leaderelection/chaosduck/main.go
+++ b/leaderelection/chaosduck/main.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"flag"
 	"log"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nak3:
LGTM

dprotaso:
can we have `disabledComponents` accept a regex?

markusthoemmes:
I thought about that @dprotaso, but eventually opted for the easier solution. I'll take another look.

markusthoemmes:
/retest

Unrelated.

Issue #2116, opened 2021-05-10
Multiple label selectors support is not be honest when creating informer factory


cdlliuy:
Per discussion in https://github.com/knative/serving/pull/11142#pullrequestreview-638700574, the label selectors  on informer factory need further improvement to avoid confusions when multiple selectors  are passed through.

https://github.com/knative/pkg/blob/b558677ab03404ed118ba3e179a7438fe2d8509a/vendor/k8s.io/client-go/informers/factory.go#L80-L85,

cdlliuy:
I would like to assign myself to fix this.

cdlliuy:
hold on,  more discusson with Julz is on-going in https://github.com/knative/serving/pull/11142#discussion_r629087910.  I will close this if it is not valid.

cdlliuy:
per discussion in https://github.com/knative/serving/pull/11142#discussion_r629277061,  the fix is still required to change the multiple label selector parameter to support a single selector only

julz:
> per discussion in knative/serving#11142 (comment), the fix is still required to change the multiple label selector parameter to support a single selector only

fwiw Im fine with leaving it as multiple selectors, as long as we fix the bug so they don't potentially clobber each other

PR #2120, opened 2021-05-12
update the multiple label selector implemention to avoid clobber  in filtered factory

cdlliuy:
# Changes

Per discussion in https://github.com/knative/pkg/issues/2116,  we decided to change the multiple label selectors in filtered informer and factories to a single one to avoid any future confusion and bugs.


julz (commented):
lgtm, let's update the PR & commit titles for git history's sake since we changed approach a bit in the end?

cdlliuy:
> lgtm, let's update the PR & commit titles for git history's sake since we changed approach a bit in the end?

@julz , updated the title

cdlliuy:
@markusthoemmes , can you help to proceed this PR further?

markusthoemmes (approved):
/lgtm
/approve

PR #2122, opened 2021-05-13
Add an import that will create fake kubeconfig for go-test runs.

n3wscott:
Adding a workaround to go build tests failing for tests that use dependency injection in `TestMain`.

Working integration in eventing: https://github.com/knative/eventing/pull/5397/checks?check_run_id=2578226975

### DIFF
--- a/injection/kubeconfig.go
+++ b/injection/kubeconfig.go
@@ -76,7 +76,7 @@ Certificate:
 ...
 */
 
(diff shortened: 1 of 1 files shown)


n3wscott:
/hold
There might be another workaround, https://github.com/knative-sandbox/actions-downstream-test/pull/10

PR #2124, opened 2021-05-17
add flag to disable test cleanup in e2e tests

psschwei:
# Changes

- add `--nocleanup` flag to disable test cleanup in e2e tests

/kind enhancement

This is in support of [Serving Issue 10001](https://github.com/knative/serving/issues/10001). The flag created here will be used to enable the functionality in the Serving repo.

```release-note
Add a `--nocleanup` flag to allow users to disable test cleanup when running e2e tests locally 
```

**Docs**

```docs
- [Usage]: https://github.com/knative/pkg/blob/add-nocleanup-flag/test/README.md#nocleanup
```
(note that the above link won't be valid until this is merged -- prior to that, text can be viewed [here](https://github.com/psschwei/pkg/blob/add-nocleanup-flag/test/README.md#nocleanup).

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -125,6 +125,7 @@ Tests importing [`knative.dev/pkg/test`](#test-library) recognize these flags:
 - [`--dockerrepo`](#specifying-docker-repo)
 - [`--tag`](#specifying-tag)
 - [`--imagetemplate`](#specifying-image-template)
+- [`--nocleanup`](#nocleanup)
 
 ### Specifying kubeconfig
 
@@ -208,6 +209,14 @@ reference to an image from the test. Defaults to
 
 ```bash
 go test ./test --imagetemplate {{.Repository}}/{{.Name}}:{{.Tag}}
+```
+
+### Nocleanup
+
+The `--nocleanup` argument lets you preserve resources created by the end-to-end tests to assist with debugging.
+
+```bash
+go test ./test --nocleanup
 ```
 
 ---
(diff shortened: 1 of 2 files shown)


psschwei:
Failed check "Test / Unit Tests (1.16.x, ubuntu-latest) (pull_request)" passed on local

```
$ go test ./metrics/ -run TestMetricsExport
ok      knative.dev/pkg/metrics 2.047s
$ go test ./metrics/ -run TestMetricsExport/OpenCensus
ok      knative.dev/pkg/metrics 1.023s
```

Failed check "Downstream / Unit Test (serving, knative) (pull_request)" failed on `TestRateLimitPerItem`, which may be a test flake a la [this](https://github.com/knative/serving/pull/11344#issuecomment-840605037)

/retest

julz:
oh fun, yeah looks like TestRateLimitPerItem is flakey somehow, on it

julz (commented):
I don't feel very strongly about this, but I'm wondering whether since the new flag isnt used in pkg/test and wont (I think?) be implemented other than by serving, it might be better to define it only in the ServingFlags, like we do with e.g. [resolvabledomain and enablealphafeatures](https://github.com/knative/serving/blob/f71282042efe73edd8e005d288ec7a7fd2173726/test/e2e_flags.go#L34)

psschwei:
> I don't feel very strongly about this, but I'm wondering whether since the new flag isnt used in pkg/test and wont (I think?) be implemented other than by serving, it might be better to define it only in the ServingFlags, like we do with e.g. [resolvabledomain and enablealphafeatures](https://github.com/knative/serving/blob/f71282042efe73edd8e005d288ec7a7fd2173726/test/e2e_flags.go#L34)


Based on a very, very quick search of the org, I think you're right that this would be a serving-specific flag right now. That said, I could envision a similar need possibly arising down the line for other repos, so maybe it makes sense to put it here in case of future need (?). But I too don't feel very strongly about this.

julz:
@markusthoemmes @dprotaso wdyt?

psschwei:
Discussed this on the May 26 Serving API WG call and the thought is to do some more cleanup here before merging this in.
/hold

PR #2127, opened 2021-05-18
Add KReference.Group field and ResolveGroup function

slinkydeveloper:

Add the `KReference.Group` field and `ResolveGroup` function. Part of the experimental feature https://github.com/knative/eventing/issues/5086

# Changes

- :gift: Add KReference.Group field and ResolveGroup function

### DIFF
--- a/apis/duck/v1/knative_reference.go
+++ b/apis/duck/v1/knative_reference.go
@@ -20,6 +20,11 @@ import (
 	"context"
 	"fmt"
 
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1lister "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	"knative.dev/pkg/apis"
 )
 
@@ -42,6 +47,11 @@ type KReference struct {
 
 	// API version of the referent.
 	APIVersion string `json:"apiVersion"`
+
+	// Group of the API, without the version of the group. This can be used as an alternative to the APIVersion, and then resolved using ResolveGroup.
+	// Note: This API is EXPERIMENTAL and might break anytime. For more details: https://github.com/knative/eventing/issues/5086
+	// +optional
+	Group string `json:"group,omitempty"`
 }
 
 func (kr *KReference) Validate(ctx context.Context) *apis.FieldError {
@@ -85,4 +95,45 @@ func (kr *KReference) SetDefaults(ctx context.Context) {
 	if kr.Namespace == "" {
 		kr.Namespace = apis.ParentMeta(ctx).Namespace
 	}
+}
+
+// ResolveGroup resolves the APIVersion of a KReference starting from the Group.
+// In order to execute this method, you need RBAC to read the CRD of the Resource referred in this KReference.
+// Note: This API is EXPERIMENTAL and might break anytime. For more details: https://github.com/knative/eventing/issues/5086
+func (kr *KReference) ResolveGroup(crdLister apiextensionsv1lister.CustomResourceDefinitionLister) error {
+	if kr.Group == "" {
+		// Nothing to do here
+		return nil
+	}
+	if kr.Group == "core" {
+		// We statically resolve it to core/v1
+		kr.APIVersion = "core/v1"
+		return nil
+	}
+
+	actualGvk := schema.GroupVersionKind{Group: kr.Group, Kind: kr.Kind}
+	pluralGvk, _ := meta.UnsafeGuessKindToResource(actualGvk)
+	crd, err := crdLister.Get(pluralGvk.GroupResource().String())
+	if err != nil {
+		return err
+	}
+
+	actualGvk.Version, err = findCRDStorageVersion(crd)
+	if err != nil {
+		return err
+	}
+
+	kr.APIVersion, kr.Kind = actualGvk.ToAPIVersionAndKind()
+
+	return nil
+}
+
+// This function runs under the assumption that there must be exactly one "storage" version
+func findCRDStorageVersion(crd *apiextensionsv1.CustomResourceDefinition) (string, error) {
+	for _, version := range crd.Spec.Versions {
+		if version.Storage {
+			return version.Name, nil
+		}
+	}
+	return "", fmt.Errorf("this CRD %s doesn't have a storage version! Kubernetes, you're drunk, go home", crd.Name)
 }
--- a/apis/duck/v1/knative_reference.go
+++ b/apis/duck/v1/knative_reference.go
@@ -64,8 +64,25 @@ func (kr *KReference) Validate(ctx context.Context) *apis.FieldError {
 	if kr.Name == "" {
 		errs = errs.Also(apis.ErrMissingField("name"))
 	}
-	if kr.APIVersion == "" {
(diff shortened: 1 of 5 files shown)


apis/duck/v1/knative_reference.go
```diff
@@ -42,6 +47,11 @@ type KReference struct {
 
 	// API version of the referent.
 	APIVersion string `json:"apiVersion"`
+
```
vaikas, line 50: I think we need to update L49 to be omitempty & +optional?

slinkydeveloper: TBH I don't know what `+optional` is useful for :smile:

slinkydeveloper: Tried to run the codegen and nothing changes using `+optional`

apis/duck/v1/knative_reference.go
```diff
@@ -54,8 +64,25 @@ func (kr *KReference) Validate(ctx context.Context) *apis.FieldError {
...
 	}
-	if kr.APIVersion == "" {
-		errs = errs.Also(apis.ErrMissingField("apiVersion"))
+	if isKReferenceGroupAllowed(ctx) {
+		if kr.APIVersion == "" && kr.Group == "" {
+			errs = errs.Also(apis.ErrMissingField("apiVersion")).
+				Also(apis.ErrMissingField("group"))
+		}
+		if kr.APIVersion != "" && kr.Group != "" {
+			errs = errs.Also(&apis.FieldError{
```
n3wscott, line 73: If they do not contradict each other, should it be an error? Seems like it would be ok to update refs to add group, then drop api in two passes if the operator is feeling nervous.

apis/duck/v1/knative_reference.go
```diff
@@ -86,3 +113,53 @@ func (kr *KReference) SetDefaults(ctx context.Context) {
 		kr.Namespace = apis.ParentMeta(ctx).Namespace
 	}
 }
+
+// ResolveGroup resolves the APIVersion of a KReference starting from the Group.
+// In order to execute this method, you need RBAC to read the CRD of the Resource referred in this KReference.
+// Note: This API is EXPERIMENTAL and might break anytime. For more details: https://github.com/knative/eventing/issues/5086
+func (kr *KReference) ResolveGroup(crdLister apiextensionsv1lister.CustomResourceDefinitionLister) error {
```
n3wscott, line 120: I would not add this code to pkg nor kref. We intend to replace this lookup code with a lookup from ClusterDuckType. Hardcoding it here seems like a mistake. I would move this to a utility package to make it clear this is one method for resolving version from group.

This method has a side effect that invalidates the validation of the KRefrence in the validate method. That seems non-ideal as well. I think it should return a new KRef that has the alternante valid Kind+APIVersion set.

apis/duck/v1/knative_reference.go
```diff
@@ -86,3 +113,53 @@ func (kr *KReference) SetDefaults(ctx context.Context) {
...
+	if err != nil {
+		return err
+	}
+
+	actualGvk.Version, err = findCRDStorageVersion(crd)
+	if err != nil {
+		return err
+	}
+
+	kr.APIVersion, kr.Kind = actualGvk.ToAPIVersionAndKind()
```
n3wscott, line 138: At this point, a call to kr.validate() fails.

apis/duck/v1/knative_reference.go
```diff
@@ -54,8 +64,25 @@ func (kr *KReference) Validate(ctx context.Context) *apis.FieldError {
...
+		if kr.APIVersion != "" && kr.Group != "" {
+			errs = errs.Also(&apis.FieldError{
+				Message: "both apiVersion and group are specified",
+				Paths:   []string{"apiVersion", "group"},
+				Details: "Only one of them must be specified",
+			})
+		}
+	} else {
+		if kr.Group != "" {
+			errs = errs.Also(apis.ErrDisallowedFields("group"))
```
n3wscott, line 81: This makes me worried about how to disable the feature if it had previously been enabled, used, and then disabled. It seems like it will brick the cluster.

apis/duck/v1/knative_reference.go
```diff
@@ -86,3 +113,53 @@ func (kr *KReference) SetDefaults(ctx context.Context) {
...
+		return err
+	}
+
+	kr.APIVersion, kr.Kind = actualGvk.ToAPIVersionAndKind()
+
+	return nil
+}
+
+// This function runs under the assumption that there must be exactly one "storage" version
+func findCRDStorageVersion(crd *apiextensionsv1.CustomResourceDefinition) (string, error) {
```
n3wscott, line 144: Feels weird to have this here. I would rather have a KRef resolving strategy package, and "StoredVersion" is a strat we can implement first. But there are other ways that work better with our duck types.

apis/duck/v1/knative_reference.go
slinkydeveloper: > This makes me worried about how to disable the feature if it had previously been enabled, used, and then disabled. It seems like it will brick the cluster.

It will, although it's a user mistake to do so... I guess it works the same for serving too, eg what happens if the user does the same thing you sad with https://knative.dev/docs/serving/feature-flags/#kubernetes-node-affinity feature? The API server will start rejecting the nodeAffinity field in the validation and will require the user to manually change their resources to use the "stable" apis.

When you're opting in for experimental features you know you can break your cluster, and you should be careful to avoid that. It's the very same concept as when you use a library which is still not stable, you rely on their methods and then the next version the api breaks, it's up to you to fix it.

apis/duck/v1/knative_reference.go
slinkydeveloper: > We intend to replace this lookup code with a lookup from ClusterDuckType. Hardcoding it here seems like a mistake

TBH I don't know what you're talking about, nor what ClusterDuckType is :smile:, can you be more specific about it?

> I would move this to a utility package to make it clear this is one method for resolving version from group.

Makes sense, which package do you propose? How about `knative.dev/pkg/krefutil`? Or maybe just `knative.dev/pkg/kref`? We could even more there some methods to convert KReference back and forth to other kube types

apis/duck/v1/knative_reference.go
slinkydeveloper: Check the last commit, now both are allowed in case they don't contradict each other

apis/duck/v1/knative_reference.go
slinkydeveloper: Solved fixing the validation: https://github.com/knative/pkg/pull/2127#discussion_r646931513

apis/duck/v1/knative_reference.go
slinkydeveloper: Check the last commit, I moved to `knative.dev/pkg/kref`

slinkydeveloper:
/retest 

Metrics ut failure.

apis/duck/v1/knative_reference.go
vaikas: FYI:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

And in particular:
```
Note that for backward compatibility, any field that has the omitempty struct tag will be considered to be optional, but this may change in the future and having the +optional comment tag is highly recommended.
```

slinkydeveloper: ok cool, adding it

slinkydeveloper:
/unhold

This has been approved

aliok (approved):
/lgtm
/approve

matzew (approved):
/lgtm
/approve

PR #2131, opened 2021-05-26
Fix revive related linting issues

markusthoemmes:
# Changes

These slipped and they're mostly because `revive` doesn't ignore exported comment related issues by default.

/assign @julz


julz (approved):
/lgtm
/approve

PR #2133, opened 2021-05-28
consolidate k8s flags to an environment package

dprotaso:


injection/config.go
```diff
@@ -17,102 +17,24 @@ limitations under the License.
...
-	return env
-}
-
 // ParseAndGetRESTConfigOrDie parses the rest config flags and creates a client or
 // dies by calling log.Fatalf.
 func ParseAndGetRESTConfigOrDie() *rest.Config {
-	env := Flags()
-
+	env := new(environment.ClientConfig)
+	env.InitFlags(flag.CommandLine)
```
dprotaso, line 32: I'm assuming this breaks injection in the use of tests

dprotaso:
eventing
> cmd/broker/filter/main.go:67:9: undefined: sharedmain.ParseAndGetConfigOrDie
> cmd/broker/ingress/main.go:78:9: undefined: sharedmain.ParseAndGetConfigOrDie

serving
> test/e2e/autotls/config/disablenscert/main.go:44:14: undefined: injection.GetRESTConfig
> test/performance/benchmarks/scale-from-zero/continuous/main.go:67:14: undefined: injection.GetRESTConfig

environment/client_config.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2020 The Knative Authors
```
n3wscott, line 2: nit: 2021

n3wscott (commented):
/lgtm

environment/client_config.go
dprotaso: lol I copied the boilerplate

dprotaso:
Gonna update boilerplate in a separate PR since it require codegen

slinkydeveloper:
@dprotaso this change broke control-protocol update: https://github.com/knative-sandbox/control-protocol/pull/78/checks?check_run_id=2709669959#step:9:17

In particular the issue is that, when the package is loaded the first time, `knative.dev/pkg/test` is loaded because of this method invocation https://github.com/knative-sandbox/control-protocol/blob/main/test/conformance/conformance_test.go#L33, which executes `initializeFlags()` that initialize flags. Then, when the main test here is executed https://github.com/knative-sandbox/control-protocol/blob/main/test/conformance/main_test.go#L54, `EnableInjectionOrDie` will end up invoking `ClientConfig.InitFlags`, re-initiliazing twice the `cluster` flag. 

Can you address this issue? Thanks

n3wscott:
@slinkydeveloper you should add  control-protocol  to the downstream tests in pkg.

dprotaso:
You'll want to pass the restconfig from the test flags into the injection framework - second argument to ` EnableInjectionOrDie`

I'll make a PR

slinkydeveloper:
TBH It doesn't seem "right" to me the fact that loading the `test` package to use the `ImagePath` method has the side effect of defining and loading the flags... Maybe is that what we need to fix, or at least guard it with a sync.Once?

dprotaso:
PR here: https://github.com/knative-sandbox/control-protocol/pull/82

dprotaso:
`ImagePath` depends on the input from the flags

dprotaso:
I agree with `knative.dev/pkg/test` having these side-effects that's why I've been slowly moving things out of that package - it's really there for backwards compability

PR #2136, opened 2021-05-31
Auto-generate webhook certs with ECDSA for FIPS compliance

markusthoemmes:
# Changes

ED25519 is not [currently compliant with FIPS regulations](https://www.federalregister.gov/documents/2019/10/31/2019-23742/request-for-comments-on-fips-186-5-and-sp-800-186), [ECDSA with SHA256 is though](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf).

This is kind of a minor security point anyway, seeing as these certificates are only used for Webhook-cert auto-creation. It's useful to have a regulatory compliant default though.

### DIFF
--- a/webhook/certificates/resources/certs.go
+++ b/webhook/certificates/resources/certs.go
@@ -18,7 +18,8 @@ package resources
 
 import (
 	"context"
-	"crypto/ed25519"
+	"crypto/ecdsa"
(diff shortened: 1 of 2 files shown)


julz (approved):
/lgtm

PR #2141, opened 2021-06-02
support multiple cleanup functions

dprotaso:
note: this change is backwards compatbile

### DIFF
--- a/test/cleanup.go
+++ b/test/cleanup.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 )
 
+type logFunc func(template string, args ...interface{})
+
 var cleanup struct {
 	once  sync.Once
 	mutex sync.RWMutex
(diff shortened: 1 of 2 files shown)


vagababov:
Probably needs a comment about the order of execution?

dprotaso:
Curious if the timing of 100 milliseconds isn't good enough

dprotaso:
well that test became ugly lol

dprotaso:
hmm.. there's a lot more uses of the `CleanupOnInterrupt` than I thought. I'll do this change gracefully

psschwei (commented):
@dprotaso Only comment I have is, should this be rebased to a single commit? (It didn't look like that was listed as a requirement in the docs anywhere, but thought I'd mention it just in case) Looks good to me otherwise.

dprotaso:
@psschwei we have prow setup to do the merging and it will squash all the commits

Thus adding new commits generally makes it easier to review vs. force pushing etc.

psschwei:
That's cool, I didn't know that
/lgtm

Issue #2111, opened 2021-05-01
Bump k8s libs to 1.20 or 1.21


nak3:
/area API
/area test-and-release
/kind feature

This is similar request to https://github.com/knative/pkg/issues/1881, but for v1.20 or v1.21.

For example, gateway-api repo stated using client-go v0.21.0 so we hit a confliction issue as pkg still uses client-go 0.19.

/cc @dprotaso @knative/technical-oversight-committee

dprotaso:
Something with GKE 1.19 made serving's e2e tests more flakey - so we haven't bump it yet. We've pinned serving to use 1.18 for now while we poke at what's going on

See: https://github.com/knative/serving/issues/11225

nak3:
@dprotaso Is it ready for update?

dprotaso:
Yup - I just need to create a 1.20 branch here: https://github.com/vdemeester/k8s-pkg-credentialprovider so serving will update cleanly

PR #2145, opened 2021-06-04
bump k8s to v1.20.7

dprotaso:
Fixes: https://github.com/knative/pkg/issues/2111


dprotaso:
hmm... that won't work cause they make a copy of gvrToList map

dprotaso:
This PR should make eventing green: https://github.com/knative/eventing/pull/5477 downstream.

Other changes are trivial string replacements with the exception of the client & a plugin

markusthoemmes:
Doublechecked the downstream failures:

* Serving has the same "different error message" thing that this has, trivial to fix
* Networking has the same.
* Client has a compile error. Would it make sense to stage that change first?
* Camel has already been dropped, so we can ignore.
* kn-plugin-admin has the same compile error as client, so I guess fixing client will fix it as well.

So this is LGTM! Thanks for unblocking that breaking change!

markusthoemmes:
FWIW, just looked at the client failure: `k8s.io/cli-runtime` has to be bumped to 0.20.7 in that repo too (it's not inherited from here) and then a few tests fail due to that dynamic client breaking change. Alerted the client team about that and now waiting for confirmation to either go ahead or to stage a fix first.

markusthoemmes (approved):
/lgtm
/approve

PR #2149, opened 2021-06-09
Lookup k8s services to ensure they exist

benmoss:
# Changes

- :bug: Fix URIResolver to verify that k8s services exist before returning their URI

/kind bug

Fixes https://github.com/knative/eventing/pull/5442

**Release Note**

```release-note
URIResolver will now error if resolving a k8s service that does not exist
```

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -158,18 +158,6 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		return nil, apierrs.NewNotFound(gvr.GroupResource(), ref.Name)
 	}
 
-	// K8s Services are special cased. They can be called, even though they do not satisfy the
-	// Callable interface.
-	// TODO(spencer-p,n3wscott) Verify that the service actually exists in K8s.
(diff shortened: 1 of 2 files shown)


benmoss:
Looks like we might be depending on this behavior in eventing tests

resolver/addressable_resolver.go
```diff
@@ -158,9 +158,18 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
...
+	lister, err := r.listerFactory(gvr)
+	if err != nil {
+		return nil, apierrs.NewNotFound(gvr.GroupResource(), "Lister")
+	}
+
+	obj, err := lister.ByNamespace(ref.Namespace).Get(ref.Name)
+	if err != nil {
+		return nil, apierrs.NewNotFound(gvr.GroupResource(), ref.Name)
+	}
+
```
dprotaso, line 170: I know this code was here before but is there not a distinction between an error and object `not found`

benmoss: Yeah, always a bit odd to see a dropped err value

PR #2160, opened 2021-06-17
Allow customize DefaultThreadsPerController value

### LINKED ISSUE #1708: Allow customize DefaultThreadsPerController value
## Expected Behavior

The **DefaultThreadsPerController** value can be customized through passing flags like  #1700 suggested or environment variable.

## Actual Behavior

https://github.com/knative/pkg/blob/master/controller/controller.go#L53
```
var (
	// DefaultThreadsPerController is the number of threads to use
	// when processing the controller's workqueue.  Controller binaries
	// may adjust this process-wide default.  For finer control, invoke
	// Run on the controller directly.
	DefaultThreadsPerController = 2
)
```
Currently, this value is fixed and can not be modified without changing the code.

zhaojizhuang:
- :gift: Add new feature

/kind <kind>

Fixes  #1708

**Release Note**

```release-note
Allow customize DefaultThreadsPerController value by flags `threads-per-controller`
```

**Docs**

```docs

```

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -57,6 +57,7 @@ var (
 	// when processing the controller's workqueue.  Controller binaries
 	// may adjust this process-wide default.  For finer control, invoke
 	// Run on the controller directly.
+	// TODO rename the const to Concurrency and deprecated this
 	DefaultThreadsPerController = 2
 )
 
@@ -203,6 +204,9 @@ type Impl struct {
 	// which are not required to complete at the highest priority.
 	workQueue *twoLaneQueue
 
+	// Concurrency - The number of workers to use when processing the controller's workqueue.
+	Concurrency int
+
 	// Sugared logger is easier to use but is not as performant as the
 	// raw logger. In performance critical paths, call logger.Desugar()
 	// and use the returned raw logger instead. In addition to the
@@ -221,6 +225,7 @@ type ControllerOptions struct { //nolint // for backcompat.
 	Logger        *zap.SugaredLogger
 	Reporter      StatsReporter
 	RateLimiter   workqueue.RateLimiter
+	Concurrency   int
 }
 
 // NewImpl instantiates an instance of our controller that will feed work to the
@@ -244,12 +249,16 @@ func NewImplFull(r Reconciler, options ControllerOptions) *Impl {
 	if options.Reporter == nil {
 		options.Reporter = MustNewStatsReporter(options.WorkQueueName, options.Logger)
 	}
+	if options.Concurrency == 0 {
+		options.Concurrency = DefaultThreadsPerController
+	}
 	return &Impl{
 		Name:          options.WorkQueueName,
 		Reconciler:    r,
 		workQueue:     newTwoLaneWorkQueue(options.WorkQueueName, options.RateLimiter),
 		logger:        options.Logger,
 		statsReporter: options.Reporter,
+		Concurrency:   options.Concurrency,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zhaojizhuang:
cc @kevball @mattmoor @dprotaso

test/e2e_flags.go
```diff
@@ -42,13 +42,15 @@ var (
 type EnvironmentFlags struct {
 	env.ClientConfig
 	testenv.TestClientConfig
+	env.ControllerConfig
```
dprotaso, line 45: These test flags shouldn't be used in any controllers so you should be able to drop the changes in this file

dprotaso (changes requested):
We have a `controller.ControllerOptions` struct where I think a `ThreadCount` property would be a better fit vs a separate `Config` struct

I think it would make sense to structure this changeset as follows

1) Allow per controller threadiness override by adding a property to `ControllerOptions`
2) If the per controller property is not set then fallback on the global property
3) Declare a flag in injection/sharedmain that'll override the default global property

zhaojizhuang:
> We have a `controller.ControllerOptions` struct where I think a `ThreadCount` property would be a better fit vs a separate `Config` struct
> 
> I think it would make sense to structure this changeset as follows
> 
> 1. Allow per controller threadiness override by adding a property to `ControllerOptions`
> 2. If the per controller property is not set then fallback on the global property
> 3. Declare a flag in injection/sharedmain that'll override the default global property

Do you mean that let the controller developer set the options (include `ThreadCount` property )? , such as https://github.com/knative/serving/blob/main/pkg/reconciler/service/controller.go#L59

dprotaso:
I think if the ThreadCount in `ControllerOptions` is not set it (ie. 0) then knative.dev/pkg/controller should just default to `DefaultThreadsPerController`

Thus the ThreadCount option is really an _override_ for controller authors that for some reason need to limit their concurrency.

I think the flag I suggest to add to `sharedmain` should override the default only `DefaultThreadsPerController` by setting the global value directly.

In terms of naming - I think we should call it `Concurrency` and not `ThreadCount` since go doesn't really have 'threads'

test/e2e_flags.go
zhaojizhuang: done

codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -253,6 +253,7 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 
 	impl := {{.controllerNewImpl|raw}}(rec, logger, ctrTypeName)
 	agentName := defaultControllerAgentName
+	impl.Concurrency = controller.DefaultThreadsPerController
```
dprotaso, line 256: We can drop this line no? Since we do the defaulting in here https://github.com/knative/pkg/pull/2160/files#diff-3c5bb5f48211873c58fcba055dcae2ac7b1958969219e06e1508d76d485dace7R252-R254

dprotaso (commented):
looks good one minor thing

codegen/cmd/injection-gen/generators/reconciler_controller.go
zhaojizhuang: @dprotaso Maybe we can't drop this, for controller author will use the `NewImpl` generated, and the `NewImpl` in `pkg/controller/controller.go` only used for test (as far as i know)

dprotaso: I'm under the impression line 254

```
impl := {{.controllerNewImpl|raw}}(rec, logger, ctrTypeName)
```

This becomes `controller.NewImpl` which has the same defaulting logic

dprotaso:
I think the linter has a legitimate issue:

```
  Error: loopclosure: loop variable ctrlr captured by func literal (govet)
```

codegen/cmd/injection-gen/generators/reconciler_controller.go
zhaojizhuang: yeah,see it , thanks, sorry for reply late

zhaojizhuang:
Downstream UT (kn-plugin-admin) failed for it's k8s version (0.19.7) https://github.com/knative-sandbox/kn-plugin-admin/pull/85

zhaojizhuang:
@dprotaso

PR #2166, opened 2021-06-23
Add CheckEndpointState method to SpoofingClient

shinigambit:
# Changes

- New method `CheckEndpointState` just for checking an endpoint's response

/kind api-change

Part of knative/serving#1178

**Release Note**

```release-note
No changes to the existing WaitForEndpointState method, these changes don't fix knative/serving#1178 but help change the way some assertions work in serving e2e tests
```

### DIFF
--- a/test/request.go
+++ b/test/request.go
@@ -120,20 +120,21 @@ func WaitForEndpointStateWithTimeout(
 	timeout time.Duration,
 	opts ...interface{}) (*spoof.Response, error) {
 
-	return waitForEndpointStateWithTimeout(ctx, kubeClient, logf, url, inState, desc, resolvable, timeout, true, opts)
+	client, rOpts, err := makeSpoofClient(ctx, kubeClient, logf, url, resolvable, timeout /* true, */, opts)
+	if err != nil {
+		return nil, err
+	}
+	return client.WaitForEndpointState(ctx, url, inState, desc, rOpts...)
 }
 
-func waitForEndpointStateWithTimeout(
+func makeSpoofClient(
 	ctx context.Context,
 	kubeClient kubernetes.Interface,
 	logf logging.FormatLogger,
 	url *url.URL,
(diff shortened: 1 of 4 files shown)


test/request.go
```diff
@@ -84,6 +84,8 @@ func EventuallyMatchesBody(expected string) spoof.ResponseChecker {
 	}
 }
 
+type endpointStateCallable = func(ctx context.Context, u *url.URL, inState spoof.ResponseChecker, desc string, r ...spoof.RequestOption) (*spoof.Response, error)
```
julz, line 87: do we think there are ever likely to be more than two possible things "implementing" this? If not I'd be very tempted to just pass down a `wait bool` parameter here

test/spoof/spoof_test.go
```diff
@@ -0,0 +1,254 @@
...
+			},
+			inState: func(resp *Response) (done bool, err error) {
+				return true, nil
+			},
+		},
+		wantErr:   false,
+		wantCalls: 1,
+	}}
+	for _, tt := range tests {
+		_, fKlient := fake.With(context.TODO(), ingress)
```
dprotaso, line 124: I would just create the k8s fake client directly vs pulling in the injection library

test/spoof/spoof.go
```diff
@@ -269,5 +287,54 @@ func (sc *SpoofingClient) WaitForEndpointState(
...
+		Body:       body,
+	}
+	ok, err := inState(resp)
+	if err != nil {
+		return resp, fmt.Errorf("response: %s did not pass checks: %w", resp, err)
+	}
+	if ok {
+		sc.logZipkinTrace(resp)
+		return resp, nil
+	}
```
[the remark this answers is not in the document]
shinigambit: It's very similar to the Poll method but I think resp is guaranteed to be non nil at this point so I just moved the log line out of the `if ok` check. Let me know if that's fine

PR #2168, opened 2021-06-24
Print proper warning logs from API warnings

markusthoemmes:
# Changes

Instead of writing to client-go's own `klog`, this writes API warnings (see https://kubernetes.io/blog/2020/09/03/warnings/) into the respective zap logger, so they'll be "proper" warnings in the log as well, which should make them easier to catch.

/assign @dprotaso @julz

### DIFF
--- a/logging/warning_handler.go
+++ b/logging/warning_handler.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logging
+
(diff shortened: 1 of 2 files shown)


logging/warning_handler.go
```diff
@@ -0,0 +1,33 @@
...
+import "go.uber.org/zap"
+
+// WarningHandler is a warning handler to forward client-go's warning logs into a zap logger.
+type WarningHandler struct {
+	Logger *zap.SugaredLogger
+}
+
+func (h *WarningHandler) HandleWarningHeader(code int, agent string, message string) {
+	// This condition is copied from K8s' default WarningLogger.
+	if code != 299 || len(message) == 0 {
```
julz, line 28: almost sure this is a silly question, but do we where 299 comes from / what it means?

julz: hah! Found it, it's apparently:

~~~~
   299 Miscellaneous persistent warning
     The warning text MAY include arbitrary information to be presented
     to a human user, or logged. A system receiving this warning MUST
     NOT take any automated action.
~~~~
from https://datatracker.ietf.org/doc/html/rfc2616#section-14.46

julz: sort of wish k8s had used a constant or something here but 🤷

julz (approved):
/lgtm

logging/warning_handler.go
dprotaso: Can we update the comment with the RFC? I feel the question will come up again

PR #2169, opened 2021-06-24
Replace reference to obsolete xip.io service

julz:
Xip.io is no more and we use sslip.io now. Updated the references in serving in https://github.com/knative/serving/pull/11589, updating here to match.


markusthoemmes (approved):
/lgtm
/approve

PR #2170, opened 2021-06-24
Add ability to filter objects on injection controller promotion

benmoss:
Currently we enqueue every object with no way to filter, which causes problems for eventing's source controller which reconciles duck CRDs.

ref: https://github.com/knative/eventing/issues/5543


markusthoemmes:
Great start, but I think it should be somewhat more generic to apply to all sorts of "global" events, i.e. resyncs. This also seems to be quite a bit of overlap with https://github.com/knative/pkg/pull/1909, so maybe get together with @lberk to hash out which way to turn?

Thanks for dealing with this :tada:

benmoss:
👍  that PR looks like a better approach, let me follow up with that

PR #2172, opened 2021-06-28
Added inverse functions for the ptr package

ellistarn:
# Changes

- :gift: Added corresponding Value functions for each ptr function. 

/kind enhancement

No issue, I just find myself needing these functions to avoid NPE errors in defaulting scenarios. This also exists within the [aws sdk](https://github.com/aws/aws-sdk-go/blob/e2d6cb448883e4f4fcc5246650f89bde349041ec/aws/convert_types.go), but I don't want to import *that* kitchen sink.

**Release Note**

```release-note
Added corresponding Value functions for each ptr function.
```

**Docs**

### DIFF
--- a/ptr/value.go
+++ b/ptr/value.go
@@ -1,16 +1,17 @@
 /*
 Copyright 2021 The Knative Authors
 
-Licensed under the Apache License, Version 2.0 (the "License"); you may not use
-this file except in compliance with the License. You may obtain a copy of the
-License at
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software distributed
-under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-CONDITIONS OF ANY KIND, either express or implied. See the License for the
-specific language governing permissions and limitations under the License.
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
 */
 
 package ptr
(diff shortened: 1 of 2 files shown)


ptr/value_test.go
```diff
@@ -0,0 +1,85 @@
+/*
+Copyright 2019 The Knative Authors
```
n3wscott, line 2: nit: 2021

ptr/value.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright 2019 The Knative Authors
```
n3wscott, line 2: nit: 2021

ptr/value_test.go
```diff
@@ -0,0 +1,85 @@
...
+*/
+
+package ptr
+
+import (
+	"testing"
+	"time"
+)
+
+func TestInt32Value(t *testing.T) {
```
n3wscott, line 24: I think you need to also add the nil cases here to get better code coverage, as that is the point you are adding the functions.

n3wscott (commented):
/ok-to-test

ptr/value_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+func TestFloat64Value_Nil(t *testing.T) {
+	want := float64(0)
+	gotValue := Float64Value(nil)
+	if want != gotValue {
+		t.Errorf("Float64() = &%v, wanted %v", gotValue, want)
+	}
+}
+
+
```
n3wscott, line 88: the linter does not like this extra line ^

ptr/value.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
```
n3wscott, line 4: the linter also does not like that this does not match what is in https://github.com/knative/pkg/blob/main/hack/boilerplate/boilerplate.go.txt

ellistarn:
Hang on -- I clobbered some of my tests =/

n3wscott:
well... come now flakes!

ellistarn:
image

ellistarn:
Thank you!

PR #2175, opened 2021-06-29
Enable the image template to fetch values from env

markusthoemmes:
Yeah, this looks a little weird, I know.

My use-case is this: Our CI system wants me to define clear dependencies on images it builds (the images are not built via `ko` but via a mechanism in said CI system). The CI system then publishes the built images pullref (sha based) into an environment variable of my choice.

By allowing the functions in this PR, this enables me to fetch these pullrefs from env via a template like

```
{{with $name := .Name}}{{ fromEnv (toEnvKey (printf "KNATIVE_SERVING_TEST_%s" $name)) }}{{end}}
```

:tada: 

/assign @dprotaso @n3wscott

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -22,6 +22,8 @@ package test
 import (
 	"bytes"
 	"flag"
+	"os"
+	"strings"
 	"text/template"
 
 	"k8s.io/client-go/tools/clientcmd"
@@ -68,7 +70,13 @@ func SetupLoggingFlags() {
 
 // ImagePath is a helper function to transform an image name into an image reference that can be pulled.
 func ImagePath(name string) string {
-	tpl, err := template.New("image").Parse(Flags.ImageTemplate)
+	tpl, err := template.New("image").Funcs(template.FuncMap{
+		"toEnvKey": func(in string) string {
+			underscores := strings.ReplaceAll(in, "-", "_")
+			return strings.ToUpper(underscores)
+		},
+		"fromEnv": os.Getenv,
+	}).Parse(Flags.ImageTemplate)
 	if err != nil {
 		panic("could not parse image template: " + err.Error())
 	}


test/e2e_flags.go
```diff
@@ -68,7 +70,13 @@ func SetupLoggingFlags() {
 
 // ImagePath is a helper function to transform an image name into an image reference that can be pulled.
 func ImagePath(name string) string {
-	tpl, err := template.New("image").Parse(Flags.ImageTemplate)
+	tpl, err := template.New("image").Funcs(template.FuncMap{
+		"toEnvKey": func(in string) string {
+			underscores := strings.ReplaceAll(in, "-", "_")
```
julz, line 75: just to ask the question: do we actually need `toEnvKey`/`fromEnv`? Another approach would be for the template to be (though I guess that ends up a pretty ugly template string that you have to build somehow so 🤷):

~~~~
{{with .Name }}
{{if eq . "readiness"}}docker.io/flibble/blobble{{end}}
{{if eq . "varlog"}}docker.io/flibble/bloop{{end}}
..
~~~~

Im wondering a bit if it's worth merging toEnvKey and fromEnv into something like `fromPrefixedEnv("KNATIVE_SERVING_TEST_", name)` since I find it a bit hard to follow the flow with multiple template functions calling each other

test/e2e_flags.go
```diff
@@ -68,7 +70,13 @@ func SetupLoggingFlags() {
...
+			return strings.ToUpper(underscores)
+		},
+		"fromEnv": os.Getenv,
```
julz, line 78: could use `os.ExpandEnv` here which would avoid needing the printf in the template

markusthoemmes:
Thanks @julz for the suggestions. I actually don't need this anymore.

PR #2176, opened 2021-06-30
Allow global resync & reconcilerImpl informers to be filtered

benmoss:
# Changes

A revised implementation of #2170 and #1909, updated and now using options as per @mattmoor's [ancient code review](https://github.com/knative/pkg/pull/1909/files#r530448985).

Also switched our reconcilers to use `controller.NewImplFull` since `NewImpl` is deprecated, and this makes setting the `GlobalResyncFilterFunc` easier too.

This came up again because of https://github.com/knative/eventing/issues/5543, so looking to reintroduce @lberk's changes.

/cc @markusthoemmes

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_controller.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_controller.go
@@ -140,6 +140,10 @@ func (g *reconcilerControllerGenerator) GenerateType(c *generator.Context, t *ty
 			Package: "knative.dev/pkg/controller",
 			Name:    "OptionsFn",
 		}),
+		"controllerGetFilterFunc": c.Universe.Type(types.Name{
+			Package: "knative.dev/pkg/controller",
+			Name:    "GetFilterFunc",
+		}),
 		"contextContext": c.Universe.Type(types.Name{
 			Package: "context",
 			Name:    "Context",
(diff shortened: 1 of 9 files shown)


codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -251,7 +261,12 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 	)
 
 
-	impl := {{.controllerNewImpl|raw}}(rec, logger, ctrTypeName)
+	options := {{.controllerControllerOptions|raw}}{
+		Logger: logger,
+		WorkQueueName: ctrTypeName,
+		GlobalResyncFilterFunc: func(o interface{}) bool { filterFunc(o) },
```
benmoss, line 267: a hacky trick here, wrapping the `filterFunc` in another function closes over the value so we can mutate it down below, passing `filterFunc` directly would pass by value

codegen/cmd/injection-gen/generators/reconciler_controller.go
```diff
@@ -225,11 +234,12 @@ func NewImpl(ctx {{.contextContext|raw}}, r Interface{{if .hasClass}}, classValu
 					return err
 				}
 				for _, elt := range all {
-					// TODO: Consider letting users specify a filter in options.
```
n3wscott, line 228: I love it when we find a need for the TODO and then implement the TODO 😄

n3wscott (commented):
This LGTM, can I ask you to add a small note in the docs for it so someone reading the docs might realize they can use this option: https://github.com/knative/pkg/blob/main/injection/README.md

/lgtm

benmoss:
Opened #2180 instead

PR #2178, opened 2021-07-01
K8s Service trailing-slash fix in URI resolver  

Shashankft9:
# Changes

- removes the trailing slash for the resolution of URI from k8s service. Same change in the test file too.

/kind bug

Fixes # 
The URI resolved for lets say a channel or a knative service does not have the trailing slash. For example, below is the status of a pingsource with a channel as sinkRef:
```
status:
  ceAttributes:
  - source: /apis/v1/namespaces/serverless/pingsources/test-ping-source
    type: dev.knative.sources.ping
  conditions:
  - lastTransitionTime: "2021-07-01T11:20:50Z"
    status: "True"
    type: Deployed
  - lastTransitionTime: "2021-07-01T11:20:50Z"
    status: "True"
    type: Ready
  - lastTransitionTime: "2021-07-01T11:20:50Z"
    status: "True"
    type: SinkProvided
  observedGeneration: 1
  sinkUri: http://my-channel-kn-channel.serverless.svc.cluster.local
  ```
  
Noticed this while working on a component that uses the `status.SinkUri` of eventing sources. Please let me know if for some reason a trailing slash for the resolution of k8s service is intended.

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -174,7 +174,7 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		url := &apis.URL{
 			Scheme: "http",
 			Host:   network.GetServiceHostname(ref.Name, ref.Namespace),
-			Path:   "/",
+			Path:   "",
 		}
 		return url, nil
 	}
--- a/resolver/addressable_resolver_test.go
+++ b/resolver/addressable_resolver_test.go
@@ -420,7 +420,7 @@ func TestGetURIDestinationV1(t *testing.T) {
 				getAddressableFromKRef(k8sServiceRef()),
 			},
 			dest:    duckv1.Destination{Ref: k8sServiceRef()},
-			wantURI: "http://testsink.testnamespace.svc.cluster.local/",
+			wantURI: "http://testsink.testnamespace.svc.cluster.local",
 		}, "ref with relative uri": {
 			objects: []runtime.Object{
 				getAddressable(),


n3wscott:
This change makes sense, I want to double check it will not break eventing's unit tests, waiting for ci

Shashankft9:
All these tests are failing: https://github.com/knative/eventing/blob/main/pkg/reconciler/subscription/subscription_test.go#L1190-L1586 because they are expecting the trailing slash by using `serviceURIWithPath`. 
So, a probable fix in eventing to change this behavior: https://github.com/knative/eventing/pull/5560

Shashankft9:
@n3wscott ping :)

vaikas:
Adding a hold since it would be nice if we also had a PR to fix the eventing-kafka-broker which seems to be failing in a similar fashion so we don't break them without a fix?
/hold
/lgtm
/approve

Once there's a fix for the PR, feel free to remove the hold :)

vaikas:
/ok-to-test
odd since the tests ran, but adding it regardless :)

Shashankft9:
@vaikas right! completely forgot to do the the fix in eventing-kafka-broker. Just created a PR in that repo as a fix.

pierDipi (commented):
Thanks @Shashankft9, @vaikas!

While a trailing slash might be annoying, this breaks existing applications, not sure if it's worth it.
Could you please explain your reasoning to do this?

Shashankft9:
@pierDipi I am working on a component that makes use of `status.SinkURI` of sources (and also uses knative pkg), in some map and string split operations, there I noticed this inconsistency with trailing slash, it was breaking few things because the code wasn't expecting a trailing slash, so I checked and it seemed like all the other eventing refs are resolved into a URI without it except k8s service - although arguably my code can be improved/changed, I thought that its an inconsistency in the knative pkg anyway

pierDipi:
Thanks, sounds good to me!

PR #2180, opened 2021-07-06
Add ability to filter objects on injection controller promotion

benmoss:
Currently we enqueue every object with no way to filter, which causes problems for eventing's source controller which reconciles duck CRDs.

I decided that since I didn't have justification for the changes in #2176 that I would feel better not making them. If we come up with a reason we need to support filtering `GlobalResync` from the constructor rather than just by using `FilteredGlobalResync` we can revisit that later.

ref: knative/eventing#5543

### DIFF
--- a/controller/options.go
+++ b/controller/options.go
@@ -41,6 +41,11 @@ type Options struct {
 
 	// Concurrency - The number of workers to use when processing the controller's workqueue.
 	Concurrency int
+
+	// PromoteFilterFunc filters the objects that are enqueued when the reconciler is promoted to leader.
+	// Objects that pass the filter (return true) will be reconciled when a new leader is promoted.
+	// If no filter is specified, all objects will be reconciled.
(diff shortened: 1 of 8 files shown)


controller/options.go
```diff
@@ -41,6 +41,9 @@ type Options struct {
 
 	// Concurrency - The number of workers to use when processing the controller's workqueue.
 	Concurrency int
+
+	// PromoteFilterFunc filters the objects that are enqueued when the reconciler is promoted to leader.
```
n3wscott, line 45: Can you add a comment on what returning true/false means in the context of the function and how it used?

Filter is always hard because true = the obj is filtered out? true = it passes the filter?

benmoss: Added more to the comment text, hopefully it's more clear now

n3wscott (approved):
/lgtm

Thanks!

Issue #2173, opened 2021-06-28
Remove stack driver exporter


dprotaso:
/area API
/area monitoring
/kind cleanup

@MontyCarter [gave the clear that the stack driver exporter can be removed](https://knative.slack.com/archives/CEPGNPHP1/p1624482429085800?thread_ts=1624296049.079200&cid=CEPGNPHP1)

dprotaso:
Sent an email to the various mailing lists
knative-users - https://groups.google.com/g/knative-users/c/YEH0Y8QAaJ0
Knative-dev - https://groups.google.com/g/knative-dev/c/OAq4EJ4LEAU

skonto:
Ok so that means we now will have opencensus and prometheus direct support only right?

dprotaso:
@skonto yes

dprotaso:
There hasn't been any objections so I'm going 'unhold' the PR

Issue #2174, opened 2021-06-28
Metric names between OpenCensus & Prometheus exporters are different


dprotaso:
/area monitoring
/kind cleanup

We currently prepend a 'domain' and the 'component' name to the metrics. This done for our OpenCensus exporter which might have been cargo cult'd from our stack driver exporter implementation. The prometheus exporter _only_ prefixes the 'component' name to the metric.

https://github.com/knative/pkg/blob/cf1994e9cc24db63cbc8a21e1a84152b732ad3f0/metrics/opencensus_exporter.go#L39

We should determine whether these should be consistent, if there should be a prefix and it's format

cc @upodroid @skonto @evankanderson @MontyCarter @pianohacker

dprotaso:
From @pianohacker re: prepending a domain prefix for prometheus  https://knative.slack.com/archives/CEPGNPHP1/p1624558123099100 

> This level of namespacing on the metric name isn't typically done for Prometheus metrics; see https://prometheus.io/docs/practices/naming/
The source (as opposed to the kind) of a Prometheus-exposed metric is intended to be distinguished by its labels (in this case, namespace). 
> In that vein, could we expose the ability to configure one or more labels on the exposed metrics?

dprotaso:
cc @evankanderson who reviewed the OpenCensus exporter changes

skonto:
Regarding Prometheus this is what I read:
```
...should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. For metrics specific to an application, the prefix is usually the application name itself. Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries. Examples:
prometheus_notifications_total (specific to the Prometheus server)
process_cpu_seconds_total (exported by many client libraries)
http_request_duration_seconds (for all HTTP requests)
```
I am not sure what value the domain adds here we already use the component name as the prefix and we know the domain for each component, it is straightforward. However, there is also a need to be able to add custom labels eg. tag metrics with special labels to avoid joining metrics info based on kube-state-metrics or other sources. I am facing this downstream as I would like to tag metrics with the `func` tag for [Boson funcs](https://github.com/boson-project/functions).

dprotaso:
@skonto @evankanderson during the operation wg meeting mentioned you wrote a doc about stable metric names and a potential deprecation policy. Do you have a link to that?

Whatever we decide with this change we'll want to roll it out gradually with a flag.

[...]

PR #2183, opened 2021-07-08
drop stackdriver metrics and tracing exporters

dprotaso:
- drop stackdriver as a tracing backend
- drop stackdriver as a metrics backend
- update deps - dropping stackdriver

Fixes: https://github.com/knative/pkg/issues/2173


metrics/config_test.go
```diff
@@ -30,27 +27,14 @@ import (
...
-
 const (
-	servingDomain          = "knative.dev/serving"
-	internalServingDomain  = "knative.dev/internal/serving"
-	eventingDomain         = "knative.dev/eventing"
-	internalEventingDomain = "knative.dev/internal/eventing"
-	customSubDomain        = "test.domain"
-	testProj               = "test-project"
-	anotherProj            = "another-project"
+	metricsDomain = "knative.dev/project"
```
skonto, line 37: What was the value of having different domains?

skonto:
@dprotaso I still see some references to the `stackdriver` name like here: https://github.com/knative/pkg/blob/952ee7db67049390a6a1067657370b711400d433/logging/config.go#L58
Shouldn't we remove it from the whole codebase?

dprotaso:
re:
```go
loggingCfg := stackdriverConfig() 
```

This just produces a logging format that stackdriver can parse easily - I'm ok with leaving this since it helps debugging with our CI/prow runs. 

Do you have suggestions on other alternatives here?

metrics/config_test.go
dprotaso: This was a stack driver thing but we're actually prefixing this to the open census metric names so I think this should keep this around until we resolve: https://github.com/knative/pkg/issues/2174


I think there's some value in repurposing this as a label in order to help disambiguate between components across different Knative projects (ie. serving vs eventing).

ie. in the future we drop the metric prefix and thus require labels to help segment metrics.

So for a golang memory metric - we'd have a component label ie. `controller`  and reuse the domain label `serving`

I'm ok to _drop_ this domain if an alternative exists - ie. if the components namespace is automatically picked up as part of the metric

skonto:
> Do you have suggestions on other alternatives here?
I can do a followup PR to cleanup anything that remains and discuss them in detail one by one. For that particular logging config func I would just rename it, if it does not do any harm.

dprotaso:
/hold 

I'll fix the downstream serving issues

dprotaso:
Downstream issue should be resolved here:
https://github.com/knative/serving/pull/11659

dprotaso:
Moving Eventing Source StatReporter out of knative.dev/pkg - https://github.com/knative/eventing/pull/5587

dprotaso:
I'm going to temporarily move the eventing sources metric keys to `pkg/source` until we can move the stats reporter into the eventing repo.

dprotaso:
last PR to fix serving downstream: https://github.com/knative/serving/pull/11679

metrics/config_observability.go
```diff
@@ -65,8 +65,8 @@ type ObservabilityConfig struct {
 	// EnableProbeRequestLog enables queue-proxy to write health check probe request logs.
 	EnableProbeRequestLog bool
 
-	// RequestMetricsBackend specifies the request metrics destination, e.g. Prometheus,
-	// Stackdriver. "None" disables all backends.
+	// RequestMetricsBackend specifies the request metrics destination, e.g. Prometheus.
+	// "None" disables all backends.
```
evankanderson, line 69: Should this have OpenCensus?

metrics/config_test.go
evankanderson: I think the namespace is likely to be picked up as part of the metric; I'm not sure if we have to / want to worry about cases like queue-proxies and sources running in a user's namespace.

I'm inclined to think that namespace + some sort of component/Resource name should be sufficient here, but I might be mistaken.

evankanderson: (With the notion that the Resource would indicate whether it was a serving or eventing construct if it was running in a user namespace, and that multi-tenant components would be running in a namespace where the name would give a clue.)

The obvious counter-example here is `mink`, where all the controllers are bundled into a single Pod, but I'm not sure that's a compelling enough case.

metrics/config_test.go
```diff
@@ -835,7 +507,6 @@ func TestUpdateExporter(t *testing.T) {
 func TestUpdateExporterFromConfigMapWithOpts(t *testing.T) {
 	setCurMetricsConfig(nil)
 	oldConfig := getCurMetricsConfig()
-	successTestsInit()
```
evankanderson, line 838: Do we still need `successTestsInit`? I've seen it removed several places here.

evankanderson (commented):
/lgtm
/approve

I'm amused that we're deleting about 10k lines of `aws-sdk-go` with this change. (Also, wow the generated proto files stackdriver pulled in were big...)

metrics/config_test.go
dprotaso: It's removed - in the diff

metrics/config_observability.go
dprotaso: I'll add it back

PR #2191, opened 2021-07-15
Removing default for KUBECONFIG to enable incluster e2e tests

### LINKED ISSUE #2190: e2e tests cannot be run inCluster (for no clear reason)
/area test-and-release

/kind bug
/kind cleanup

-->

## Expected Behavior

e2e tests should be run in cluster as well as outside a cluster

## Actual Behavior
e2e tests default the KUBECONFIG which makes it impossible to run in cluster. 

Unset the KUBECONFIG before running the cluster, and you will see that the tests still look for `.kube/config` blocking the InCluster client to be created. 

## Additional Info

Issue created in `serving`
https://github.com/knative/serving/issues/11686

salaboy:
# Changes

Removing default setting to KUBECONFIG in e2e flags. This allows e2e testing to be done from inside a Kubernetes Cluster. 

/kind bug 
/kind cleanup

Fixes #2190

closes #2190

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -24,7 +24,6 @@ import (
 	"flag"
 	"text/template"
 
-	"k8s.io/client-go/tools/clientcmd"
 	env "knative.dev/pkg/environment"
 	testenv "knative.dev/pkg/test/environment"
 	"knative.dev/pkg/test/logging"
@@ -49,13 +48,6 @@ func initializeFlags() *EnvironmentFlags {
 
 	f.ClientConfig.InitFlags(flag.CommandLine)
 	f.TestClientConfig.InitFlags(flag.CommandLine)
-
-	// We want to do this defaulting for tests only. The flags are reused between tests
-	// and production code and we want to make sure that production code defaults to
-	// the in-cluster config correctly.
-	if f.Kubeconfig == "" {
-		f.Kubeconfig = clientcmd.RecommendedHomeFile
-	}
 
 	return f
 }


salaboy:
@markusthoemmes I will appreciate feedback on this one, as I am not entirely sure to understand the comments above the if statement. Is there a reason why tests cannot be run incluster? Because this check here is definitely blocking that possibility, unless there is a way to bypass that check.

test/e2e_flags.go
```diff
@@ -50,13 +49,6 @@ func initializeFlags() *EnvironmentFlags {
 	f.ClientConfig.InitFlags(flag.CommandLine)
 	f.TestClientConfig.InitFlags(flag.CommandLine)
 
-	// We want to do this defaulting for tests only. The flags are reused between tests
-	// and production code and we want to make sure that production code defaults to
-	// the in-cluster config correctly.
-	if f.Kubeconfig == "" {
-		f.Kubeconfig = clientcmd.RecommendedHomeFile
```
julz, line 57: doesn't this mean that this will no longer properly default the flag when running locally?

markusthoemmes: So, checking the code, I think this should work, as the fallback is precisely this path as well.

julz: yeah, looks like the fallback was added after this code, which explains why it was added, and why it's now redundant 🤞

salaboy:
@julz @markusthoemmes tests seems to be failing but unrelated with the change.. I will figure out how to reproduce that.. or should we just retrigger the tests

dprotaso:
re-ran and the unit tests ran fine

/lgtm
/approve

PR #2195, opened 2021-07-19
Addressable resolver now accepts custom resolvers

lionelvillard:
# Changes

- :gift: The addressable resolver can now be configured to run custom resolvers before falling back to the default duck-type based URI resolution.
- Needed for https://github.com/knative/eventing/issues/5593
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -38,16 +38,21 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
+type URIResolverFunc func(ctx context.Context, ref *corev1.ObjectReference) (*apis.URL, error)
+
 // URIResolver resolves Destinations and ObjectReferences into a URI.
 type URIResolver struct {
 	tracker       tracker.Interface
 	listerFactory func(schema.GroupVersionResource) (cache.GenericLister, error)
+	resolvers     []URIResolverFunc
 }
 
 // NewURIResolver constructs a new URIResolver with context and a callback
 // for a given listableType (Listable) passed to the URIResolver's tracker.
-func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *URIResolver {
-	ret := &URIResolver{}
+func NewURIResolver(ctx context.Context, callback func(types.NamespacedName), resolvers ...URIResolverFunc) *URIResolver {
+	ret := &URIResolver{
+		resolvers: resolvers,
+	}
 
 	ret.tracker = tracker.New(callback, controller.GetTrackerLease(ctx))
 
@@ -146,6 +151,17 @@ func (r *URIResolver) URIFromKReference(ctx context.Context, ref *duckv1.KRefere
 func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.ObjectReference, parent interface{}) (*apis.URL, error) {
 	if ref == nil {
 		return nil, apierrs.NewBadRequest("ref is nil")
+	}
+
+	// try custom resolvers first
+	for _, resolver := range r.resolvers {
+		url, err := resolver(ctx, ref)
+		if err != nil {
+			return nil, err
+		}
+		if url != nil {
+			return url, nil
+		}
 	}
 
 	gvr, _ := meta.UnsafeGuessKindToResource(ref.GroupVersionKind())
--- a/resolver/addressable_resolver_test.go
+++ b/resolver/addressable_resolver_test.go
@@ -18,6 +18,7 @@ package resolver_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 
@@ -379,10 +380,11 @@ func TestGetURIDestinationV1Beta1(t *testing.T) {
 
 func TestGetURIDestinationV1(t *testing.T) {
 	tests := map[string]struct {
-		objects []runtime.Object
-		dest    duckv1.Destination
-		wantURI string
-		wantErr string
+		objects         []runtime.Object
+		dest            duckv1.Destination
+		customResolvers []resolver.URIResolverFunc
+		wantURI         string
+		wantErr         string
 	}{"nil everything": {
 		wantErr: "destination missing Ref and URI, expected at least one",
 	}, "Happy URI with path": {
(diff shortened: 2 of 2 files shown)


lionelvillard:
@markusthoemmes @vaikas Can you PTAL, ideally before tomorrow pkg release? I believe this PR is completely harmless and if it turns out that eventing#5593 does not work for users, I'll submit a PR to revert this change.

markusthoemmes (commented):
Looks okay implementation wise. Not sure we should push this in just before a release cut though, given the discussion state of the referenced issue. I'll defer to eventing peeps for that.

lionelvillard:
@vaikas WDYT?

lionelvillard:
now that the release happened, can we get this in? 🙏 @markusthoemmes

resolver/addressable_resolver.go
```diff
@@ -38,16 +38,21 @@ import (
...
+
 // URIResolver resolves Destinations and ObjectReferences into a URI.
 type URIResolver struct {
 	tracker       tracker.Interface
 	listerFactory func(schema.GroupVersionResource) (cache.GenericLister, error)
+	resolvers     []URIResolverFunc
 }
 
 // NewURIResolver constructs a new URIResolver with context and a callback
 // for a given listableType (Listable) passed to the URIResolver's tracker.
```
n3wscott, line 51: Please add documentation for the new argument `resolvers`.

resolver/addressable_resolver.go
```diff
@@ -38,16 +38,21 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
+type URIResolverFunc func(ctx context.Context, ref *corev1.ObjectReference) (*apis.URL, error)
```
n3wscott, line 41: Please document `URIResolverFunc`

n3wscott: I do wonder if we need a different return signature here, we can set the resolve up with more than one func, so the func needs to be able to say "I passed on this one, but it is not an error":

`(bool, *apis.URL, error)`

Then if the func resolves, it is `true, aURL, nil`
If the func passes it is `false, nil, nil`
If it hit a real error it is `true, nil, anError`

resolver/addressable_resolver.go
```diff
@@ -148,6 +153,17 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		return nil, apierrs.NewBadRequest("ref is nil")
 	}
 
+	// try custom resolvers first
+	for _, resolver := range r.resolvers {
+		url, err := resolver(ctx, ref)
+		if err != nil {
+			return nil, err
+		}
+		if url != nil {
```
n3wscott, line 162: here is where the returned bool would come in handy and you don't have to guess if a nil url with a nil error means skip. Which is not documented behaviour and not expected.

resolver/addressable_resolver_test.go
```diff
@@ -532,13 +534,28 @@ func TestGetURIDestinationV1(t *testing.T) {
 		}, "notFound k8s service": {
 			dest:    duckv1.Destination{Ref: k8sServiceRef()},
 			wantErr: fmt.Sprintf("services %q not found", addressableName),
+		}, "with sample resolver": {
```
n3wscott, line 537: I would like to see a test with two custom resolvers injected, and the first one passes.

resolver/addressable_resolver.go
```diff
@@ -38,16 +38,21 @@ import (
...
 	tracker       tracker.Interface
 	listerFactory func(schema.GroupVersionResource) (cache.GenericLister, error)
+	resolvers     []URIResolverFunc
 }
 
 // NewURIResolver constructs a new URIResolver with context and a callback
 // for a given listableType (Listable) passed to the URIResolver's tracker.
-func NewURIResolver(ctx context.Context, callback func(types.NamespacedName)) *URIResolver {
-	ret := &URIResolver{}
+func NewURIResolver(ctx context.Context, callback func(types.NamespacedName), resolvers ...URIResolverFunc) *URIResolver {
```
julz, line 52: Have you considered using  composition here rather than adding logic to the existing struct? The URIResolverFunc seems like it probably wants to be an interface eventually anyway (once it gets parameterised and has an http client in it), so I wonder whether we could keep NewURIResolver as it is, and then add eg NewLookasideURIResolver(ctx, callback, fallback). That would also make it easier to - for example - opt out of the fallback case or alter the logic for when a resolver does or doesn't fall back.

julz: Ugh I just realised we use the struct directly rather than a Resolver interface in loads of places so we'd need to change all of those to use an interface. Bah. Oh well.

resolver/addressable_resolver.go
lionelvillard: done!

resolver/addressable_resolver.go
lionelvillard: sounds good.

resolver/addressable_resolver.go
lionelvillard: revolved?

resolver/addressable_resolver_test.go
lionelvillard: done and added a bonus test

lionelvillard:
should be good now @n3wscott

resolver/addressable_resolver.go
julz: Yeah, I think there's a potentially nice refactor here, but it's way more of a can of worms than I thought. Either way I wouldn't block the PR on it 👍

resolver/addressable_resolver.go
```diff
@@ -153,6 +159,14 @@ func (r *URIResolver) URIFromObjectReference(ctx context.Context, ref *corev1.Ob
 		return nil, apierrs.NewBadRequest("ref is nil")
 	}
 
+	// try custom resolvers first
+	for _, resolver := range r.resolvers {
+		handled, url, err := resolver(ctx, ref)
```
n3wscott, line 164: we likely want to check the error here before handled. Looks like err is lots when handled is false.

I am not sure what is right, but it could do with a comment to say if we can ignore err if handled is false here.

lionelvillard: added a comment. Please TAL.

lionelvillard (commented):
clicked on the wrong button

lionelvillard:
@n3wscott can you please TAL?

PR #2197, opened 2021-07-22
Use default Kubeconfig loading rules

benmoss:
This mimics the behavior of tools like kubectl, should work for both in-cluster and out-of-cluster usages

- :bug: Fix defaulting to `~/.kube/config` without an explicitly set KUBECONFIG

/kind bug

### DIFF
--- a/environment/client_config.go
+++ b/environment/client_config.go
@@ -17,7 +17,6 @@ limitations under the License.
 package environment
 
 import (
-	"errors"
 	"flag"
 	"fmt"
 	"math"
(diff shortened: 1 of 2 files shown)


environment/client_config.go
```diff
@@ -59,47 +58,29 @@ func (c *ClientConfig) GetRESTConfig() (*rest.Config, error) {
...
+	if c.Kubeconfig != "" {
+		loadingRules.ExplicitPath = c.Kubeconfig
+	}
 	if c.Cluster != "" {
 		overrides.Context = clientcmdapi.Context{Cluster: c.Cluster}
 	} else if c.ServerURL != "" {
 		overrides.ClusterInfo = clientcmdapi.Cluster{Server: c.ServerURL}
 	}
 
 	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
```
dprotaso, line 73: unsure if this error `err` has context about what it was doing - ie. could not create valid kubeconfig.

If it doesn't can we wrap it prior to returning

test/clients.go
```diff
@@ -41,13 +41,18 @@ func NewSpoofingClient(ctx context.Context, client kubernetes.Interface, logf lo
 
 // BuildClientConfig builds the client config specified by the config path and the cluster name
 func BuildClientConfig(kubeConfigPath string, clusterName string) (*rest.Config, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	overrides := clientcmd.ConfigOverrides{}
+
+	if kubeConfigPath != "" {
+		loadingRules.ExplicitPath = kubeConfigPath
+	}
```
dprotaso, line 49: what's this changed for?

environment/client_config.go
dprotaso, line 73: For clarity what's the priority and fallback order now?

benmoss: Wrong lines?

benmoss: - A kubeconfig set by flags
- A kubeconfig set by `KUBECONFIG`
- ~/.kube/config
- in-cluster config

test/clients.go
benmoss: This is what makes it so that if a user provides a kubeconfig via flags it will still get used

environment/client_config.go
dprotaso: yeah maybe this is applicable to the block

```go
if err != nil {
  	return nil, err
}
```

I guess my point is - if this function returns low level error messages (ie. yaml syntax error) should we wrap it

```go
return nil, fmt.Errorf("failed to create REST config: %w", err)
```

PR #2198, opened 2021-07-22
include a flag to disable metric prefixes

dprotaso:
the default continues to include the prefix

Part of: https://github.com/knative/pkg/issues/2174


dprotaso:
I'm going to get rid of that downstream serving test that tests logic that exists in knative.dev/pkg

dprotaso:
/hold cancel
ok done: https://github.com/knative/serving/pull/11711

metrics/config_observability.go
```diff
@@ -79,13 +79,22 @@ type ObservabilityConfig struct {
...
 	// when the metrics backend is opencensus.
 	MetricsCollectorAddress string
+
+	// EnableDeprecatedMetricPrefix retains the legacy behaviour of prefixing our metric names
+	// with either the component name and/or metric domain
+	//
+	// When enabled:
+	// - OpenCensus exporter prefixes the domain and component
+	// - Prometheus exporter prefixes the component name (as a prom namespace)
+	EnableDeprecatedMetricPrefix bool
```
skonto, line 89: Since this is a breaking change and I see there is a plan [here](https://github.com/knative/pkg/issues/2174#issuecomment-875009789) for mirgation, next question would be if there is a plan for removing this flag or users should just keep this flag on.

metrics/prometheus_exporter.go
```diff
@@ -43,7 +43,11 @@ func (emptyPromExporter) ExportView(viewData *view.Data) {
 
 //nolint: unparam // False positive of flagging the second result of this function unused.
 func newPrometheusExporter(config *metricsConfig, logger *zap.SugaredLogger) (view.Exporter, ResourceExporterFactory, error) {
-	e, err := prom.NewExporter(prom.Options{Namespace: config.component})
+	opts := prom.Options{}
+	if config.enableDeprecatedMetricPrefix {
+		opts.Namespace = config.component
```
skonto, line 48: Why we set the Namespace to be equal to a component name here?

metrics/config_observability_test.go
```diff
@@ -35,22 +35,24 @@ func TestObservabilityConfiguration(t *testing.T) {
...
-			"metrics.request-metrics-backend-destination": "opencensus",
-			"profiling.enable":                            "true",
+			EnableProbeReqLogKey:                "true",
+			"logging.enable-var-log-collection": "true",
+			ReqLogTemplateKey:                   `{"requestMethod": "{{.Request.Method}}"}`,
+			"logging.revision-url-template":     "https://logging.io",
+			EnableReqLogKey:                     "true",
+			"profiling.enable":                  "true",
+			requestMetricsBackendKey:            "opencensus",
+			enableDeprecatedMetricPrefixKey:     "false",
```
skonto, line 55: Should we test `enableDeprecatedMetricPrefixKey` true here or e2e test is enough?

metrics/prometheus_exporter.go
dprotaso: this is the legacy behaviour - which I want to preserve with this flag

metrics/config_observability.go
dprotaso: Probably worth surfacing this topic on the thread.

dprotaso:
bump @skonto @evankanderson

evankanderson:
/assign @MontyCarter 

I think Monty already said that this wouldn't mess up Google, so I'm all in favor.

skonto:
@dprotaso I am fine with this as long as there is an option to define the component name etc. 
Right now for metrics like `source_name_event_count` ambiguity/conflict will be created when all metrics will be gathered at the Prometheus side if we completely remove the prefix.

/lgtm

dprotaso:
Yeah the component name would have to move to a label. I'm just introducing this flag so folks can set it explicitly to `true` prior to us changing the default in the future

MontyCarter:
Sorry for the delay here.  I've switched teams; you should probably update your contact on metrics to @ZhiminXiang.

With the ability to keep the legacy behavior in this PR:
/lgtm

Looking forward, Google has already switched to a collector based setup; metric name/label rewrites are no problem.

With regards to the component name, as long as this is available as a label, we can rewrite things in the collector config.  

With regards to domain portion of MetricPrefix, our domain value changes with component.  It really would be nice to retain the ability to read `domain` into the config and pipe that through to a label if we get rid of support for MetricPrefix.  

Acceptable paths forward for us would include:
 - Keeping MetricsPrefix behavior as it is in this PR.
 - Retaining MetricsPrefix legacy behavior, but make the default false (we must opt in).
 - Continuing to read domain and component into config, but passing these as labels instead.

skonto:
@dprotaso gentle ping, been a while, planning to continue this?

PR #2201, opened 2021-07-29
Add a new mechanism for requeuing a key.

mattmoor:
# Changes

This is modeled after some of the semantics available in controller-runtime's `Result` type, which allows reconcilers to indicate a desire to reprocess a key either immediately or after a delay.

I worked around our lack of this when I reworked Tekton's timeout handling logic by exposing a "snooze" function to the Reconciler that wrapped `EnqueueAfter`, but this feels like a cleaner API for folks to use in general, and is consistent with our `NewPermanentError` and `NewSkipKey` functions in terms of influencing the queuing behaviors in our reconcilers via wrapped errors.

You can see some discussion of this here: https://knative.slack.com/archives/CA4DNJ9A4/p1627524921161100

/kind enhancement

**Release Note**

```release-note
Introduce a way for Knative-style controllers to requeue keys by returning a wrapped error from Reconcile[Kind].  See NewRequeue{Immediately,After} for more details.
```

**Docs**

Shouldn't have user-facing docs impact.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -555,6 +555,12 @@ func (c *Impl) handleErr(err error, key types.NamespacedName, startTime time.Tim
 		c.workQueue.Forget(key)
 		return
 	}
+	if ok, delay := IsRequeueKey(err); ok {
+		c.workQueue.AddAfter(key, delay)
+		c.logger.Debugf("Requeuing key %s (by request) after %v (depth: %d)", safeKey(key), delay, c.workQueue.Len())
+		return
+	}
+
 	c.logger.Errorw("Reconcile error", zap.Duration("duration", time.Since(startTime)), zap.Error(err))
 
 	// Re-queue the key if it's a transient error.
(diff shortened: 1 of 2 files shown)


vaikas:
UT failed with:
```
--- FAIL: TestAdmissionValidResponseForResource (0.21s)
    logger.go:130: 2021-07-29T21:15:09.052Z	ERROR	webhook/webhook.go:196	failed to fetch secret	{"error": "secret \"webhook-certs\" not found"}
    admission_integration_test.go:139: Failed to get response Get "https://0.0.0.0:43137/bazinga": remote error: tls: unrecognized name
    admission_integration_test.go:176: Admit was called before informers had synced.
    logger.go:130: 2021-07-29T21:15:09.052Z	INFO	webhook/webhook.go:235	Starting to fail readiness probes...
    ```

mattmoor:
I haven't touched that and can't rerun it because I'm not an editor.

vaikas:
Yeah, I'm rerunning them :)
/lgtm
/approve
/hold

Just adding a hold so the other folks from the thread have a chance to take a look. If they don't have concerns, I'll remove the hold tmw morning.
@markusthoemmes

mattmoor:
Looks like we want codegen to not emit an event for these too, so I'll have a small follow-up shortly.

PR #2204, opened 2021-07-31
Support the webhook serving over non-TLS.

mattmoor:
# Changes

This reworks the way we setup the TLS Config for the webhook, so that when the `opts.SecretName` is omitted we serve over non-TLS.

/kind enhancement

**Release Note**

N/A

**Docs**

N/A


mattmoor:
Ok, this should be RFAL.  I was able to use this to run a Knative-style webhook as a ksvc 😁 

/assign @vaikas @n3wscott

PR #2207, opened 2021-08-09
[WIP] Allow specifying operations of a webhook rule for a resource

pierDipi:
Allow specifying operations of a webhook rule for a resource.

# Changes

- Allow specifying operations of a webhook rule using `OperationTypes` extension interface

/kind enhancement


pierDipi:
Hi @markusthoemmes, @julz, @dprotaso can you please review this PR?

julz:
can you say a bit more about the context / use case etc? (or a link to an issue if there is one maybe)

pierDipi:
I don't have a specific issue, basically, I need to update a resource (`setDefaults`) only when it is created and it's unnecessary call the webhook on updates.

dprotaso:
> I don't have a specific issue, basically, I need to update a resource (setDefaults) only when it is created and it's unnecessary to call the webhook on updates.

This is problematic and assumes only one actor in the system. In theory you could have some other component in the system modifying your object on `UPDATE`. So I would recommend validating the object.

pierDipi:
The field that I want to change is immutable.

dprotaso:
> Yes, the field that I want to change is immutable.

Can you share more context about what's happening and why?

pierDipi:
I need to change the `Volumes` field of a `Pod` (and related `VolumeMounts`) because I need to mount a different `ConfigMap` for each `Pod` created by a `Deployment`.

dprotaso:
What's immutable in that scenario?

pierDipi:
`Volumes` cannot be changes on raw pods

pierDipi:
https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement

> Pod updates may not change fields other than spec.containers[\*].image, spec.initContainers[\*].image, spec.activeDeadlineSeconds or spec.tolerations. For spec.tolerations, you can only add new entries.

pierDipi:
Closing this, for now, @dprotaso suggested solving this by using callbacks.

PR #2209, opened 2021-08-16
Sink two codegen passes below loop.

mattmoor:
I noticed doing some tinkering that these were running on every iteration, but the inputs don't change.  This sinks the calls to below the loop, and reduces things to a single call for each.

/kind cleanup
/assign @n3wscott 

**Release Note**

```release-note

```

**Docs**

```docs

```


mattmoor:
/retest

OpenCensus flake...

n3wscott (approved):
/lgtm
/approve

PR #2210, opened 2021-08-16
Sketch out an alternate way of injecting clients/informers

mattmoor:
One of the long-standing motivations for the injection-based controller architecture has been to enable us to synthesize different backing implementations, which can be made available to controller constructors by substituting what `foo.Get(ctx)` accesses.

In this change I have implemented codegen that sets up a new `injection.Dynamic` with an alternate implementation of clients and listers all backed by `k8s.io/client-go/dynamic.Interface`.  This is currently not usable via `knative.dev/pkg/injection/sharedmain`, but can be used directly by replacing `injection.Default.SetupInformers` with:

```go
	// Tell it which dynamic client to use (not automatically set up!)
	injCtx := context.WithValue(ctx, dynamicclient.Key{}, /* the dynamic client to use */)
	// Set up the injected context based on the above dynamic client.
	injCtx = injection.Dynamic.SetupDynamic(injCtx)
```

The first commit has the changes to `injection/` and `codegen/` to support this (including the manually crafted `namespacedkube` implementation for secrets).  The second commit has the results of code-generation, and a handful of `*_expansion.go` files needed to fully satisfy `kubernetes.Interface`.

As far as testing this... with what's here, we can generate a build-able `kubernetes.Interface`, I have also been playing with testing this in a downstream tekton repository, and the main gap running e2e testing is some of the event-handling logic, which lives in `client_expansion.go` and is currently NYI (I left all `*_expansion.go` skeletal to try to keep the scope slightly narrower).

/kind feature
/assign @n3wscott


mattmoor:
The eventing failure is real.  For packages like `duckv1` without any types market with `// +genclient` we should skip the `DuckV1()` method.  I'll push an update shortly with a fix.

mattmoor:
I believe the other two issues should be resolved by:
 * https://github.com/knative-sandbox/eventing-rabbitmq/pull/375, and
 * https://github.com/knative-sandbox/eventing-autoscaler-keda/pull/169
 
 Where these repos were using older K8s clients.

mattmoor:
They were not, but it turns out their funniness is a result of the dance we're doing around kubebuilder layouts downstream.  I believe these should now be fixed by:
 * https://github.com/knative-sandbox/eventing-rabbitmq/pull/378, and
 * https://github.com/knative-sandbox/eventing-autoscaler-keda/pull/171

🤞

mattmoor:
Woot, finally clean (failure is the coverage report, which always seems to fail)!

codegen/cmd/injection-gen/generators/filtered_informer.go
```diff
@@ -129,6 +179,20 @@ func withInformer(ctx {{.contextContext|raw}}) ({{.contextContext|raw}}, []{{.co
...
+func withDynamicInformer(ctx {{.contextContext|raw}}) {{.contextContext|raw}} {
+	untyped := ctx.Value({{.factoryLabelKey|raw}}{})
+	if untyped == nil {
+		{{.loggingFromContext|raw}}(ctx).Panic(
+			"Unable to fetch labelkey from context.")
+	}
+	labelSelectors := untyped.([]string)
+	for _, selector := range labelSelectors {
+		inf := &wrapper{client: {{ .clientGet|raw }}(ctx), selector: selector}
+		ctx = context.WithValue(ctx, Key{Selector: selector}, inf)
```
n3wscott, line 191: you will want to register this usage of `context.WithValue` in the `m` map to prevent rando context imports.

codegen/cmd/injection-gen/generators/filtered_informer.go
```diff
@@ -138,4 +202,55 @@ func Get(ctx {{.contextContext|raw}}, selector string) {{.informersTypedInformer
 	}
 	return untyped.({{.informersTypedInformer|raw}})
 }
+
+type wrapper struct {
+	client    {{.clientSetInterface|raw}}
+{{ if .Namespaced }}
+	namespace string
+{{ end }}
+    selector string
```
n3wscott, line 211: nit: this is missing a tab

codegen/cmd/injection-gen/generators/informer.go
```diff
@@ -104,17 +149,23 @@ func (g *injectionGenerator) GenerateType(c *generator.Context, t *types.Type, w
...
 func withInformer(ctx {{.contextContext|raw}}) ({{.contextContext|raw}}, {{.controllerInformer|raw}}) {
 	f := {{.factoryGet|raw}}(ctx)
-	inf := f.{{.group}}().{{.version}}().{{.type|publicPlural}}()
+	inf := f.{{.groupGoName}}().{{.versionGoName}}().{{.type|publicPlural}}()
 	return context.WithValue(ctx, Key{}, inf), inf.Informer()
 }
 
+func withDynamicInformer(ctx {{.contextContext|raw}}) {{.contextContext|raw}} {
+	inf := &wrapper{client: {{ .clientGet|raw }}(ctx)}
+	return context.WithValue(ctx, Key{}, inf)
```
n3wscott, line 166: here again, `context.WithValue` should come from a registered type

codegen/cmd/injection-gen/generators/informer.go
```diff
@@ -124,4 +175,49 @@ func Get(ctx {{.contextContext|raw}}) {{.informersTypedInformer|raw}} {
 	}
 	return untyped.({{.informersTypedInformer|raw}})
 }
+
+type wrapper struct {
+	client    {{.clientSetInterface|raw}}
```
n3wscott, line 180:
```suggestion
	client    {{ .clientSetInterface|raw }}
```

n3wscott: nit, half the file does that way and half does not add spaces, so idk, I don't care too much.

n3wscott (commented):
I like the direction, a couple nits.

/approve

codegen/cmd/injection-gen/generators/filtered_informer.go
mattmoor: Done

codegen/cmd/injection-gen/generators/filtered_informer.go
mattmoor: Done

codegen/cmd/injection-gen/generators/informer.go
mattmoor: Done

codegen/cmd/injection-gen/generators/informer.go
mattmoor: Yeah, the squished way was making my eyes bleed.  Perhaps we should follow up with a formatting pass?

mattmoor (commented):
I incorporated the feedback as a separate commit (in the middle) and there were no changes when codegen ran. 👍

n3wscott:
I think this is real interesting and could be used to solve the heavy mix of duck type and strongly typed clients in reconcilers, and our fun and hard to find cache mixing bugs. 

/lgtm

PR #2211, opened 2021-08-17
Drop redundant pointers and decoders

markusthoemmes:
# Changes

- :broom: Drop unnecessary double-pointers in decoding logic.
- :broom: Replace `decoder.Decode` with `json.Unmarshal` where they are equivalent anyway. This also avoids unnecessary byte buffers.

/kind cleanup

**Release Note**

```release-note
NONE
```

**Docs**

```docs

```

/assign @julz


julz (approved):
/lgtm
/approve

PR #2212, opened 2021-08-17
Deprecate ReadOnlyFinalizer interface over OnDeletionInterface

markusthoemmes:
# Changes

We've determined `ReadOnlyFinalizer` to be unsafe a while ago and hence introduced the (supposed to be safe) alternative `OnDeletionInterface`. We should nudge people to use the latter and completely fade out the former.

Note: Neither knative nor sandbox nor Tekton is using ReadOnlyFinalizer anymore, so we might also think about "just" dropping it.

/kind deprecation

**Release Note**

```release-note
Deprecate ReadOnlyFinalizer interface over OnDeletionInterface.
```

/assign @julz @n3wscott

### DIFF
--- a/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
+++ b/codegen/cmd/injection-gen/generators/reconciler_reconciler.go
@@ -268,11 +268,13 @@ type ReadOnlyInterface interface {
 // controller finalizing {{.type|raw}} if they want to process tombstoned resources
 // even when they are not the leader.  Due to the nature of how finalizers are handled
 // there are no guarantees that this will be called.
-// DEPRECATED: Use reconciler.OnDeletionInterface instead.
+//
+// Deprecated: Use reconciler.OnDeletionInterface instead.
 type ReadOnlyFinalizer interface {
 	// ObserveFinalizeKind implements custom logic to observe the final state of {{.type|raw}}.
 	// This method should not write to the API.
-	// DEPRECATED: Use reconciler.ObserveDeletion instead.
+	//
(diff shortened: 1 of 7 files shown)


codegen/cmd/injection-gen/generators/reconciler_reconciler.go
```diff
@@ -268,9 +268,11 @@ type ReadOnlyInterface interface {
 // controller finalizing {{.type|raw}} if they want to process tombstoned resources
 // even when they are not the leader.  Due to the nature of how finalizers are handled
 // there are no guarantees that this will be called.
+// DEPRECATED: Use reconciler.OnDeletionInterface instead.
```
julz, line 271: These should be `Deprecated:` (the canonical form is case-sensitive)

markusthoemmes:
Ack on the deprecated bit. How do we feel about straight up dropping it? I think there's no real safe usage for it (and I've verified us and Tekton are not using it anywhere either anymore). Feels safe enough?

julz:
I can certainly see the argument that given all uses of it are wrong it's better to not compile than encourage something that won't work. Feels bad to potentially break people on a non-major-version update obviously, though. To state the obvious: the advantage of landing the deprecation comment first and letting it sit for a release is at least that way people get a warning with a clear hint about what to do instead, though you could argue the release notes can do that anyway. I'm ok to go with whatever you decide on this one.

markusthoemmes:
/retest

Nooooot related.

julz (approved):
/lgtm

PR #1935, opened 2020-12-02
Remove the unused file

vagababov:
Everyone is actually using the aliased option

/assign @tcnghia mattmoor

PR #2213, opened 2021-08-17
Revert "Remove the unused file (#1935)"

markusthoemmes:
This reverts commit e45bbefd1d63e0bc82c88984cab408441378da07.

The commit was done inadvertently, we actually wanted to switch to the alias that got deleted. So let's do that now!

/assign @julz

### DIFF
--- a/network/error_handler.go
+++ b/network/error_handler.go
@@ -25,8 +25,8 @@ import (
 
 // ErrorHandler sets up a handler suitable for use with the ErrorHandler field on
 // httputil's reverse proxy.
-// TODO(mattmoor): Move the implementation into handlers/error.go once downstream consumers
-// have adopted the alias.
+//
+// Deprecated: Use handler.Error instead.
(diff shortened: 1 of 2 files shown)


markusthoemmes:
/retest

OpenCensus stuff.

julz:
worth adding an explicit "Deprecated:" comment on the old method to encourage people to move to the right one this time?

markusthoemmes:
Don't think so, as that'll break the existing usages (in Knative anyway). I planned to send the fixing PR and then remove this right after the bump.

julz:
Hm, don't totally understand how a deprecation comment would break people but removing the method won't? (I'm misunderstanding what's going on here aren't I? 😊)

julz:
(I'm totally fine just lgtm-ing as-is if we're planning to remove the method once we've bumped serving btw, just making sure I understand the plan before stamping!)

julz (approved):
/lgtm

PR #2216, opened 2021-08-17
Use consistent case for "Deprecated" comments

julz:
Not the most important thing ever (to say the least), but the canonical prefix to use for `Deprecated:` warnings in godoc is title case (and should be in its own paragraph).

# Changes

- :broom: make deprecation comments use consistent case.

/kind cleanup

### DIFF
--- a/apis/interfaces.go
+++ b/apis/interfaces.go
@@ -53,7 +53,8 @@ type Listable interface {
 }
 
 // Annotatable indicates that a particular type applies various annotations.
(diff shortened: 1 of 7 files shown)


julz:
Surely not related

/retest

julz:
oh fun, that caused the linter to notice an actual use of a deprecated method (in a test).

/hold

n3wscott:
`Deprecated:` is the only key word that go uses, the capital version is incorrect 😄 

/lgtm
/approve

PR #2217, opened 2021-08-18
Add length limit for the generated names used in tests

yanweiguo:
`ObjectNameForTest` is used to generate names for objects (like service) used in tests. It calls `kmeta.ChildName` which allows max name length is 63 (supported by k8s). However some non-k8s implementation has a smaller limit than k8s.

This PR proposes to limit the length of generated name to 50, which is sufficient in knative tests.

### DIFF
--- a/test/helpers/name.go
+++ b/test/helpers/name.go
@@ -26,11 +26,12 @@ import (
 )
 
 const (
-	letterBytes    = "abcdefghijklmnopqrstuvwxyz"
-	randSuffixLen  = 8
-	sep            = '-'
-	sepS           = "-"
-	testNamePrefix = "Test"
+	letterBytes     = "abcdefghijklmnopqrstuvwxyz"
+	randSuffixLen   = 8
+	nameLengthLimit = 50
+	sep             = '-'
+	sepS            = "-"
+	testNamePrefix  = "Test"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


markusthoemmes:
To hook into this, since we have a similar problem down at Openshift: What exactly is this fixing? Is the limitation of your system (I assume Cloud Run?) imposed on the names themselves or are you using a different domainTemplate (i.e. $name-$ns) and this fixes that?

yanweiguo:
> To hook into this, since we have a similar problem down at Openshift: What exactly is this fixing? Is the limitation of your system (I assume Cloud Run?) imposed on the names themselves or are you using a different domainTemplate (i.e. $name-$ns) and this fixes that?

The max length of Cloud Run supports is less than 63.

test/helpers/name.go
```diff
@@ -52,7 +53,14 @@ func ObjectPrefixForTest(t named) string {
...
 func ObjectNameForTest(t named) string {
-	return kmeta.ChildName(ObjectPrefixForTest(t), string(sep)+RandomString())
+	prefix := ObjectPrefixForTest(t)
+	suffix := string(sep) + RandomString()
+	limit := nameLengthLimit - len(suffix)
+	if len(prefix) < limit {
+		limit = len(prefix)
+	}
+
+	return kmeta.ChildName(prefix[:limit], suffix)
```
julz, line 63: does ChildName actually do anything here? IIUC ChildName truncates the name iff the prefix plus the suffix are bigger than 63 characters, but in this case we've already limited the prefix to `50 - the length of the suffix`, so - I think - the input to this method is always less than 50 characters

yanweiguo: Good caught. Removed.

yanweiguo:
@markusthoemmes @julz ping

PR #2220, opened 2021-08-20
Have the addressable resolver take a tracker.

mattmoor:
I noticed in a few places downstream that reconcilers were creating both trackers and Addressable resolvers, which seems superfluous.  As part of examining the way we use the tracker, I'm experimenting with changing this to just take a tracker.

/kind cleanup

**Release Note**

```release-note
resolver.NewURIResolver now takes a tracker.Interface instead of a callback.
```

**Docs**

```docs

```


mattmoor:
PSA: I'm opening this to survey what breaks, I don't plan to check this in as-is.

PR #2222, opened 2021-08-21
Introduce `NewContext`, deprecate `NewImplFull`.

mattmoor:
Our generated `NewImpl` methods have long taken `context.Context`, but despite many iterations the forms we expose from our `controller` package never have.  This change contains several elements:
1. Expose a new `NewContext` method that takes `context.Context` in addition to the current `NewImplFull` signature.
2. Call `NewContext` instead of the deprecated `NewImpl` from our generated controller code.
3. Call `NewContext` from all our webhook reconcilers.
4. Introduce a `tracker.Interface` on `controller.Impl` now that we can extract the tracker lease from `ctx` (for use with https://github.com/knative/eventing/pull/5651, and https://github.com/knative/pkg/pull/2220 downstream)

/kind cleanup

**Release Note**

```release-note
Introduce controller.NewContext and deprecate controller.NewImplFull
```

**Docs**

```docs

```

/assign @n3wscott

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/controller.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/controller.go
@@ -50,7 +50,7 @@ const (
 // NewImpl returns a controller.Impl that handles queuing and feeding work from
 // the queue through an implementation of controller.Reconciler, delegating to
 // the provided Interface and optional Finalizer methods. OptionsFn is used to return
-// controller.Options to be used by the internal reconciler.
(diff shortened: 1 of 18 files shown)


mattmoor:
The code style failure is a timeout, the serving downstream failure is a flake, and codecov/project is just always red.  If nothing else fails, then this should be RFAL 👀

mattmoor:
/retest

MetricsExporter flake...

mattmoor:
Same test 🙃 
/retest

n3wscott (approved):
/lgtm
/approve

PR #2224, opened 2021-08-21
This implements the EventExpansion needed to emit K8s events.

mattmoor:
With this change, I was able to pass some downstream e2e tests that check for event emission with this path enabled, which previously failed.

The relevant hand-rolled bits in Kubernetes client-go is here: https://github.com/kubernetes/client-go/blob/35bf219cc60981cdf318b711e2630cc05e0a83e5/kubernetes/typed/core/v1/event_expansion.go#L49

I opted to avoid copy/pasting the generated code, and instead used a trick that let me call into the generated code.

Going through this exercise also (likely) uncovered a Kubernetes bug: https://github.com/kubernetes/kubernetes/issues/104495

/kind enhancement
/assign @n3wscott 

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/client/injection/kube/client/client_expansion.go
+++ b/client/injection/kube/client/client_expansion.go
@@ -18,6 +18,7 @@ package client
 
 import (
 	context "context"
+	fmt "fmt"
 
 	certificatesv1beta1 "k8s.io/api/certificates/v1beta1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


mattmoor:
/retest

MetricsExporter...

vagababov (approved):
/lgtm

PR #2226, opened 2021-08-22
Make the configmap update logging Debugf

mattmoor:
/kind cleanup

**Release Note**

```release-note
Logging when configmap updates occur is now at the "debug" log level.
```

**Docs**

```docs

```


mattmoor:
I'm ignoring both failures.  golangci-lint keeps timing out, and the codecov/project thing just always fails.

PR #2227, opened 2021-08-23
Update prometheus dep to 0.30.0

markusthoemmes:
This update was blocked for a while as the dependency chain hadn't picked up a breaking change. It has now though, so yay!

/assign @evankanderson


evankanderson (commented):
/lgtm
/approve

Issue #2229, opened 2021-08-24
When using config `registriesSkippingTagResolving`, it has problem with space, I think it can be fixed.


jwcesign:
when I edit configmap of config-deployment with `kubectl edit configmap config-deployment -nknative-serving`, the content with space, as follows:
~~~
registriesSkippingTagResolving: kind.local,ko.local,dev.local, gcr.io
~~~
for the img with `gcr.io`, it doesn't work, because there is a space with `gcr.io`

so maybe this can be fixed as follows, to delete the space of head and tail.
~~~
diff --git a/vendor/knative.dev/pkg/configmap/parse.go b/vendor/knative.dev/pkg/configmap/parse.go
index 4c280757b..03ee9cb47 100644
--- a/vendor/knative.dev/pkg/configmap/parse.go
+++ b/vendor/knative.dev/pkg/configmap/parse.go
@@ -141,7 +141,11 @@ func AsDuration(key string, target *time.Duration) ParseFunc {
 func AsStringSet(key string, target *sets.String) ParseFunc {
        return func(data map[string]string) error {
                if raw, ok := data[key]; ok {
-                       *target = sets.NewString(strings.Split(raw, ",")...)
+                       rawCopy := strings.Split(raw, ",")
+                       for i, v := range rawCopy {
+                               rawCopy[i] = strings.TrimSpace(v)
+                       }
+                       *target = sets.NewString(rawCopy...)
                }
                return nil
        }
~~~

jwcesign:
BTW, I think it hasn't got side effects for other config parse, because it just delete the space of head and tail.

julz:
Hi @jwcesign - seems reasonable to me, did you want to create a PR for this or is this supposed to be a `/good-first-issue` for someone else to PR?

jwcesign:
@julz Very pleasure to create a PR for me.

markusthoemmes:
SGTM, please fire away!

PR #2230, opened 2021-08-24
Trim the space in configmap/registriesSkippingTagResolving's url. (#2…

jwcesign:
# Changes

🎁 Trim the space in configmap/registriesSkippingTagResolving's url. Because many develpers do like "a.com, b.com", it's easy to have this mistake.

Related to https://github.com/knative/pkg/issues/2229
Related PR

### DIFF
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -141,7 +141,11 @@ func AsDuration(key string, target *time.Duration) ParseFunc {
 func AsStringSet(key string, target *sets.String) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
-			*target = sets.NewString(strings.Split(raw, ",")...)
+			splitted := strings.Split(raw, ",")
+			for i, v := range splitted {
+				splitted[i] = strings.TrimSpace(v)
+			}
+			*target = sets.NewString(splitted...)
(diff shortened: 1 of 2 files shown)


jwcesign:
@googlebot I signed it!

markusthoemmes:
Thanks! The implementation looks good to me! :partying_face: Have you taken a look at adding a respective unit-test as well? It'd be nice to add a case that removes the spaces.

jwcesign:
@markusthoemmes  I did test with ko in real environment.

jwcesign:
@markusthoemmes I added test code

configmap/parse.go
```diff
@@ -141,7 +141,11 @@ func AsDuration(key string, target *time.Duration) ParseFunc {
 func AsStringSet(key string, target *sets.String) ParseFunc {
 	return func(data map[string]string) error {
 		if raw, ok := data[key]; ok {
-			*target = sets.NewString(strings.Split(raw, ",")...)
+			urlArr := strings.Split(raw, ",")
```
markusthoemmes, line 144: nit: `urlArr` isn't quiet accurate here. The variable doesn't contain URLs per se. Personally, I would've chosen `splitted` in this case.

PR #2243, opened 2021-08-29
add reason for happy condition when use MarkTrueWithReason

yuzp1996:
# Changes

- :gift:  add reason for happy condition when use method MarkTrueWithReason

/kind  enhancement

Fixes #

When user use MarkTrueWithReason to set their condition the happy condition can not be set with reason while I think this is not a good choice. So I hope reason can be  also be set for happy condiition if I use the method MarkTrueWithReason.

**Release Note**

```release-note

```

**Docs**

```docs

```


yuzp1996:
@googlebot I signed it!

yuzp1996:
@googlebot I fixed it.

PR #2244, opened 2021-08-29
add reason for happy condition when use method MarkTrueWithReason

yuzp1996:
# Changes

- :gift:  add reason for happy condition when use method MarkTrueWithReason

/kind  enhancement

Fixes #

When user use MarkTrueWithReason to set their condition the happy condition can not be set with reason while I think this is not a good choice. So I hope reason can be  also be set for happy condiition if I use the method MarkTrueWithReason.

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/apis/condition_set.go
+++ b/apis/condition_set.go
@@ -262,7 +262,7 @@ func (r conditionsImpl) MarkTrue(t ConditionType) {
 		Status:   corev1.ConditionTrue,
 		Severity: r.severity(t),
 	})
-	r.recomputeHappiness(t)
+	r.recomputeHappiness(t, "")
 }
 
 // MarkTrueWithReason sets the status of t to true with the reason, and then marks the happy condition to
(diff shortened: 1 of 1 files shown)


apis/condition_set.go
```diff
@@ -295,6 +295,7 @@ func (r conditionsImpl) recomputeHappiness(t ConditionType) {
 		r.SetCondition(Condition{
 			Type:     r.happy,
 			Status:   corev1.ConditionTrue,
+			Reason: reason,
```
n3wscott, line 298: normally the happy condition does not have a reason. The reason it is happy is because _everything_ else is happy. It is not a lot of signal to propagate the last flop of ready=true from a dependent to happy.

yuzp1996: OK!   Is it means that only when happy condition's status is false, there will need a reason?  Otherwise, no reason will need?

yuzp1996: Is there any document or contract about how to use conditions like when to set reason or not.  Is it just convention?

Thank you very much !

dprotaso: > OK! Is it means that only when happy condition's status is false, there will need a reason? Otherwise, no reason will need?

Yeah - to Scott's point there really isn't an need to propagate a reason why the `Ready` Condition is happy - it should be inferred from the dependent child conditions

> Is there any document or contract about how to use conditions like when to set reason or not? Is it just convention?

Usually the Type is descriptive enough so you wound't need an additional reason string when the condition is `True`. 

`Reason` & `Message` are for situations when a condition cannot be satisfied - ie. Status=`False` and we want to convey additional information

dprotaso:
As per the comment above it isn't necessary to include a `Reason` when the conditions are `Happy`

Going to close this out because we want to continue that convention

yuzp1996:
Thanks for all your reply very much!

PR #2246, opened 2021-08-30
Added invalid value field error with details

senthilnathan:
# Changes

- Added new error type (Invalid Value Error with Details)
- Needed for https://github.com/knative/serving/pull/11883

/kind enhancement

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -323,22 +323,13 @@ func ErrInvalidArrayValue(value interface{}, field string, index int) *FieldErro
 	return ErrInvalidValue(value, CurrentField).ViaFieldIndex(field, index)
 }
 
-// ErrInvalidValue constructs a FieldError for a field that has received an
-// invalid value.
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -332,6 +332,16 @@ func ErrInvalidValue(value interface{}, fieldPath string) *FieldError {
 	}
 }
 
+// ErrInvalidValueWithDetails constructs a FieldError for a field that has received an
+// invalid value including Details.
+func ErrInvalidValueWithDetails(value interface{}, fieldPath string, details string) *FieldError {
```
dprotaso, line 337: I think it would be simpler to modify `ErrInvalidValue` to take an optional variadic input for the details strings - ie. see `ErrInvalidKeyName`

https://github.com/knative/pkg/blob/dabb207d45983cfd03bd10985639b42c0b3db404/apis/field_error.go#L374-L380

PR #2248, opened 2021-08-31
Delete deprecated ObserveFinalizeKind API

markusthoemmes:
# Changes

Shaking the :evergreen_tree:, here's hoping nobody actually implements it and if they do, imma make them not do it :joy:

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
@@ -73,20 +73,6 @@ type ReadOnlyInterface interface {
 	ObserveKind(ctx context.Context, o *v1.CustomResourceDefinition) reconciler.Event
 }
 
-// ReadOnlyFinalizer defines the strongly typed interfaces to be implemented by a
-// controller finalizing v1.CustomResourceDefinition if they want to process tombstoned resources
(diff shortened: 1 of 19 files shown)


lionelvillard:
/hold

are you sure? This is breaking a lot of downstream repos

dprotaso:
blah sorry - yeah I missed that.

@markusthoemmes we should leave the older interfaces around until the changes that update the generated code proliferate a bit

markusthoemmes:
This actually looks like the downstream test action is wrong :thinking:. After an update-codegen, everything should be fine.

markusthoemmes:
Ahhhh very good point @dprotaso, thanks! That actually explains all the breaks, I think.

markusthoemmes:
@dprotaso @lionelvillard kept the opposing constant in for now. Once pkg updates have rolled through, we can delete that as well.

antoineco:
Uh-oh, I [do use that](https://github.com/triggermesh/aws-event-sources/blob/13d3cd242c94c32ff7539c541fa83ed5cd0e9416/pkg/adapter/awssnssource/reconciler.go#L85) 😊 

I believe a sound alternative is to use an event handler on Delete, is that what you folks do?

markusthoemmes:
@antoineco you'll want to switch over to the solution implemented via https://github.com/knative/pkg/issues/2021

PR #2249, opened 2021-09-01
allow unknown metadata fields

dprotaso:
Fixes:

### DIFF
--- a/webhook/json/decode.go
+++ b/webhook/json/decode.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package json
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+)
+
+var (
+	emptyMeta  = []byte(`:{}`)
+	metaPrefix = []byte(`{"metadata"`)
+	metaSuffix = []byte(`}`)
+)
+
+var (
+	// Unmarshal is an alias for json.Unmarshal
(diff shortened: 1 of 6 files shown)


dprotaso:
/assign @julz @markusthoemmes @shinigambit 
/hold 

I'm going to test downstream

webhook/json/decode.go
```diff
@@ -0,0 +1,131 @@
...
+			if v == '{' {
+				level += 1
+			} else if v == '}' {
+				level -= 1
+			}
+		case string:
+			if v == "metadata" && level == 1 {
+				start = dec.InputOffset()
+				x := struct{}{}
+				if err = dec.Decode(&x); err != nil {
```
markusthoemmes, line 120: Is this guaranteed to move the decoder to the end of the entire metadata object? Even if there's sub objects like `annotations` for example? Is that worth an extra test?

markusthoemmes (commented):
I love this :partying_face: . Had something similar in mind but didn't dare doing the parsing shenanigans. Great stuff!

webhook/json/decode.go
dprotaso: > Is this guaranteed to move the decoder to the end of the entire metadata object?

Yeah - it's decoding the metadata value into the empty struct.

> Even if there's sub objects like annotations for example? Is that worth an extra test?

It's easy enough to add so we can catch regressions if someone changes the implementation

dprotaso:
> I love this 🥳 . Had something similar in mind but didn't dare doing the parsing shenanigans. Great stuff!

The biggest gain over the naive implementation is being able to bail out early after we have the metadata byte offsets.

webhook/json/decode.go
```diff
@@ -0,0 +1,131 @@
...
+	level := 0
+
+	var (
+		dec = json.NewDecoder(bytes.NewReader(bites))
+		t   json.Token
+	)
+
+	for {
+		t, err = dec.Token()
+		if err == io.EOF { //nolint
```
markusthoemmes, line 102: can we not do the `errors.Is` thing? (I guess this is asking for that?)

dprotaso: You could but it's not necessary - https://pkg.go.dev/encoding/json#Decoder.Token states it will return `io.EOF`

markusthoemmes (approved):
/lgtm
/approve

dprotaso:
@chizhg do we have the latest cherry pick plugin? It should support multiple invocations

https://github.com/kubernetes/test-infra/issues/21714

PR #2250, opened 2021-09-01
fix check calls being unreliable

shinigambit:
Add back `DefaultErrorRetryChecker` into the `CheckEndpointState` calls so it retries for issues caused by the network.

Problem introduced in knative/serving/pull/11680

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -313,8 +313,8 @@ func (sc *SpoofingClient) Check(req *http.Request, inState ResponseChecker, erro
 					return false, nil
 				}
 			}
+			sc.Logf("NOT Retrying %s: %v", req.URL.String(), err)
 		}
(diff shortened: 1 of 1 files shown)


markusthoemmes (approved):
/lgtm
/approve

test/spoof/spoof.go
```diff
@@ -295,12 +295,30 @@ func (sc *SpoofingClient) endpointState(
...
+		if err != nil {
+			for _, checker := range errorRetryCheckers {
+				retry, newErr := checker(err)
+				if retry {
+					sc.Logf("Retrying %s: %v", req.URL.String(), newErr)
+					return false, nil
+				}
+			}
+		}
 		sc.Logf("NOT Retrying %s: %v", req.URL.String(), err)
```
markusthoemmes, line 317: This log line should move into the `if err != nil` block I believe.

markusthoemmes (approved):
/lgtm
/approve

PR #2251, opened 2021-09-01
Source validation fixes

gabo1208:
# Changes

- :broom: Added general validation to knative sources.spec.ceOverrides.extensions in general

related to eventing https://github.com/knative/eventing/issues/5638


apis/duck/v1/source_types_test.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2019 The Knative Authors
+
```
lionelvillard, line 3: can you replace by 2021

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,50 @@ type SourceList struct {
...
+	if len(key) < 1 {
+		return apis.ErrInvalidKeyName(
+			key,
+			"",
+			"CloudEvents attribute names MUST NOT be empty",
+		)
+	}
+
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
```
lionelvillard, line 210: only lower-case letters or digits are allowed.

gabo1208: Yup, should i comment that? this is the exact condition that the cloudevents go sdk uses

lionelvillard: really? hmm maybe a bug in the SDK?

gabo1208: Here is the go-sdk line: https://github.com/cloudevents/sdk-go/blob/main/v2/event/extensions.go#L51
Im waiting for a response from them to see if its a bug or a feature :p

apis/duck/v1/source_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors
+Copyright 2021 The Knative Authors
```
n3wscott, line 2: don't change the year of the file

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,50 @@ type SourceList struct {
 
 	Items []Source `json:"items"`
 }
+
+func (s *Source) Validate(ctx context.Context) *apis.FieldError {
+	if s == nil {
+		return nil
+	}
+	return ValidateSource(ctx, *s).ViaField(apis.CurrentField)
```
n3wscott, line 179: `ValidateSource` is a useless method, just call `src.Spec.Validate(ctx).ViaField("spec")`directly.

apis/duck/v1/source_types.go
n3wscott: the go sdk allows users to set mixed case and it fixes it for them in the same way the http header lib does.

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,50 @@ type SourceList struct {
...
+			"CloudEvents attribute names MUST NOT be empty",
+		)
+	}
+
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+			return apis.ErrInvalidKeyName(
+				key,
+				"",
+				"CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set",
```
n3wscott, line 214: it is not clear that this is the message knative would like to show users. maybe something more like : `keys are expected to be lower-case letters ('a' to 'z') or digits ('0' to '9')` The fact that this is a cloudevents rule does not matter to knative so much. It will confuse people.

apis/duck/v1/source_types.go
lionelvillard: what happens when there is a name collision, for example `myExt` vs `myext`?

lionelvillard:
/lgtm
Thanks!

apis/duck/v1/source_types_test.go
```diff
@@ -0,0 +1,91 @@
...
+			"keys MUST NOT be empty",
+		),
+	}, {
+		name: "extension name too long",
+		src: &Source{Spec: SourceSpec{
+			CloudEventOverrides: &CloudEventOverrides{
+				Extensions: map[string]string{"nameLongerThan20Characters": "test"},
+			},
+		}},
+		want: nil,
```
antoineco, line 58: The 20-chars limit is just a recommendation. It is valid to have extensions that are longer than 20 characters.

apis/duck/v1/source_types_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}, {
+		name: "invalid extension name ",
```
antoineco, line 60: Trailing space.

apis/duck/v1/source_types_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}} {
+		t.Run(tt.name, func(t *testing.T) {
+			tt := tt
+			t.Parallel()
+			got := tt.src.Validate(context.TODO())
+			if tt.want != nil && got.Error() != tt.want.Error() {
+				t.Errorf("Unexpected error want:\n%+s\ngot:\n%+s", tt.want, got)
+			}
+
+			if tt.want == nil && tt.want != got {
```
antoineco, line 86: I feel that `tt.want == nil && got != nil` would read more logically here.

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,45 @@ type SourceList struct {
...
+			"keys MUST NOT be empty",
+		)
+	}
+
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+			return apis.ErrInvalidKeyName(
+				key,
+				"",
+				"keys MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set",
```
antoineco, line 209: Your condition does accept upper-case letters (which is valid according to the spec), so the error message should be updated.

antoineco (changes requested):

apis/duck/v1/source_types.go
antoineco: @lionelvillard the validation in the Go SDK happens [inside `SetExtension()`](https://github.com/cloudevents/sdk-go/blob/82f2b61ecde41fd0577969cc58a7c2a18eeda250/v2/event/eventcontext_v1.go#L76-L81). At this point, `myExt` and `myext` can coexist like Scott mentioned.

When you call `Send()` via a CloudEvent client that uses the HTTP protocol, `SetExtension()` [overrides any existing header](https://github.com/cloudevents/sdk-go/blob/82f2b61ecde41fd0577969cc58a7c2a18eeda250/v2/protocol/http/write_request.go#L136) during the conversion. And because extensions are stored as a map, the header's value will be random.

apis/duck/v1/source_types_test.go
gabo1208: Yup this test is here just to ensure it passes, and maybe in the future to verify we return a warning at least :)

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,45 @@ type SourceList struct {
...
+	for key := range ceOverrides.Extensions {
+		if err := validateExtensionName(key); err != nil {
+			return err.ViaField("extensions")
+		}
+	}
+	return nil
+}
+
+func validateExtensionName(key string) *apis.FieldError {
+	if len(key) < 1 {
```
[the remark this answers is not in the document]
gabo1208: I'm using the same expresion as the cloudeventes library, but ok ok. Makes more sense to me this way also.

apis/duck/v1/source_types_test.go
antoineco: Sounds reasonable!
Would you then consider renaming the test to `long extension name is valid` or something like that?

gabo1208:
Fixes #2273

apis/duck/v1/source_types_test.go
gabo1208: Sure 👍🏼

antoineco (approved):

antoineco (approved):

apis/duck/ABOUT.md
```diff
@@ -302,7 +302,11 @@ is used with duck types:
                 ...
         }
         logger := c.Logger.Named("Revisions")
-        impl := controller.NewImpl(c, logger, "Revisions")
+        impl := controller.NewContext(
+          ctx,
+          c,
+          controller.ControllerOptions{WorkQueueName: "Revisions", Logger: logger},
```
n3wscott, line 308: is this change valid? Seems like a merge conflict.

apis/duck/v1/source_types.go
n3wscott: knative can choose to be more strict

apis/duck/v1/source_types.go
```diff
@@ -170,3 +171,45 @@ type SourceList struct {
...
+			"keys MUST NOT be empty",
+		)
+	}
+
+	for _, c := range key {
+		if !((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')) {
+			return apis.ErrInvalidKeyName(
+				key,
+				"",
+				"keys MUST consist of lower and upper-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set",
```
n3wscott, line 209: this error still feels out of place for the validation webhook, this is the spec language, but we don't need to do that in the webhook, I would say this: 
```suggestion
				"keys are expected to be alphanumeric",
```

apis/duck/ABOUT.md
gabo1208: Yup now the new version of the pkg have this format on the imp initialization

And for the comment below, for now we could stick to the same cloudevents validations

n3wscott (approved):

PR #2252, opened 2021-09-02
Avoid double-resyncs without leader election.

mattmoor:
tl;dr Without leader election enabled, we've been suffering from double resyncs, this fixes that.

Essentially when the informers are started, prior to starting the controllers, they fill the controller's workqueue with ~the world.  Once we start the controllers, the leader election code calls `Promote` on a `Bucket` so that it starts processing them as their leader, and then is requeues every key that falls into that `Bucket`.

When leader election is disabled, we use the `UniversalBucket`, which always owns everything.  This means that the first pass through the workqueue was already hitting every key, so re-enqueuing every key results in superfluous processing of those keys.

This change makes the `LeaderAwareFuncs` elide the call to `PromoteFunc` when the `enq` function passed is `nil`, and makes the `unopposedElector` have a `nil` `enq` field when we explicitly pass it the `UniversalBucket` because leader election is disabled.

/kind cleanup

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -941,15 +941,15 @@ func (cr *countingLeaderAwareReconciler) Reconcile(ctx context.Context, key stri
 }
 
 func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
-	promoted := make(chan struct{})
 	r := &countingLeaderAwareReconciler{
 		LeaderAwareFuncs: reconciler.LeaderAwareFuncs{
 			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) error {
-				close(promoted)
+				t.Error("Promote should not be called when no leader election is enabled.")
 				return nil
 			},
 		},
 	}
+
 	impl := NewContext(context.TODO(), r, ControllerOptions{
 		Logger:        TestLogger(t),
 		WorkQueueName: "Testing",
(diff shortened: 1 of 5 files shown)


mattmoor:
`TestMetricsExport/OpenCensus` in the actions leg that failed

dprotaso:
Can we add some unit tests for the `leaderelection` and `reconciler` package changes?

mattmoor:
Done, RFAL.

controller/controller_test.go
```diff
@@ -968,13 +968,10 @@ func TestStartAndShutdownWithLeaderAwareNoElection(t *testing.T) {
...
 
 	select {
-	case <-promoted:
-		// We expect to be promoted immediately, since there is no
-		// ElectorBuilder attached to the context.
 	case <-doneCh:
 		t.Fatal("StartAll finished early.")
 	case <-time.After(10 * time.Second):
-		t.Error("Timed out waiting for StartAll.")
+		// Give it some time to run.
```
n3wscott, line 974: the loss of the error message might be a problem? This will fall through to the cancel at line 977 and you might not understand if or why it timed out.

n3wscott (commented):
one nit/question, otherwise LGTM

controller/controller_test.go
mattmoor: This is now the normal path for the test.  Previously `PromoteFunc` was called, which triggered another path.  Now this is waiting to effectively make sure it isn't called.

PR #2253, opened 2021-09-02
Widen the error checker interface to allow response error checking

markusthoemmes:
# Changes

This'll allow us to remove `RetryingRouteInconsistency` in Serving to push "patcheable" functionality deeper into the stack. Shaking the :evergreen_tree:  to see what this'd break.

### DIFF
--- a/test/spoof/spoof.go
+++ b/test/spoof/spoof.go
@@ -68,6 +68,14 @@ type ResponseChecker func(resp *Response) (done bool, err error)
 // If an error should be retried, it should return true and the wrapped error to explain why to retry.
 type ErrorRetryChecker func(e error) (retry bool, err error)
 
+// ResponseRetryChecker is used to determine if a response should be retried or not.
+// If a response should be retried, it should return true and an error to explain why to retry.
+//
+// This is distinct from ResponseChecker in that it shall be used to retry responses,
+// where the HTTP request was technically successful (it returned something) but indicates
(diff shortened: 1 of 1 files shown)


test/spoof/spoof.go
```diff
@@ -66,7 +66,7 @@ type ResponseChecker func(resp *Response) (done bool, err error)
 
 // ErrorRetryChecker is used to determine if an error should be retried or not.
 // If an error should be retried, it should return true and the wrapped error to explain why to retry.
-type ErrorRetryChecker func(e error) (retry bool, err error)
+type ErrorRetryChecker func(resp *Response, e error) (retry bool, err error)
```
dprotaso, line 69: Maybe introduce a new type `ResponseRetryChecker` ? And have Do and Poll accept both `ErrorRetryChecker` and `ResponseRetryChecker` ?

markusthoemmes: We could, though tbh. I dunno how valuable that'd be as we'd have to do type casting funkiness and whatnot to tell the two apart. Do you feel like that's going to be clearer?

markusthoemmes:
/retest

Unrelated.

markusthoemmes:
Doing https://github.com/knative/pkg/pull/2264 first which will cut down on the surface area of this significantly.

nak3:
LGTM!

PR #2264, opened 2021-09-02
Deduplicate Check and Poll functions on the spoof client

markusthoemmes:
# Changes

We want the same retrying behavior for both on the base requests, but want to check `inState` only once. So we `Do` (which polls with an always-passing check) and then check what we receive again if it matches what we actually want. This greatly reduces code duplication.

/assign @shinigambit


shinigambit:
This is great!
/lgtm

PR #2265, opened 2021-09-02
when testing downstream use the correct release branch

dprotaso:
additionally, we re-merge upstream in case new changes appear in HEAD

Fixes: https://github.com/knative/pkg/pull/2265

### DIFF
--- a/.github/workflows/knative-downstream.yaml
+++ b/.github/workflows/knative-downstream.yaml
@@ -80,18 +80,33 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1.16.x
+
     - name: Install Dependencies
       run: |
         go get github.com/google/go-licenses
-    - name: Checkout Upstream
+
+    - name: Check out code
       uses: actions/checkout@v2
       with:
+        fetch-depth: 0
         path: ./src/knative.dev/${{ github.event.repository.name }}
+
+    - name: Merge upstream
+      if: github.event_name == 'pull_request'
+      shell: bash
+      working-directory: ./src/knative.dev/${{ github.event.repository.name }}
+      run: |
+         git remote add upstream https://github.com/${{ github.repository }}.git
+         git fetch upstream ${{ github.base_ref }}
+         git pull --no-rebase --no-commit upstream ${{ github.base_ref }}
+
     - name: Checkout Downstream
       uses: actions/checkout@v2
       with:
+        ref: ${{ github.base_ref }}
         repository: ${{ matrix.org }}/${{ matrix.repo }}
         path: ./src/knative.dev/${{ matrix.repo }}
+
     - name: Test Downstream
       uses: knative-sandbox/actions-downstream-test@v1
       with:
--- a/.github/workflows/knative-downstream.yaml
+++ b/.github/workflows/knative-downstream.yaml
@@ -85,7 +85,7 @@ jobs:
       run: |
         go get github.com/google/go-licenses
 
-    - name: Check out code
+    - name: Checkout upstream at ${{ github.base_ref }}
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/knative-downstream.yaml
```diff
@@ -80,18 +80,33 @@ jobs:
...
+
     - name: Install Dependencies
       run: |
         go get github.com/google/go-licenses
-    - name: Checkout Upstream
+
+    - name: Check out code
       uses: actions/checkout@v2
       with:
+        fetch-depth: 0
```
n3wscott, line 91: this is the default, no need to set it to zero

.github/workflows/knative-downstream.yaml
```diff
@@ -80,18 +80,33 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1.16.x
+
     - name: Install Dependencies
       run: |
         go get github.com/google/go-licenses
-    - name: Checkout Upstream
+
+    - name: Check out code
```
n3wscott, line 88:
```suggestion
    - name: Checkout Upstream at ${{ github.base_ref }}
```

dprotaso:
oh weird - something broke
/hold

dprotaso:
looks like fetch-depth default is 1 and not 0 - so we'll probably want 0

dprotaso:
@n3wscott PTAL

.github/workflows/knative-downstream.yaml
```diff
@@ -80,19 +80,34 @@ jobs:
...
-    - name: Checkout Upstream
+
+    - name: Checkout upstream at ${{ github.base_ref }}
       uses: actions/checkout@v2
       with:
+        fetch-depth: 0
         path: ./src/knative.dev/${{ github.event.repository.name }}
-    - name: Checkout Downstream
+
+    - name: Merge upstream
```
n3wscott, line 94: Why is using ref in the checkout not enough?

.github/workflows/knative-downstream.yaml
```diff
@@ -80,19 +80,34 @@ jobs:
...
+      if: github.event_name == 'pull_request'
+      shell: bash
+      working-directory: ./src/knative.dev/${{ github.event.repository.name }}
+      run: |
+         git remote add upstream https://github.com/${{ github.repository }}.git
+         git fetch upstream ${{ github.base_ref }}
```
n3wscott, line 100: What is baseref in the case of a pull request?

.github/workflows/knative-downstream.yaml
```diff
@@ -80,19 +80,34 @@ jobs:
...
+         git pull --no-rebase --no-commit upstream ${{ github.base_ref }}
```
n3wscott, line 101: I think what you are going to be doing is checking out the fork, and then pulling the forks branch that the pr is based on and what you are trying to do is checkout the original repo and try to merge the fork onto it? I am not sure this is needed, if the repo has broken the API for the downstream, it should have been reported, and there would be a conflict?

.github/workflows/knative-downstream.yaml
```diff
@@ -80,19 +80,34 @@ jobs:
...
+
     - name: Install Dependencies
       run: |
         go get github.com/google/go-licenses
-    - name: Checkout Upstream
+
+    - name: Checkout upstream at ${{ github.base_ref }}
       uses: actions/checkout@v2
       with:
+        fetch-depth: 0
```
n3wscott, line 91: Remove the default?

n3wscott:
I am not sure this is checking the HEAD repo out, but I could.be wrong. It seems it is merging the base branch back onto the pr from the fork repo? Please confirm.

.github/workflows/knative-downstream.yaml
dprotaso: after a PR sits for a few days we want to retest with any changes `HEAD`

dprotaso: especially if there are failed checks - this what prow does as well

.github/workflows/knative-downstream.yaml
dprotaso: Whatever the target branch is - ie. `main` or `release-*`

.github/workflows/knative-downstream.yaml
dprotaso: the default is `1` not 0

.github/workflows/knative-downstream.yaml
dprotaso: This pulls in the latest commits from `HEAD` into the PRs changes. We already do this prior to running unit tests etc.

dprotaso:
> It seems it is merging the base branch back onto the pr from the fork repo? Please confirm.

Upstream is here:

`git remote add upstream https://github.com/${{ github.repository }}.git`


repository is where the event occurred - https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

dprotaso:
bump @n3wscott

n3wscott (commented):
/lgtm
/approve

PR #2267, opened 2021-09-02
[release-0.24] when testing downstream use the correct release branch

dprotaso:
Verifies - https://github.com/knative/pkg/pull/2265

and useful for this supported release branch


dprotaso:
I'm going to close this out - there are more changes required in older releases that we don't need

ie. chmod permission on due to go-license tooling changes

PR #2272, opened 2021-09-03
Check Knative CustomResourceDefinitions stored version

mgencur:
# Changes

- Additional check that can be reused in post-upgrade tests
- Can be reused downstream to verify migration from one version to another
- Example usage in Knative Eventing is available at https://github.com/knative/eventing/pull/5705

We have recently run into an issue after migrating from Knative Eventing 0.22 to 0.23. The v1beta1 version of apiserversources.sources.knative.dev was completely removed and it was not possible to upgrade because sometimes in the past the "migrator" did not properly migrate these resources (between 0.19.2 and 0.20.0). Adding these checks will ensure that the migration happens. 

/kind enhancement

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/test/migrate/checks.go
+++ b/test/migrate/checks.go
@@ -1,0 +1,63 @@
+package migrate
+
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckStoredVersions verifies that all status.storedVersions from Knative CRDs are listed in the spec
+// with storage: true. It means the CRDs have been migrated and previous/unused API versions
+// can be safely removed from the spec.
+func CheckStoredVersions(ctx context.Context, apiextensions *apiextensionsv1.ApiextensionsV1Client) error {
+	crdClient := apiextensions.CustomResourceDefinitions()
+
+	crdList, err := crdClient.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to fetch crd list %w", err)
+	}
+
+	var (
+		failed bool
+		errMsg strings.Builder
+	)
+	for _, crd := range crdList.Items {
+		if strings.Contains(crd.Name, "knative.dev") {
+			for _, stored := range crd.Status.StoredVersions {
+				for _, v := range crd.Spec.Versions {
+					if stored == v.Name && !v.Storage {
+						failed = true
+						fmt.Fprintf(&errMsg, "\"%s\" is invalid: spec.versions.storage must be true for \"%s\" or "+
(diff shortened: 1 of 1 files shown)


test/migrate/checks.go
```diff
@@ -0,0 +1,63 @@
...
+
+// CheckStoredVersions verifies that all status.storedVersions from Knative CRDs are listed in the spec
+// with storage: true. It means the CRDs have been migrated and previous/unused API versions
+// can be safely removed from the spec.
+func CheckStoredVersions(ctx context.Context, apiextensions *apiextensionsv1.ApiextensionsV1Client) error {
+	crdClient := apiextensions.CustomResourceDefinitions()
+
+	crdList, err := crdClient.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to fetch crd list %w", err)
```
markusthoemmes, line 37:
```suggestion
		return fmt.Errorf("unable to fetch crd list: %w", err)
```

test/migrate/checks.go
```diff
@@ -0,0 +1,63 @@
...
+		errMsg strings.Builder
+	)
+	for _, crd := range crdList.Items {
+		if strings.Contains(crd.Name, "knative.dev") {
+			for _, stored := range crd.Status.StoredVersions {
+				for _, v := range crd.Spec.Versions {
+					if stored == v.Name && !v.Storage {
+						failed = true
+						fmt.Fprintf(&errMsg, "\"%s\" is invalid: spec.versions.storage must be true for \"%s\" or "+
+							"version %s must be removed from status.storageVersions\n", crd.Name, v.Name, v.Name)
```
markusthoemmes, line 51: I'm wondering if we should make this take `testing.T` and call `t.Errorf` for each of the fails. That'd avoid the awkwardness of having an error which is actually multiple errors here.

Also: Use `%q` for quoted strings to avoid the escape stuff.

markusthoemmes:
/assign @dprotaso 

This should be right up your alley!

test/migrate/checks.go
mgencur: Alright, makes sense. I guess.

test/migrate/checks.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"strings"
+	"testing"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CheckStoredVersions verifies that all status.storedVersions from Knative CRDs are listed in the spec
+// with storage: true. It means the CRDs have been migrated and previous/unused API versions
```
dprotaso, line 29: This is not the case

You can have N `storedVersions` technically but you can only have a single version in the spec with `storage: true`

It's on us to migrate resources to newer versions and then update the status block removing the older version

dprotaso: If your intent is to use this after migration post-install job - then I'd probably name this something more explicit.

maybe something link

```
ExpectSingleStoredVersion
```

test/migrate/checks.go
```diff
@@ -0,0 +1,52 @@
...
+// can be safely removed from the spec.
+func CheckStoredVersions(t *testing.T, apiextensions *apiextensionsv1.ApiextensionsV1Client) {
+	t.Helper()
+	crdClient := apiextensions.CustomResourceDefinitions()
```
dprotaso, line 33: nit: maybe accept the CRD client vs the wider api extensions client

test/migrate/checks.go
```diff
@@ -0,0 +1,52 @@
...
+
+	crdList, err := crdClient.List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		t.Fatal("Unable to fetch crd list:", err)
+	}
+
+	for _, crd := range crdList.Items {
+		if strings.Contains(crd.Name, "knative.dev") {
```
dprotaso, line 41: consider making this a param since knative.dev/pkg is consumed by tekton etc.

test/migrate/checks.go
mgencur: OK. Hopefully the latest changes address your comments. Sorry it took so long. I didn't notice your comments.

PR #2276, opened 2021-09-06
Declare `DefaultErrorRetryChecker` as `ErrorRetryChecker` type

nak3:
This patch changes `DefaultErrorRetryChecker` to `ErrorRetryChecker`
type before adding in the interface array.

Without it, the type of DefaultErrorRetryChecker is still `func(error) (bool, error)`.
https://play.golang.org/p/CLlGbwGQeMu

/cc @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

I'll add some test coverage to this since I screwed it up :sweat:

PR #2278, opened 2021-09-06
Drop Retrying and EventuallyMatchesBody checks

markusthoemmes:
# Changes

As per title, these are no longer in-use anywhere.

/assign @julz

### DIFF
--- a/test/request.go
+++ b/test/request.go
@@ -21,7 +21,6 @@ package test
 import (
 	"context"
 	"net/url"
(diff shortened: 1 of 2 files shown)


julz:
is the downstream unit failure on operator legit?

markusthoemmes:
Goes away when https://github.com/knative/operator/pull/758 merges.

julz (approved):
/lgtm

PR #2279, opened 2021-09-07
Enable using `duckv1.WithPod` with our webhook infra.

mattmoor:
Today you can't use `duckv1.WithPod` to author webhooks because it doesn't implement the `Validate` or `SetDefaults` methods, and that makes sense since they are (by definition) a generic encapsulation of a number of types.

However, if we could infuse `ctx` with appropriate callbacks for `Validate` and `SetDefaults` then folks can use our types, but infuse `ctx` with callbacks that perform the appropriate validation.

What I have in mind is something along these lines:

```go
return defaulting.NewAdmissionController(ctx,
	// Name of the resource webhook.
	"foo.bar.dev",

	// The path on which to serve the webhook.
	"/defaulting",

	// The resources to default.
	map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
		appsv1.SchemeGroupVersion.WithKind("Deployment"):  &duckv1.WithPod{},
		appsv1.SchemeGroupVersion.WithKind("ReplicaSet"):  &duckv1.WithPod{},
		appsv1.SchemeGroupVersion.WithKind("StatefulSet"): &duckv1.WithPod{},
		appsv1.SchemeGroupVersion.WithKind("DaemonSet"):   &duckv1.WithPod{},
		batchv1.SchemeGroupVersion.WithKind("Job"):        &duckv1.WithPod{},
	},

	// A function that infuses the context passed to Validate/SetDefaults with custom metadata.
	func(ctx context.Context) context.Context{
		return duckv1.WithPodSpecDefaulter(ctx, myFancyLogic)
	},

	// Whether to disallow unknown fields.
	false,
)
```

It is roughly equivalent for validation.

/kind enhancement

I was looking at this through the lens of @dlorenc `cosigned` webhook, which currently has to futz around with a mix of concrete types and dynamic clients to achieve a similar result.  I'm hoping that with this we can just write some logic against `duckv1.WithPod` that captures all of these types.

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/apis/duck/v1/podspec_defaults.go
+++ b/apis/duck/v1/podspec_defaults.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"context"
+)
+
+// PodSpecDefaulter is a callback to validate a PodSpecable.
+type PodSpecDefaulter func(context.Context, *WithPod)
+
+// SetDefaults implements apis.Defaultable
+func (wp *WithPod) SetDefaults(ctx context.Context) {
+	if psd := GetPodSpecDefaulter(ctx); psd != nil {
+		psd(ctx, wp)
+	}
+}
+
+// psvKey is used for associating a PodSpecDefaulter with a context.Context
+type psdKey struct{}
+
+func WithPodSpecDefaulter(ctx context.Context, psd PodSpecDefaulter) context.Context {
+	return context.WithValue(ctx, psdKey{}, psd)
+}
+
+// GetPodSpecDefaulter extracts the PodSpecDefaulter from the context.
+func GetPodSpecDefaulter(ctx context.Context) PodSpecDefaulter {
+	untyped := ctx.Value(psdKey{})
+	if untyped == nil {
+		return nil
+	}
+	return untyped.(PodSpecDefaulter)
+}
--- a/apis/duck/v1/podspec_defaults_test.go
+++ b/apis/duck/v1/podspec_defaults_test.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


apis/duck/v1/podspec_defaults.go
```diff
@@ -0,0 +1,47 @@
...
+func (wp *WithPod) SetDefaults(ctx context.Context) {
+	if psd := GetPodSpecDefaulter(ctx); psd != nil {
+		psd(ctx, wp)
+	}
+}
+
+// psvKey is used for associating a PodSpecDefaulter with a context.Context
+type psdKey struct{}
+
+func WithPodSpecDefaulter(ctx context.Context, psv PodSpecDefaulter) context.Context {
```
n3wscott, line 36: super nit, 
```suggestion
func WithPodSpecDefaulter(ctx context.Context, psd PodSpecDefaulter) context.Context {
```

apis/duck/v1/podspec_defaults.go
```diff
@@ -0,0 +1,47 @@
...
+	return context.WithValue(ctx, psdKey{}, psv)
```
n3wscott, line 37:
```suggestion
	return context.WithValue(ctx, psdKey{}, psd)
```

apis/duck/v1/podspec_defaults.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// GetPodSpecDefaulter extracts the PodSpecDefaulter from the context.
+func GetPodSpecDefaulter(ctx context.Context) PodSpecDefaulter {
+	untyped := ctx.Value(psdKey{})
+	if untyped == nil {
+		return nil
+	}
+	return untyped.(PodSpecDefaulter)
```
n3wscott, line 46: do you want to check the type here with `typed, ok := ...` and not return `typed` if not ok?

apis/duck/v1/podspec_defaults_test.go
```diff
@@ -0,0 +1,99 @@
...
+		name: "no change",
+		with: func(ctx context.Context) context.Context {
+			return WithPodSpecDefaulter(ctx, func(ctx context.Context, wp *WithPod) {
+			})
+		},
+		want: p.DeepCopy(),
+	}, {
+		name: "no busybox",
+		with: func(ctx context.Context) context.Context {
+			return WithPodSpecDefaulter(ctx, func(ctx context.Context, wp *WithPod) {
```
n3wscott, line 62: this looks very useful.

This might also help with in-memory defaulters for reconciler diffs when we are trying to figure out when to reapply a manifest.

n3wscott (approved):
lgtm except for the nit, looks like copy/pasta from the validator version.

apis/duck/v1/podspec_defaults.go
mattmoor: The private `psdKey` makes it impossible for anything outside of this package to associate garbage here, so I'm not worried.

PR #2281, opened 2021-09-09
Add more context to reconcile error debug logs

julz:
We were debugging some slow queue processing behaviour, and the fact these logs dont have a trace_id or key to associate them with other logs / the failing item, and to distinguish them from other parallel items, made things harder than they needed to be to follow on a busy system.

# Changes

- 🧹 Make sure the context (key, trace id) is included in the debug logs for the non-happy path in controller.go
- 🧹 Fix up error message prefix in EnqueueAfter to match the actual method it's in

/kind enhancement

**Release Note**

```release-note
Debug logs in controller now properly associate messages with trace id and key
```

**Docs**

```docs

```

/assign @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

julz:
Metrics test flake, again. Putting that on my list to look at (again :)).

/retest

PR #2285, opened 2021-09-15
Source validation fixes

gabo1208:
Little bug fix to knative/eventing#5638

### DIFF
--- a/apis/duck/v1/source_types_test.go
+++ b/apis/duck/v1/source_types_test.go
@@ -29,8 +29,12 @@ func TestSourceValidate(t *testing.T) {
 		src  *Source
 		want *apis.FieldError
 	}{{
-		name: "empty source validation",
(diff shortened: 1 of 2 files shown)


apis/duck/v1/source_types_test.go
```diff
@@ -32,6 +32,10 @@ func TestSourceValidate(t *testing.T) {
 		name: "empty source validation",
 		src:  nil,
 		want: nil,
+	}, {
+		name: "empty source spec validation",
+		src:  &Source{Spec: SourceSpec{}},
```
n3wscott, line 37: Can you also add a test where `.spec` is nil?

mattmoor:
I don't see this cherry-picked into `release-0.26`, and it is causing https://github.com/knative/eventing/pull/5744 to fail

PR #2293, opened 2021-09-22
Continual tests logging within test scope

mgencur:
Fixes an issue with logging of Continual upgrade tests which interleaves with other logs and cannot be easily connected to a test that produces the logs. An example run that has this problem is [here](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_serving/11988/pull-knative-serving-upgrade-tests/1437861332329172992/build-log.txt) (note especially logs from autoscale.go which are scattered across other logs.

# Changes

- :bug: Fix bug
- Use internal logger that is buffering the logs from continual tests - both Start and Verify phases
- Print the complete log in t.Log before the "Verify" phase. Print the rest of background logs to t.Log when the Verify phase ends in case a background routine is still writing logs while being in the Verify phase. 
- If the Start phase fails and there are some logs they're printed to t.Log immediately without waiting for the Verify phase.
- The logs are then printed within the test scope as can be seen in a test PR here: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_serving/12027/pull-knative-serving-upgrade-tests/1473586664545718272

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


markusthoemmes:
> Print the complete log in t.Log once the test finishes (at "Verify" phase). This is in turn printed to stdout by testing.T only if the test fails.

Does that mean that the loglines will be out of order and we'll have to backtrack their order based on the timestamps logged?

mgencur:
@markusthoemmes no, if you look at the test run the logs are ordered, the timestamps go chronologically. https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_serving/12027/pull-knative-serving-upgrade-tests/1440597482730950656/build-log.txt
Do you see a problem in the logs?

markusthoemmes:
`autoscaler.go:65: Induced error` supposedly should actually be at the end of the log rather than at the beginning? If anything in the test itself logs to `t.Log` it'll be out of order with the buffered logs, I think.

mgencur:
Actually, if the Verify phase logs something via t.Log/Error. it will be printed first. And then the whole output of background operations is printed (but the output in itself is ordered)

test/upgrade/testing_operations_test.go
```diff
@@ -38,15 +35,12 @@ func newConfig(t *testing.T) (upgrade.Configuration, fmt.Stringer) {
...
-	}
-	ws := zapcore.NewMultiWriteSyncer(buf, os.Stdout)
-	core := zapcore.NewCore(encoder, ws, zap.DebugLevel)
-	return zap.New(core).WithOptions(), buf
+	buf := upgrade.NewThreadSafeBuffer()
+	core := zapcore.NewCore(
+		zapcore.NewConsoleEncoder(ec),
+		zapcore.NewMultiWriteSyncer(zapcore.AddSync(buf), os.Stdout),
+		zap.DebugLevel)
+	return zap.New(core), buf
```
markusthoemmes, line 43: Could this reuse `NewInMemoryLoggerBuffer` and could we make the ThreadSafeBuffer stuff non-exported as a consequence?

test/upgrade/functions.go
```diff
@@ -160,3 +165,39 @@ func (s *simpleBackgroundOperation) Setup() func(c Context) {
 func (s *simpleBackgroundOperation) Handler() func(bc BackgroundContext) {
 	return s.handler
 }
+
+func NewThreadSafeBuffer() *ThreadSafeBuffer {
+	return &ThreadSafeBuffer{
+		Buffer: bytes.Buffer{},
+		Mutex:  sync.Mutex{},
```
markusthoemmes, line 172: You don't actually need to write this out I believe.

test/upgrade/functions.go
```diff
@@ -110,6 +114,7 @@ func handleStopEvent(
 	bc.Log.Infof("%s have received a stop event: %s", wc.Name, se.Name())
 	defer close(se.Finished)
 	wc.OnStop(se)
+	se.T.Log(bc.LogBuffer)
```
markusthoemmes, line 117: Should this log every line on its own rather than the whole package in one line?

Additionally (because of the ordering "issue") should we log "Background logs:" and "Background logs end" explicitly, to clearly denote which logs are coming from the background process?

test/upgrade/types.go
```diff
@@ -84,10 +86,12 @@ type Context struct {
...
 // which end user should use to obtain a testing.T for error reporting. Until
 // StopEvent is sent user may use zap.SugaredLogger to log state of execution if
-// necessary.
+// necessary. The logs are stored in a ThreadSafeBuffer and flushed to the test
+// output when the test fails.
 type BackgroundContext struct {
-	Log  *zap.SugaredLogger
-	Stop <-chan StopEvent
+	Log       *zap.SugaredLogger
+	LogBuffer *ThreadSafeBuffer
```
cardil, line 93: I don't like adding this internal thing to public struct as public field. As minimal effort, to should be `logBuffer` if this needs to be here at all.

test/upgrade/steps.go
```diff
@@ -63,17 +63,20 @@ func (se *suiteExecution) startContinualTests(num int) {
...
 				handler := operation.Handler()
 				go func() {
-					bc := BackgroundContext{Log: l, Stop: operation.stop}
+					bc := BackgroundContext{Log: logger, LogBuffer: buffer, Stop: operation.stop}
 					handler(bc)
 				}()
 
 				se.failed = se.failed || t.Failed()
 				if se.failed {
+					t.Log(buffer)
```
cardil, line 79: Each line should be logged by it self, instead of whole buffer in one go.

Also, I think we should log no mater that test fail or not. That's the job of `go test -v` to display those logs or not.

test/upgrade/skip_at_verification_test.go
```diff
@@ -23,31 +23,31 @@ import (
 )
 
 func TestSkipAtBackgroundVerification(t *testing.T) {
+	log, buf := newExampleZap()
```
cardil, line 26: Why all those changes to this test? It seams not required to me at all.

test/upgrade/functions_test.go
```diff
@@ -120,22 +120,26 @@ func TestSuiteExecuteWithComplete(t *testing.T) {
...
-	txt := expectedTexts(suite, fp)
-	txt.append(upgradeTestRunning, upgradeTestSuccess)
-	txt.append(
+	expected := expectedTexts(suite, fp)
+	expected.append(upgradeTestRunning, upgradeTestSuccess)
+	expected.append(
 		"Installing Serving stable 0.17.1",
 		"Installing Eventing stable 0.17.2",
-		"Running Serving continual test",
-		"Stopping and verify of Eventing continual test",
```
cardil, line 129: Could you explain why those texts are not expected any more? I presume that's connected to the other my comment - logging only on failure.

test/upgrade/functions.go
```diff
@@ -160,3 +165,39 @@ func (s *simpleBackgroundOperation) Setup() func(c Context) {
 func (s *simpleBackgroundOperation) Handler() func(bc BackgroundContext) {
 	return s.handler
 }
+
+func NewThreadSafeBuffer() *ThreadSafeBuffer {
```
cardil, line 169: All those `ThreadSafeBuffer` related functions doesn't need to be public

cardil (changes requested):

test/upgrade/steps.go
mgencur: I agree with therest, but why log every line separately?

test/upgrade/skip_at_verification_test.go
mgencur: The background operations were using a logger that doesn't have the right logs anymore. The buffer now has them. So I'm passing the buffer/logger directly. The goal of this test is to check that the tests are/aren't skipped.

test/upgrade/functions_test.go
mgencur: The tests are setting up a logger/buffer at the beginning and passing that to `upgrade.Configuration`. Later they expect the messages to be there in the buffer. But the framework now uses a different logger/buffer for the background operations which are only logged to testing.T when the test fails. So, when the test fails the messages should be on stdout (through testing.T), when the test passes the messages are not printed..
If we want to specifically verify these messages I can pass a separate logger like I did in the TestSkipAtBackgroundVerification

mgencur:
Pushed updates. The example of resulting logs is in this run: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_serving/12027/pull-knative-serving-upgrade-tests/1440679478693990400 (this is without the last commit which removes the extra space)

mgencur:
@cardil @markusthoemmes is there anything else to do with this PR or is this ready for merging from your POV? Can we agree that the logs look better than before? Instead of being interleaved with all other logs they're now in the place of the given test.
I've answered the questions and addressed the feedback mostly (except for logging each line separately which I'm missing the reason for). Thanks

test/upgrade/steps.go
cardil: As we are not logging them when they occur, we should save emit time. Then log those log entries with that emit time. That's to ensure we can see how much time given operations take.

test/upgrade/functions_test.go
cardil: Now, the logs are written on the output in successful state as well - after other my comment.

So, with that change, we should see expected logs in the end.

test/upgrade/skip_at_verification_test.go
```diff
@@ -23,31 +23,31 @@ import (
...
 					},
 					func(c upgrade.Context) {
-						c.Log.Info("Verify 2")
+						log.Info("Verify 2")
 					},
 				),
 			},
 		},
 	}
-	log, buf := newExampleZap()
```
cardil, line 50: I don't see a valid point why those changes are needed here in this test. Now we are logging, no matter if test fail or success. This test should work as it was set up before.

cardil (changes requested):

test/upgrade/steps.go
cardil: Sorry about this. You are still using zap logger, and he adds the times.

So, it should be okay to print all of them at once.

test/upgrade/skip_at_verification_test.go
mgencur: We print the output to t.Log in the end but the test was checking that the original in memory logger had the logs which is no longer true. The internal logger has the logs and they're printed to t.Log

mgencur: I believe the point of this test is to check what is and is not executed (it's about skipping the test). And for this purpose it's enough to pass a separate logger like I'm going with this change.

test/upgrade/functions_test.go
mgencur: They're written to t.Log, not to the original logger which this test is checking. So, one way to test it would be similar to what I did for the skipped test below - pass a separate logger for the purpose of the test.

mgencur:
Hey, added a new commit that should address comments from @cardil 
Here's an example run on knative/serving side: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_serving/12027/pull-knative-serving-upgrade-tests/1445651684154937344

mgencur:
@cardil Hey, it's been 1.5 months since I pushed the latest changes. Anything else you want to add/comment?

cardil:
> @cardil Hey, it's been 1.5 months since I pushed the latest changes. Anything else you want to add/comment?

Sorry for the hold up. Review is here: https://github.com/mgencur/pkg/pull/1 - it still has problems, though.

mgencur:
@cardil thanks for the review ! After looking at the other PR with the code review it looks like almost nothing from my PR is left there and it's like a complete rewrite. I'm not sure why/how it performs better at this point except for keeping the tests closer to the original due to loggers delegating to testing.T. But I suppose that's the main part of the review. Perhaps, in this case I'd suggest that you open a new PR against knative/pkg and I can close mine (even though not happy about that). Somebody else should then review it.
Regarding the `t.Skip` or `t.Fail`, I don't think they're blockers. Skipping tests should happen at the beginning when there's nothing logged in the background yet. On the other hand `t.Fail` was not recommended even before this PR because (I believe!) it breaks the whole flow of the upgrade framework. And that's also why all tests so far have been using `t.Error`, not t.Fail. So, I don't see it as a critical drawback.

cardil:
> _it looks like almost nothing from my PR is left there and it's like a complete rewrite_

Not entirely. A lot of your code is still there - like to log configuration stuff, thread safe buffer. I did really add the forwarding to `testing.T` sink stuff, and some cosmetics. I think we could push this further as co-authors, at this point :smiley:.

> _I'm not sure why/how it performs better at this point except for keeping the tests closer to the original due to loggers delegating to `testing.T`_

A couple of points I had problems with these PRs was the change of tests, to not include expected messages. We should stick to tests as they are describing the current behavior, on which many projects now rely on. Also, by feeding directly to `testing.T` we will always report messages in the correct test case report. `testing.T` is taking care of that for us.



> _Regarding the `t.Skip` or `t.Fail`, I don't think they're blockers. Skipping tests should happen at the beginning when there's nothing logged in the background yet._

Unfortunately, that's not true. Even if you skip test at the beginning of verify phase, you will lose all logs that are there. They are collecting in the background. The in memory buffer is created in setup phase, before upgrade/downgrade is taking place. So, numerous logs will be lost.

> _On the other hand `t.Fail` was not recommended even before this PR because (I believe!)_

The `t.Fatal` or `t.Fail` are completely fine with current upgrade framework. What was problematic, was `panic` (from [`ensure.NoError(error)`](https://github.com/wavesoftware/go-ensure/blob/3824fe95969c1376f10cf2fe855f56abb69c1b84/errors.go#L12)). Couple of months ago we had removed the `panic` from the framework, in favor of `t.Fatal`, and now it reports properly.

test/upgrade/logger.go
```diff
@@ -0,0 +1,104 @@
...
+	// This is for backwards compatibility.
+	if c.Log != nil {
+		return c.Log.Sugar(), nil
+	}
+
+	testName := tb.Name()
+	testingTeeBees[testName] = tb
+	cfg := c.Config
+	outs := []string{fmt.Sprintf("%s://Log/%s", testingTScheme, testName)}
+	if len(cfg.OutputPaths) == 1 && cfg.OutputPaths[0] == "stdout" {
```
mgencur, line 51: Due to this line I have to specifically set `OutputPaths` to stdout because the `zap.NewDevelopmentConfig` by default writes to stderr for both OutputPaths and ErrorOutputPaths. If I don't override the OutputPaths everything shows twice in the console log (the logger then writes to both stderr and t.Error).
```
config := zap.NewDevelopmentConfig()
	config.OutputPaths = []string{"stdout"}
	return pkgupgrade.Configuration{T: t, LogConfig: pkgupgrade.LogConfig{Config: config}}
```

The question is: Should we use the `errOuts` if the cfg.OutputPaths is configured for stderr? ...
```
if len(cfg.OutputPaths) == 1 && cfg.OutputPaths[0] == "stderr" {
   cfg.OutputPaths = errOuts
}
```
```

Comment (mgencur) [reply test/upgrade/logger.go:51]: This part of code would then look like this:
```
outs := []string{fmt.Sprintf("%s://Log/%s", testingTScheme, testName)}
	errOuts := []string{fmt.Sprintf("%s://Error/%s", testingTScheme, testName)}
	if len(cfg.OutputPaths) == 1 {
		switch cfg.OutputPaths[0] {
		case "stdout":
			cfg.OutputPaths = outs
		case "stderr":
			cfg.OutputPaths = errOuts
		}
	} else {
		cfg.OutputPaths = append(cfg.OutputPaths, outs...)
	}
	if len(cfg.ErrorOutputPaths) == 1 && cfg.ErrorOutputPaths[0] == "stderr" {
		cfg.ErrorOutputPaths = errOuts
	} else {
		cfg.ErrorOutputPaths = append(cfg.ErrorOutputPaths, errOuts...)
	}
```

Comment (mgencur) [reply test/upgrade/logger.go:51]: Perhaps it would even make sense to treat "stderr" and "stdout" specifically and do this:
For both ErrorOutputPaths and OutputPaths: if any of the paths is "stderr" , replace it with `errOuts`, if any of the paths is "stdout", replace it with `outs`, and leave the rest as it is (no matter if we have 1 or more registered paths):
```
cfg.OutputPaths = replaceStdOutErr(testName, cfg.OutputPaths)
cfg.ErrorOutputPaths = replaceStdOutErr(testName, cfg.ErrorOutputPaths)
...
func replaceStdOutErr(testName string, paths []string) []string {
	out := fmt.Sprintf("%s://Log/%s", testingTScheme, testName)
	errOut := fmt.Sprintf("%s://Error/%s", testingTScheme, testName)
	var newPaths []string
	for _, path := range paths {
		switch path {
		case "stdout":
			newPaths = append(newPaths, out)
		case "stderr":
			newPaths = append(newPaths, errOut)
		default:
			newPaths = append(newPaths, path)
		}
	}
	return newPaths
}
```

**Review by cardil — CHANGES_REQUESTED** [review]
I've done review and manual checking of this. It looks good, but I have 2 minor requests.

They are committed in this PR: https://github.com/mgencur/pkg/pull/2

 * https://github.com/mgencur/pkg/pull/2/commits/6bbe3edf565c3928da69f13f765fe9b2e6b49d20 uses better config defaults, as creating default empty configuration will work (Golang best practices)
 * https://github.com/mgencur/pkg/pull/2/commits/f644bbbdb84760c37bb3cdd1226a57fa3f5acf06 fixes the problem of double output you've mentioned here https://github.com/knative/pkg/pull/2293#discussion_r774441736

File: test/upgrade/logger.go (hunk 10 above)
Comment (cardil) [reply test/upgrade/logger.go:51]: I think this should be fixed. I proposed such change in my 2nd review PR: https://github.com/mgencur/pkg/pull/2 (this commit https://github.com/cardil/knative-pkg/commit/f644bbbdb84760c37bb3cdd1226a57fa3f5acf06)

**Comment by mgencur** [comment]
@cardil thanks for the review! I'm testing the changes you suggested. The solution for "double output" seems to work fine. But the solution for config defaults has a problem: the zap.Config is passed by reference so the `logger() function in logger.go always references the same config and appends `fmt.Sprintf("%s://Log/%s", testingTScheme, testName)` to it. The result is printing every message multiple times as you can see in [here](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_serving/12027/pull-knative-serving-upgrade-tests/1485894745992466432/build-log.txt) (look at the bottom). I think the config needs to be passed by value so that logger() always uses a new cfg and doesn't append to the existing one. What do you think about this solution?: https://github.com/mgencur/pkg/commit/d8f109f7016d6396953f4a1928884df5d8dfa90e  (I'm testing it now...)

**Comment by cardil** [comment]
I've got a reproduction of that double output bug you've found in here. Please attach it for completeness.

https://github.com/knative/pkg/commit/3bba2da852228ee9342e85c8a5cc43f30ff8872c

**Comment by mgencur** [comment]
@cardil this unit test hangs if the fix (the last commit: "Do not pass zap.Config by reference") is not there. I believe we should have a test that either succeeds or fails but doesn't hang. I was trying to find the cause of that but without success. Anyway, this test is something that doesn't need to be in this PR.

**Comment by cardil** [comment]
At @mgencur:

> this unit test hangs if the fix (the last commit: "Do not pass zap.Config by reference") is not there. I believe we should have a test that either succeeds or fails but doesn't hang. I was trying to find the cause of that but without success. Anyway, this test is something that doesn't need to be in this PR.

That's true. This is super weird behavior. I did some debugging, and I saw the goroutine stack's looks like the code is waiting for the stop event. This shouldn't happen. It properly reports error if you limit failure point to first 3 steps: https://github.com/mgencur/pkg/blob/86d8875cfb1f62277b09d3a60010624c639f7748/test/upgrade/execute_failures_test.go#L29 

```go
for i := 1; i <= 3; i++ {
```

I feel a little unsure about merging this PR, without knowing the root cause. I know you are tired of this PR, but surly it's better to debug this here, then on someone else's PR in Serving or Eventing, right?

> If you're planning some changes - can you add the test in another PR along the changes? Let's unblock this PR.

Apart from this issue, there are some other failing checks in this PR. We should merge with those failures not addressed.

cardil:
The https://github.com/knative/pkg/issues/2399 is culprit of one test failure (code do not compile).

mgencur:
The downstream failure "Downstream Knative / Unit Test (serving, knative)" is caused by this PR: https://github.com/knative/pkg/pull/2395 . When I use `addressable_resolver.go` from before this commit the unit tests in Serving pass (tried locally). The PR above seems to have same failures as my own PR.

pierDipi:
> The downstream failure "Downstream Knative / Unit Test (serving, knative)" is caused by this PR: #2395 . When I use `addressable_resolver.go` from before this commit the unit tests in Serving pass (tried locally). The PR above seems to have same failures as my own PR. Edit: The same PR also causes the failure in "Downstream Knative / Unit Test (eventing, knative)" and also Downstream Knative / Unit Test (eventing-kafka-broker, knative-sandbox)

I'm aware of this and I was waiting for the release since pkg is cut one week before the release, downstream repositories can't use pkg@main

cardil:
I'm also seeing this test failing, https://github.com/knative/pkg/issues/2401. Strangely, it doesn't being reported in prow unit tests or Github's unit test check.

mgencur:
Hey @cardil , I think the cause of the hanging is following: The tests share the same zap.Config and it keeps adding either `testing-t://Log` or `testing-t://Error`. As soon as one of the tests adds `testing-t:Error`, any following call to `t.Log` or `t.Debug` will write to all the outputs including the previously added Error - which writes to t.Error - and it ends the current function. The framework is then not able to send the stop event.

mgencur:
I think the scenario with injected failure and shared zap.Config is not testable because t.Error will interrupt the test unexpectedly (in OnWait of the continual test we're calling t.Log.Debug)

cardil:
@mgencur Yes. Awesome work on analysis! That's the case, for sure. But, I think you can include the test as well because it works properly when the `zap.Config` is ensured to be copied. Your 86d8875 ensure that the case.

mgencur:
Even now, the test only passes when it's run with `-v` and fails when running without it. I guess that's because you stopped using the buffer and only check the test output which is only complete with `-v` :-/

cardil:
Right. I'll try fixing the test to pass always, even without `-v`.

I'm okay with merging it as it is, but maybe we should wait until the release, as @pierDipi said?

pierDipi:
> Right. I'll try fixing the test to pass always, even without `-v`.
> 
> I'm okay with merging it as it is, but maybe we should wait until the release, as @pierDipi said?

The release was done yesterday, so we're good.

cardil:
I don't get it @pierDipi. We are supposed to merge this without those passing checks? Or, you are going to prepare a fix? Or, we should wait until those failing projects consume new knative-pkg via automated update of dependencies?

pierDipi:
We can merge it since these tests are already failing.

I started fixing issues from serving, but the PR is still open https://github.com/knative/serving/pull/12560

cardil (approved):
/lgtm
/approve

dprotaso (commented):
This whole package is really hard to review and I think part of it is because we're mixing two loggers.  Related I think we can simplify the need for this PR if we map continous tests to golang tests differently.

Take serving for example which has the following structure

```
test/upgrade.TestServingUpgrades
test/upgrade.TestServingUpgrades/InstallingBase/*
test/upgrade.TestServingUpgrades/ContinualTests
test/upgrade.TestServingUpgrades/ContinualTests/SetupProbeTest
test/upgrade.TestServingUpgrades/PreUpgradeTests/*
test/upgrade.TestServingUpgrades/UpgradeWith/*
test/upgrade.TestServingUpgrades/PostUpgradeTests/*
test/upgrade.TestServingUpgrades/DowngradeWith/*
test/upgrade.TestServingUpgrades/PostDowngradeTests/*
test/upgrade.TestServingUpgrades/VerifyContinualTests
test/upgrade.TestServingUpgrades/VerifyContinualTests/ProbeTest
```

The continous probe test is 'split' across two golang tests instances - `ContinualTests/SetupProbeTest`  & `VerifyContinualTests/ProbeTest`.  Because of this split we don't have access to a test logger during the background probing. Thus I believe that's why we have a separate logger.

The alternative would be to run the continual tests in a single `t.Run()` - then the test logger is accessible during the life of the entire 'logical' test
ie.
```
test/upgrade.TestServingUpgrades/Continual/ProbeTest
```

In this golang test instance we would run setup, the background probing and verify.

mgencur:
> The continous probe test is 'split' across two golang tests instances - ContinualTests/SetupProbeTest & VerifyContinualTests/ProbeTest. Because of this split we don't have access to a test logger during the background probing. Thus I believe that's why we have a separate logger.

Yes. That's why we have two separate loggers.

> The alternative would be to run the continual tests in a single t.Run() - then the test logger is accessible during the life of the entire 'logical' test ie.

The continual test needs to run in background and while it's running the "upgrade" needs to happen. I tried this approach before but hit issues with ordering of tests and parallelization (especially the t.Parallel) feature. We discussed that briefly with @cardil - an option would be to get rid of nested `t.Run` and make the structure of the upgrade framework flat - use go routines and synchronization etc. But that would basically require a complete rewrite of the upgrade framework.
This current PR already took enormous time and we arrived at the point where it works. I believe the current way of logging stuff from continual tests is a big improvement and will make parsing the logs easier for everybody.

cardil:
@dprotaso The real life tests that uses this framework are complex. Take a look, for example, at this test suite configuration:

https://github.com/openshift-knative/serverless-operator/blob/ed2b76b759db261925e0fc6bd0d1788dc18b69e6/test/upgrade/upgrade_test.go#L45-L73

Example test report: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-knative_serverless-operator/1390/pull-ci-openshift-knative-serverless-operator-main-4.9-upgrade-tests-aws-ocp-49/1486800035646541824/artifacts/upgrade-tests-aws-ocp-49/serverless-e2e/artifacts/build-IzqjZquW/junit_5fDI8OZv.xml

We've got multiple subtests in various places, all of them are running in parallel during the upgrades. As @mgencur is saying, what you are proposing will mean removing the test structure `t.Run`, making the reporting much, much more complex.

This PR targets the reporting as the current approach uses the zap logger, so various continual tests outputs are interwoven with each other. That would be the case, if we do what you are suggesting.

cardil (approved):
Failing downstream tests are not related.

/lgtm

dprotaso:
Hey I wasn't suggesting we flatten the entire structure - you can use `t.Parallel()` to coordinate 

Here's an example how - https://go.dev/play/p/shyAxjbTZbk

PR #2294, opened 2021-09-22
Update very old Google API dependencies

markusthoemmes:
# Changes

As per title, these dependencies are **very** old and we should probably track more recent releases.


markusthoemmes:
/hold

To check all builds.

markusthoemmes:
The net-istio failure is not related.

PR #2302, opened 2021-09-30
introduce new case convention for leader election config map

dprotaso:
Part of https://github.com/knative/serving/issues/7300

Majority of serving uses lowercase-hyphens - ie. some-property over
someProperty

This changes the leader election config map parsing to support formats
and to prioritize the lowercase-hyphen case

### DIFF
--- a/leaderelection/config.go
+++ b/leaderelection/config.go
@@ -46,9 +46,14 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
 	config := defaultConfig()
 
 	if err := cm.Parse(data,
+		// Parse legacy keys first
 		cm.AsDuration("leaseDuration", &config.LeaseDuration),
 		cm.AsDuration("renewDeadline", &config.RenewDeadline),
 		cm.AsDuration("retryPeriod", &config.RetryPeriod),
+
+		cm.AsDuration("lease-duration", &config.LeaseDuration),
+		cm.AsDuration("renew-deadline", &config.RenewDeadline),
+		cm.AsDuration("retry-period", &config.RetryPeriod),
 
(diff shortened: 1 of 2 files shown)


leaderelection/config.go
```diff
@@ -46,10 +46,15 @@ func NewConfigFromMap(data map[string]string) (*Config, error) {
...
 
 	if err := cm.Parse(data,
+		// Parse legacy keys first
 		cm.AsDuration("leaseDuration", &config.LeaseDuration),
 		cm.AsDuration("renewDeadline", &config.RenewDeadline),
 		cm.AsDuration("retryPeriod", &config.RetryPeriod),
 
+		cm.AsDuration("lease-duration", &config.LeaseDuration),
+		cm.AsDuration("renew-deadline", &config.RenewDeadline),
+		cm.AsDuration("retry-period", &config.RetryPeriod),
```
markusthoemmes, line 56: Same comment as in Serving: Should we be "smart" in our `cm.AsX` functions and auto-transform the keys between camelCase and kebab-case to allow both styles?

dprotaso: I'm mixed on it - I figure a post-install job would help with the migration

julz (approved):
/lgtm
/approve

I figure we can maybe deprecate the existing keys for a few releases (or migrate them in a post-release job as @dprotaso suggests) and then drop the fallback code here, so I think I agree we should do this the explicit way rather than adding long-term support for both to pkg

/hold in case @markusthoemmes still has concerns

markusthoemmes:
I'm fine in general, just making sure we've also thought about the operator use-case.

dprotaso:
blah metrics
/retest

PR #2304, opened 2021-10-05
Allow tracking Cluster-scoped resources.

mattmoor:
Previously we required namespace to be non-empty, but it never is for Cluster-scoped resources.

This change elides validation for missing namespaces, assuming cluster-scoping.

/kind enhancement

**Release Note**

```release-note
tracker now supports Cluster-scoped resources.
```

**Docs**
```docs

```

### DIFF
--- a/tracker/enqueue_test.go
+++ b/tracker/enqueue_test.go
@@ -334,15 +334,6 @@ func TestBadObjectReferences(t *testing.T) {
 		},
 		match: "Kind",
 	}, {
-		name: "Missing Namespace",
-		objRef: corev1.ObjectReference{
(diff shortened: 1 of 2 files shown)


tracker/enqueue_test.go
```diff
@@ -333,15 +333,6 @@ func TestBadObjectReferences(t *testing.T) {
 			Name:       "kaniko",
 		},
 		match: "Kind",
-	}, {
-		name: "Missing Namespace",
```
n3wscott, line 337: could you please add a test into `TestAllowedObjectReferences` with the lack of a namespace?

PR #2305, opened 2021-10-05
Provide the webhook infrastructure with the raw request context.

mattmoor:
Today, we can use `Path()` of `/foo/` (trailing slash) to support prefix-matched webhooks, but unfortunately the request context is lost when `Admit()` or `Convert()` is called.

This ensures that information flows through associated with context for anyone who would like this metadata for additional processing.

/kind enhancement

**Release Note**

```release-note
Webhooks can now access the raw http.Request via `apis.GetHTTPRequest`.
```

**Docs**

```docs

```

### DIFF
--- a/apis/contexts.go
+++ b/apis/contexts.go
@@ -18,6 +18,7 @@ package apis
 
 import (
 	"context"
+	"net/http"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


n3wscott (approved):

mattmoor:
I wanted to see the path.  If you use `/foo/` for the path, then you can receive requests for `/foo/bar/baz`

mattmoor:
> That's very subtle

It's a "feature" of Go.

> how are you using that

I'm playing with a consolidated webhook that delegates logic based on the path that comes in.

Issue #2320, opened 2021-10-20
knative.dev/pkg broken with k8s 0.22 libraries


wlynch:
knative.dev/pkg is incompatible with k8s 0.22 libraries - looks like it's due to changes that cause the generated knative clients to no longer satisfy the upstream interfaces.

[Example](https://gist.github.com/wlynch/b3095516a1f44e7b8afb55905cff878b), which results in the following:

```sh
$ go get .
# knative.dev/pkg/client/injection/kube/client
../../../../../../pkg/mod/knative.dev/pkg@v0.0.0-20211019132235-ba2b2b1bf268/client/injection/kube/client/client.go:1236:9: cannot use &wrapAppsV1DeploymentImpl{...} (type *wrapAppsV1DeploymentImpl) as type "k8s.io/client-go/kubernetes/typed/apps/v1".DeploymentInterface in return argument:
        *wrapAppsV1DeploymentImpl does not implement "k8s.io/client-go/kubernetes/typed/apps/v1".DeploymentInterface (missing ApplyScale method)
../../../../../../pkg/mod/knative.dev/pkg@v0.0.0-20211019132235-ba2b2b1bf268/client/injection/kube/client/client.go:1253:5: cannot use (*wrapAppsV1DeploymentImpl)(nil) (type *wrapAppsV1DeploymentImpl) as type "k8s.io/client-go/kubernetes/typed/apps/v1".DeploymentInterface in assignment:
        *wrapAppsV1DeploymentImpl does not implement "k8s.io/client-go/kubernetes/typed/apps/v1".DeploymentInterface (missing ApplyScale method)
../../../../../../pkg/mod/knative.dev/pkg@v0.0.0-20211019132235-ba2b2b1bf268/client/injection/kube/client/client.go:1400:5: cannot use (*wrapAppsV1ReplicaSetImpl)(nil) (type *wrapAppsV1ReplicaSetImpl) as type "k8s.io/client-go/kubernetes/typed/apps/v1".ReplicaSetInterface in assignment:
        *wrapAppsV1ReplicaSetImpl does not implement "k8s.io/client-go/kubernetes/typed/apps/v1".ReplicaSetInterface (missing ApplyScale method)
[...]

PR #1731, opened 2020-09-22
Enable unparam linter to catch unused parameters

markusthoemmes:
Super useful linter to catch dead code hidden in parameters and return values.

PR #2327, opened 2021-10-26
Extend Apply verb to extensions, Bump to K8s 1.22.

mattmoor:
This was motivated by the linked issue, and does the work to upgrade PKG to 1.22 libs.

Fixes: https://github.com/knative/pkg/issues/2320

/kind enhancement
/hold

HOLDing this until we're clear that we *should* be bumping to 1.22 libs

**Release Note**

```release-note
Bump to K8s 1.22 libs
```

**Docs**
```docs

```


antoineco:
@mattmoor I didn't look into the details, but is anything stopping us from updating the code generator (add the `Apply` method) without actually pulling 1.22 dependencies into Knative yet?

This would allow an upgrade in two steps, and downstream could update their generated code ahead of time while upstream takes the time to do what's necessary.

Prometheus [recently bumped their dependencies](https://github.com/prometheus/prometheus/blob/v2.31.1/go.mod#L74-L76), which means it is now difficult to depend both on Knative and (up-to-date) Prometheus.

mattmoor:
I paged out a lot of the context here, but I think there was something (new type?).  You are welcome to try splitting pieces off of this.

antoineco:
Right, the new `ApplyScale()` method has the following [signature](https://github.com/kubernetes/kubernetes/blob/v1.22.3/staging/src/k8s.io/client-go/kubernetes/typed/apps/v1/deployment.go#L59):

```go
// import applyconfigurationsautoscalingv1 "k8s.io/client-go/applyconfigurations/autoscaling/v1"

ApplyScale(
	ctx context.Context,
	objectName string,
	scale *applyconfigurationsautoscalingv1.ScaleApplyConfiguration,
	opts metav1.ApplyOptions,
) (*autoscalingv1.Scale, error)
```

The type [`applyconfigurationsautoscalingv1.ScaleApplyConfiguration`](https://github.com/kubernetes/kubernetes/blob/v1.22.3/staging/src/k8s.io/client-go/applyconfigurations/autoscaling/v1/scale.go#L27-L34) was first introduced in Kubernetes `v1.22.0` 🧐 

So no choice but to update the libs.

Do you have cycles to rebase or would you prefer me to take over in a separate PR?

mattmoor:
I'm happy to rebase if we're 👍  on bumping, I just wanted to minimize that number if it was going to be a while.

antoineco:
@markusthoemmes based on your experience bumping k8s from 1.20 to 1.21, do you remember any specific issue downstream in the process?

markusthoemmes:
Not that I recall. Once we bump here, we're fine downstream usually.

nak3:
1.1 was released. @dprotaso Is this ready to go now?

mattmoor:
I can try to dust off this PR over the weekend if we are cool with the bump next release.

mattmoor:
Walking thru failures:
 * Tekton - all legs: failures applying patch they are carrying
 * Eventing: downstream has a problem with a breaking `trace` change: https://github.com/knative/pkg/runs/4661167676?check_suite_focus=true#step:7:491
 * Eventing Kafka (both): has same problem as above: https://github.com/knative/pkg/runs/4661168049?check_suite_focus=true#step:7:336
 * Eventing KEDA: downstream has K8s pinned https://github.com/knative-sandbox/eventing-autoscaler-keda/blob/04b969d0a9afcd2f9a73ff0aa5a7326130323918/go.mod#L23-L30
 * Eventing Redis: downstream has an OTel conflict: https://github.com/knative/pkg/runs/4661168191?check_suite_focus=true#step:7:160

I'll TAL at what we need to fix Eventing, and I already started a thread in Tekton slack to pre-flight bumping their dependency to fix this.

mattmoor:
From the OTel CHANGELOG:
> The concept of a remote and local Span stored in a context is unified to just the current Span. Because of this "go.opentelemetry.io/otel/trace".RemoteSpanContextFromContext is removed as it is no longer needed. Instead, "go.opentelemetry.io/otel/trace".SpanContextFromContex can be used to return the current Span. If needed, that Span's SpanContext.IsRemote() can then be used to determine if it is remote or not. (#1731)

mattmoor:
The eventing issue seems resolved, so I expect the kafka repos to be unblocked once their dependency on eventing is bumped (both have 9+ day old PRs with e2e failures/flakes)

mattmoor:
Ok, so KEDA is fixed by: https://github.com/knative-sandbox/eventing-autoscaler-keda/pull/223

This leaves the 2x Kafka failures, 3x Tekton failures, and the codecov leg, which always seems to fail.

I'm inclined to try to get this merged...

nak3:
LGTM :tada:

mattmoor:
cc @n3wscott for codegen changes

go.mod
```diff
@@ -41,11 +44,11 @@ require (
 	google.golang.org/grpc v1.42.0
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
-	k8s.io/api v0.21.4
-	k8s.io/apiextensions-apiserver v0.21.4
-	k8s.io/apimachinery v0.21.4
-	k8s.io/client-go v0.21.4
-	k8s.io/code-generator v0.21.4
+	k8s.io/api v0.22.2
```
dprotaso, line 47: probably want to bump to the latest patch now `v0.22.5`

dprotaso (changes requested):
we should bump to the latest patch which is `0.22.5`

mattmoor:
@dprotaso done.  I'm going to stage a PR to tekton/pipelines to confirm we'll be good there, but this is RFAL.

PR #2339, opened 2021-11-03
Disable controllers using a command-line flag

### LINKED ISSUE #2338: Disable controllers using a flag
I would like to disable controllers using a command-line flag

## Proposal

- Add a `disable-controllers` flag that is a comma-separated list of controller names.
- Add [1] to main.go

[1]:

```golang
// MainNamed runs the generic main flow for controllers and webhooks.
func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {

	disabledControllers := flag.String("disable-controllers", "", "Comma-separated list of disabled controllers.")

	// HACK: This parses flags, so the above should be set once this runs.
	cfg := injection.ParseAndGetRESTConfigOrDie()

	enabledCtors := enabledControllers(strings.Split(*disabledControllers, ","), ctors)

	MainWithConfig(ctx, component, cfg, enabledCtors...)
}
```

where `NamedControllerConstructor` is

```golang
// NamedControllerConstructor is a ControllerConstructor with an associated name.
type NamedControllerConstructor struct {
	// Name is the name of associated with the controller returned by ControllerConstructor.
	Name string
	// ControllerConstructor is a constructor for a controller.
	ControllerConstructor ControllerConstructor
}
```

pierDipi:
This patch allows disabling controllers when using a new `Main` function
called `MainNamed` which takes a list of `NamedControllerConstructor`.

# Changes

- Disable controllers using a command-line flag

/kind feature

Fixes #2338

**Release Note**

```release-note
Add MainNamedthat allows disabling controllers using a command-line flag (`disable-controllers`)
```

**Docs**

```docs
None
```

### DIFF
--- a/injection/interface.go
+++ b/injection/interface.go
@@ -103,6 +103,14 @@ type DynamicInterface interface {
 
 type ControllerConstructor func(context.Context, configmap.Watcher) *controller.Impl
 
+// NamedControllerConstructor is a ControllerConstructor with an associated name.
+type NamedControllerConstructor struct {
+	// Name is the name of associated with the controller returned by ControllerConstructor.
+	Name string
+	// ControllerConstructor is a constructor for a controller.
+	ControllerConstructor ControllerConstructor
+}
+
 var (
 	// Check that impl implements Interface
 	_ Interface = (*impl)(nil)
--- a/injection/interface.go
+++ b/injection/interface.go
@@ -105,7 +105,7 @@ type ControllerConstructor func(context.Context, configmap.Watcher) *controller.
 
 // NamedControllerConstructor is a ControllerConstructor with an associated name.
 type NamedControllerConstructor struct {
-	// Name is the name of associated with the controller returned by ControllerConstructor.
+	// Name is the name associated with the controller returned by ControllerConstructor.
 	Name string
 	// ControllerConstructor is a constructor for a controller.
 	ControllerConstructor ControllerConstructor
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -116,7 +116,7 @@ var (
 
 // MainNamed runs the generic main flow for controllers and webhooks.
 //
-// In addition to the MainWithConfig flow, it defines a `disabled-controllers` flag that allows to disable controllers
+// In addition to the MainWithConfig flow, it defines a `disabled-controllers` flag that allows disabling controllers
 // by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -132,7 +132,7 @@ func MainNamed(ctx context.Context, component string, ctors ...injection.NamedCo
 
 func enabledControllers(disabledControllers []string, ctors []injection.NamedControllerConstructor) []injection.NamedControllerConstructor {
 	disabledControllersSet := sets.NewString(disabledControllers...)
-	activeCtors := make([]injection.NamedControllerConstructor, 0, len(ctors)-len(disabledControllers))
+	activeCtors := make([]injection.NamedControllerConstructor, 0, len(ctors))
 	for _, ctor := range ctors {
 		if disabledControllersSet.Has(ctor.Name) {
 			log.Printf("Disabling controller %s", ctor.Name)
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -115,6 +115,9 @@ var (
 )
 
 // MainNamed runs the generic main flow for controllers and webhooks.
+//
+// In addition to the MainWithConfig flow, it defines a `disabled-controllers` flag that allows to disable controllers
+// by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
 	disabledControllers := pflag.StringSlice("disable-controllers", []string{}, "Comma-separated list of disabled controllers.")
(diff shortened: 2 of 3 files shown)


injection/sharedmain/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+			disabledControllers: []string{"foo"},
+			ctors: []injection.NamedControllerConstructor{
+				{Name: "foo"},
+				{Name: "bar"},
+			},
+			want: 1,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
julz, line 55: super nit, but do you need this if you don't call t.Parallel? (Maybe you want to add a `t.Parallel()` inside the run?)

injection/sharedmain/main.go
```diff
@@ -111,6 +113,38 @@ var (
...
+	cfg := injection.ParseAndGetRESTConfigOrDie()
+
+	enabledCtors := enabledControllers(strings.Split(*disabledControllers, ","), ctors)
+
+	MainWithConfig(ctx, component, cfg, enabledCtors...)
+}
+
+func enabledControllers(disabledControllers []string, ctors []injection.NamedControllerConstructor) []injection.ControllerConstructor {
+
+	disabledControllersSet := sets.NewString()
```
julz, line 131: seems like a shame not to not be able to use `sets.NewString(disabledControllers...)` here, do we definitely need to support spaces in the flag input? (looks like pflag.StringSlice doesn't strip spaces so it may make sense to stay consistent with that..)

injection/sharedmain/main.go
```diff
@@ -111,6 +113,38 @@ var (
 	WebhookMainWithConfig  = MainWithConfig
 )
 
+// MainNamed runs the generic main flow for controllers and webhooks.
+func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
+
+	disabledControllers := flag.String("disable-controllers", "", "Comma-separated list of disabled controllers.")
```
julz, line 119: can we use [pflag.StringSlice](https://pkg.go.dev/github.com/spf13/pflag#StringSlice) here?

odacremolbap (commented):
Looking good.

I'm wondering:

- should a warning be logged if a disabled name does not match any of the named controllers?
- if all controllers are disabled, shouldn't we `Fatal`?

There is also a suggestion at the code review to create a `MainWithOptions` to rule them all.

injection/sharedmain/main.go
```diff
@@ -111,6 +113,38 @@ var (
 	WebhookMainWithConfig  = MainWithConfig
 )
 
+// MainNamed runs the generic main flow for controllers and webhooks.
+func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
```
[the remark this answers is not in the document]
pierDipi: The HA flag will be removed soon since HA has been stable for a really long time.

I'm not sure how what you're suggesting should work.

What `WithNamedControllerConstructor(ctors ...injection.NamedControllerConstructor) Option` should be doing?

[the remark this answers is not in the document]
pierDipi: I don't really like pushing and pulling things from the context in general, especially if we can have proper typing in the interface.

pierDipi:
> should a warning be logged if a disabled name does not match any of the named controllers?

I was looking at what Kubernetes does and they don't log this info.

odacremolbap (approved):

pierDipi:
Hi @julz, when you have time please take a look, thanks!

injection/interface.go
```diff
@@ -103,6 +103,14 @@ type DynamicInterface interface {
 
 type ControllerConstructor func(context.Context, configmap.Watcher) *controller.Impl
 
+// NamedControllerConstructor is a ControllerConstructor with an associated name.
+type NamedControllerConstructor struct {
+	// Name is the name of associated with the controller returned by ControllerConstructor.
```
julz, line 108:
```suggestion
	// Name is the name associated with the controller returned by ControllerConstructor.
```

injection/sharedmain/main_test.go
```diff
@@ -0,0 +1,61 @@
...
+			want:                0,
+		},
+		{
+			name:                "two",
+			disabledControllers: []string{"foo"},
+			ctors: []injection.NamedControllerConstructor{
+				{Name: "foo"},
+				{Name: "bar"},
+			},
+			want: 1,
```
julz, line 51: I think I'd probably want to have this assert that we got the expected list rather than just the size of it (seems a bit too possible for this to accidentally pass with e.g. `{"foo"}` rather than `{"bar"}`)

injection/sharedmain/main.go
```diff
@@ -111,6 +114,32 @@ var (
 	WebhookMainWithConfig  = MainWithConfig
 )
 
+// MainNamed runs the generic main flow for controllers and webhooks.
```
julz, line 117: should probably say something about the difference between this method and the others here

julz (commented):
lgtm code-wise (just a couple little nits we might as well fix). Could you say a little more about [the use case behind the flag](https://github.com/knative/pkg/issues/2338#issuecomment-958968728)?

injection/sharedmain/main_test.go
pierDipi: OK, to do that we need to return `[]NamedControllerContructor`, meaning that we need to walk the controller constructor list twice, one for filtering named controllers (so this test can verify the list) and one for converting `[]NamedControllerContructor` to `[]ControllerContructor`.

pierDipi:
The use case is to disable controllers in order to safely rollout competing controllers (reconciling the same resource).

injection/sharedmain/main.go
```diff
@@ -111,6 +114,43 @@ var (
...
+	cfg := injection.ParseAndGetRESTConfigOrDie()
+
+	enabledCtors := enabledControllers(*disabledControllers, ctors)
+
+	MainWithConfig(ctx, component, cfg, toControllerConstructors(enabledCtors)...)
+}
+
+func enabledControllers(disabledControllers []string, ctors []injection.NamedControllerConstructor) []injection.NamedControllerConstructor {
+	disabledControllersSet := sets.NewString(disabledControllers...)
+	activeCtors := make([]injection.NamedControllerConstructor, 0, len(ctors)-len(disabledControllers))
```
dprotaso, line 135: This can panic

dprotaso:
I'm a bit confused about your use case since [there wasn't much discussion](https://github.com/knative/pkg/issues/2338#issuecomment-958968728). Is this for testing?

pierDipi:
> I'm a bit confused about your use case since [there wasn't much discussion](https://github.com/knative/pkg/issues/2338#issuecomment-958968728). Is this for testing?

This is the use case: https://github.com/knative/pkg/pull/2339#issuecomment-963165162

dprotaso:
But doesn't this approach require you to stop the entire controller?

dprotaso:
Can you be more concrete with the example - I'm just envisioning this in serving where we have [8 controllers in a single process](
https://github.com/knative/serving/blob/4bdd867387aa8ed428c360dc94dbd133035289a5/cmd/controller/main.go#L21-L28).

So if the intent is the disable one of those controllers as a way to migrate _who_ is responsible for reconciling that resource then this flag wouldn't work since you'd bring down 7 other controllers. For testing this is fine but if you're doing this in production that's scary.

There might be a way to do an 'online' transition by extending some of the leader election. Right now that has an assumption that it's the same deployment I believe.

pierDipi:
Is there anything left here @julz @dprotaso ?

dprotaso:
Thanks for the example.

The last thing to point out is that even though this 'disables' the controller from reconciling resources - the informers for that controller are still being started. This happens because injection builds it's DAG at compile time and all registered informers are created & sync'd prior to creating the controllers.

So I just wanted to confirm whether this is an issue for your scenario? And likewise to poll others whether the flag as is makes sense.

pierDipi:
On Mon, Nov 15, 2021 at 9:42 PM Dave Protasowski
***@***.***> wrote:
>
> Thanks for the example.
>
> The last thing to point out is that even though this 'disables' the controller from reconciling resources - the informers for that controller are still being started. This happens because injection builds it's DAG at compile time and all registered informers are created & sync'd prior to creating the controllers.
>
> So I just wanted to confirm whether this is an issue for your scenario? And likewise to poll others whether the flag as is makes sense.
>

Yes, that's OK.

PR #2341, opened 2021-11-03
Build downstream tests with go 1.17

markusthoemmes:
As per title. Some of them have started failing because of a type mismatch caused by Golang 1.16 vs 1.17.

/assign @julz


julz (approved):
/lgtm

PR #2347, opened 2021-11-11
Remove unclear comment for `makeSpoofClient()` arg

nak3:
This patch makes a tiny change to remove unclear comment for `makeSpoofClient()` arg.

The `timeout` arg is not bool but `time.Duration` so this comment is confusing.

/cc @shinigambit


markusthoemmes (approved):
/lgtm
/approve

PR #2350, opened 2021-11-16
codegen - support multiple annotation keys

dprotaso:
- allow multiple values for comment tag keys
- add OR filter chain
- update codegen to support multiple annotation keys

Part of: https://github.com/knative/serving/issues/12103

### DIFF
--- a/codegen/cmd/injection-gen/generators/comment_parser.go
+++ b/codegen/cmd/injection-gen/generators/comment_parser.go
@@ -19,25 +19,38 @@ import "strings"
 
 // Adapted from the k8s.io comment parser https://github.com/kubernetes/gengo/blob/master/types/comments.go
 
+// CommentsTags maps marker prefixes to a set of tags containing keys and values
+type CommentTags map[string]CommentTag
+
+// CommentTags maps keys to a list of values
+type CommentTag map[string][]string
+
 // ExtractCommentTags parses comments for lines of the form:
 //
-//   'marker' + ':' "key=value,key2=value2".
+//   "marker" + "prefix" + ':' + "key=value,key2=value2".
 //
-// Values are optional; empty map is the default.  A tag can be specified more than
+// Values are optional; empty map is the default. A tag can be specified more than
 // one time and all values are returned.  If the resulting map has an entry for
 // a key, the value (a slice) is guaranteed to have at least 1 element.
 //
 // Example: if you pass "+" for 'marker', and the following lines are in
 // the comments:
-//   +foo:key=value1,key2=value2
+//   +foo:key=value1,key2=value2,key=value3
 //   +bar
 //
 // Then this function will return:
-//   map[string]map[string]string{"foo":{"key":value1","key2":"value2"}, "bar": nil}
+//   map[string]map[string]string{
+//     "foo":{
+//      "key":  []string{"value1", "value3"},
+//      "key2": []string{"value2"}
+//     },
+//     "bar": {},
+//  }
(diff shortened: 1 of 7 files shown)


dprotaso:
/assign @n3wscott @nak3 
/hold cancel

Looks like networking => net-* is good and things build properly

nak3:
/cc @julz @markusthoemmes 

It looks good to me. But I don't have much experience around this code.

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -19,55 +19,71 @@ import "strings"
...
 // Example: if you pass "+" for 'marker', and the following lines are in
 // the comments:
-//   +foo:key=value1,key2=value2
+//   +foo:key=value1,key2=value2,key=value3
 //   +bar
 //
 // Then this function will return:
-//   map[string]map[string]string{"foo":{"key":value1","key2":"value2"}, "bar": nil}
+//   map[string]map[string]string{
+//     "foo":{
```
n3wscott, line 43: Can you add a prefix example here too?

codegen/cmd/injection-gen/generators/comment_parser.go
```diff
@@ -19,55 +19,71 @@ import "strings"
...
+	out := CommentTags{}
+
 	for _, line := range lines {
 		line = strings.TrimSpace(line)
 		if len(line) == 0 || !strings.HasPrefix(line, marker) {
 			continue
 		}
 
 		options := strings.SplitN(line[len(marker):], ":", 2)
+		prefix := options[0]
```
n3wscott, line 61: the comment above seems to think that "marker" + "prefix" is the annotation key now. But I do not see the marker first part getting used?

Oh I see, the marker is not needed, just to find the prefix and tags.

aaah, ok the marker becomes just "+", and prefix is the tag value. 


I think this might give us too many tags? It will return all tags in the file and the original version of this will filter the tags for just what we want. But I am not thinking of a downside to returning all the tags. It just causes the downstream to filter more CommentTags.

n3wscott (approved):
/lgtm

/hold

I think the comment could be updated to be more clear the `prefix` is "+", but LGTM and unhold if you disagree.

codegen/cmd/injection-gen/generators/comment_parser.go
dprotaso: The comment is a bit confusing

the marker is `+` the prefix is `foo` and `bar` in the example

Maybe I'll update the comment to make it more obvious

codegen/cmd/injection-gen/generators/comment_parser.go
dprotaso: > It will return all tags in the file and the original version of this will filter the tags for just what we want.

Not quiet - prior this function would convert

`+foo:key=value1,key=value2` to `foo: {key:value2}` 

now it converts that to `foo: {key:[value1, value2]}`

dprotaso: > It just causes the downstream to filter more CommentTags.

To clarify downstream has to iterate over the `CommentTag[key]` array while before it returned a string

codegen/cmd/injection-gen/generators/comment_parser.go
dprotaso: done

n3wscott:
/lgtm

Thanks Dave!!

PR #2351, opened 2021-11-17
Add retry checker for DNS failure from Ingress

nak3:
This patch fixes the recent flake in `service_to_service_test.go` on Kind/Istio.
istio-ingressgateway returns 502 with no such host error when it fails to resolve the cluster domain.

/cc @markusthoemmes

### DIFF
--- a/test/spoof/error_checks.go
+++ b/test/spoof/error_checks.go
@@ -21,6 +21,7 @@ package spoof
 import (
 	"errors"
 	"net"
+	"net/http"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


nak3:
https://github.com/knative/serving/pull/12294 runs this patch and seems error could be avoided.

markusthoemmes (approved):
/lgtm
/approve

dprotaso:
> This patch fixes the recent flake in service_to_service_test.go on Kind/Istio.

I guess to clarify this isn't a flake as  it consistently failed in Kind running on GitHub actions. Thus I wonder if the retry is masking some other issue.

The other concerns I have, and let me know if this is just paranoia, is that by including this error checker by default we could mask issues in other ingresses and wouldn't know about it

dprotaso:
To be more specific this only failed on GitHub actions with kind 1.22 & net-istio

nak3:
I was also concerned but it retries only for the no DNS record in the internal DNS temporary. 
I think it is difficult for ingress/coredns to guarantee that `*.svc.cluster.local` record is registered as soon as k8s svc is created so we need to retry it :thinking:

dprotaso:
Then maybe as part of probing we should ensure that DNS works as expected.

K8s 1.22 introduces a newer CoreDNS version - ie. with endpoint slice support thus maybe istio isn't playing well with it

markusthoemmes:
Sounds fair enough! Shall we track this and make sure we do our due diligence here?

PR #2354, opened 2021-11-19
kmap

dprotaso:
- move some functional map operations to it's own package `kmap`
- include a new accessor helper for migrating key names
- use kmap methods in various knative.dev/pkg subpackages
- add deprecation notices

This PR has two purposes:
- It sets up the `configmap.Parse` methods to be moved here 
  - https://github.com/knative/serving/issues/9957
- The accessor struct will help us manage multiple annotation keys 
   - https://github.com/knative/serving/issues/12103

It's important that this package has no external dependencies (especially k8s) to help with queue proxy size issues

### DIFF
--- a/kmap/accessor.go
+++ b/kmap/accessor.go
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    https://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
 Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


kmap/accessor.go
```diff
@@ -0,0 +1,58 @@
...
+limitations under the License.
+*/
+
+package kmap
+
+// Accessor is a utility struct for getting values from a map
+// given a list of keys
+//
+// This is to help the migration/renaming of annotations & labels
+type Accessor struct {
```
mattmoor, line 23: Maybe "OrderedLookup" would be a clearer name?

mattmoor (commented):
Generally LGTM.

dprotaso:
/hold going to update some godoc

unit test failure is some other flake

mattmoor (approved):

PR #2358, opened 2021-11-23
Remove unused recorder field

julz:
This field of the metricsConfig isn't set anywhere (and isn't exported), so it's doing nothing. Might as well simplify the code a bit by removing it.


julz:
Unrelated

/retest

odacremolbap (approved):

PR #2360, opened 2021-11-25
Drop unnecessary list creation from bucketer

markusthoemmes:
# Changes

As per title. It's unnecessary to actually create a list of the set to then only return the first element. Since the element's order is random anyway, might as well use `PopAny` in this case.

/assign @julz

### DIFF
--- a/hash/bucketer.go
+++ b/hash/bucketer.go
@@ -98,8 +98,7 @@ func (bs *BucketSet) Owner(key string) string {
 	}
 	bs.mu.RLock()
 	defer bs.mu.RUnlock()
-	l := ChooseSubset(bs.buckets, 1 /*single query wanted*/, key)
-	ret := l.UnsortedList()[0]
+	ret, _ := ChooseSubset(bs.buckets, 1 /*single query wanted*/, key).PopAny()
 	bs.cache.Add(key, ret)
 	return ret
 }
--- a/hash/hash_test.go
+++ b/hash/hash_test.go
@@ -35,11 +35,11 @@ func ExampleChooseSubset_selectOne() {
 
 	tasks := sets.NewString("task1", "task2", "task3")
 
-	ret := ChooseSubset(tasks, 1, "my-key1")
-	fmt.Println(ret.UnsortedList()[0])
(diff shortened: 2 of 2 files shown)


hash/bucketer.go
```diff
@@ -98,8 +98,7 @@ func (bs *BucketSet) Owner(key string) string {
 	}
 	bs.mu.RLock()
 	defer bs.mu.RUnlock()
-	l := ChooseSubset(bs.buckets, 1 /*single query wanted*/, key)
-	ret := l.UnsortedList()[0]
+	ret, _ := ChooseSubset(bs.buckets, 1 /*single query wanted*/, key).PopAny()
```
julz, line 101: just to say this out loud and check it's what we want: one downside of PopAny - it [deletes the item from the set](https://github.com/kubernetes/apimachinery/blob/v0.22.4/pkg/util/sets/byte.go#L191). I think that's OK because we don't use the set again here (and ChooseSubset doesn't return cached results), but it would be a problem if e.g. ChooseSubset cached. Also iiuc ChooseSubset with 1 as the second argument always results in a 1-element set, so there may not be much difference here between the two cases, but I can see an argument that PopAny is clearer in any event.

markusthoemmes: Thought through the exact same thingy and contemplated adding a new function to the set implementation (GetAny or whatever).
We can replace this with a quick `for` loop that immediately breaks if you think we should.

julz: (Or PeekAny maybe to be consistent with the set wording?). I was actually sort of toying with suggesting a non-subset `Choose` that would avoid the set entirely for the case where you only want 1 thing returned, but 🤷 we're _probably_ not calling this enough to justify that, although it would make the usage here a bit nicer and potentially let us drop some allocations (don't think this is used in hot path anyway, though). I'll stamp this for now and we can always decide to do one of the above things later.

julz (approved):
/lgtm

dprotaso:
note: This changes seems to have a weird side-effect and it's causing failures in a bunch of PRs where we're updating the package

PR #2362, opened 2021-11-26
Expose Queue proxy request metrics reporting period

skonto:
# Changes

- Exposes the reporting period used by Prometheus and Opencensus backend destinations.
- Required for fixing https://github.com/knative/serving/issues/12069

/kind api-change


metrics/config_observability.go
```diff
@@ -100,6 +103,7 @@ func NewObservabilityConfigFromConfigMap(configMap *corev1.ConfigMap) (*Observab
 		cm.AsBool(EnableReqLogKey, &oc.EnableRequestLog),
 		cm.AsBool(EnableProbeReqLogKey, &oc.EnableProbeRequestLog),
 		cm.AsString("metrics.request-metrics-backend-destination", &oc.RequestMetricsBackend),
+		cm.AsInt("metrics.request-metrics-reporting-period-seconds", &oc.RequestMetricsReportingPeriodSeconds),
```
skonto, line 106: Not using internal key `metrics.reporting-period-seconds` to target specifically queue proxy.

metrics/config.go
```diff
@@ -41,8 +41,6 @@ const (
 	DomainEnv = "METRICS_DOMAIN"
 
 	// The following keys are used to configure metrics reporting.
-	// See https://github.com/knative/serving/blob/main/config/config-observability.yaml
```
skonto, line 44: link is dead.

metrics/config_observability.go
```diff
@@ -69,6 +69,9 @@ type ObservabilityConfig struct {
 	// OpenCensus. "None" disables all backends.
 	RequestMetricsBackend string
 
+	// RequestMetricsReportingPeriodSeconds specifies the request metrics reporting period in sec at queue proxy, eg 1
+	RequestMetricsReportingPeriodSeconds int
```
julz, line 73: Should this be defaulted in the `defaultConfig` or is 0 going to be default & valid value?

metrics/config_observability.go
julz: Wonder if it's worth adding this context to the actual field / as a comment here -- if it's important enough for a reviewer here to know it might be something a future reader of the code might want to know too

metrics/config_observability.go
skonto: I know that this is an implementation detail but the reporting period in the metricsConfig has the following definition:

```
	// reportingPeriod specifies the interval between reporting aggregated views.
	// If duration is less than or equal to zero, it enables the default behavior.
	reportingPeriod time.Duration
```
So the same concept applies here. If a zero value is passed the default underlying behavior is triggered. Probably need to add a comment here.

metrics/config_observability.go
skonto: Yes would be helpful.

metrics/config_observability.go
skonto: I added a comment to clarify default behavior.

skonto: Added a default value.

metrics/config_observability.go
skonto: Gave a second though to this. We dont need that this is meant for queue proxy. We can do this at the side where we use the field. For example ` metrics.request-metrics-backend-destination` use in Serving is documented here:

https://github.com/knative/serving/blob/b70e8979d8b61e5d0dfa6eb3b9387cc789adb437/config/core/configmaps/observability.yaml#L99-102

skonto:
Gentle ping @julz @dprotaso

nak3:
It is probably better to create a PoC against serving with this change? At least it will be easier to review.

metrics/config.go
pierDipi: The new link is https://github.com/knative/serving/blob/main/config/core/configmaps/observability.yaml, can we re-add it?

pierDipi (commented):
LGTM, one nit.

skonto:
@nak3 ok I will do one to show how this is going to be used, the idea is to have that attribute configurable so we can fix that bug.

skonto:
Regarding [codecov](https://app.codecov.io/gh/knative/pkg/):

> codegen/cmd/injection-gen/generators 1.51%
> injection 32%
> tracing 57.99% 
> test 52.23%

Could we skip test and codegen, not sure what the settings are (dont have access) and open two issue for improving tracing and injection? /cc @pierDipi @nak3

skonto:
@nak3 here is the [poc](https://github.com/knative/serving/compare/main...skonto:reporting_qp?expand=1), setting the new attribute influences the default reporting at the qp exporter. I verified it again playing with the collector.

nak3:
Thank you.

> Note: The default exporter reporting-period values are defined per backend [here](https://github.com/knative/serving/blob/0b09cdda41ac308ea586fc2d2c86af4a6d2f87d5/vendor/knative.dev/pkg/metrics/config.go#L218). Prometheus: 5min, opencensus 1min. If the user sets metrics.reporting-period-seconds the new attribute we introduce in this PR will override that value at the queue proxy side. However, metrics.reporting-period-seconds will be set for other components like activator, autoscaler etc if the user adds that in the configmap.

Umm... I think it is a little bit complicated to understand. I understood as the following table:

component   | default       | when `metrics.reporting-period-seconds` in cm is 5s |
------------|---------------|-----------------------------------------------------|
queue-proxy | 5min or 1min  |  5s                                                 |
activator   | ?             |  ?                                                  |

But then, what is the value for `defaultRequestReportingPeriodSeconds`? activator (and other)'s default?

skonto:
@nak3 this attribute does not affect the other components:



component | default | when metrics.reporting-period-seconds in cm is 5s
-- | -- | --
queue-proxy | 10s | 5s
activator | 5min or 1min | 5min or 1min

nak3:
I understand. Thank you. (`request-metrics-reporting-period-seconds` vs `reporting-period-seconds` is bad for eyes, though :sweat_smile: )

So, this PR has changes, to be precise:
1. exposing `request-metrics-reporting-period-seconds` to set queue-proxy's reporting period via configmap.
2. decreasing the queue-proxy's default reporting period from 1min or 5min to 10s.

Is it alright without having two separated PRs? I am asking because I'm not sure how much the default change affects the current users and so I'm wondering the separated PRs will help troubleshooting/reverting/etc..

skonto:
> Is it alright without having two separated PRs?

I dont mind to do so.

> request-metrics-reporting-period-seconds vs reporting-period-seconds is bad for eyes,

I am open to suggestions :)

nak3:
/lgtm

Alright, then LGTM. (though I think I don't have the power to merge it. :sweat_smile:  )

skonto:
@nak3 will do the two PRs.

dprotaso:
/retest

@skonto is there anything blocking this change?

skonto:
@dprotaso no just need some cycles to update. Will do.

skonto:
Actually there was a mistake above the default reporting periods are: Prometheus: sec, opencensus 1min

skonto:
```
2022/07/27 13:06:48 http: TLS handshake error from 127.0.0.1:54210: tls: no certificates configured
--- FAIL: TestConversionValidResponse (0.21s)
    logger.go:130: 2022-07-27T13:06:48.345Z	ERROR	webhook/webhook.go:148	failed to fetch secret	{"error": "secret \"webhook-certs\" not found"}
    conversion_integration_test.go:117: Failed to get response Get "https://0.0.0.0:34381/bazinga": remote error: tls: unrecognized name
    logger.go:130: 2022-07-27T13:06:48.345Z	INFO	webhook/webhook.go:240	Starting to fail readiness probes...
```

skonto:
Will retest this passes locally:

> metrics: TestMetricsExport/OpenCensus expand_less	5s
> {Failed  === RUN   TestMetricsExport/OpenCensus
>     e2e_test.go:194: Created exporter at localhost:12345
> {"level":"info","ts":1658927027.88901,"logger":"fallback","caller":"metrics/opencensus_exporter.go:56","msg":"Created OpenCensus exporter with config:","config":{}}
>     e2e_test.go:232: Timeout reading input
>     e2e_test.go:238: Unexpected OpenCensus exports (-want +got):
>           []metrics.metricExtract(Inverse(Sort, []string{
>           	"knative.dev/project/testComponent/global_export_counts<>:2",
>           	"knative.dev/project/testComponent/resource_global_export_count<>:2",
>           	`knative.dev/project/testComponent/testing/value<project="p1",rev`...,
>         - 	`knative.dev/project/testComponent/testing/value<project="p1",revision="r2">:1`,
>           }))
>     --- FAIL: TestMetricsExport/OpenCensus (5.05s)
> }

skonto:
Tekton tests need an update downstream :shrug: :
```
# github.com/tektoncd/pipeline/pkg/pod
Error: pkg/pod/pod.go:293:18: assignment mismatch: 2 variables but changeset.Get returns 1 value
```

dprotaso:
/retest 

you're hitting flakes

skonto:
@dprotaso gentle ping.

metrics/config_observability.go
```diff
@@ -127,3 +139,27 @@ func ConfigMapName() string {
 	}
 	return "config-observability"
 }
+
+// Use the same as `metrics.reporting-period-seconds` for the default
+// of `metrics.request-metrics-reporting-period-seconds`
+func getDefaultRequestMetricsReportingPeriod(data map[string]string) (*int, error) {
```
dprotaso, line 145: Don't think it's necessary to return a `*int` - I think an `int` would suffice since we always return a default value

dprotaso (commented):
minor nit
/test all

skonto:
@dprotaso I think this is ready.

PR #2365, opened 2021-11-29
Add ConfigMap field parsing for int16 and uint16 types

pierDipi:
- Support ConfigMap field parsing for int16 and uint16 types


markusthoemmes (approved):
/lgtm
/approve

PR #2366, opened 2021-11-29
Drop unused test/cmd package

markusthoemmes:
# Changes

I haven't found any usage of this package through knative or knative-sandbox and it gets rid of a weird dep... so yay?

/assign @julz @dprotaso


dprotaso:
cc @cardil - I recall you were looking to replace some bash scripts with Go - is there any utility in keeping this?

cc @kvmware

cardil:
+1 from me for removal. This isn't related to what we should be doing with knative/hack ie. we shouldn't shell out left and right. Also, there are already packages that can do that better.

PR #2377, opened 2021-12-14
Tidy helper to avoid need for nosec comment

julz:
Currently we fmt.Sprintf a command string only to immediately split it
back in to args, which causes gosec to complain about unsanitised inputs
to exec.Command. Refactoring this to directly pass an args array is
clearer, more secure, and avoids needing the nosec comment.

### DIFF
--- a/test/monitoring/monitoring.go
+++ b/test/monitoring/monitoring.go
@@ -74,21 +74,11 @@ func Cleanup(pid int) error {
 // PortForward sets up local port forward to the pod specified by the "app" label in the given namespace
 func PortForward(logf logging.FormatLogger, podList *v1.PodList, localPort, remotePort int, namespace string) (int, error) {
 	podName := podList.Items[0].Name
-	portFwdProcess, err := executeCmdBackground(logf, "kubectl", "port-forward", podName, fmt.Sprintf("%d:%d", localPort, remotePort), "-n", namespace)
-	if err != nil {
+	cmd := exec.Command("kubectl", "port-forward", podName, fmt.Sprintf("%d:%d", localPort, remotePort), "-n", namespace)
+	if err := cmd.Start(); err != nil {
 		return 0, fmt.Errorf("failed to port forward: %w", err)
 	}
 
-	logf("Running %s port-forward in background, pid = %d", podName, portFwdProcess.Pid)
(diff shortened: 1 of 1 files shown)


markusthoemmes (commented):
Love it ❤️

test/monitoring/monitoring.go
```diff
@@ -86,12 +83,10 @@ func PortForward(logf logging.FormatLogger, podList *v1.PodList, localPort, remo
 	return portFwdProcess.Pid, nil
 }
 
-// RunBackground starts a background process and returns the Process if succeed
-func executeCmdBackground(logf logging.FormatLogger, format string, args ...interface{}) (*os.Process, error) {
-	cmd := fmt.Sprintf(format, args...)
+// executeCmdBackground starts a background process and returns the Process if succeed
+func executeCmdBackground(logf logging.FormatLogger, cmd string, args ...string) (*os.Process, error) {
```
markusthoemmes, line 87: If this is really only used in that one place, maybe we should just drop the function at this point? The wrapper is **really** thin.

julz: yeah, let me see the downstream tests pass with this and I'll tidy up a bit more

julz: hm actually I was hoping the downstream would run the eventing e2es with this for me, but looks like they don't anyway, bah

test/monitoring/monitoring.go
```diff
@@ -75,9 +74,7 @@ func Cleanup(pid int) error {
 // PortForward sets up local port forward to the pod specified by the "app" label in the given namespace
 func PortForward(logf logging.FormatLogger, podList *v1.PodList, localPort, remotePort int, namespace string) (int, error) {
 	podName := podList.Items[0].Name
```
markusthoemmes, line 76: FWIW, totes not your fault but because I've seen this now: This interface is quite awkward. Why pass a whole podList if all you want is just the name of the first pod anyway 🤔 .

julz: heh, very good question! Let's tidy that up separately, though

julz:
test in eventing looks good: https://github.com/knative/eventing/pull/5994

markusthoemmes (approved):
/lgtm
/approve

PR #2378, opened 2021-12-14
Add nil checks for Source types validation to be pointer receiver based

pierDipi:
It's better to have nil checks inside the Validate method of the
type itself instead of the type that holds a field of that type
because this allows using the types in other contexts without
the types that hold a field.

/kind enhancement


pierDipi:
```
=== RUN   TestSpoofingClient_WaitForEndpointState/Non_matching_response_triggers_more_requests
    spoof_test.go:198: Expected Transport to be invoked 3 time but got invoked 2
    --- FAIL: TestSpoofingClient_WaitForEndpointState/Non_matching_response_triggers_more_requests (0.00s)
```
/retest

matzew (approved):
/lgtm
/approve

PR #2379, opened 2021-12-15
using flags, since we already use that via ParseAndGetRESTConfigOrDie

matzew:

# Changes

- as per title
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -24,9 +24,8 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 	"time"
-
-	"github.com/spf13/pflag"
 
 	"go.uber.org/automaxprocs/maxprocs" // automatically set GOMAXPROCS based on cgroups
 	"go.uber.org/zap"
@@ -120,13 +119,12 @@ var (
 // by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
-	disabledControllers := pflag.StringSlice("disable-controllers", []string{}, "Comma-separated list of disabled controllers.")
+	disabledControllers := flag.String("disable-controllers", "", "Comma-separated list of disabled controllers.")
 
 	// HACK: This parses flags, so the above should be set once this runs.
 	cfg := injection.ParseAndGetRESTConfigOrDie()
 
-	enabledCtors := enabledControllers(*disabledControllers, ctors)
-
+	enabledCtors := enabledControllers(strings.Split(*disabledControllers, ","), ctors)
 	MainWithConfig(ctx, component, cfg, toControllerConstructors(enabledCtors)...)
 }
 


injection/sharedmain/main.go
```diff
@@ -58,6 +57,7 @@ import (
 
 func init() {
 	maxprocs.Set()
+	flag.StringVar(&disabledControllers, "disable-controllers", "", "Comma-separated list of disabled controllers.")
```
pierDipi, line 60: Can we define this inside the `MainNamed` function?

If a controller is not using MainNamed and there is a `disable-controllers` flag defined we would get an error of undefined flag.
With a global flag, we silently define the flag for every importer.

injection/sharedmain/main.go
```diff
@@ -112,6 +111,7 @@ func Main(component string, ctors ...injection.ControllerConstructor) {
 var (
 	WebhookMainWithContext = MainWithContext
 	WebhookMainWithConfig  = MainWithConfig
+	disabledControllers    string
```
pierDipi, line 114: This can be function local to `MainNamed`

injection/sharedmain/main.go
```diff
@@ -120,12 +120,11 @@ var (
 // by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
-	disabledControllers := pflag.StringSlice("disable-controllers", []string{}, "Comma-separated list of disabled controllers.")
-
 	// HACK: This parses flags, so the above should be set once this runs.
 	cfg := injection.ParseAndGetRESTConfigOrDie()
+	flag.StringVar(&disabledControllers, "disable-controllers", "", "Comma-separated list of disabled controllers.")
```
pierDipi, line 125: This should be before `ParseAndGetRESTConfigOrDie`

pierDipi: Also
```suggestion
	disabledControllers := flag.String("disable-controllers", "", "Comma-separated list of disabled controllers.")
```

injection/sharedmain/main.go
```diff
@@ -120,12 +119,14 @@ var (
 // by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
-	disabledControllers := pflag.StringSlice("disable-controllers", []string{}, "Comma-separated list of disabled controllers.")
-
+	var (
+		disabledControllers string
+	)
 	// HACK: This parses flags, so the above should be set once this runs.
```
pierDipi, line 125: This comment should be before like 127 `cfg := injection.ParseAndGetRESTConfigOrDie()`

injection/sharedmain/main.go
```diff
@@ -120,12 +119,14 @@ var (
 // by name.
 func MainNamed(ctx context.Context, component string, ctors ...injection.NamedControllerConstructor) {
 
-	disabledControllers := pflag.StringSlice("disable-controllers", []string{}, "Comma-separated list of disabled controllers.")
-
+	var (
+		disabledControllers string
+	)
+	flag.StringVar(&disabledControllers, "disable-controllers", "", "Comma-separated list of disabled controllers.")
```
pierDipi, line 125:
```suggestion
	disabledControllers := flag.String("disable-controllers", "", "Comma-separated list of disabled controllers.")
```

injection/sharedmain/main.go
```diff
@@ -120,12 +119,14 @@ var (
...
 	// HACK: This parses flags, so the above should be set once this runs.
 	cfg := injection.ParseAndGetRESTConfigOrDie()
 
-	enabledCtors := enabledControllers(*disabledControllers, ctors)
+	enabledCtors := enabledControllers(strings.Split(disabledControllers, ","), ctors)
```
julz, line 129: Im not sure I understand why we want to do the split ourselves here rather than using the pflag.StringSlice library which seems more concise / avoids the extra local var, and is a perfectly valid flag.Var?

pierDipi: pflag used only for this flag doesn't work.

matzew: yeah, was not working, therefore changing :man_shrugging:

julz: I don't follow, isn't pflag just a standard flag.Var so should work fine with flag library? 🤔 Is there a failing test or something for this

julz: (I'm probably missing something obvious about flag/pflag, but I'd love to understand before we merge it why this doesn't work)

pierDipi: Maybe we need this import everywhere we use `flag`?
```golang
import flag "github.com/spf13/pflag"
```

julz: ahh I looked at the code - seems like pflag.* adds the var to its own flag set, not the one package flag uses, and it doesn't expose the internal flag.Var externally afaict, annoyingly

matzew:
/close

@julz I will revisit this, and come back w/ more details, or a different PR

thanks for looking

PR #2389, opened 2022-01-12
Simplify URI creation in source validation tests

pierDipi:
As per title.

Replace:
```golang
				URI: func() *apis.URL {
					u, _ := apis.ParseURL("https://localhost")
					return u
				}(),
```
with:
```golang
				URI: apis.HTTP("localhost"),
```


evankanderson (commented):
/lgtm
/approve

PR #2390, opened 2022-01-12
adds WithNamespace ability to contexts

gab-satchi:
# Changes

- :bug: Fixes issue where namespace can be misrepresented in the context. Depending on the client, namespace may not be present in the metadata. With these changes, we attach the namespace directly from the incoming request. 

/kind bug

Fixes part of the issue raised in knative/eventing: [#4514](https://github.com/knative/eventing/issues/4514)

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/apis/contexts.go
+++ b/apis/contexts.go
@@ -141,6 +141,22 @@ func IsWithinParent(ctx context.Context) bool {
 	return ok
 }
 
+type namespaceKey struct{}
+// WithNamespace attaches the request's namespace. Intended to be used
+// where defaulting depends on the namespace (eventing channels, brokers)
+// and the client may not provide it in the resource ObjectMeta
+func WithNamespace(ctx context.Context, ns string) context.Context {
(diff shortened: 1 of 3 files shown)


gab-satchi:
I didn't modify one location where the parentMeta is still used: in [knative_reference](https://github.com/knative/pkg/blob/main/apis/duck/v1/knative_reference.go#L107-L112). I'm unsure where else this reference may be getting used and whether it should also be modified in this PR.

dprotaso:
Looking at the original issue - is this not a k8s bug? I'd expected namespaced resources hitting the webhook to have valid metadata

dprotaso:
/hold 

cc @shinigambit looks like there's an issue with the merge upstream action for non-members of the org?

gab-satchi:
> Looking at the original issue - is this not a k8s bug? I'd expected namespaced resources hitting the webhook to have valid metadata

Yeah I do think it's better solved in k8s. I'll investigate there

Issue #2383, opened 2022-01-03
LeaderElection defaults too aggressive?


evankanderson:
## Expected Behavior

After installing, other operations on my cluster work as before.

## Actual Behavior

I sometimes get the following message:

```
I0103 14:16:13.438479   52363 request.go:668] Waited for 1.045323069s due to client-side throttling, not priority and fairness, request: GET:https://192.168.1.205:6443/apis/networking.antrea.tanzu.vmware.com/v1beta1?timeout=32s
```

According to [the Kubernetes documentation](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/), there is are `--max-requests-inflight` and `--max-mutating-requests-inflight` settings that limit the total number of API requests in flight. It looks like adding abut 50 leases which are renewed every 2s may have pushed me over that limit.  (This is a cluster which also has other workloads on it.)

## Steps to Reproduce the Problem

1. Enable audit logging on the cluster (apiserver flag). This is probably more likely to trigger on a cluster with fewer apiserver processes.
2. Load Knative onto an existing cluster (I have a [FluxCD](https://fluxcd.io/) deployment already on the cluster along with Rook and FluentBit).
3. Count occurrences of `"requestURI"` that contain `/apis/coordination.k8s.io/v1/namespaces/knative-serving/leases/`

## Additional Info

It seems like we might want different "classes" of leases for serving-path (autoscaler, serverlessservice) and control-plane components (the default). Looking at `kube-node-lease` as an example that is updated less often, it appears that they use a 10s "refresh interval" and a 40s lease duration. This would reduce the lease traffic by a factor of 5 over our defaults (2s / 15s), with a fairly similar safety margin in terms of lease renewal opportunities.

evankanderson:
https://github.com/knative/serving/pull/12499 indicates that the following defaults wouldn't break serving tests:

```yaml
data:
  lease-duration: "60s"
  renew-deadline: "40s"
  retry-period: "10s"
```

PR #2391, opened 2022-01-12
Adjust leaderelection defaults

evankanderson:
Fixes: #2383

Also adds documentation for the `leaderelection` package, because I went
looking and found it wasn't there.

Verified that these settings do not seem to interfere with chaos duck integration tests in serving: https://github.com/knative/serving/pull/12499

/assign @mattmoor @dprotaso


mattmoor (commented):
LGTM, will defer to @dprotaso to stamp

dprotaso:
/lgtm
/approve

Looks like there's an unnecessary downstream (knative.dev/serving) test asserting the defaults in this package you'll want to remove

PR #2402, opened 2022-01-26
:bug: `TestDialWithBackoff` work without special environment

### LINKED ISSUE #2401: `TestDialWithBackoff` from `network/transports_test.go` fails by not recognizing expected error outside of CI
## Expected Behavior

Test should pass without errors on any environment.

## Actual Behavior

Test fails with:

```shell
$ gotestsum -f testname -- -race -count=1 -run 'TestDialWithBackoff' ./network
=== RUN   TestDialWithBackoff
    transports_test.go:91: Error = dial tcp4 198.18.0.254:8888: connect: network is unreachable, want: timed out dialing(...)
--- FAIL: TestDialWithBackoff (0.00s)
FAIL network.TestDialWithBackoff (0.00s)
FAIL network

=== Failed
=== FAIL: network TestDialWithBackoff (0.00s)
    transports_test.go:91: Error = dial tcp4 198.18.0.254:8888: connect: network is unreachable, want: timed out dialing(...)

DONE 1 tests, 1 failure in 0.489s
```

1. `gotestsum -f testname -- -race -count=1 -run 'TestDialWithBackoff' ./network`

## Additional Info

The test is using a special address `198.18.0.254:8888` that looks to work on the CI environment. I think a good idea would be to refactor tests to work on every environment, cutting this external dependency.

cardil:
# Changes

- :bug: `TestDialWithBackoff` work without special environment

/kind bug

Fixes #2401

### DIFF
--- a/network/transports.go
+++ b/network/transports.go
@@ -55,6 +55,9 @@ var backOffTemplate = wait.Backoff{
 	Steps:    15,
 }
 
+// ErrTimeoutDialing when the timeout is reached after set amount of time.
+var ErrTimeoutDialing = errors.New("timed out dialing")
+
 // DialWithBackOff executes `net.Dialer.DialContext()` with exponentially increasing
 // dial timeouts. In addition it sleeps with random jitter between tries.
 var DialWithBackOff = NewBackoffDialer(backOffTemplate)
@@ -110,7 +113,7 @@ func dialBackOffHelper(ctx context.Context, network, address string, bo wait.Bac
 		return c, nil
 	}
 	elapsed := time.Since(start)
-	return nil, fmt.Errorf("timed out dialing after %.2fs", elapsed.Seconds())
+	return nil, fmt.Errorf("%w after %.2fs", ErrTimeoutDialing, elapsed.Seconds())
 }
 
 func newHTTPTransport(disableKeepAlives, disableCompression bool, maxIdle, maxIdlePerHost int) http.RoundTripper {
--- a/network/transports_test.go
+++ b/network/transports_test.go
@@ -111,8 +111,20 @@ func testDialWithBackoffConnectionRefused(tlsConf *tls.Config) func(t *testing.T
 
 func testDialWithBackoffTimeout(tlsConf *tls.Config) func(t *testing.T) {
 	return func(t *testing.T) {
-		// Timeout. Use non-routable IP. See: https://stackoverflow.com/a/31581323/844449
-		c, err := dialer(context.TODO(), tlsConf)("10.0.0.0:81")
+		// Create a listening socket with backlog 1, then occupy the backlog to force a timeout.
+		closer, addr, err := listenOne()
+		if err != nil {
+			t.Fatal("Unable to create listener:", err)
+		}
+		defer closer()
+		c1, err := net.Dial("tcp4", addr.String())
+		if err != nil {
+			t.Fatalf("Unable to connect to server on %s: %s", addr, err)
+		}
+		defer c1.Close()
+
+		// Since the backlog is full, the next request must time out.
+		c, err := dialer(context.TODO(), tlsConf)(addr.String())
 		if err == nil {
 			closeOrFail(t, c)
 			t.Error("Unexpected success dialing")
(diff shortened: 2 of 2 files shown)


cardil:
The unit test failures [1486341657388388352](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/2402/pull-knative-pkg-unit-tests/1486341657388388352) look unrelated:

network/transports.go
```diff
@@ -54,6 +54,9 @@ var backOffTemplate = wait.Backoff{
 	Steps:    15,
 }
 
+// ErrTimeoutDialing when the timeout is reached after set amount of time.
+var ErrTimeoutDialing = errors.New("timed out dialing")
```
dprotaso, line 58: Rather than introduce a new error maybe we can return `context.DeadlineExceeded`

dprotaso (changes requested):

network/transports_test.go
```diff
@@ -68,35 +74,77 @@ func TestHTTPRoundTripper(t *testing.T) {
...
+func testDialWithBackoffTimeout(t *testing.T) {
 	// Make the test short.
 	bo := backOffTemplate
-	bo.Steps = 2
+	bo.Steps = 1
+	bo.Duration = time.Millisecond
 
-	// Timeout. Use special testing IP address.
-	c, err = dialBackOffHelper(context.Background(), "tcp4", "198.18.0.254:8888", bo, sleepTO)
+	// Timeout. Use non-routable IP. See: https://stackoverflow.com/a/31581323/844449
```
evankanderson, line 103: One of the [documentation IP ranges](https://datatracker.ietf.org/doc/html/rfc5737#section-3) (or a localhost IP like `127.200.100.0`) might make more sense than private IP space (and all are better than some real IP address belonging to Oracle, so thanks for fixing that!).

cardil: Localhost's range isn't a good idea. We may hit, by chance, the port being open.

network/transports_test.go
```diff
@@ -68,35 +74,77 @@ func TestHTTPRoundTripper(t *testing.T) {
...
-	// Nobody's listening on a random port. Usually.
-	c, err := DialWithBackOff(context.Background(), "tcp4", "127.0.0.1:41482")
-	if err == nil {
-		c.Close()
-		t.Error("Unexpected success dialing")
+	t.Parallel()
+	t.Run("ConnectionRefused", testDialWithBackoffConnectionRefused)
+	t.Run("Timeout", testDialWithBackoffTimeout)
+	t.Run("Success", testDialWithBackoffSuccess)
+}
```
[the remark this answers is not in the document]
cardil: In my opinion, why I wrote it like this, it's more organized. In reports as well.

The subtests could be executed as easily as root level ones. And we can ensure the test will run fast thanks to parallelization.

network/transports.go
cardil: This error type is introduced to not change the final error message. See the change in line 97:

```golang
return nil, fmt.Errorf("%w after %.2fs", ErrTimeoutDialing, elapsed.Seconds())
```

By introducing this error, consumers of this package could properly do the error checking using types:


```golang
if errors.Is(err, network.ErrTimeoutDialing) {
  // do some
}
```

instead of current dodgy string checking:


```golang
if string.Contains(err.Error(), "timed out dialing")) {
  // do some
}
```

network/transports_test.go
```diff
@@ -68,35 +74,77 @@ func TestHTTPRoundTripper(t *testing.T) {
...
+
+func findUnusedPortOrFail(tb testing.TB) int {
+	tb.Helper()
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	if err != nil {
+		tb.Fatal(err)
+	}
+
+	l, err := net.ListenTCP("tcp", addr)
```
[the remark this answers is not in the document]
cardil: Done

network/transports_test.go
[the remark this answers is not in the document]
cardil: Also, now those cases are reused by regular and TLS version of the test.

evankanderson:
Rather than using a likely-non-routable IP, how about using `Listen`'s backlog to force a timeout? (This is similar to the localhost IP range suggestion, but more likely to consistently work)

https://github.com/cardil/knative-pkg/pull/1

Note that both this PR and the current state in main don't work on Windows, with similar errors:

```
--- FAIL: TestDialWithBackoff (0.17s)
    --- FAIL: TestDialWithBackoff/ConnectionRefused (0.09s)
        transports_test.go:107: Unexpected error: timed out dialing after 0.09s
--- FAIL: TestDialTLSWithBackoff (0.19s)
    --- FAIL: TestDialTLSWithBackoff/ConnectionRefused (0.09s)
        transports_test.go:107: Unexpected error: timed out dialing after 0.09s
```

As a side note, putting all of these in one test makes it harder (for example) to run just the `DialWithBackoffTimout` test from VSCode -- you can click on "run test" above methods named `TestFoo` to run just that test, but in this case, you have to run the `TestDialWithBackoff` tests as a whole.

cardil:
> Rather than using a likely-non-routable IP, how about using `Listen`'s backlog to force a timeout? [cardil#1](https://github.com/cardil/knative-pkg/pull/1)

Interesting. I'll continue on this next week.

> Note that both this PR and the current state in main don't work on Windows, with similar errors

I did run those tests multiple times, but apparently they also failed on Prow. It looks to me like your solution could help. Thanks! I'll take a look next week.

> As a side note, putting all of these in one test makes it harder to run just one test

That's a bummer. I use Goland, and there I could easily execute those subtests. VSCode should also have this feature.

Screenshot from 2022-04-22 22-40-59

evankanderson (commented):
/lgtm
/approve

PR #2414, opened 2022-02-01
Using flags instead of pflag 

matzew:

# Changes

- coming back to this, we must use `flags` instead of `pflags`, since this is not working. It  seems like pflag.* adds the var to its own flag set, not the one package flag uses, and it doesn't expose the internal flag.Var externally - hence this fix
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


pierDipi (approved):
/lgtm

PR #2423, opened 2022-02-06
allow overriding the default grace period of 45 seconds

tzneal:
This allows users to configure a faster restart of their
webhook if desired while retaining the current behavior.

# Changes

- :gift: allow users to change the webhook drain timeout

/kind enhancement

**Release Note**

```release-note
webhook.Options now contains a GracePeriod parameter to allow modifying the http server drain time from the default of 45 seconds.
```

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -34,7 +34,6 @@ import (
 	"golang.org/x/sync/errgroup"
 	admissionv1 "k8s.io/api/admission/v1"
 	"knative.dev/pkg/logging"
-	"knative.dev/pkg/network"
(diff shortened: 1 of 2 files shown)


webhook/webhook.go
```diff
@@ -126,7 +125,6 @@ func New(
 		Options:     *opts,
 		Logger:      logger,
 		synced:      cancel,
-		gracePeriod: network.DefaultDrainTimeout,
```
pierDipi, line 129: We should set to `opts.GracePeriod` the previous default value `network.DefaultDrainTimeout` when `opts.GracePeriod` isn't set.

PR #2425, opened 2022-02-11
changeset: Look in packed-refs if ref file not found.

wlynch:
# Changes

🎁  changeset: Look in packed-refs file if ref file not found.

If you have a lot of refs in a repo, git will eventually start
condensing them into a single file instead of letting them sprawl out in
separate directories on the filesystem. This change lets the changeset
package use the packed-refs file as a fall back in the event that the
ref file is not found.

See https://git-scm.com/docs/git-pack-refs for more details on packed
refs.

/kind feature

Bumped into this while working on some Tekton PRs - the refs got stuffed into `packed-refs` and calls relying on `changeset.Get()` started failing / causing log noise because the `refs/heads/main` file no longer existed. Seems small enough that it didn't warrant its own issue, but happy to open one if desired!

**Release Note**

```release-note
* changeset now supports reading from Git `packed-refs` files.
```

**Docs**

 n/a


wlynch:
Did some testing - runtime/debug probably won't be super useful until 1.18, particularly for local changes.

The main module info for something built not at a semver tag looks like:

```
Main:{Path:knative.dev/pkg Version:(devel) Sum: Replace:<nil>}
```

What we really want are the [vcs tags included in 1.18](https://tip.golang.org/doc/go1.18#go-command) which look like:

```
Settings:[
 {Key:-compiler Value:gc}
 {Key:CGO_ENABLED Value:1}
 {Key:CGO_CFLAGS Value:}
 {Key:CGO_CPPFLAGS Value:} 
 {Key:CGO_CXXFLAGS Value:}
 {Key:CGO_LDFLAGS Value:} 
 {Key:GOARCH Value:amd64} 
 {Key:GOOS Value:linux} 
 {Key:GOAMD64 Value:v1} 
 {Key:vcs Value:git} 
 {Key:vcs.revision Value:62b1122c86b2f032117da832332037ff511c2b74} 
 {Key:vcs.time Value:2022-02-11T23:48:33Z} 
 {Key:vcs.modified Value:true}
]
```

Issue #2436, opened 2022-02-22
Controllers (and related) should relinquish their leases when shutting down


dprotaso:
## Expected Behavior

When a controller is holding various K8s leases it should clear the `holderIdentity` when shutting down.

## Actual Behavior

The controller process doesn't wait for the leases to be released before exiting. The resulting behaviour is non-deterministic but the cascading of errors have been more pronounced with GKE 1.21 and the fact that we [increased our default lease duration](https://github.com/knative/pkg/pull/2391).

For example the API audit logs will show `leases.update` with a `Gateway Timeout` status. Serving has seen upgrade tests become flaky because when one controller shuts down there's a wait for the new controller to start reconciliation of resources.

Note: this isn't limited to controllers but any component that uses `sharedmain`

PR #2437, opened 2022-02-22
Wait for reconciler/controllers to return prior to exiting the sharedmain process

dprotaso:
Fixes: https://github.com/knative/pkg/issues/2436

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"golang.org/x/sync/errgroup"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
@@ -440,6 +441,11 @@ func (c *Impl) EnqueueKeyAfter(key types.NamespacedName, delay time.Duration) {
 	}
 }
 
+// Run runs the controller with it's configured Concurrency
+func (c *Impl) Run(ctx context.Context) error {
+	return c.RunContext(ctx, c.Concurrency)
+}
+
 // RunContext starts the controller's worker threads, the number of which is threadiness.
 // If the context has been decorated for LeaderElection, then an elector is built and run.
 // It then blocks until the context is cancelled, at which point it shuts down its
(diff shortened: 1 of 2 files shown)


injection/sharedmain/main.go
```diff
@@ -266,7 +266,9 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 		wh.InformersHaveSynced()
 	}
 	logger.Info("Starting controllers...")
-	go controller.StartAll(ctx, controllers...)
+	eg.Go(func() error {
+		return controller.StartAll(ctx, controllers...)
+	})
```
dprotaso, line 271: note: the other changes are just a minor clean up. Running this in an error group which `Wait`s below is the fix

dprotaso:
Looks like eventing keda was using the deprecated method - https://github.com/knative-sandbox/eventing-autoscaler-keda/pull/243

controller/controller.go
```diff
@@ -778,18 +771,17 @@ func WaitForCacheSyncQuick(stopCh <-chan struct{}, cacheSyncs ...cache.InformerS
...
-		wg.Add(1)
-		concurrency := ctrlr.Concurrency
-		go func(c *Impl) {
-			defer wg.Done()
-			c.RunContext(ctx, concurrency)
-		}(ctrlr)
+	for _, controller := range controllers {
+		c := controller
+		eg.Go(func() error {
+			return c.Run(egCtx)
```
nader-ziada, line 781: can't you run the `c.RunContext` directly?

dprotaso: no - err group doesn't pass through a `ctx` to the function - otherwise we could

nader-ziada:
looks good, but tests failing

dprotaso:
The failures are known flakes/unrelated - excluding the `keda` change

nader-ziada:
> The failures are known flakes/unrelated - excluding the `keda` change

I was just looking at the `keda` one, can we ignore that?

dprotaso:
I created a PR fix which is linked above

nader-ziada:
/retest

for the flaky knaitve one

PR #2445, opened 2022-02-28
Use GetAny from bucket map instead of sorting the list 

### LINKED ISSUE #2364: root cause failures caused by https://github.com/knative/pkg/pull/2360
The PR https://github.com/knative/pkg/pull/2360 seems to have broke a lot of things downstream. Given that the breakage was non-obvious (see: https://github.com/knative/pkg/pull/2360#issuecomment-979295081) it would be good to figure out why the change was disruptive - and potentially create a test to prevent this in the future.

nader-ziada:
# Changes

- Use getany item from bucket map instead of sorting the list

/kind enhancement

Fixes #2364 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/hash/bucketer.go
+++ b/hash/bucketer.go
@@ -98,10 +98,20 @@ func (bs *BucketSet) Owner(key string) string {
 	}
 	bs.mu.RLock()
 	defer bs.mu.RUnlock()
-	l := ChooseSubset(bs.buckets, 1 /*single query wanted*/, key)
-	ret := l.UnsortedList()[0]
-	bs.cache.Add(key, ret)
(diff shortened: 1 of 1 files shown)


nader-ziada:
I think the problem with using `PopAny` is that it would [delete](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go#L191) the item it returned, which resulted in the following calls to this func getting empty set of buckets which made the controller mess up (somehow) the LeaderAwareFuncs and the reconciler would not find any items. Replaced the PopAny with a GetAny that doesn't delete and testing with serving

nader-ziada:
tried the same change in serving https://github.com/knative/serving/pull/12692 and all e2e passed

PR #2460, opened 2022-03-11
Take pointer to map in AsOptionalMap for consistency

pierDipi:
- Other `As*` functions take pointers.
- This avoids having to initialize the map passed to the function

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

### DIFF
--- a/configmap/parse.go
+++ b/configmap/parse.go
@@ -247,15 +247,19 @@ func Parse(data map[string]string, parsers ...ParseFunc) error {
 
 // AsOptionalMap parses the data into the target as a map[string]string, if it exists.
 // The map is represented as a list of key-value pairs with a common prefix.
-func AsOptionalMap(prefix string, target **map[string]string) ParseFunc {
+func AsOptionalMap(prefix string, target *map[string]string) ParseFunc {
+	if target == nil {
+		panic("target cannot be nil")
+	}
+
 	return func(data map[string]string) error {
 		for k, v := range data {
 			if strings.HasPrefix(k, prefix) && len(k) > len(prefix)+1 {
 				if *target == nil {
 					m := make(map[string]string, 2)
-					*target = &m
+					*target = m
 				}
-				(**target)[k[len(prefix)+1: /* remove dot `.` */]] = v
+				(*target)[k[len(prefix)+1: /* remove dot `.` */]] = v
 			}
 		}
 		return nil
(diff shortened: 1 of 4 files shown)


pierDipi:
@nak3 with this we don't have to update serving or networking.

/cc @dprotaso @nak3

configmap/parse.go
```diff
@@ -247,11 +247,15 @@ func Parse(data map[string]string, parsers ...ParseFunc) error {
 
 // AsOptionalMap parses the data into the target as a map[string]string, if it exists.
 // The map is represented as a list of key-value pairs with a common prefix.
-func AsOptionalMap(prefix string, target map[string]string) ParseFunc {
+func AsOptionalMap(prefix string, target **map[string]string) ParseFunc {
```
dprotaso, line 250: So the reason we have `AsOptional*` helpers is for when the null value have some meaning.  ie. bool - null is the default (set by the operator), vs. an explicit `true` or `false`

But I don't think we need this for a map - since nil/empty maps behave the same when doing safe guarded key lookups (ie. `v, ok := m[k]`)

Thus I think we can drop the double pointer.

My other suggestion is making the function name more explicit that we're doing this special prefix parsing. `AsMap` implies we are taking a string literal and turn it into a go map. (similar to the other `As[Type]` functions)

Maybe `CollectMapEntriesWithPrefix`?

```suggestion
func CollectMapEntriesWithPrefix(prefix string, target *map[string]string) ParseFunc {
```

pierDipi: or `CollectEntriesStartingWith`?

pierDipi: Done

configmap/parse.go
```diff
@@ -245,13 +245,21 @@ func Parse(data map[string]string, parsers ...ParseFunc) error {
...
 }
 
-// AsOptionalMap parses the data into the target as a map[string]string, if it exists.
+// CollectMapEntriesWithPrefix parses the data into the target as a map[string]string, if it exists.
 // The map is represented as a list of key-value pairs with a common prefix.
-func AsOptionalMap(prefix string, target map[string]string) ParseFunc {
+func CollectMapEntriesWithPrefix(prefix string, target *map[string]string) ParseFunc {
+	if target == nil {
+		panic("target cannot be nil")
+	}
```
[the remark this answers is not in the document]
pierDipi:
```golang
Parse(data,
  CollectMapEntriesWithPrefix("prefix", nil), // panic here vs panic when the "Parse" loop runs
)
```
it seems a nit :smile:

PR #2471, opened 2022-03-22
Fix coverage settings

### LINKED ISSUE #2457: Coverage CI check is failing
As described in this [comment](https://github.com/knative/pkg/pull/2362#issuecomment-1061832852) coverage check is failing. Some dirs could be omitted and others improved?

skonto:
Fixes #2457. Looking at the report [here](https://app.codecov.io/gh/knative/pkg/) a few folders have low coverage leading to a low coverage overall. This PR [ignores them](https://docs.codecov.com/docs/ignoring-paths) so that CI run against PRs dont fail, however coverage should pass the threshold for the ones that are not generated code eg. tracing, injection.

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -18,6 +18,10 @@ comment:
   behavior: default
 ignore:
   - "**/zz_generated*.go" # Ignore generated files.
+  - "codegen/**/*" #TODO: fix coverage
+  - "injection/**/*" #TODO: fix coverage
+  - "test/**/*"
+  - "tracing/**/*" #TODO: fix coverage
   - "client"
   - "hack"
   - "reconciler/testing"


.codecov.yaml
```diff
@@ -18,6 +18,10 @@ comment:
   behavior: default
 ignore:
   - "**/zz_generated*.go" # Ignore generated files.
+  - "codegen/**/*" #TODO: fix coverage
+  - "injection/**/*" #TODO: fix coverage
+  - "test/**/*"
+  - "tracing/**/*" #TODO: fix coverage
```
dprotaso, line 24: We should make issues for these

.codecov.yaml
```diff
@@ -18,6 +18,10 @@ comment:
   behavior: default
 ignore:
```
dprotaso, line 19: another question - what does ignore actually do?

ie. it would be good to know the coverage for these packages and not ignore them.

We only want to ignore these packages when calculating the total coverage across this repo

.codecov.yaml
skonto: Will create those thanks.

PR #2472, opened 2022-03-25
Sets mutating webhook ReinvocationPolicy to IfNeeded

### LINKED ISSUE #1195: Mutating webhooks should set the ReinvocationPolicy to IfNeeded
Not sure what the default value is here, but we should make sure that our mutating webhooks are configured to get reinvoked until the mutations close.

cc @dprotaso

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

# Changes
Fixes #1195

tested this by patching the change into serving and verified that after deployment both mutating webhooks (domainmapping and serving) had the appropriate `reinvocationPolicy`

**Release Note**

```release-note
Mutating webhooks have ReinvocationPolicy set to IfNeeded, meaning the webhook may be called again as part of the admission evaluation if the object being admitted is modified by other admission plugins after the initial webhook call.

For more information, see: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy
```


psschwei:
Some of the downstream Knative failures seem to be needing to update Handler -> ProbeHandler (as part of the k8s version bump). 

Not sure about the Tekton ones

dprotaso:
/hold

Will unhold after pkg cut. `IfNeeded` might have an impact on performance that could affect some downstream e2e tests that are sensitive to timings

PR #2483, opened 2022-04-07
Fix `InitialBuckets()` for statefulSetBuilder's electors

tcnghia:
Some are for StatefulSet and only owns their own

# Changes

- :bug: Fix the bug where all `statefulSetBuilder`'s electors get promote to own the universal bucket.

**Release Note**

```release-note
NONE
```

**Docs**

```docs
NONE
```

### DIFF
--- a/leaderelection/context.go
+++ b/leaderelection/context.go
@@ -272,8 +272,12 @@ func (ue *unopposedElector) Run(ctx context.Context) {
 }
 
 func (ue *unopposedElector) InitialBuckets() []reconciler.Bucket {
+	bkt := ue.bkt
+	if bkt == nil {
+		bkt = reconciler.UniversalBucket()
+	}
 	return []reconciler.Bucket{
-		reconciler.UniversalBucket(),
+		bkt,
 	}
 }
 


leaderelection/context.go
```diff
@@ -272,8 +272,12 @@ func (ue *unopposedElector) Run(ctx context.Context) {
 }
 
 func (ue *unopposedElector) InitialBuckets() []reconciler.Bucket {
+	bkt := ue.bkt
+	if bkt == nil {
+		bkt = reconciler.UniversalBucket()
+	}
 	return []reconciler.Bucket{
-		reconciler.UniversalBucket(),
+		bkt,
```
dprotaso, line 280: I think you can simply return `ue.bkt` since the universal bucket is configured here:

https://github.com/knative/pkg/blob/8db11d0f9fedd8ac90e283e9d9b97148ebe64ac7/leaderelection/context.go#L101-L109

The only notable thing is you get a double enqueue of resources on startup - since the statefulset unopposed elector is configured with an enq function

https://github.com/knative/pkg/blob/8db11d0f9fedd8ac90e283e9d9b97148ebe64ac7/leaderelection/context.go#L219-L223

So I wonder if we should drop the extra promotion here:
https://github.com/knative/pkg/blob/8db11d0f9fedd8ac90e283e9d9b97148ebe64ac7/leaderelection/context.go#L269-L272

or pass `nil` for the enq function

tcnghia: I see, we could dedup simply by checking if the `enq` is `nil` in RunContext()

tcnghia: Ok based on what you both say here I am going to simplify the `InitialBuckets()` func as recommended and let 2x `la.Promote()` be since there is no side-effect.

tcnghia: @vaikas no there hasn't been a release cut since then. release cut is next week :)

tcnghia: @dprotaso I simplified the `InitialBuckets()` func as suggested.

PR #2485, opened 2022-04-07
bump our min k8s version to 1.22

dprotaso:


evankanderson (commented):
Do we need to update Prow configs as well?

/lgtm

dprotaso:
Docs update: https://github.com/knative/docs/pull/4913

dprotaso:
> Do we need to update Prow configs as well?

Prow defaults to `latest` in stable which is has been 1.22 for a while. Serving I believe is the only repo that pin's it's version but we bumped that to 1.22 a few weeks ago (https://github.com/knative/serving/pull/12758)

PR #2493, opened 2022-04-15
Ignore migration of CRDs that aren't installed

aavarghese:
Signed-off-by: aavarghese <avarghese@us.ibm.com>

For more info: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2089#issuecomment-1099512070

# Changes

- :broom: Change returning error when `Get` CRD fails to log error and return no error so that migration does not crash when CRDs are not installed in the system

/kind enhancement


apiextensions/storageversion/migrator.go
```diff
@@ -55,11 +56,13 @@ func NewMigrator(d dynamic.Interface, a apixclient.Interface) *Migrator {
...
 // versions
 func (m *Migrator) Migrate(ctx context.Context, gr schema.GroupResource) error {
+	logger := logging.FromContext(ctx)
 	crdClient := m.apixClient.ApiextensionsV1().CustomResourceDefinitions()
 
 	crd, err := crdClient.Get(ctx, gr.String(), metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("unable to fetch crd %s - %w", gr, err)
+		logger.Infof("Ignoring resource migration - unable to fetch crd %s - %w", gr, err)
+		return nil
```
pierDipi, line 65: The ignore should only happen when `apierrors.IsNotFound(err)` otherwise migration errors aren't visibles.

pierDipi: Is it possible to add a test?

aavarghese: Yes. Definitely. Thank you!

lionelvillard:
/lgtm

the failures seem to be unrelated to this PR.

pierDipi:
> I would update the downstream storage migration job definition to include only custom resource definitions that exist.
> 
> I worry that introducing this change would cause serious issues to fail silently. ie. I typed the name of my CRD incorrectly and I'm assuming the migration job is running successfully when it's not.

This means maintaining multiple jobs definition with the consequences of breaking users' workflow, in addition, once these jobs are created and tested changes are very rare so the risk of mistyping a resource is very low.

This is a problem with eventing core too, where `InMemoryChannel` is assumed to be there but it's not part of the core resources nor it's a resource that people use for production.
https://github.com/knative/eventing/blob/3e24a78037e9cefb5f81aeccc9e51855286b9eef/config/post-install/storage-version-migrator.yaml#L51

Can we make this behavior an opt-in with a flag?

dprotaso:
> This means maintaining multiple jobs definition with the consequences of breaking users' workflow

How is it breaking?

> This is a problem with eventing core too, where InMemoryChannel is assumed to be there but it's not part of the core resources nor it's a resource that people use for production.

If the in-memory channel is optional I would create an optional post-install job that migrates that CRDs storage version.

> Can we make this behavior an opt-in with a flag?

I'm ok with this being opt-in but be warned ;)

aavarghese:
@dprotaso @pierDipi thank you for the detailed comments.
I've added a new env var to opt-in to this ignore CRD migration change. Could you please TAL again? Thank you!

apiextensions/storageversion/migrator.go
```diff
@@ -38,6 +40,10 @@ type Migrator struct {
 	apixClient    apixclient.Interface
 }
 
+type MigrateConfig struct {
+	NoFailIfCRDFound bool `envconfig:"MIGRATE_CRD_IF_FOUND_NO_FAILURE" default:"false"`
+}
+
```
dprotaso, line 46: It's not obvious that consuming this API will read env-vars so let's add a `IgnoreNotFound` bool to the `Migrator` struct and do the lookup in the `main` of the cmd at https://github.com/knative/pkg/blob/45c37c266b74eed0c2b32f537220d5da7cdfe077/apiextensions/storageversion/cmd/migrate/main.go#L34

apiextensions/storageversion/migrator.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/kelseyhightower/envconfig"
```
dprotaso, line 23: envconfig seems overweight for just looking up a single env var 

It's probably simpler to use `os.Getenv` and `strconv.ParseBool`

aavarghese: Sure. Added e1741aed98464514cdf2d55b7a592eed6624ebc3

apiextensions/storageversion/migrator.go
aavarghese: @dprotaso ok to change the signature of [NewMigrator](https://github.com/knative/pkg/pull/2493/files#diff-9a19a801068606c877c96a2e53cac2bc639abcd220540f160aa90b582c0a6471R44) to add this new bool ?

apiextensions/storageversion/cmd/migrate/main.go
```diff
@@ -50,9 +56,15 @@ func main() {
 		logger.Fatal(err)
 	}
 
+	ignoreNotFound, err := strconv.ParseBool(os.Getenv(IgnoreNotFound))
+	if err != nil {
+		logger.Infof("Error getting environment variable %s for migrate %w", IgnoreNotFound, err)
+	}
+
```
dprotaso, line 63: ignore the error since the result will return `false` in that instance. By default we don't expect this to be set

```suggestion
	ignoreNotFound, _ := strconv.ParseBool(os.Getenv(IgnoreNotFound))

```

apiextensions/storageversion/cmd/migrate/main.go
```diff
@@ -31,6 +33,10 @@ import (
 	"knative.dev/pkg/signals"
 )
 
+const (
+	IgnoreNotFound = "MIGRATE_CRD_IF_FOUND_NO_FAILURE"
```
dprotaso, line 37: This name is really confusing and not obvious

```suggestion
	IgnoreNotFoundEnvKey = "IGNORE_NOT_FOUND"
```

apiextensions/storageversion/migrator.go
```diff
@@ -29,20 +29,23 @@ import (
...
-	dynamicClient dynamic.Interface
-	apixClient    apixclient.Interface
+	dynamicClient  dynamic.Interface
+	apixClient     apixclient.Interface
+	ignoreNotFound bool
 }
 
 // NewMigrator will return a new Migrator
-func NewMigrator(d dynamic.Interface, a apixclient.Interface) *Migrator {
+func NewMigrator(d dynamic.Interface, a apixclient.Interface, i bool) *Migrator {
```
dprotaso, line 44: Since `IgnoreNotFound` is optional I think the best approach would be a public property on the struct and you can set it _after_ calling `NewMigrator`

ie 
```go
migrator := NewMigrator(d, a)
migrator.IgnoreNotFound = true
```

This would ensure API compatibility - and we don't break existing consumers of this package

apiextensions/storageversion/migrator.go
```diff
@@ -55,10 +58,14 @@ func NewMigrator(d dynamic.Interface, a apixclient.Interface) *Migrator {
 // Finally the migrator will update the CRD's status and drop older storage
 // versions
 func (m *Migrator) Migrate(ctx context.Context, gr schema.GroupResource) error {
+	logger := logging.FromContext(ctx)
```
dprotaso, line 61: There's no logger on the context this ~will panic~ return the default logger which is not the same as the one in the `cmd`. Maybe follow the same approach as `IgnoreNotFound` and it can be an optional property

aavarghese: Done - thanks for the review with suggestions!!

dprotaso (commented):
Apologies - I realize In hindsight the simpler change would be to leave the `Migrator` alone and move all this logic to the `cmd/migrate/main.go` ?

ie.  modify this if block

https://github.com/knative/pkg/blob/1f681e126af66141941167c0a5aa44772089d82f/apiextensions/storageversion/cmd/migrate/main.go#L64-L66

Then there's no need for the two extra properties on the `Migrator` struct etc.

apiextensions/storageversion/migrator_test.go
```diff
@@ -167,6 +169,30 @@ func TestMigrate_Errors(t *testing.T) {
...
+	err := m.Migrate(context.Background(), fakeGR)
+	if err == nil {
+		t.Error("Migrate should have returned an error")
+	}
+	if !ignoreNotFound || !apierrs.IsNotFound(err) {
+		t.Error("Unexpected error:", err)
+	}
+
+}
+
```
dprotaso, line 195: We can delete this test now

apiextensions/storageversion/cmd/migrate/main.go
```diff
@@ -62,6 +71,10 @@ func main() {
 	for _, gr := range grs {
 		logger.Info("Migrating group resource ", gr)
 		if err := migrator.Migrate(ctx, gr); err != nil {
+			if ignoreNotFound && apierrs.IsNotFound(err) {
+				logger.Infof("Ignoring resource migration - unable to fetch crd %s - %w", gr, err)
```
dprotaso, line 75: `%w` only works for wrapping errors but here we are logging so we should use `%s`.

Otherwise the text is garbled in the log output - ie. https://go.dev/play/p/jyVUTGUxMJU

```suggestion
				logger.Infof("Ignoring resource migration - unable to fetch crd %s - %s", gr, err)
```

PR #2496, opened 2022-04-22
:gift: Deterministic exit code for zap's Fatal for `sharedmain`

### LINKED ISSUE #2495: For `sharedmain` zap's Fatal should deterministically calculate the POSIX return code
/area test-and-release
/kind feature

## Actual Behavior

The _Zap_'s _Fatal_ method is calling `os.Exit(1)` regardless of a message. This isn't the best _UX_ choice. The _POSIX_ return code could be used for quick assessment about a process failure. When the retcode is always equal `1` for any error, this can't be utilized.

Tools like Docker or Kubernetes report back the return code of the process. By using a deterministic algorithm to calculate the _POSIX_ return code, a quick debugging is possible. Users could indicate if the process is failing because of the same reason, without looking into logs.

## Expected Behavior

The best would be to rely on _Fatal_'s message or on an _error_ object if it's attached as a _Field_.

## Additional Info

Related to https://github.com/uber-go/zap/issues/1086

cardil:
# Changes

- :gift: Deterministic exit code for zap's Fatal for `sharedmain`

/kind enhancement

Fixes #2495

### DIFF
--- a/logging/exiting_zapcore.go
+++ b/logging/exiting_zapcore.go
@@ -1,0 +1,80 @@
+/*
+Copyright 2022 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logging
+
+import (
+	"errors"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"knative.dev/pkg/system/posix/exit"
+	"knative.dev/pkg/system/posix/retcode"
+)
+
+// ExitingZapcore is a zap.Option which will exit the process with deterministic
+// POSIX retcode for every Fatal+ invocation.
+var ExitingZapcore = zap.WrapCore(func(core zapcore.Core) zapcore.Core {
+	return retcodeCore{base: core}
+})
+
+type retcodeCore struct {
+	base   zapcore.Core
+	fields []zapcore.Field
+}
+
+func (r retcodeCore) Enabled(level zapcore.Level) bool {
+	return r.base.Enabled(level)
+}
+
+func (r retcodeCore) With(fields []zapcore.Field) zapcore.Core {
+	return retcodeCore{
+		base:   r.base.With(fields),
+		fields: fields,
+	}
+}
+
+func (r retcodeCore) Check(entry zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
+	if entry.Level >= zapcore.DPanicLevel {
+		return ce.AddCore(entry, r)
+	}
+	return r.base.Check(entry, ce)
+}
+
+func (r retcodeCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	if err := r.base.Write(entry, fields); err != nil {
+		return err
+	}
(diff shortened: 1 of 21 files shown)


dprotaso:
zap's godoc says the Fatal exit code is `1`

See: https://pkg.go.dev/go.uber.org/zap#Logger.Fatal

Are you seeing something different?

dprotaso:
I'm not particularly sold on this change - as the exit code is sorta random depending on the log message.

Plus a range of exit codes as meant to reserved:
https://man7.org/linux/man-pages/man1/exit.1p.html#RATIONALE

logging/exiting_zapcore.go
```diff
@@ -0,0 +1,80 @@
...
+	return r.base.Check(entry, ce)
+}
+
+func (r retcodeCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	if err := r.base.Write(entry, fields); err != nil {
+		return err
+	}
+	code := r.calculateRetcode(entry, fields)
+	_ = r.Sync()
+	exit.Exit(code)
```
evankanderson, line 63: Do you want to call [`go.uber.org/zap/internal/exit`](https://github.com/uber-go/zap/blob/9e61a39e216ee8b356cef2dbfb0e7872a9292e74/internal/exit/exit.go#L32) to enable working with zap's testing spy?

evankanderson: It looks like you've copied the `zap` spy into the `pkg` codebase, though I don't understand why we need that rather than adopting / improving the zap one.

system/posix/retcode/retcode.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package retcode
+
+import "hash/crc32"
+
+// Calc will calculate an POSIX retcode from an error.
```
evankanderson, line 21: I'm not sure why we want to do this -- it looks like we're collapsing all errors into a range of code 1-255.

https://unix.stackexchange.com/a/418802 suggests that the range 1-125 might be more reasonable, or 1 - (2^31-1) if you're willing to work with `waitid` and some of the newer posix calls.

system/posix/retcode/retcode_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"knative.dev/pkg/system/posix/retcode"
+)
+
+var (
+	errExampleError = errors.New("example error")
+)
+
+func TestCalc(t *testing.T) {
+	cases := testCases()
```
evankanderson, line 32: Why not inline this declaration ([as we do elsewhere](https://github.com/knative/pkg/blob/main/apis/url_test.go#L28))?

```golang

cases := []struct {
  name string
  err error
  want int
}{{
  name: "nil",
  err: nil,
  want: 0,
}, {
  ...
}}

File: logging/exiting_zapcore_test.go [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+	})
+
+	assert.Check(t, ex.Exited)
+	assert.Equal(t, ex.Code, 129)
+
+	logBytes, err := ioutil.ReadFile(logfile)
+	assert.NilError(t, err)
+	logs := string(logBytes)
+	assert.Check(t, cmp.Contains(logs, `"message":"foo"`))
+	assert.Check(t, cmp.Contains(logs, `"error":"bar"`))
```
evankanderson, line 54: I'm slightly concerned about adding a new dependency here to save a line of code:
```golang
if ! strings.Contains(logs, `"message":"foo"`) {
  t.Errorf("Expected %q to contain message tag", logs)
}
```

(To put it simply, the cost of adding the dependency seems higher than the value that it adds in terms of clarity.)

rhuss:
Where do we stand with this PR ? Seems to be dormant since quite some time.

dprotaso:
Going to close this out

Issue #2497, opened 2022-04-23
Add support for AdmissionReview warnings


mattmoor:
/area API
/kind feature

Kubernetes added support for [Warnings](https://kubernetes.io/blog/2020/09/03/warnings/) in the `AdmissionReview` response in Kubernetes 1.19.  The intent of this is to enable folks to give API users advance warning when a field they are using is headed towards deprecation, for instance I'd have loved to have this when we were deprecating build support in the v1alpha1 Service shape.  This also would have been very useful as we changed behaviors like disabling service links by default.

I'd suggest that we extend `apis.FieldError` to contain a "level" attribute of some sort: https://github.com/knative/pkg/blob/50410e0b833abc1a464d334b9e52e2c873dafcd4/apis/field_error.go#L31-L36

... and probably adapt the aggregated internal storage `errors []FieldError` to something more like `diagnostics map[string]FieldError` (this is private, so it should be safe).

In the webhook where we consume this: https://github.com/knative/pkg/blob/50410e0b833abc1a464d334b9e52e2c873dafcd4/webhook/resourcesemantics/validation/validation_admit.go#L81-L83

... we would then add some sensitivity to this new functionality.

---

I believe that the Tekton folks will find this interesting as they head towards things like PipelineResources deprecation and v1 (cc @vdemeester @bobcatfish @imjasonh @afrittoli )

I believe that the `cosigned` folks will find this interesting as a mechanism for surfacing a "Warn" mode on ClusterImagePolicy resources (cc @vaikas @hectorj2f).  I think this would give @vpnachev a solid path for what he wants in https://github.com/knative/pkg/issues/2494

I also think this would have (and will be) enormously useful for surfacing diagnostics to users in Knative itself! (cc @dprotaso )

---

I'd be happy to throw together a change with the picture forming in my head for this, if that's help illustrate things for folks.

hectorj2f:
+1 🙂

vaikas:
Yeah, when I saw that warning mode, was mosdef thinking that would be warning for cosigned webhooks mode for policies failing instead of trying, seems like they described our use case to a T :)

```
webhook implementations adding a "complain" mode, where they return warnings instead of rejections, to allow trying out a policy to verify it is working as expected before starting to enforce it
```

mattmoor:
Ok, I threw together the linked PR: https://github.com/knative/pkg/pull/2498

[...]

PR #2498, opened 2022-04-24
Add support for admission webhook warnings.

mattmoor:
- :gift: Add new feature
/kind enhancement

This extends `apis.FieldError` to support designating certain FieldErrors as "warnings" (or explicitly as "errors", however, this is the default for back-compat).

You can turn an `apis.FieldError` into a warning using: `fe.At(apis.WarningLevel)` or force it into an error using: `fe.At(apis.ErrorLevel)`.

You can get the errors at a particular diagnostic level using: `fe.Filter(apis.WarningLevel)`.

This change also hooks this into the admission webhook infrastructure to support surfacing the "warning" level `apis.FieldError`s via the `Warnings` section of the `AdmissionResponse`.

Fixes: #2497

**Release Note**

```release-note
Knative admission webhooks can now surface warnings (e.g. deprecations, "warn mode" policies).
```

**Docs**

```docs
```

/hold

Holding this until more folks can get 👀 on it.

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -28,18 +28,55 @@ import (
 // a problem with the current field itself.
 const CurrentField = ""
 
+// DiagnosticLevel is used to signal the severity of a particular diagnostic
+// in the form of a FieldError.
+type DiagnosticLevel string
+
+const (
+	// ErrorLevel is used to signify fatal/blocking diagnostics, e.g. those
+	// that should block admission in a validating admission webhook.
+	ErrorLevel DiagnosticLevel = "Error"
+
+	// WarningLevel is used to signify information/non-blocking diagnostics,
+	// e.g. those that should be surfaced as warnings in a validating admission
+	// webhook.
+	WarningLevel DiagnosticLevel = "Warning"
+)
+
+// Matches checks whether the provided diagnostic level matches this one,
+// including the special case where an empty DiagnosticLevel equals ErrorLevel.
+func (l DiagnosticLevel) Matches(r DiagnosticLevel) bool {
+	switch {
+	case l == ErrorLevel && r == "":
+		return true
+	case r == ErrorLevel && l == "":
+		return true
+	default:
+		return l == r
+	}
+}
+
 // FieldError is used to propagate the context of errors pertaining to
 // specific fields in a manner suitable for use in a recursive walk, so
 // that errors contain the appropriate field context.
 // FieldError methods are non-mutating.
 // +k8s:deepcopy-gen=true
 type FieldError struct {
+	// Message holds the main diagnostic message carried by this FieldError
 	Message string
-	Paths   []string
+
+	// Paths holds a list of paths to which this diagnostic pertains
+	Paths []string
+
+	// Level holds the severity of the diagnostic.
+	// If empty, this defaults to ErrorLevel.
+	Level DiagnosticLevel
+
 	// Details contains an optional longer payload.
 	// +optional
 	Details string
-	errors  []FieldError
+
+	errors []FieldError
 }
 
 // FieldError implements error
@@ -60,6 +97,7 @@ func (fe *FieldError) ViaField(prefix ...string) *FieldError {
 	// along using .Also().
 	newErr := &FieldError{
 		Message: fe.Message,
+		Level:   fe.Level,
 		Details: fe.Details,
 	}
 
@@ -105,6 +143,54 @@ func (fe *FieldError) ViaKey(key string) *FieldError {
 // ViaFieldKey is the short way to chain: err.ViaKey(bar).ViaField(foo)
 func (fe *FieldError) ViaFieldKey(field, key string) *FieldError {
 	return fe.ViaKey(key).ViaField(field)
+}
+
+// At is a way to alter the level of the diagnostics held in this FieldError.
+//    ErrMissingField("foo").At(WarningLevel)
+func (fe *FieldError) At(l DiagnosticLevel) *FieldError {
+	if fe == nil {
+		return nil
+	}
+	// Copy over message and details, paths will be updated and errors come
+	// along using .Also().
+	newErr := &FieldError{
+		Message: fe.Message,
+		Level:   l,
+		Details: fe.Details,
+		Paths:   fe.Paths,
+	}
+
+	for _, e := range fe.errors {
+		newErr = newErr.Also(e.At(l))
+	}
+	return newErr
+}
+
+// Filter is a way to access the set of diagnostics having a particular level.
+//    if err := x.Validate(ctx).Filter(ErrorLevel); err != nil {
+//       return err
+//    }
+func (fe *FieldError) Filter(l DiagnosticLevel) *FieldError {
+	if fe == nil {
+		return nil
+	}
+	var newErr *FieldError
+	if l.Matches(fe.Level) {
+		newErr = &FieldError{
+			Message: fe.Message,
+			Level:   fe.Level,
+			Details: fe.Details,
+			Paths:   fe.Paths,
+		}
+	}
+
+	for _, e := range fe.errors {
+		newErr = newErr.Also(e.Filter(l))
+	}
+	if newErr.isEmpty() {
+		return nil
+	}
+	return newErr
 }
 
 // Also collects errors, returns a new collection of existing errors and new errors.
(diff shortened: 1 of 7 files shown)


mattmoor:
I think most of the actions failures are unrelated, and I can't rerun the ones that look like flakes.

apis/field_error.go
```diff
@@ -28,18 +28,55 @@ import (
...
+	WarningLevel DiagnosticLevel = "Warning"
+)
+
+// Matches checks whether the provided diagnostic level matches this one,
+// including the special case where an empty DiagnosticLevel equals ErrorLevel.
+func (l DiagnosticLevel) Matches(r DiagnosticLevel) bool {
+	switch {
+	case l == ErrorLevel && r == "":
+		return true
+	case r == ErrorLevel && l == "":
```
vaikas, line 52: I was having a hard time understanding how this case could happen. Mayhaps add a comment for it?

webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -77,8 +77,16 @@ func (ac *reconciler) Admit(ctx context.Context, request *admissionv1.AdmissionR
 		return webhook.MakeErrorStatus("decoding request failed: %v", err)
 	}
 
-	if err := validate(ctx, resource, request); err != nil {
-		return webhook.MakeErrorStatus("validation failed: %v", err)
+	errors, warnings := validate(ctx, resource, request)
+	if warnings != nil {
+		// If there were warnings, then keep processing things, but augment
+		// whatever AdmissionResponse we send with the warnings.
+		defer func() {
```
vaikas, line 84: Just curious why this is defer instead of just resp.Warnings = ...

vaikas (commented):
Awesomesauce!

webhook/resourcesemantics/validation/validation_admit.go
mattmoor: It's mostly for plumbing.  This way of writing it catches whatever we return and sets `Warnings` on it.

If we were just to set it this way, then all of the return paths below here actually would overwrite it since `return webhook.MakeErrorStatus` is really the equivalent of:
```go
// Overwrite resp entirely, including Warnings.
resp = webhook.MakeErrorStatus(...)
return resp
```

The named return variable and defer let us get a crack at the actual returned value, and alter it.

apis/field_error.go
evankanderson: Do you want to simplify the code (but uglify the state) by making `ErrorLevel = "" // zero-value for for backwards compatibility`?

apis/field_error.go
```diff
@@ -274,6 +361,9 @@ func merge(errs []*FieldError) []*FieldError {
 	// Sort the flattened map.
 	sort.Slice(newErrs, func(i, j int) bool {
 		if newErrs[i].Message == newErrs[j].Message {
+			if newErrs[i].Details == newErrs[j].Details {
+				return newErrs[i].Level < newErrs[j].Level
```
evankanderson, line 365: This _happens_ to sort `Error` before `Warning` -- I don't know if we want to be more explicit about this, or leave it as a happy coincidence.

webhook/resourcesemantics/validation/validation_admit.go
evankanderson: It might be worth a comment that the regular form Ville thought of doesn't work here because of the function returns below.

I also wonder whether `validate` should return a list of api.FieldErrors and this function separates them, but that seems like six of one and a half-dozen of the other.

evankanderson (commented):
Looks pretty nice, lovely!

There is a unrelated code scanner warning about log entries from user input on line 168 of `validation_admit.go`. I'm not quite sure how to turn it into a comment or otherwise handle it -- it's logging the operation name from the admission request (presumably from apiserver, though someone could call the admit hook manually, I suppose).

apis/field_error.go
mattmoor: We do this so that the ordering is deterministic, otherwise (I'm guessing map ordering) the lines can flip-flop.

webhook/resourcesemantics/validation/validation_admit.go
mattmoor: Added a comment about this.

For the return value, I did it this way because of the `errMissingNewObject` path, but we could always make this a field error (and/or do this refactor later, since this is private).

mattmoor (commented):
I'm not sure how to disable the code scanning warning via a comment either 🤔

mattmoor:
Ok, gonna remove the hold, since the reception is fairly positive.

/unhold

mattmoor:
I rebased this, since there should be fewer action failures at HEAD (down to 1).

Since this clears the `/lgtm` I am going to remove the hold as well.

/unhold

mattmoor:
@evankanderson @dprotaso any other comments on this?

apis/field_error.go
evankanderson: (Still wondering about this)

apis/field_error.go
```diff
@@ -28,18 +28,55 @@ import (
...
+	// Message holds the main diagnostic message carried by this FieldError
 	Message string
-	Paths   []string
+
+	// Paths holds a list of paths to which this diagnostic pertains
+	Paths []string
+
+	// Level holds the severity of the diagnostic.
+	// If empty, this defaults to ErrorLevel.
+	Level DiagnosticLevel
```
evankanderson, line 73: Do we want `Level` to be a string or an int (using https://pkg.go.dev/golang.org/x/tools/cmd/stringer to convert to string when needed)?

evankanderson: i.e.

```golang

type Level int

// go:generate stringer -type=Level
const (
  ErrorLevel Level = iota
  DiagnosticLevel
)

File: apis/field_error.go (hunk 3 above)
Comment (evankanderson) [reply apis/field_error.go:365]: My concern was that adding an "Alert" level would sort "Alert", "Error", "Warning".

I'm also wondering if we want to sort by Level before Message, but I suspect that most of the time there will only be a few errors/warnings.

File: apis/field_error.go [hunk 5]
```diff
@@ -285,7 +375,11 @@ func merge(errs []*FieldError) []*FieldError {
 
 // key returns the key using the fields .Message and .Details.
 func key(err *FieldError) string {
-	return fmt.Sprintf("%s-%s", err.Message, err.Details)
+	l := err.Level
+	if l == "" {
+		l = ErrorLevel
+	}
```
evankanderson, line 381: If you switch to Stringer, I think you could remove these lines.

evankanderson (commented):
/lgtm

One suggestion on https://pkg.go.dev/golang.org/x/tools/cmd/stringer, but not required to implement.

apis/field_error.go
mattmoor: Hmm, `iota` isn't the zero value (AFAIK), so I'm not sure this WAI with the back-compat goal...  Maybe I'm missing something? 🤔

apis/field_error.go
mattmoor: At first I thought you meant `fmt.Stringer`, but I see you mean a tool (which I assume auto-generates the method, TIL).  I think for this to work, `ErrorLevel` would have to be explicitly `0` and then subsequent values would start with `iota`, or something like that... 🤔

evankanderson: [`iota` will start at 0](https://github.com/golang/go/wiki/Iota), so this would mean `ErrorLevel` was 0 and `WarningLevel` was 1.

Since we don't serialize these values, if we wanted to insert an "AlarmLevel" between "Error" and "Warning", we could redefine "Warning" to be 2 and "Alarm" to be 1. (I can't imagine that we'll need multiple levels of warnings, but I've implemented a lot of code I never imagined I'd need to write...)

apis/field_error.go
evankanderson: `iota` starts at the zero value at the beginning of each const block, and then increases for each subsequent ConstSpec (line) to the end of the block.

apis/field_error.go
mattmoor: Interesting, I always thought of it like an "epsilon" value, but perhaps it was designed to start at zero for cases exactly like this.  I'll try to experiment with this in a bit 🤞

mattmoor: https://github.com/knative/pkg/pull/2511

PR #2505, opened 2022-05-01
Support fetching configs from `ctx`.

mattmoor:
This enables callers of `sharedmain` to infuse `ctx` with a `logging.Config` or `leaderelection.Config` instead of relying on the API server for it.

- :gift: Add new feature

/kind enhancement

**Release Note**

```release-note
The logging and leaderelection configurations can now be configured with sharedmain.With{Logging,LeaderElection}Config respectively.
```

**Docs**

```docs

```

/assign @dprotaso @n3wscott


injection/sharedmain/main.go
```diff
@@ -59,10 +59,25 @@ func init() {
 	maxprocs.Set()
 }
 
-// GetLoggingConfig gets the logging config from either the file system if present
-// or via reading a configMap from the API.
+type lcfg struct{}
+
+// WithLoggingConfig associates a logging configuration with the context.  This
+// configuration takes priority in GetLoggingConfig (below).
+func WithLoggingConfig(ctx context.Context, cfg *logging.Config) context.Context {
```
dprotaso, line 66: Might make sense to move these helpers (With*, Get*) to `pkg/logging`.

Though the API lookup should probably remain in `sharedmain`

mattmoor: Done for both

mattmoor:
@dprotaso LMK if there is anything else on this.

Issue #2502, opened 2022-04-27
[flaky] networking.TestDialTLSWithBackoffSuccess


dprotaso:
Failed run: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_pkg/2500/unit-tests_pkg_main/1519305375889756160
```
=== RUN   TestDialTLSWithBackoffSuccess
2022/04/27 13:21:09 http: TLS handshake error from 127.0.0.1:46468: EOF
2022/04/27 13:21:09 http: TLS handshake error from 127.0.0.1:46472: EOF
2022/04/27 13:21:09 http: TLS handshake error from 127.0.0.1:46470: EOF
    transports_test.go:167: Dial error = timed out dialing after 0.22s
2022/04/27 13:21:09 http: TLS handshake error from 127.0.0.1:46474: write tcp 127.0.0.1:42715->127.0.0.1:46474: use of closed network connection
--- FAIL: TestDialTLSWithBackoffSuccess (0.23s)
```

/assign @cardil

cardil:
This is the problem of parallel tests that are choosing the same ports as free ones. Sometimes, most likely on contested servers, the OS is taking longer to allow running the port again.

We can:

- prolong the back off period https://github.com/knative/pkg/blob/2d552be030f69a4b9f74e49b5ab57256590dae74/network/transports_test.go#L194-L195
- pre-allocate free ports before the tests, and use them withing. This requires to integrate tests within a single group, and use `t.Run` to invoke the specific cases.

IDK, which is better?!?

dprotaso:
Let's prolong the backoff to begin with.

Looking at the httptest.Server it uses `127.0.0.1:0` but internal `Serve` call is on a goroutine so things could race.

Oddly this test fails consistently on my machine.

```
$ go test
--- FAIL: TestDialWithBackoffTimeout (0.00s)
    transports_test.go:134: Unexpected success dialing
```

You can sorta repro slower machines by limiting concurrency to a single OS thread - ie.  `GOMAXPROCS=1`

cardil:
Looking at the fail you posted, it looks to me like the problem will not be solved by prolonging the backoff. The tests are executed in parallel, and clearly `TestDialWithBackoffTimeout` connected successfully to other test port, which succeeded.

```
$ go test
--- FAIL: TestDialWithBackoffTimeout (0.00s)
    transports_test.go:134: Unexpected success dialing
```

[...]

PR #2510, opened 2022-05-03
Fix timeout test on darwin and log dial destination on timeout.

evankanderson:
Fixes: #2502

This adds extra logging so we can determine if a dial timeout like in #2502 was
hitting the expected port.

/assign @cardil
/cc @dprotaso

### DIFF
--- a/network/transports.go
+++ b/network/transports.go
@@ -113,7 +113,7 @@ func dialBackOffHelper(ctx context.Context, network, address string, bo wait.Bac
 		return c, nil
 	}
 	elapsed := time.Since(start)
-	return nil, fmt.Errorf("%w after %.2fs", ErrTimeoutDialing, elapsed.Seconds())
+	return nil, fmt.Errorf("%w %s after %.2fs", ErrTimeoutDialing, address, elapsed.Seconds())
 }
 
 func newHTTPTransport(disableKeepAlives, disableCompression bool, maxIdle, maxIdlePerHost int) http.RoundTripper {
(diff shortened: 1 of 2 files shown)


evankanderson:
FML; the second (backlog) argument to `listen` is handled _differently_ on darwin (MacOS) and Linux. On Mac, a backlog of `1` requires a single connection to enter the connection refused state for new connections, while on Linux, a backlog of `1` requires _two_ connections to enter the connection refused state. Even better, a backlog of `0` on MacOS seems to mean some large default, while it means _only allow one connection_ on Linux.

I'm going to switch to an infinite loop with a break statement in the backoff test.

dprotaso (approved):
tested on linux and OS X and they both passed

dprotaso:
One thing I saw with this pkg if you set `GOMAXPROCS=1` 

You'll get the log output
```
2022/05/03 15:16:00 http: TLS handshake error from 127.0.0.1:58772: read tcp 127.0.0.1:40673->127.0.0.1:58772: use of closed network connection
```

But the test passes so 🤷‍♂️ - so unsure

/lgtm

evankanderson:
> One thing I saw with this pkg if you set `GOMAXPROCS=1` 
> 
> You'll get the log output
> ```
> 2022/05/03 15:16:00 http: TLS handshake error from 127.0.0.1:58772: read tcp 127.0.0.1:40673->127.0.0.1:58772: use of closed network connection
> ```
> 
> But the test passes so 🤷‍♂️ - so unsure

This is because for the TLS timeout test, we're connecting to a TLS server and then hanging up before completing the TLS handshake. The error is correct, coming from the server, and is benign in this case.

PR #2511, opened 2022-05-03
Switch to iota/fmt.Stringer-based diagnostic levels.

mattmoor:
See discussion in: https://github.com/knative/pkg/pull/2498

This allows for a slightly cleaner backwards-compatible implementation 🎉 

This drops the `Matches` method, since we can now simply use `==`.

/kind cleanup
/assign @evankanderson 

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -30,29 +30,28 @@ const CurrentField = ""
 
 // DiagnosticLevel is used to signal the severity of a particular diagnostic
 // in the form of a FieldError.
-type DiagnosticLevel string
+type DiagnosticLevel int
 
 const (
 	// ErrorLevel is used to signify fatal/blocking diagnostics, e.g. those
 	// that should block admission in a validating admission webhook.
-	ErrorLevel DiagnosticLevel = "Error"
+	ErrorLevel DiagnosticLevel = iota
 
 	// WarningLevel is used to signify information/non-blocking diagnostics,
 	// e.g. those that should be surfaced as warnings in a validating admission
 	// webhook.
-	WarningLevel DiagnosticLevel = "Warning"
+	WarningLevel
 )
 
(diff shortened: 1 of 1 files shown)


mattmoor:
cc @dprotaso @vaikas

apis/field_error.go
```diff
@@ -30,29 +30,28 @@ const CurrentField = ""
...
+	case ErrorLevel:
+		return "Error"
+	case WarningLevel:
+		return "Warning"
+
 	default:
-		return l == r
+		return fmt.Sprintf("<UNKNOWN: %d>", dl)
 	}
 }
```
evankanderson, line 56: I think using `//go:generate stringer -type=DiagnosticLevel` will generate this code.

evankanderson (commented):
/lgtm
/hold

If you want to switch to `go generate`.

apis/field_error.go
mattmoor: I didn't do this because I wasn't sure it was worth adding `go generate` to the mix of things folks have to do 🤷 

I also figured it would do something like `"ErrorLevel"`, and `"WarningLevel"`, and I have a mild preference for this.

Neither is particular strong, but I also don't think it's saving us too much code either.

apis/field_error.go
```diff
@@ -175,7 +174,7 @@ func (fe *FieldError) Filter(l DiagnosticLevel) *FieldError {
 		return nil
 	}
 	var newErr *FieldError
-	if l.Matches(fe.Level) {
+	if l == fe.Level {
```
dprotaso, line 177: so fast

mattmoor:
/unhold

We can switch to codegen later if we think it's worth it and non-invasive.

PR #2518, opened 2022-05-15
Abstract how our webhook implementations target things.

mattmoor:
This change introduces a `targeter.Interface` abstraction that we've been
using downstream to abstract the core logic of webhooks from the details
of how delivery to them is handled.

The most immediate benefit of this refactoring is that each of our webhooks
no longer needs direct knowledge of `caBundle` and the secret logic, which
is consolidated in the provided implementations.  So for examples, folks
could now provide an implementation of `targeter.Interface` that receives
its `caBundle` information in a wholly different way (e.g. mounted from the
filesystem).

This encapsulation of how the `WebhookClientConfig` is produced also enables
folks to produce these configurations in more imaginative ways. For example,
I have included an implementation that allows folks to host a webhook with a
`BasePath()`, produce configs for "names" under this `BasePath()` and
then extract the "name" being handled when receiving a webhook request.

This encapsulation also lets folks produce `WebhookClientConfig` blocks that
don't even need `caBundle` and leverage `URL` blocks, if they are hosting
things with public-CA TLS-terminated endpoints (e.g. as a `ksvc`).

# Changes

/kind cleanup

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/webhook/targeter/targeter.go
+++ b/webhook/targeter/targeter.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2022 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package targeter
+
+import (
+	"context"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	apixv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Interface encapsulates how webhooks target us.
+type Interface interface {
+	// BasePath returns the base path on which we serve.
+	// If the path ends in a '/' then it will receive requests for all paths
+	// with that prefix.
+	// If the path does not end in a '/' then it will match exactly this path.
+	BasePath() string
+
+	// WebhookClientConfig returns the targeting config for this webhook.
+	// The path in the configured WebhookClientConfig should match, or have a
+	// prefix matching the value returned by BasePath above.
+	WebhookClientConfig(context.Context) (*admissionregistrationv1.WebhookClientConfig, error)
+
+	// AddEventHandler registers informer events as needed by this Interface
+	AddEventHandlers(context.Context, func(interface{}))
+}
+
+// SwitchClientConfig converts an admission webhook config to an API extension
+// webhook config.  These are effectively type aliases, but Kubernetes redefines
+// the type for unknown reasons.
+func SwitchClientConfig(in *admissionregistrationv1.WebhookClientConfig) *apixv1.WebhookClientConfig {
+	out := &apixv1.WebhookClientConfig{
+		URL:      in.URL,
+		CABundle: in.CABundle,
+	}
+	if in.Service != nil {
+		out.Service = &apixv1.ServiceReference{
+			Namespace: in.Service.Namespace,
+			Name:      in.Service.Name,
+			Path:      in.Service.Path,
+			Port:      in.Service.Port,
+		}
+	}
+	return out
+}
--- a/webhook/configmaps/configmaps.go
+++ b/webhook/configmaps/configmaps.go
@@ -19,11 +19,9 @@ package configmaps
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"reflect"
 
-	"go.uber.org/zap"
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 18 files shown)


webhook/targeter/targeter.go
```diff
@@ -0,0 +1,60 @@
...
+	// If the path does not end in a '/' then it will match exactly this path.
+	BasePath() string
+
+	// WebhookClientConfig returns the targeting config for this webhook.
+	// The path in the configured WebhookClientConfig should match, or have a
+	// prefix matching the value returned by BasePath above.
+	WebhookClientConfig(context.Context) (*admissionregistrationv1.WebhookClientConfig, error)
+
+	// AddEventHandler registers informer events as needed by this Interface
+	AddEventHandlers(context.Context, func(interface{}))
```
dprotaso, line 40: This is confusing since there's really no contract defined here - ie. in the implementations (dynamic, fixed) it's actually enqueuing the secret name and we're sorta of taking for granted that the webhook type implementations ignore that and use a fixed name.

This leaks the informer abstraction and can probably be simplified by having something like `SetChangeHandler(func)`

webhook/targeter/targeter.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"context"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	apixv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+// Interface encapsulates how webhooks target us.
+type Interface interface {
```
dprotaso, line 27: This seems more like a `ClientConfigProvider`  that could live in the `webhook` package

dprotaso (commented):
I think there's something here but I'm not sold on the implementation as is.

Generally I am +1 on something that provides a WebhookClientConfig. We can probably call
that a `ClientConfigProvider` and it can live in the `webhook` package. +1 on it abstracting
caBundle generation/lookup.

Do you have an example usage of the DispatchingInterface? My guess is you're using it to
support multiple webhooks within a single MutatingWebhookConfiguration. If that's the case
why not just have the `GetConfig` include the name of the webhook?

Likewise seeing duplicate `AddEventHandler` code in both Fixed and Dynamic is an
indication things can be more composable.

ie. a rough sketch of interfaces/structs

```go

type ClientConfigProvider interface {
	Get(context.Context, webhookName string) ClientConfig
	AddOnChangedHandler(func())
}

type CABundleProvider interface { 
	Get(context.Context, Options{WebhookName, URL, ServiceReference}) ([]byte, error)

	// Unsure if the callback should just include the updated CA thus allowing you to skip calling the Get
	AddOnChangedHandler(func())
}

type Fixed struct {
	path string
	provider CABundleProvider
	callback func()
	...
}

func NewFixed(path, options, caBundleProvider) Fixed {
	f := Fixed{}
	caBundleProvider.AddOnChangedHandler(f.bundleChanged)
	return f
}

func (f *Fixed) bundleChanged() {
	if f.callback != nil {
		f.callback()
	}
}

func (f *Fixed) SetOnChangeHandler(funk func()) {
	f.callback = funk
}

type Dynamic struct {
	NamedProviders map[string]ClientConfigProvider
}

func (d *Dynamic) SetOnChangeHandler(funk func()) {
	for _, provider := range d.NamedProviders {
	   provider.SetOnChangedHandler(funk)
	}
} 
```

webhook/targeter/targeter.go
```diff
@@ -0,0 +1,60 @@
...
+	WebhookClientConfig(context.Context) (*admissionregistrationv1.WebhookClientConfig, error)
+
+	// AddEventHandler registers informer events as needed by this Interface
+	AddEventHandlers(context.Context, func(interface{}))
+}
+
+// SwitchClientConfig converts an admission webhook config to an API extension
+// webhook config.  These are effectively type aliases, but Kubernetes redefines
+// the type for unknown reasons.
+func SwitchClientConfig(in *admissionregistrationv1.WebhookClientConfig) *apixv1.WebhookClientConfig {
```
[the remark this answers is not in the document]
mattmoor: Yeah every time I sat down to think of a better name / solution I just got grumpy they forked the type and so I went with this and moved on.  Happy to rename it if we can come up with something less horrible.

mattmoor:
> Do you have an example usage of the DispatchingInterface? My guess is you're using it to
support multiple webhooks within a single MutatingWebhookConfiguration. If that's the case
why not just have the GetConfig include the name of the webhook?

Nothing public yet, but yes.  It isn't about multiple webhooks within a single MutatingWebhookConfiguration, it's sort of the opposite.  We have N webhook configurations pointed at `/base/{name}` and want to be able to dispatch to an appropriate configuration based on the request path.  This is a simple case, it gets more fun from there.

The problem with including name directly in the interface for fetching the client config is that it assumes one particular implementation, to spitball a little bit:
1. For `NewFixed` the `name` parameter would be meaningless (it's the same config for all names).
2. (Devil's advocate a bit) Where `name` is sufficient for reading named configuration with a particular type, it is insufficient for say a polymorphic namespaced resource where I might want `(kind, namespace, name)` to determine the config to process.

I don't believe we're leveraging `NewDynamic` downstream as-is, since I think I wanted a few changes not easily upstreamed, so I am happy to remove it if it's confusing things, but I thought it might help illustrate a second implementation.  The main thing I'm after is a good `targeter.Interface` (or w/e we call it) to encapsulate how we slot in different logic for producing these configs.

I'm not in any particular rush here, so marinate on it a bit, and maybe I can set up some time next week to hop on a call (I can check if the WG call slot works, if that's best).

dprotaso:
I guess the `name` I was referring to was some combination of the webhook `metadata.name` and the `webhooks[*].name` and for conversion webhooks that would be just the `metadata.name`

That's the final destination for these webhook configs. So for me it's easy to reason our webhooks scan the current webhook spec and delegate to the helper with some additional context.

> (Devil's advocate a bit) Where name is sufficient for reading named configuration with a particular type, it is insufficient for say a polymorphic namespaced resource where I might want (kind, namespace, name) to determine the config to process.

Given this - `Get` probably needs the _entire context_ of the target webhook - 
ie. type & object meta, webhook name for `*WebhookConfiguration` and the CRD definition for conversion webhooks


> We have N webhook configurations pointed at /base/{name} and want to be able to dispatch to an appropriate configuration based on the request path. This is a simple case, it gets more fun from there.

But what's putting the `Name` on the context? I want to avoid these hidden APIs semantics on `context` objects. Thus I hope improvements don't continue this pattern.

> For NewFixed the name parameter would be meaningless (it's the same config for all names).

Yup - I think that's fine - ie. serving's webhook uses the same config for validation, defaulting and configmaps etc.

Issue #2533, opened 2022-06-21
CVE-2022-1996: Vulnerability in github.com/emicklei/go-restful-v2.15.0


JorTurFer:
/area API
/kind bug

## Expected Behavior

The dependency `github.com/emicklei/go-restful-v2.15.0` is affected by the vulnerability [CVE-2022-1996](https://vuln.whitesourcesoftware.com/vulnerability/CVE-2022-1996). To patch it, an update to v3.8.0 is needed

## Actual Behavior

Not having vulnerabilities in dep

## Additional Info

We have been notified by Whitesource-Bolt about this vulnerability https://github.com/kedacore/keda/issues/3156

cr22rc:
Hi @JorTurFer @psschwei  Help me, I thought the fix for this was go-restful v3.8.0 
https://nvd.nist.gov/vuln/detail/CVE-2022-1996

The go.mod still has github.com/emicklei/go-restful v2.9.5+incompatible /
Only reason I ask is I'm being pushed to close some vulnerabilities in downstream code.
Thanks.

psschwei:
So the problem is that the current version of `k8s.io/code-generator` (0.23.8) is using an outdated version of `k8s.io/kube-openapi`  (v0.0.0-20211115234752-e816edb12b65) which doesn't use the latest `github.com/emicklei/go-restful`.

psschwei:
We could fix it by temporarily replacing the version of `k8s.io/kube-openapi` with `v0.0.0-20220627174259-011e075b9cb8` which includes the fix. Assuming that's permitted (cc @evankanderson  or @dprotaso )

As of right now, it doesn't look like the next patch release will include the fix either (it's not in v0.23.9-rc.0 at the moment).

evankanderson:
It looks like this is a fix to CORS processing. Do we ship or expose either openapi (in general) or CORS (specifically) functionality in `knative.dev/pkg`?

psschwei:
There was only one [CORS related commit](https://github.com/knative/pkg/pull/294/files), but it appears that was removed a while ago (v0.11).

Also looks like kube-openapi is only an indirect dependency (it used to be used by pkg, but was [dropped two years ago](https://github.com/knative/pkg/commit/9792d28c6197ba20d88fb613154b8b484b95dd43)).

dprotaso:
We don't. use openapi-gen from code-generator so I think we're fine

```
$ go mod why -m github.com/emicklei/go-restful
[...]

PR #2544, opened 2022-07-06
update k8s.io/kube-openapi to fix CVE

psschwei:
Signed-off-by: Paul S. Schweigert <paulschw@us.ibm.com>

# Changes

This is a temp fix to bump the version of `k8s.io/kube-openapi` to address [CVE-2022-1996](https://vuln.whitesourcesoftware.com/vulnerability/CVE-2022-1996). Not sure if this is something we actually do, so if not feel free to close this.

xref #2533 

/assign @evankanderson @dprotaso


dprotaso:
looks like there are legit build failures - 

I don't believe we're using go-restful in a way that exercises that CVE so we could say this is a false positive if bumping is tricky

evankanderson:
Per my comment here:

https://github.com/knative/pkg/issues/2533#issuecomment-1176650322

I don't think this CVE affects any of the live code in this repo, though it might affect the test harnesses.

evankanderson:
I don't see the bump in the referenced package in the `go.mod` changes, should I?

psschwei:
> I don't see the bump in the referenced package in the `go.mod` changes, should I?

No, because I put the wrong package  in the PR body :facepalm: should've been `kube-openapi`

psschwei:
Based on @evankanderson 's comments here in the PR, @dprotaso's comment [in the issue](https://github.com/knative/pkg/issues/2533#issuecomment-1176718524), and the fact that `kube-openapi` doesn't even seem to call the cors.Filter method that's impacted by the CVE, I think we're good to close this.

PR #2547, opened 2022-07-15
Fix #2115, allow selective verbs / subresources for webhooks.

### LINKED ISSUE #2115: [webhook] allow webhooks to opt out of validating status updates
ie. serving doesn't have validation logic for the `status` block since we expect only our controllers update it. But our webhooks are invoked for status updates

https://github.com/knative/pkg/blob/132b1e71088b721a05f447b18e4459ffe8828169/webhook/resourcesemantics/defaulting/defaulting.go#L148-L152

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

# Changes
Currently when we create the validatingwebhookconfiguration, there's couple of wonky things with it:
1. Always includes status as a subresource and there are cases where we do not need it (#2115 )
2. Always includes all the verbs in the configuration even though they will later on get filtered, but the call to webhook is still made.

Also the API is widely used so I wanted to investigate what could be a smallest change using what we currently have to make this configurable.

So, the idea is that we have a way to customize additional validation logic via `Callbacks`, so if we utilize that struct we do not have to change the signature, but can piggyback off of it to provide ways to customize the actual webhook configuration without breaking any existing code.

Anyways, wanted to just put the smallest amount of code to see what folks think. obvs, need to write tests, etc. but just wanted to see what folks think before going too much deeper.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/webhook/resourcesemantics/interface.go
+++ b/webhook/resourcesemantics/interface.go
@@ -30,25 +30,21 @@ type GenericCRD interface {
 	runtime.Object
 }
 
-// GenericCRDWithConfig is the interface definition that allows us to perform
-// the generic CRD actions like deciding whether to increment generation and so
-// forth. You can further customiz with SupportedVerbs and SupportedSubResources
-// which
-type GenericCRDWithConfig struct {
-	GenericCRD
+// VerbLimited defines which Verbs you want to have the webhook invoked on.
+type VerbLimited interface {
+	// SupportedVerbs define which operations (verbs) webhook is called on.
+	SupportedVerbs() []admissionregistrationv1.OperationType
+}
 
-	// supportedVerbs are the verbs registered for the callback.
-	// If left empty, configures, Create, Update, and Delete
-	SupportedVerbs []admissionregistrationv1.OperationType
-
-	// supportedSubResources are the subresources that will be registered
+// SubResourceLimited defines which subresources you want to have the webhook
+// invoked on. For example "status", "scale", etc.
+type SubResourceLimited interface {
+	// SupportedSubResources are the subresources that will be registered
 	// for the resource validation.
-	// To get the main resource and status registered (old behaviour), you
-	// leave this empty.
 	// If you wanted to add for example scale validation for Deployments, you'd
 	// do:
 	// []string{"", "/status", "/scale"}
 	// And to get just the main resource, you would do:
 	// []string{""}
-	SupportedSubResources []string
+	SupportedSubResources() []string
 }
--- a/webhook/resourcesemantics/interface.go
+++ b/webhook/resourcesemantics/interface.go
@@ -17,6 +17,7 @@ limitations under the License.
 package resourcesemantics
 
 import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/apis"
 )
@@ -27,4 +28,27 @@ type GenericCRD interface {
 	apis.Defaultable
 	apis.Validatable
 	runtime.Object
+}
+
+// GenericCRDWithConfig is the interface definition that allows us to perform
+// the generic CRD actions like deciding whether to increment generation and so
+// forth. You can further customiz with SupportedVerbs and SupportedSubResources
+// which
+type GenericCRDWithConfig struct {
+	GenericCRD
+
+	// supportedVerbs are the verbs registered for the callback.
+	// If left empty, configures, Create, Update, and Delete
+	SupportedVerbs []admissionregistrationv1.OperationType
+
+	// supportedSubResources are the subresources that will be registered
+	// for the resource validation.
+	// To get the main resource and status registered (old behaviour), you
+	// leave this empty.
+	// If you wanted to add for example scale validation for Deployments, you'd
+	// do:
+	// []string{"", "/status", "/scale"}
+	// And to get just the main resource, you would do:
+	// []string{""}
+	SupportedSubResources []string
 }
--- a/webhook/resourcesemantics/validation/reconcile_config.go
+++ b/webhook/resourcesemantics/validation/reconcile_config.go
@@ -105,7 +105,8 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 
 		// If SupportedVerbs has not been given, provide the legacy defaults
 		// of Create, Update, and Delete
-		supportedVerbs := []admissionregistrationv1.OperationType{admissionregistrationv1.Create,
+		supportedVerbs := []admissionregistrationv1.OperationType{
+			admissionregistrationv1.Create,
 			admissionregistrationv1.Update,
 			admissionregistrationv1.Delete,
 		}
@@ -114,13 +115,14 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 			logging.FromContext(ctx).Debugf("Using custom Verbs")
 			supportedVerbs = vl.SupportedVerbs()
 		}
+		logging.FromContext(ctx).Debugf("Registering verbs: %s", supportedVerbs)
+
 		resources := []string{}
 		// If SupportedSubResources has not been given, provide the legacy
 		// defaults of main resource, and status
 		if srl, ok := config.(resourcesemantics.SubResourceLimited); ok {
 			logging.FromContext(ctx).Debugf("Using custom SubResources")
 			for _, subResource := range srl.SupportedSubResources() {
-				logging.FromContext(ctx).Debugf("Adding custom SubResource: %s", subResource)
 				if subResource == "" {
 					// Special case the actual plural if given
 					resources = append(resources, plural)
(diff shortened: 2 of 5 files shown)


webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -36,25 +36,51 @@ import (
 var errMissingNewObject = errors.New("the new object may not be nil")
 
 // Callback is a generic function to be called by a consumer of validation
+// To preserve backwards compatibility, it also doubles up as a way to configure
+// which verbs and which subresources are configured in the
+// validatingwebhookconfiguration. For configuring which verbs and subresources
+// should get validated, create the Callback and pass it to
```
dprotaso, line 42: I don't think we should utilize the struct in this way

dprotaso: I think we need an additional `New` method that's more generation and allows for more options

vaikas:
Created a new `New` method. I created a new struct, take a looksie @dprotaso

webhook/resourcesemantics/interface.go
```diff
@@ -28,3 +29,26 @@ type GenericCRD interface {
...
+	// supportedSubResources are the subresources that will be registered
+	// for the resource validation.
+	// To get the main resource and status registered (old behaviour), you
+	// leave this empty.
+	// If you wanted to add for example scale validation for Deployments, you'd
+	// do:
+	// []string{"", "/status", "/scale"}
+	// And to get just the main resource, you would do:
+	// []string{""}
+	SupportedSubResources []string
```
mattmoor, line 53: An alternative to this, which requires wrapping our typical objects in this struct, might be to define these as accessor methods in a separate interface and use type assertions?

e.g.
```
type VerbLimited interface {
    SupportedVerbs() []admissionregistrationv1.OperationType
}

type SubResourceLimited iterface {
   SupportedSubResources() []string
}
```

Then you can have the logic say:
```go
supportedVerbs := []string{/* what we do today*/}
if vl, ok := blah.(apis.VerbLimited); ok {
  supportedVerbs = vl.SupportedVerbs()
}
```

dprotaso: > The one downside is that then reusing esp. duck types will get a bit more complicated since the code defines what verbs/subresources are configured and changing them will require code changes.

Can you elaborate on this?

vaikas: Well, maybe I'm just confused about how this looks from the callers point of view. Aka, the person creating the webhook config in their main.

Say, there's a predefined `Resource`, like here (say it would be PodSpecable):
```
	handlers = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
		{
			Group:   "pkg.knative.dev",
			Version: "v1alpha1",
			Kind:    "Resource",
		}: &Resource{},
}
```

And then I want to say, nah, I actually don't want "status" updates for it. `Resource` is defined somewhere else, so I can't just add these methods to it in some other place (like my code that's using pkg types). So, I guess I'd have to define something like:
type ResourceCustom struct {
   GenericCRD
}

func (r *ResourceCustom) SupportedSubResources() []string {
  return []string{""} // just the main resource
}

And now the handlers call would be like so:
```
	handlers = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
		{
			Group:   "pkg.knative.dev",
			Version: "v1alpha1",
			Kind:    "Resource",
		}: &ResourceCustom{GenericCRD: &Resource{}},
}

@mattmoor @dprotaso is that what y'all are thinking, or am I missing something obvs?

**Comment by vaikas** [comment]
FYI. here's a working example demonstrating how it works with before / after. As in an answer to my question above :) 
https://github.com/sigstore/policy-controller/pull/165

**Comment by vaikas** [comment]
Not sure what the Tekton test is all about:
```
go: finding module for package knative.dev/pkg/pool
github.com/tektoncd/chains/pkg/chains imports
	github.com/sigstore/cosign/pkg/cosign imports
	knative.dev/pkg/pool: module knative.dev/pkg@latest found (v0.0.0-20220819090049-2e08ca63a922, replaced by /home/runner/work/pkg/pkg/upstream), but does not contain package knative.dev/pkg/pool

```

File: webhook/resourcesemantics/validation/reconcile_config.go [hunk 3]
```diff
@@ -100,23 +100,51 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
 	logger := logging.FromContext(ctx)
 
 	rules := make([]admissionregistrationv1.RuleWithOperations, 0, len(ac.handlers))
-	for gvk := range ac.handlers {
+	for gvk, config := range ac.handlers {
 		plural := strings.ToLower(flect.Pluralize(gvk.Kind))
 
+		// If SupportedVerbs has not been given, provide the legacy defaults
+		// of Create, Update, and Delete
+		supportedVerbs := []admissionregistrationv1.OperationType{admissionregistrationv1.Create,
```
mattmoor, line 108: nit: Can you line-break this?  I think it could be easy to misread this as just Update/Delete with current wrapping.

webhook/resourcesemantics/validation/reconcile_config.go
```diff
@@ -100,23 +100,51 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
...
+			admissionregistrationv1.Update,
+			admissionregistrationv1.Delete,
+		}
+
+		if vl, ok := config.(resourcesemantics.VerbLimited); ok {
+			logging.FromContext(ctx).Debugf("Using custom Verbs")
```
mattmoor, line 114: Print the verbs?

webhook/resourcesemantics/validation/validation_admit_test.go
```diff
@@ -645,7 +645,7 @@ func TestNewResourceAdmissionController(t *testing.T) {
 
 	NewAdmissionController(
 		ctx, testResourceValidationName, testResourceValidationPath,
-		handlers,
+		nil,
```
mattmoor, line 648: I'm curious why you need these changes?

mattmoor (approved):
/lgtm
/approve
/hold

Couple superficial comments, but this LGTM!  Hold for the nits and to see if @dprotaso has anything else.

webhook/resourcesemantics/validation/validation_admit_test.go
vaikas: leftover from earlier iterations, thanks!

webhook/resourcesemantics/validation/reconcile_config.go
```diff
@@ -100,23 +100,51 @@ func (ac *reconciler) reconcileValidatingWebhook(ctx context.Context, caCert []b
...
+			supportedVerbs = vl.SupportedVerbs()
+		}
+		resources := []string{}
+		// If SupportedSubResources has not been given, provide the legacy
+		// defaults of main resource, and status
+		if srl, ok := config.(resourcesemantics.SubResourceLimited); ok {
+			logging.FromContext(ctx).Debugf("Using custom SubResources")
```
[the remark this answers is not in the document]
vaikas: Below on L123?

webhook/resourcesemantics/validation/reconcile_config.go
vaikas: yep, done!

webhook/resourcesemantics/validation/reconcile_config.go
vaikas: yep!

webhook/resourcesemantics/validation/reconcile_config.go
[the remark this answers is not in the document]
vaikas: But, I just print them all now in one line for easier reading.

vaikas:
UT failed with:
```
--- FAIL: TestDialTLSWithBackoffSuccess (0.32s)
FAIL network.TestDialTLSWithBackoffSuccess (0.32s)
PASS test/imports
PASS test/interactive.TestNewCommand (0.00s)
PASS test/interactive.TestAddArgs (0.00s)
PASS test/interactive.TestCommandStringer (0.00s)

```
/test unit-tests_pkg_main

mattmoor:
/lgtm
/approve
/unhold

🎉 PKG cut last week, so we have another 5 to roll this back if we have regrets, but let's get it baking.

vaikas:
Oops, forgot to cut&paste the previous failures for retrying the unit tests:
```
--- FAIL: TestDialTLSWithBackoffSuccess (0.38s)
FAIL network.TestDialTLSWithBackoffSuccess (0.38s)
PASS test/imports
PASS test/helpers.TestAppendRandomString (0.00s)
PASS test/helpers.TestMakeK8sNamePrefix/abcd123 (0.00s)
PASS test/helpers.TestMakeK8sNamePrefix/AbCdef (0.00s)
```

dprotaso:
Been dealing with the release and haven't had a chance to comment

I think rather than baking in SupportedVerbs into our go types we allow types to declare
 `OnCreate` `OnUpdate` `OnDelete` as a way to signal the lifecycle events you care about

With the current solution you're sorta bleeding `admissionregistration` into various go types

Issue #2427, opened 2022-02-15
Use runtime/debug in our changeset package 


dprotaso:
Did some testing - runtime/debug probably won't be super useful until 1.18, particularly for local changes.

The main module info for something built not at a semver tag looks like:

```
Main:{Path:knative.dev/pkg Version:(devel) Sum: Replace:<nil>}
```

What we really want are the [vcs tags included in 1.18](https://tip.golang.org/doc/go1.18#go-command) which look like:

```
Settings:[
 {Key:-compiler Value:gc}
 {Key:CGO_ENABLED Value:1}
 {Key:CGO_CFLAGS Value:}
 {Key:CGO_CPPFLAGS Value:} 
 {Key:CGO_CXXFLAGS Value:}
 {Key:CGO_LDFLAGS Value:} 
 {Key:GOARCH Value:amd64} 
 {Key:GOOS Value:linux} 
 {Key:GOAMD64 Value:v1} 
 {Key:vcs Value:git} 
 {Key:vcs.revision Value:62b1122c86b2f032117da832332037ff511c2b74} 
 {Key:vcs.time Value:2022-02-11T23:48:33Z} 
 {Key:vcs.modified Value:true}
]
```

_Originally posted by @wlynch in https://github.com/knative/pkg/issues/2425#issuecomment-1039273410_

dprotaso:
We should switch over when we start using go1.18 for building

PR #2548, opened 2022-07-16
Bump min go version to  1.18 && read build info from embedded binary

dprotaso:
Fixes: https://github.com/knative/pkg/issues/2427

- :broom: Bump go version to 1.18
- :broom: Read commit info from embedded binary 
  - https://pkg.go.dev/runtime/debug#ReadBuildInfo
  - https://tip.golang.org/doc/go1.18#go-command

This should let us drop `HEAD` and `refs` in our `kodata` folders

/kind cleanup

**Release Note**

```release-note
changeset.Get now reads from vcs information embedded by the golang compiler
```


dprotaso:
cc @wlynch

dprotaso:
unit tests flaked (metrics)

cc @imjasonh for tekton pipeline downstream failure -

dprotaso:
One option is we fail silently if the `vcs.revision` is missing - which would happen in `go test` scenarios

imjasonh:
> One option is we fail silently if the `vcs.revision` is missing - which would happen in `go test` scenarios or `go run somefile.go`

Or if the binaries/images are built with `-buildvcs=false`, which some downstream might want to do if, e.g., Git isn't available in the CI environment, or they want to omit source information for some other reason.

I'll prepare a PR to fix Tekton's usage though, it doesn't look like these tests need to use `changeset.Get` anyway.

changeset/commit.go
```diff
@@ -17,113 +17,45 @@ limitations under the License.
...
-// KO_DATA_PATH. If it fails, it returns the error it gets.
+// Get returns the 'vcs.revision' property from the embedded build information
+// This function will return an error if value is not a valid Git SHA
+//
+// The result will have a '-dirty' suffix if the workspace was not clean
 func Get() (string, error) {
-	data, err := readFileFromKoData(commitIDFile)
-	if err != nil {
-		return "", err
+	info, ok := readBuildInfo()
```
imjasonh, line 35: Might be worth wrapping this in a `sync.Once` since this will likely get called lots of times by a long-running controller, and the result won't change after the first time.

dprotaso:
Ok added some tweaks

- Use a sync.Once to prevent double parsing
- Support SHA256
- Allow non-SHA revision strings

changeset/commit.go
```diff
@@ -17,113 +17,66 @@ limitations under the License.
...
-}
 
-// readFileFromKoData tries to open the file with given name under KO_DATA_PATH.
-// The file is expected to be wrapped into the container from /kodata by ko.
-// If it fails, returns the error it gets.
-func koDataFile(filename string) (*os.File, error) {
-	koDataPath := os.Getenv(koDataPathEnvName)
-	if koDataPath == "" {
-		return nil, fmt.Errorf("%q does not exist or is empty", koDataPathEnvName)
+	if shaRegexp.MatchString(revision) {
```
imjasonh, line 73: Technically, we don't need this check 😎 

(not necessarily advice to remove it, just a funny coincidence)

Actually, I wonder if this should just be `if len(revision) >= 7`, so folks using Subversion or whatever can still enjoy this goodness. I'm not sure there aren't other places we assume Git in and around here, but just in case.

imjasonh (approved):

changeset/commit.go
```diff
@@ -17,113 +17,66 @@ limitations under the License.
...
+	readBuildInfo = debug.ReadBuildInfo
+)
 
-// Get tries to fetch the first 7 digitals of GitHub commit ID from HEAD file in
-// KO_DATA_PATH. If it fails, it returns the error it gets.
+// Get returns the 'vcs.revision' property from the embedded build information
+// This function will return an error if value is not a valid Git SHA
+//
+// The result will have a '-dirty' suffix if the workspace was not clean
 func Get() (string, error) {
```
imjasonh, line 43: Can we have a `changeset.MustGet` that panics if it can't detect the version? I'm updating some callsites to get this once at `init`-time, and they're all just `log.Fatal`ing if this returns an error. Or should this return `Unknown` if it can't read build info, and just never return an error?

dprotaso: > Or should this return Unknown if it can't read build info, and just never return an error?

I was thinking about this - let me do it in a follow up

changeset/commit.go
dprotaso: I think I'd leave it as - since I don't want to assume truncating will work for different VCS - ie svn is numeric so we'd be alternating the commit id

dprotaso: A further improvement could be to do this only for vcs that use SHA's for revisions -ie git, hg etc.

PR #2549, opened 2022-07-18
Drop error from changeset.Get signature

dprotaso:
We simply return 'unknown' if we can't read the build info. We already return 'unknown' if the vcs.revision key is missing from the build settings

Discussion: https://github.com/knative/pkg/pull/2548#discussion_r923510432

/assign @imjasonh


imjasonh (approved):
Nice.

PR #2558, opened 2022-07-24
Fix issue of tracing.SetupDynamicPublishing

### LINKED ISSUE #2557: An issue with tracing.SetupDynamicPublishing
/area monitoring
/kind bug

## Expected Behavior

Tracing of pingsource doesn't work.

## Actual Behavior

Tracing of pingsource should work.

1. Set up pingsource following the [guidance](https://knative.dev/docs/eventing/sources/ping-source/)
2. Set up tracing following the [guidance](https://knative.dev/docs/eventing/accessing-traces/#configuring-tracing)
3. Set up zipkin according to the [guidance](https://knative.dev/docs/serving/accessing-traces/#zipkin)
4. Go to the web interface of zipkin, and we will not find any tracing of pingsource

## Additional Info

This issue is related with eventing issue https://github.com/knative/eventing/issues/6456

The pingsource tracing only works after I do a manually configmap `config-tracing` update. I think the reason is caused by [SetupDynamicPublishing](https://github.com/knative/pkg/blob/main/tracing/setup.go#L45) which should call `oct.ApplyConfig(cfg)` at the first time. Currently SetupDynamicPublishing doesn't call ApplyConfig(cfg) directly and `ApplyConfig(cfg)`  is only called when configmap is updated.

I propose to change the interface of SetupDynamicPublishing from
```
func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName, tracingConfigName string) error
```
to
```
func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName string, configmap *corev1.ConfigMap) error
```
and invoke a oct.ApplyConfig(cfg) after oct is created and then set up watcher.

SetupDynamicPublishing is not used in serving but used several times in eventing. These usages should be changed accordingly if the interface is changed.

daisy-ycguo:
# Changes
- :bug: Add tracing.SetupDynamicPublishingWithInitialValue to apply trace config

Fixes #2557

### DIFF
--- a/tracing/setup.go
+++ b/tracing/setup.go
@@ -21,6 +21,7 @@ import (
 
 	"go.uber.org/zap"
 
+	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/tracing/config"
 )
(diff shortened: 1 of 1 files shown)


dprotaso:
The downstream tests shows this change breaks the API.

My recommendation would be to create a new constructor ie. `SetupDynamicPublishingWithDefault`  (or something better)

daisy-ycguo:
> 

Thank you. I think it's a good suggestion and I will submit a new version.

daisy-ycguo:
/assigh @dprotaso

I think "downstream-tekton" could be override since it is caused by tekton using the old pkg version.

Issue #2561, opened 2022-07-26
Support Jaeger tracing backends


haubenr:
At the moment knative only supports tracing to Zipkin backends. It would be good to have more options, e.g. supporting Jaeger tracing backends as well (natively, without the need of external converters etc.)

/area tracing
/kind feature

## Expected Behavior

knative support Jaeger tracing backends.

## Actual Behavior

knative only supports tracing to Zipkin backends.

## Steps to Reproduce the Problem

n/a

## Additional Info

This issue is a follow-up to these issues that have been closed due to inactivity:

https://github.com/knative/pkg/issues/1076
https://github.com/knative/pkg/issues/1077

dprotaso:
I think I'd like to support jaeger by exporting Otel traces. Jaeger as of 1.35 can directly receive them.

We do need an OpenCensus => OTel exporter - which seems to have existed before but was removed

I prodded the issue here to see if it's unblocked & possible to add back a tracing bridge 

https://github.com/open-telemetry/opentelemetry-go/issues/2808

dprotaso:
Note: we want to eventually swap to OTel instead of OpenCensus https://github.com/knative/pkg/issues/855

haubenr:
Sure, sounds like the right move.

dprotaso:
We've migrated to OpenTelemetry so this isn't an issue anymore (should be out in the next release). Jaeger can also now accept OTLP traces!

PR #2566, opened 2022-08-03
SetupStaticPublishing and SetupDynamicPublishing returns Tracer with Shutdown function

### LINKED ISSUE #2475: Make sure OpenCensusTracer from tracing package can be correctly shut down (flushed)
The problem is that currently the components that use [SetupStaticPublishing](https://github.com/knative/pkg/blob/main/tracing/setup.go#L32) or [SetupDynamicPublishing](https://github.com/knative/pkg/blob/main/tracing/setup.go#L45) can drop some Spans/Traces when being shut down or evicted (e.g. during upgrades).
The goal is to prevent dropping buffered traces before they're sent as indicated by [upstream issue](https://github.com/census-instrumentation/opencensus-go/issues/862) in OpenCensus.

The OpenCensusTracer has a [Finish](https://github.com/knative/pkg/blob/main/tracing/opencensus.go#L84) function that can be called to properly [Close](https://github.com/knative/pkg/blob/main/vendor/github.com/openzipkin/zipkin-go/reporter/http/http.go#L71) the OpenZipkin http reporter and flush buffered traces. However, the SetupStaticPublishing/SetupDynamicPublishing functions don't make the OpenCensusTracer available so that anyone can call the Finish function. Another problem is that the Finish function [passes nil](https://github.com/knative/pkg/blob/main/tracing/opencensus.go#L92) which triggers nil pointer exception when calling the Finish function.

I've verified that calling the Finish function on the returned OpenCensusTracer sends all remaining traces and makes them available in Zipkin backend. It is only required to pass `&config.Config{}` on [this line](https://github.com/knative/pkg/blob/main/tracing/opencensus.go#L92) instead of nil.

It would be good to enforce calling Finish before components are shutdown as long as there's no sleep time in that component by default before it is shut down or evicted (I think some components have the sleep time such as Knative Serving Activator).

Related discussion [here](https://github.com/knative/eventing/pull/6219#discussion_r836446571) and [here](https://github.com/knative/eventing/pull/6219#discussion_r835613964)

## Expected Behavior

It is possible to flush traces on demand (possibly before shutting down a component).

## Actual Behavior

Traces that were buffered before the component is shutdown are not properly sent to tracing backend.

mgencur:
…acer

* The caller can then call Shutdown() to properly flush buffered Spans
before shutdown

This is incompatible change that will require updates in many repositories. I can either prepare those PRs so that they can be merged when this change propagates to those repositories or I can keep the old implementation and also provide a new impl that returns the OpenCensusTracer. Whatever works best.
Initially, I will provide an example PR for knative/eventing that will show the required updates.

# Changes

/kind enhancement

Fixes #2475

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -89,7 +89,7 @@ func (oct *OpenCensusTracer) Finish() error {
 	}
 
 	for _, configOpt := range oct.configOptions {
-		if err = configOpt(nil); err != nil {
+		if err = configOpt(&config.Config{}); err != nil {
 			return err
 		}
 	}
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -17,6 +17,7 @@ limitations under the License.
 package tracing
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -81,7 +82,7 @@ func (oct *OpenCensusTracer) ApplyConfig(cfg *config.Config) error {
 	return nil
 }
 
-func (oct *OpenCensusTracer) Finish() error {
+func (oct *OpenCensusTracer) Shutdown(ctx context.Context) error {
 	err := oct.acquireGlobal()
 	defer octMutex.Unlock()
 	if err != nil {
@@ -89,8 +90,13 @@ func (oct *OpenCensusTracer) Finish() error {
 	}
 
 	for _, configOpt := range oct.configOptions {
-		if err = configOpt(nil); err != nil {
-			return err
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			if err = configOpt(nil); err != nil {
+				return err
+			}
 		}
 	}
 	globalOct = nil
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -89,7 +89,7 @@ func (oct *OpenCensusTracer) Finish() error {
 	}
 
 	for _, configOpt := range oct.configOptions {
-		if err = configOpt(&config.Config{}); err != nil {
+		if err = configOpt(nil); err != nil {
 			return err
 		}
 	}
@@ -140,6 +140,9 @@ func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
 //   "host" is a endpoint IP like activator-service's endpoint IP.
 func WithExporterFull(name, host string, logger *zap.SugaredLogger) ConfigOption {
 	return func(cfg *config.Config) error {
+		if cfg == nil {
+			return errors.New("nil config")
+		}
 		var (
 			exporter trace.Exporter
 			closer   io.Closer
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -140,40 +140,40 @@ func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
 //   "host" is a endpoint IP like activator-service's endpoint IP.
 func WithExporterFull(name, host string, logger *zap.SugaredLogger) ConfigOption {
 	return func(cfg *config.Config) error {
-		if cfg == nil {
-			return errors.New("nil config")
-		}
 		var (
 			exporter trace.Exporter
 			closer   io.Closer
 		)
-		switch cfg.Backend {
-		case config.Zipkin:
-			// If host isn't specified, then zipkin.NewEndpoint will return an error saying that it
-			// can't find the host named ''. So, if not specified, default it to this machine's
-			// hostname.
-			if host == "" {
-				n, err := os.Hostname()
-				if err != nil {
-					return fmt.Errorf("unable to get hostname: %w", err)
+		if cfg != nil {
+			switch cfg.Backend {
+			case config.Zipkin:
+				// If host isn't specified, then zipkin.NewEndpoint will return an error saying that it
+				// can't find the host named ''. So, if not specified, default it to this machine's
+				// hostname.
+				if host == "" {
+					n, err := os.Hostname()
+					if err != nil {
+						return fmt.Errorf("unable to get hostname: %w", err)
+					}
+					host = n
 				}
-				host = n
-			}
-			if name == "" {
-				name = host
-			}
-			zipEP, err := zipkin.NewEndpoint(name, host)
-			if err != nil {
-				logger.Errorw("error building zipkin endpoint", zap.Error(err))
-				return err
-			}
-			reporter := httpreporter.NewReporter(cfg.ZipkinEndpoint)
-			exporter = oczipkin.NewExporter(reporter, zipEP)
-			closer = reporter
+				if name == "" {
+					name = host
+				}
+				zipEP, err := zipkin.NewEndpoint(name, host)
+				if err != nil {
+					logger.Errorw("error building zipkin endpoint", zap.Error(err))
+					return err
+				}
+				reporter := httpreporter.NewReporter(cfg.ZipkinEndpoint)
+				exporter = oczipkin.NewExporter(reporter, zipEP)
+				closer = reporter
 
-		default:
-			// Disables tracing.
+			default:
+				// Disables tracing.
+			}
 		}
+
 		if exporter != nil {
 			trace.RegisterExporter(exporter)
 		}
--- a/tracing/opencensus.go
+++ b/tracing/opencensus.go
@@ -82,6 +82,11 @@ func (oct *OpenCensusTracer) ApplyConfig(cfg *config.Config) error {
 	return nil
 }
 
+// Deprecated: Use Shutdown.
+func (oct *OpenCensusTracer) Finish() error {
+	return oct.Shutdown(context.Background())
+}
+
 func (oct *OpenCensusTracer) Shutdown(ctx context.Context) error {
 	err := oct.acquireGlobal()
 	defer octMutex.Unlock()
--- a/tracing/setup.go
+++ b/tracing/setup.go
@@ -32,12 +32,12 @@ type Tracer interface {
 	Shutdown(ctx context.Context) error
 }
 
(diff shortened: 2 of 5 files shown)


tracing/opencensus.go
```diff
@@ -140,6 +140,9 @@ func WithExporter(name string, logger *zap.SugaredLogger) ConfigOption {
 //   "host" is a endpoint IP like activator-service's endpoint IP.
 func WithExporterFull(name, host string, logger *zap.SugaredLogger) ConfigOption {
 	return func(cfg *config.Config) error {
+		if cfg == nil {
+			return errors.New("nil config")
+		}
```
mgencur, line 145: Prevent the `Finish()` function from throwing `panic`. More info in https://github.com/knative/eventing/pull/6219#discussion_r837277395

mgencur: Ehm. This will not work. This function needs to reach [this line](https://github.com/knative/pkg/blob/main/tracing/opencensus.go#L182) when calling Finish even if the passed config is nil.

mgencur: Pushed another commit that fixes that.

dprotaso:
We want to move away from OpenCensus - I don't think we should be returning the concrete type.

See: https://github.com/knative/pkg/issues/2561#issuecomment-1195806631

mgencur:
@dprotaso The downstream repositories are still using OpenCensus.
Looks like opentelemetry mostly uses something like this: `Shutdown(ctx context.Context) error`, E.g. [here](https://github.com/open-telemetry/opentelemetry-go/blob/e99a0ac0b5397c9997449131f7efa7a19442d1bf/sdk/trace/provider.go#L239). Would it sound better to return a type that implements this function? Not sure what other options we have.

mgencur:
@dprotaso I've pushed a new commit and provided the Tracer type with the generic Shutdown function. WDYT?

tracing/setup.go
```diff
@@ -25,24 +26,38 @@ import (
...
+// SetupPublishingWithDynamicConfig sets up trace publishing for the process, by watching a
+// ConfigMap for the configuration. The caller should call .Shutdown() on
+// the returned OpenCensusTracer before shutdown to make sure all traces are properly flushed.
+// Note that other pieces still need to generate the traces, this
 // just ensures that if generated, they are collected appropriately. This is normally done by using
 // tracing.HTTPSpanMiddleware as a middleware HTTP handler. The configuration will be dynamically
 // updated when the ConfigMap is updated.
-func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName, tracingConfigName string) error {
+func SetupPublishingWithDynamicConfig(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName, tracingConfigName string) (Tracer, error) {
 	oct := NewOpenCensusTracer(WithExporter(serviceName, logger))
```
daisy-ycguo, line 61: Please also take a new method [SetupDynamicPublishingWithInitialValue](https://github.com/knative/pkg/blob/main/tracing/setup.go#L76) into consideration when you make the changes.

tracing/setup.go
```diff
@@ -25,24 +26,38 @@ import (
...
 // handler. The configuration will not be dynamically updated.
-func SetupStaticPublishing(logger *zap.SugaredLogger, serviceName string, cfg *config.Config) error {
+func SetupPublishingWithStaticConfig(logger *zap.SugaredLogger, serviceName string, cfg *config.Config) (Tracer, error) {
 	oct := NewOpenCensusTracer(WithExporter(serviceName, logger))
 	if err := oct.ApplyConfig(cfg); err != nil {
-		return fmt.Errorf("unable to set OpenCensusTracing config: %w", err)
+		return nil, fmt.Errorf("unable to set OpenCensusTracing config: %w", err)
 	}
-	return nil
+	return oct, nil
```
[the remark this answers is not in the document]
mgencur: I don't think so. Components that are creating the tracer will do some work and they need to shutdown the tracer at the end (before their own shutdown). And they do it by calling Shutdown throught the tracing.Tracer interface. So the components will now depend only on the pkg/tracing package.
If you have other thoughts how this could be done please let me know. 
There's not https://github.com/knative/eventing/pull/6474 which shows how this would be used.

tracing/opencensus.go
```diff
@@ -81,16 +82,26 @@ func (oct *OpenCensusTracer) ApplyConfig(cfg *config.Config) error {
...
 	if err != nil {
 		return errors.New("finish called on OpenTracer which is not the global OpenCensusTracer")
 	}
 
 	for _, configOpt := range oct.configOptions {
-		if err = configOpt(nil); err != nil {
-			return err
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
dprotaso, line 100: Is there any reason why we skip performing the shutdown when the context is cancelled?

tracing/opencensus.go
```diff
@@ -144,33 +155,36 @@ func WithExporterFull(name, host string, logger *zap.SugaredLogger) ConfigOption
...
 			closer   io.Closer
 		)
-		switch cfg.Backend {
-		case config.Zipkin:
-			// If host isn't specified, then zipkin.NewEndpoint will return an error saying that it
-			// can't find the host named ''. So, if not specified, default it to this machine's
-			// hostname.
-			if host == "" {
-				n, err := os.Hostname()
+		if cfg != nil {
```
dprotaso, line 158: curious why the config can be nil now when before it wasn't.

tracing/setup.go
```diff
@@ -26,24 +27,38 @@ import (
 	"knative.dev/pkg/tracing/config"
 )
 
-// SetupStaticPublishing sets up trace publishing for the process. Note that other
-// pieces still need to generate the traces, this just ensures that if generated, they are collected
+type Tracer interface {
+	// Shutdown allows for proper cleanup/flush.
+	Shutdown(ctx context.Context) error
```
dprotaso, line 32: looks good

tracing/setup.go
```diff
@@ -102,5 +123,11 @@ func SetupDynamicPublishingWithInitialValue(logger *zap.SugaredLogger, configMap
...
 		tracerUpdater)
 	configStore.WatchConfigs(configMapWatcher)
-	return nil
+	return oct, nil
+}
+
+// Deprecated: Use SetupPublishingWithDynamicConfigAndInitialValue.
+func SetupDynamicPublishingWithInitialValue(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName string, configm *corev1.ConfigMap) error {
+	_, err := SetupPublishingWithDynamicConfigAndInitialValue(logger, configMapWatcher, serviceName, configm)
+	return err
```
dprotaso, line 132: We probably don't need this variation of the method that just returns an `error` since it was just added yesterday.

tracing/opencensus.go
mgencur: The `Finish/Shutdown` function is passing nil to the config option on shutdown. But so far this function was only called in tests and those tests never used this function `WithExporterFull`. From now on the Shutdown function will be called with the config option provided by WithExportedFull so it needs to check if the passed config is nil to prevent `panic`.

tracing/opencensus.go
mgencur: It's a common way to allow for interrupting a long running functions through the context. OpenTelemetry uses this approach as well for interrupting the Shutdown function.

tracing/setup.go
```diff
@@ -26,24 +27,38 @@ import (
...
-// ConfigMap for the configuration. Note that other pieces still need to generate the traces, this
+// SetupPublishingWithDynamicConfig sets up trace publishing for the process, by watching a
+// ConfigMap for the configuration. The caller should call .Shutdown() on
+// the returned OpenCensusTracer before shutdown to make sure all traces are properly flushed.
+// Note that other pieces still need to generate the traces, this
 // just ensures that if generated, they are collected appropriately. This is normally done by using
 // tracing.HTTPSpanMiddleware as a middleware HTTP handler. The configuration will be dynamically
 // updated when the ConfigMap is updated.
-func SetupDynamicPublishing(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName, tracingConfigName string) error {
+func SetupPublishingWithDynamicConfig(logger *zap.SugaredLogger, configMapWatcher configmap.Watcher, serviceName, tracingConfigName string) (Tracer, error) {
```
daisy-ycguo, line 61: Since we are changing the interface and we want to change all the invocation of `SetupDynamicPublishing` to `SetupPublishingWithDynamicConfig`, could we change to `SetupPublishingWithDynamicConfigAndInitialValue` directly? As I reported in issue https://github.com/knative/pkg/issues/2557, `SetupDynamicPublishing` does not work properly since it doesn't call oct.ApplyConfig(cfg) at the first time.

If you agree, I think we don't need method `SetupPublishingWithDynamicConfig`. We just need method `SetupPublishingWithDynamicConfigAndInitialValue`. 

@dprotaso your thoughts?

mgencur: Since https://github.com/knative/eventing/issues/6456 is not yet implemented and `SetupDynamicPublishingWithInitialValue` is not referenced yet I think it makes sense to do this:
* Keep the current PR as it is and only have the latest/newest `SetupPublishingWithDynamicConfigAndInitialValue`
* If the current PR is merged soon enough the new function can be used directly in a PR for knative/eventing
* If there is a PR for knative/eventing merged before the current one I will put back the backwards-compatible function here so as to not break downstream

dprotaso:
FYI - `knative.dev/pkg` cut is tomorrow - if you plan on making any more edits tomorrow let me know and I can wait as long as it's straight forward

mgencur:
@dprotaso @daisy-ycguo Thanks for reviews and merging!

PR #2567, opened 2022-08-11
Allow configuration of threads-per-controller

SaschaSchwarze0:
# Changes

The Knative controller infrastructure uses by default two threads per controller. This setting is not configurable in Knative itself while [Tekton allows to configure it](https://github.com/tektoncd/pipeline/blob/v0.38.2/cmd/controller/main.go#L46). [I was playing around with that setting in Knative](https://knative.slack.com/archives/C0186KU7STW/p1659968402805789) and saw massively better numbers to bring Knative up when it reconciles everything in the system.

- :gift: introduce `threads-per-controller` flag to configure the number of go routines per controller

/kind enhancement

**Release Note**

```release-note
You can now provide a `threads-per-controller` command line flag to configure the number of go routines per thread, the default is 2
```

**Docs**

Please let me know if this needs to be documented now given this is "only" in the library and needs to be picked up by the serving repo first.

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -165,6 +166,14 @@ func toControllerConstructors(namedCtors []injection.NamedControllerConstructor)
 // MainWithContext runs the generic main flow for controllers and
 // webhooks. Use MainWithContext if you do not need to serve webhooks.
 func MainWithContext(ctx context.Context, component string, ctors ...injection.ControllerConstructor) {
+	// Allow configuration of threads per controller
+	if val, ok := os.LookupEnv("K_THREADS_PER_CONTROLLER"); ok {
+		threadsPerController, err := strconv.Atoi(val)
+		if err != nil {
+			log.Fatalf("failed to parse value %q of K_THREADS_PER_CONTROLLER: %v\n", val, err)
+		}
+		controller.DefaultThreadsPerController = threadsPerController
+	}
 
 	// TODO(mattmoor): Remove this once HA is stable.
 	disableHighAvailability := flag.Bool("disable-ha", false,


injection/sharedmain/main.go
```diff
@@ -165,6 +165,8 @@ func toControllerConstructors(namedCtors []injection.NamedControllerConstructor)
 // MainWithContext runs the generic main flow for controllers and
 // webhooks. Use MainWithContext if you do not need to serve webhooks.
 func MainWithContext(ctx context.Context, component string, ctors ...injection.ControllerConstructor) {
+	// Allow configuration of threads per controller
+	flag.IntVar(&controller.DefaultThreadsPerController, "threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
```
n3wscott, line 169: Not a huge fan of flags, there is a config map that the control can pull config out of. Maybe that is a better place?

SaschaSchwarze0: Hi @n3wscott, thank you for the feedback. I can see how the leader election configuration is read from a ConfigMap here, https://github.com/knative/pkg/blob/main/injection/sharedmain/main.go#L88-L104. But looking through the names of the existing ConfigMaps, I do not see a good candidate. Are you suggesting to introduce an additional ConfigmMap ? And if yes, how should we name it ? Would it only contain the threads-per-controller, or do you envision that related things that are today configurable using flags, like qps and burst, would be in there as well ? So maybe a `config-tuning` ?

n3wscott: Ok good point. Perhaps then we could introduce an ENV VAR instead of a flag? something like `K_THREADS_PER_CONTROLLER` or something similar?

SaschaSchwarze0: Okay, as you wish. I put it at the same location. Let me know if you want it elsewhere.

SaschaSchwarze0:
> [SaschaSchwarze0](https://github.com/SaschaSchwarze0) requested review from [n3wscott](https://github.com/n3wscott) and removed request for [mattmoor](https://github.com/mattmoor) 27 seconds ago

Hm, no? I requested re-review from n3wscott but have no permission to add or remove somebody without the use of the bot.

/cc mattmoor

n3wscott (approved):

injection/sharedmain/main.go
```diff
@@ -165,6 +166,14 @@ func toControllerConstructors(namedCtors []injection.NamedControllerConstructor)
 // MainWithContext runs the generic main flow for controllers and
 // webhooks. Use MainWithContext if you do not need to serve webhooks.
 func MainWithContext(ctx context.Context, component string, ctors ...injection.ControllerConstructor) {
+	// Allow configuration of threads per controller
+	if val, ok := os.LookupEnv("K_THREADS_PER_CONTROLLER"); ok {
+		threadsPerController, err := strconv.Atoi(val)
+		if err != nil {
+			log.Fatalf("failed to parse value %q of K_THREADS_PER_CONTROLLER: %v\n", val, err)
+		}
+		controller.DefaultThreadsPerController = threadsPerController
```
mattmoor, line 175: FWIW it is a global variable so that binary entrypoints can use the following to surface it as a flag, but I'm also totally fine with having an environment variable for it as well.
```
flag.StringVar(&controller.DefaultThreadsPerController, ...)
```

n3wscott: It was a baked in flag and _those kinds_ flags have caused so many footguns I asked for an env var. And yes, because it is exposed, an application can choose to use a flag if they wanted. +1

n3wscott:
this unit test failure looks like a flake to me. 

```
=== RUN   TestDialWithBackoffTimeout
    transports_test.go:150: Unexpected success dialing
--- FAIL: TestDialWithBackoffTimeout (0.05s)
```

/test unit-tests

The other failures are version mismatches with pkg and the downstream.

injection/sharedmain/main.go
mattmoor: I wasn't suggesting a flag in `package sharedmain` (yes, library = footgun), I was suggesting a flag in the consumer's `package main` (<-- binary entrypoint).

PR #2491, opened 2022-04-12
timer set to nil after draining period

Shashankft9:
# Changes

- :bug: sets timer to nil once the draining period wait is over

/kind bug

Fixes https://github.com/knative/serving/issues/12571

**Release Note**

```release-note

```

**Docs**

```docs

```

PR #2572, opened 2022-08-15
allow drainer to be reset

dprotaso:
Replaces https://github.com/knative/pkg/pull/2491

We need a reset so that downstream for serving to work around containers that are failing their liveness probe see: https://github.com/knative/serving/issues/12571

### DIFF
--- a/network/handlers/drain.go
+++ b/network/handlers/drain.go
@@ -70,11 +70,10 @@ type Drainer struct {
 	// after Drain is called before it may return.
 	QuietPeriod time.Duration
 
-	// once is used to initialize timer
-	once sync.Once
-
 	// timer is used to orchestrate the drain.
 	timer timer
+
+	ch chan struct{}
 
 	// HealthCheckUAPrefixes are the additional user agent prefixes that trigger the
 	// drainer's health check
@@ -106,7 +105,7 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	d.reset()
+	d.resetTimer()
 	d.Inner.ServeHTTP(w, r)
 }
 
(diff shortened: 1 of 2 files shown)


network/handlers/drain.go
```diff
@@ -145,8 +167,24 @@ func (d *Drainer) isHealthCheckRequest(r *http.Request) bool {
...
 
+func (d *Drainer) Reset() {
+	d.Lock()
+	defer d.Unlock()
+
+	if d.timer.Stop() {
+		d.timer = nil
+	} else {
+		drainTimer(d.timer.tickChan())
+	}
```
dprotaso, line 178: need an nil check here

network/handlers/drain.go
```diff
@@ -70,12 +70,11 @@ type Drainer struct {
 	// after Drain is called before it may return.
 	QuietPeriod time.Duration
 
-	// once is used to initialize timer
-	once sync.Once
-
 	// timer is used to orchestrate the drain.
 	timer timer
 
+	ch chan struct{}
```
nader-ziada, line 77: can you add a comment here on this field to make more clear

network/handlers/drain.go
```diff
@@ -145,8 +167,26 @@ func (d *Drainer) isHealthCheckRequest(r *http.Request) bool {
...
+
+	if d.ch != nil {
+		close(d.ch)
+		d.ch = nil
+	}
+}
+
 // reset resets the drain timer to the full amount of time.
-func (d *Drainer) reset() {
+func (d *Drainer) resetTimer() {
```
nader-ziada, line 190: this comment should move the new `reset` func

network/handlers/drain.go
```diff
@@ -115,19 +114,42 @@ func (d *Drainer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	})
+		d.ch = ch
+		d.timer = timer
+		return ch
+	}()
+
+	<-ch
+}
+
+func drainTimer(tc <-chan time.Time) {
```
nader-ziada, line 149: I don't get this, is it just resetting the timer when a signal comes on the channel?

dprotaso: this drains the timer's channel - if the `Stop` call didn't stop the timer there may be an event here and we drain it.

PR #2574, opened 2022-08-17
fix double close of channel

dprotaso:


dprotaso:
whoops
/cherry-pick release-1.7

PR #2576, opened 2022-08-19
WaitForServiceEndpoints doesn't take into account not found endpoints

pierDipi:
`Endpoints` are created asynchronously so it might not exists as soon
as a sertvice is created.

We're occasionally getting:
```
did not reach the desired number of endpoints, got: 0: endpoints "trans-pod" not found
```

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>


matzew (approved):
/lgtm
/approve

PR #2582, opened 2022-08-23
split warnings. Fix issue: 2581

### LINKED ISSUE #2581: Warnings are silently dropped by k8s due to newlines.
/kind bug

## Expected Behavior
When creating a warning like this:

```
warnField := apis.ErrGeneric(fmt.Sprintf("failed policy: %s", failingPolicy), "image").ViaFieldIndex(field, i)
warnField.Details = warnDetails
errs = errs.Also(warnField).At(apis.WarningLevel)
```
I would expect this to work. However, this turns into something like this:
```
{
	"kind": "AdmissionReview",
	"apiVersion": "admission.k8s.io/v1",
	"response": {
		"uid": "f3ebec5c-6302-4efb-9ef8-49215353a566",
		"allowed": true,
		"warnings": ["failed policy: image-policy-keyless: spec.containers[0].image\nregistry.local:5000/sigstore/demo@sha256:a420f0e3a1f8c20c6be3caaca3d6601139806b7dc5d7e4256a912675cfd319c6 signature keyless validation failed for authority authority-0 for registry.local:5000/sigstore/demo@sha256:a420f0e3a1f8c20c6be3caaca3d6601139806b7dc5d7e4256a912675cfd319c6: Get \"https://registry.local:5000/v2/\": dial tcp: lookup registry.local on 10.12.0.10:53: no such host; Get \"http://registry.local:5000/v2/\": dial tcp: lookup registry.local on 10.12.0.10:53: no such host"]
	}
}
```
Note the `\n` that we use to construct the error message by joining on `\n`

I'd argue this is not very well spelled out here what's allowed and not allowed:
https://github.com/kubernetes/api/blob/master/admission/v1/types.go#L144

But, digging deeper, the the problem is that k8s silently ignores this:
https://github.com/kubernetes/apiserver/blob/b10345148a460f9ed3dc576871a547953d021adf/pkg/endpoints/filters/warning_test.go#L83-L85

I would expect warnings to get created by our pkg to get wired through to `kubectl` properly.

## Actual Behavior
The warning message gets dropped on the floor, and nothing is passed to `kubectl`.

1.  Create the warning like above that includes `details`, or anything that causes a `join` to happen.
2. 
3.

## Additional Info
Thanks @mattmoor for second set of eyes for digging into this :)

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

# Changes
🐛 Fix #2581 

If warnings contain newlines, split on them, and create a new warning for each of them.

/kind <kind>

Fixes #

**Release Note**

```release-note
* Fixes a bug (#2581) where (most if not all) warnings were silently ignored by k8s apiserver due to having newlines on them.
```

**Docs**

```docs

```

### DIFF
--- a/webhook/admission.go
+++ b/webhook/admission.go
@@ -133,9 +133,7 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
 		if reviewResponse.Warnings != nil {
 			cleanedWarnings := make([]string, 0, len(reviewResponse.Warnings))
 			for _, w := range reviewResponse.Warnings {
-				for _, piece := range strings.Split(w, "\n") {
-					cleanedWarnings = append(cleanedWarnings, piece)
-				}
+				cleanedWarnings = append(cleanedWarnings, strings.Split(w, "\n")...)
 			}
 			reviewResponse.Warnings = cleanedWarnings
 		}
--- a/webhook/admission.go
+++ b/webhook/admission.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


JAORMX (changes requested):
Mind adding a test for this?

vaikas:
Flaky test again.
```
=== FAIL: network TestDialTLSWithBackoffSuccess (0.22s)
2022/08/23 18:48:05 http: TLS handshake error from 127.0.0.1:48412: EOF
2022/08/23 18:48:05 http: TLS handshake error from 127.0.0.1:48424: EOF
2022/08/23 18:48:05 http: TLS handshake error from 127.0.0.1:48438: EOF
    transports_test.go:183: Dial error = timed out dialing 127.0.0.1:44127 after 0.22s
2022/08/23 18:48:05 http: TLS handshake error from 127.0.0.1:48446: write tcp 127.0.0.1:44127->127.0.0.1:48446: use of closed network connection

```

/test unit-tests_pkg_main

mattmoor (approved):
/lgtm
/approve

Thanks!

JAORMX (commented):
thanks for adding tests! lgtm (too late, but that's fine)

PR #2583, opened 2022-08-23
Make request body available in admission context

elfotografo007:
Signed-off-by: Andrés Torres <andrest@vmware.com>

# Changes

- :bug: Make request body available in admission context

/kind bug

Fixes #

The `admissionHandler` is consuming the request body:
https://github.com/knative/pkg/blob/5fbe0c3f016d75ebea3a15ca9727ce52e1abb931/webhook/admission.go#L93

If the `Admit` function of the `AdmissionController` later gets the request using `apis.GetHTTPRequest` and tries to read `Request.Body`, an EOF error will be returned. This PR addresses this issue by making the entire `Request.Body` available again for an additional read operation.

The use case is that I need to parse again the `AdmissionRequest` from `Request.Body` to perform further checks in the validating/mutating phase.

**Release Note**

```release-note

```

**Docs**

```docs

```


vaikas:
Sweet!

vaikas:
la sigh, known flake:
```
=== FAIL: network TestDialTLSWithBackoffSuccess (0.23s)
2022/08/25 23:30:58 http: TLS handshake error from 127.0.0.1:59496: EOF
2022/08/25 23:30:58 http: TLS handshake error from 127.0.0.1:59504: EOF
2022/08/25 23:30:58 http: TLS handshake error from 127.0.0.1:59514: EOF
    transports_test.go:183: Dial error = timed out dialing 127.0.0.1:36721 after 0.22s

```

/ok-to-test
/retest

dprotaso:
> The use case is that I need to parse again the AdmissionRequest from Request.Body to perform further checks in the validating/mutating phase.

Would it make more sense for the `AdmissionRequest` go object to be on the context - instead of having the raw request body and parsing it again yourself?

PR #2585, opened 2022-08-26
ValidatesCreates also calls SetDefault if defaultable

gab-satchi:
- 🧹 Create validation calls SetDefaults if available.

/kind enhancement

There's a case with Eventing sugar controller where it creates an empty Broker and relies on the defaulting config map to fill in the details. After defaulting the Broker will become valid. The tests were failing when defaulting wasn't called on the newly created Broker object.

### DIFF
--- a/reconciler/testing/reactions.go
+++ b/reconciler/testing/reactions.go
@@ -48,6 +48,12 @@ func InduceFailure(verb, resource string) clientgotesting.ReactionFunc {
 
 func ValidateCreates(ctx context.Context, action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
 	got := action.(clientgotesting.CreateAction).GetObject()
+	o, ok := got.(apis.Defaultable)
+	if ok {
+		o.SetDefaults(ctx)
+		got, _ = o.(runtime.Object)
+	}
+
 	obj, ok := got.(apis.Validatable)
 	if !ok {
 		return false, nil, nil


pierDipi (commented):
Isn't better to let test authors and callers decide when to call `SetDefaults` and when not?

reconciler/testing/reactions.go
```diff
@@ -48,6 +48,12 @@ func InduceFailure(verb, resource string) clientgotesting.ReactionFunc {
 
 func ValidateCreates(ctx context.Context, action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
 	got := action.(clientgotesting.CreateAction).GetObject()
+	o, ok := got.(apis.Defaultable)
+	if ok {
+		o.SetDefaults(ctx)
+		got, _ = o.(runtime.Object)
+	}
+
```
pierDipi, line 56: Another alternative is to have a Defaulting reactor that calls SetDefaults on defaultable objects that are created as well as updated and in eventing we can choose to opt-in the SetDefault call by using it only for the sugar reconciler.

gab-satchi:
@pierDipi Do you mean a `SkipValidation/SkipDefaulting` flag on TestRow like [this](https://github.com/knative/pkg/blob/7c4fef1af593ca351ccd27537c3b3c3b06e4adef/reconciler/testing/table.go#L87-L89).

Or do you mean calling `SetDefaults` should be left up to the tests?

reconciler/testing/reactions.go
gab-satchi: I explored the Reactor route as well. But in eventing, we use [prependReactors](https://github.com/knative/eventing/blob/abe0a570af624f98b9c2bf93cf921ba8225b201a/pkg/reconciler/testing/factory.go#L107-L117) every time and the Defaulting reactor would get called after the `ValidateCreates`.

gab-satchi: I think the bigger problem with the Reactor route is that the object doesn't get passed along the Reaction chain. So if we have a `SetDefaults` and `ValidateCreates`. Any modifications in SetDefaults won't be reflected when ValidateCreates is called. If SetDefaults returns true for `handled`, it will stop calling any other Reactors which will defeat the purpose of the validation.

gab-satchi: I'm sending out an update to my other PR that would avoid these changes

PR #2589, opened 2022-09-11
fix lint warnings

dprotaso:
- resolve k8s deprecation warnings
- drop use of ioutil


psschwei:
Looks like this one needs the update-codegen script run

psschwei (commented):
/lgtm

PR #2591, opened 2022-09-12
Allow logstream to stream logs from user namespace

mgencur:
* When streaming logs from user namespace, disable filtering lines so all lines are printed.

* More information on when this comes handy is available in Serving PR: https://github.com/knative/serving/pull/13293, together with usage example.
It will be very handy also in knative/eventing upgrade tests where multiple pods are started and the Sender pod is removed before the test ends.

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/test/logstream/interface.go
+++ b/test/logstream/interface.go
@@ -34,8 +34,6 @@ type (
 	// Canceler is the type of function returned when a logstream is started to be
 	// deferred so that the logstream can be stopped when the test is complete.
 	Canceler = logstreamv2.Canceler
-	// Callback is the type of generic function for printing log lines.
-	Callback = logstreamv2.Callback
 )
 
 type ti interface {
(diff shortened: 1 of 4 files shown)


mgencur:
The test failures look unrelated.

test/logstream/interface.go
```diff
@@ -31,7 +32,10 @@ import (
 
 // Canceler is the type of a function returned when a logstream is started to be
 // deferred so that the logstream can be stopped when the test is complete.
-type Canceler = logstreamv2.Canceler
+type (
+	Canceler = logstreamv2.Canceler
+	Callback = logstreamv2.Callback
```
skonto, line 37: I guess we need to document `Callback` too here.

mgencur: Oh, I see.

mgencur: Pushed.

test/logstream/v2/stream.go
```diff
@@ -34,21 +34,25 @@ import (
...
+		kc:          c,
+		namespaces:  namespaces,
+		keys:        make(map[string]Callback, 1),
+		filterLines: filterLines,
+		podPrefixes: podPrefixes,
 	}
 }
 
-func FromNamespace(ctx context.Context, c kubernetes.Interface, namespace string) Source {
+func FromNamespace(ctx context.Context, c kubernetes.Interface, namespace string, filterLines bool, podPrefixes ...string) Source {
```
dprotaso, line 48: This is a breaking API change. You could extend this method without breaking by using variadic functional options

see option 3 - https://golang.cafe/blog/golang-functional-options-pattern.html

test/logstream/v2/stream.go
```diff
@@ -34,21 +34,25 @@ import (
 	"knative.dev/pkg/ptr"
 )
 
-func FromNamespaces(ctx context.Context, c kubernetes.Interface, namespaces []string) Source {
+func FromNamespaces(ctx context.Context, c kubernetes.Interface, namespaces []string, filterLines bool, podPrefixes ...string) Source {
```
dprotaso, line 37: I don't have a good suggestion on what to do here that's not a breaking change.

test/logstream/interface.go
```diff
@@ -41,53 +46,69 @@ type ti interface {
...
 
-	return stream.Start(t)
+	return sysStream.Start(t, t.Logf)
+}
+
+// StartForUserNamespace begins streaming the logs from custom namespaces.
+// Filtering of log lines is disabled in this case so all lines will be printed
+// through the provided callback.
+// It returns a Canceler which must be called before the test completes.
+func StartForUserNamespace(t ti, callback Callback, namespace string, podPrefixes ...string) Canceler {
```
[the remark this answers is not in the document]
mgencur: OK. Good to know.

test/logstream/v2/stream.go
mgencur: This is addressed in the latest commit.

test/logstream/v2/stream.go
mgencur: Thanks. I have tried what you're suggesting and pushed a new commit. Please let me know what you think. I think this new way of constructing the logstream could replace the "FromNamespace" and "FromNamespaces" functions so we could mark them as deprecated (which I haven't done yet).

mgencur: I have also vendored the changes into https://github.com/knative/serving/pull/13293 so the exaple usage can be seen there.

mgencur:
Rebased. Just to be sure this works after some time.

@dprotaso Could you review this please? There's a commit which haven't been reviewed. See [comment](https://github.com/knative/pkg/pull/2591#discussion_r973972854)

mgencur:
@dprotaso Thank you!

PR #2596, opened 2022-09-20
Avoid second lister fetch during finalizer reconciliation.

mattmoor:
I noticed that the finalizer manipulation code was performing a second fetch of the resource from the lister cache to determine the differences in finalizers when the finalizers are being modify immediately above each callsite.

This changes both call sites to simply pass through the "desired finalizers" as a new parameter (instead of assigning them), which lets us use the unmodified resource and avoid the redundant lister cache fetch.

/kind cleanup

**Release Note**

```release-note
Eliminate a redundant lister cache fetch in the knative reconciler codegen.
```

**Docs**

```docs

```

cc @vaikas @dprotaso @tcnghia


vaikas (approved):
/lgtm
/approve

PR #2597, opened 2022-09-21
Define a global for `reconciler.DefaultTimeout`.

mattmoor:
Serving currently uses a fixed constant for the context timeout on reconciliation, which is causing some bad interactions with certain policy webhooks.

Based on feedback on https://github.com/knative/serving/pull/13323 I am hoisting this into a shared variable, so that folks can turn it into a flag, so it is configurable.

```release-note

```

```docs

```

/assign @vaikas

### DIFF
--- a/reconciler/reconcile_common.go
+++ b/reconciler/reconcile_common.go
@@ -30,6 +30,11 @@ import (
 
 const failedGenerationBump = "NewObservedGenFailure"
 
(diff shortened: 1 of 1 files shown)


evankanderson (commented):
/lgtm
/approve

evankanderson:
... and the kitchen sink grows. 😁

PR #2601, opened 2022-09-28
changeset: allow override the revision through ldflags

vdemeester:
# Changes

The main reason for this is to support cases where the `go build` happens on a non git folder but we want to pass a revision anyway.

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

/kind enhancement
/cc @imjasonh @dprotaso 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/changeset/commit.go
+++ b/changeset/commit.go
@@ -39,7 +39,10 @@ var (
 // The result will have a '-dirty' suffix if the workspace was not clean
 func Get() string {
 	once.Do(func() {
-		rev = get()
+		if rev == "" {
+			rev = get()
+		}
+		// It has been set through ldflags, do nothing
 	})
 
 	return rev
--- a/changeset/commit_test.go
+++ b/changeset/commit_test.go
@@ -73,6 +73,7 @@ func TestGet(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
+			rev = ""
 			once = sync.Once{}
 			readBuildInfo = func() (info *debug.BuildInfo, ok bool) {
 				return c.info, c.ok


changeset/commit.go
```diff
@@ -31,6 +31,12 @@ var (
 	once      sync.Once
 
 	readBuildInfo = debug.ReadBuildInfo
+
+	// Revision contains the revision of the binary.
+	// It is exported so that it can be overriden using ldflags.
+	// By default, it will use 'vcs.revision' property from the embedded build information
+	// Note: Use changeset.Get() to get it instead of through this variable.
```
imjasonh, line 38: Can ldflags only set an exported var? Can we just unexport this?

vdemeester: hum that was my initial though (because it's been a *long* time since I did that…). Let me test it out.

vdemeester: ah seems like it works with unexported, let me update this.

imjasonh:
Do we still need this if we go with https://github.com/tektoncd/pipeline/pull/5573 ?

vdemeester:
> Do we still need this if we go with [tektoncd/pipeline#5573](https://github.com/tektoncd/pipeline/pull/5573) ?

Yes, this is unrelated with tektoncd/pipeline#5573 👼🏼. Downstream, we are building tekton (and I guess knative as well), through a cachito cache/proxy, which means, when we build the project, we are not in a git repository (we have the information about it but we are not). Hence proposing this to be able to set the correct revision in that case 🙃.

imjasonh (approved):

vdemeester:
@dprotaso it would be something like

```
go build -ldflags="-X 'knative.dev/pkg/changeset.rev={the-commit-sha-or-something}'"
```

PR #2611, opened 2022-10-17
Filter reactor FieldErrors on validation to only error-level alerts.

evankanderson:
See https://github.com/knative/serving/pull/13399#issuecomment-1280033374

When we added warning-level `api.FieldErrors`, we missed filtering the warnings from `ValidateCreates` and `ValidateUpdates`.
For https://github.com/knative/serving/pull/13399, we want to be able to issue warnings on insecure default values for containers, but we don't want to force all our existing tests to use secure default values (and we want to validate that the insecure values will still work for controllers, etc).


mattmoor (approved):
/lgtm
/approve

PR #2612, opened 2022-10-17
deprecate pkg/ptr in favor of k8s.io/utils/pointer

imjasonh:
Mentioned here: https://github.com/knative-sandbox/eventing-rabbitmq/pull/954#discussion_r996188932

`ptr.Time` doesn't have an analogue in k8s-pointer, and it has [a few usages](https://cs.github.com/?scopeName=All+repos&scope=&q=ptr.Time+knative.dev%2Fpkg%2Fptr) that we'll want to get rid of if we proceed with deprecating this package.

/kind cleanup

**Release Note**

```release-note
Deprecate knative.dev/pkg/ptr in favor of k8s.io/utils/pointer, update funcs to point to those methods
```

cc @dprotaso @vaikas

### DIFF
--- a/depcheck/depcheck_test.go
+++ b/depcheck/depcheck_test.go
@@ -66,7 +66,10 @@ func TestExample(t *testing.T) {
 			"golang.org/x/sync/errgroup",
 		},
 		"knative.dev/pkg/ptr": {
+			"fmt",
+			"reflect",
 			"time",
+			"k8s.io/utils/pointer",
 		},
 	})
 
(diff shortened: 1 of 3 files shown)


dprotaso:
Seems like we should just use a generic function instead ;)

evankanderson:
You've got a bunch of warnings of usage of the deprecated package in this repo -- want to clean those up at the same time?

I'm inclined to stick with the Kubernetes `pointer` package and let that switch over to generics at the moment, unless there's already a generics-ready package.

PR #2613, opened 2022-10-20
Support istio *List types where Items is an array of pointer types

dprotaso:
Generally K8s types have a list type where the 'Items' member is an
array of structs.

ie. https://pkg.go.dev/k8s.io/api@v0.25.3/apps/v1#DeploymentList

```
type DeploymentList struct {
	...
	// Items is the list of Deployments.
	Items []Deployment `json:"items" protobuf:"bytes,2,rep,name=items"`
}
```

Istio is an exception where the list contains pointers to structs

```
type GatewayList struct {
	...
	Items       []*Gateway `json:"items" protobuf:"bytes,2,rep,name=items"`
}
```

To hint that the list types are pointers you can now pass the flag
`--lister-has-pointer-elem`.

Ideally we could infer this info by inspecting the types but
unfortunately the generator doesn't load this information

### DIFF
--- a/codegen/cmd/injection-gen/args/args.go
+++ b/codegen/cmd/injection-gen/args/args.go
@@ -29,6 +29,7 @@ type CustomArgs struct {
 	ExternalVersionsInformersPackage string
 	ListersPackage                   string
 	ForceKinds                       string
+	ListerHasPointerElem             bool
 }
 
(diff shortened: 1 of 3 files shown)


dprotaso:
`network: TestDialTLSWithBackoffSuccess`

/retest

AngeloDanducci (approved):

PR #2618, opened 2022-10-26
Add FieldError#WrappedErrors to allow dependent code to have access (readonly)

tcnghia:
# Changes

- :gift: Add FieldError#GetErrors to allow dependent code to have access (readonly)

/kind enhancement

**Release Note**

```release-note
NONE
```

**Docs**
```docs
NONE
```

### DIFF
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -256,8 +256,8 @@ func (fe *FieldError) normalized() []*FieldError {
 	return errors
 }
 
-// WrapErrors returns the value of the nested errors.
-func (fe *FieldError) WrapErrors() []FieldError {
+// WrappedErrors returns the value of the nested errors.
+func (fe *FieldError) WrappedErrors() []FieldError {
 	if fe == nil {
 		return nil
 	}
--- a/apis/field_error.go
+++ b/apis/field_error.go
@@ -83,11 +83,12 @@ var _ error = (*FieldError)(nil)
 
 // ViaField is used to propagate a validation error along a field access.
 // For example, if a type recursively validates its "spec" via:
-//   if err := foo.Spec.Validate(); err != nil {
-//     // Augment any field paths with the context that they were accessed
-//     // via "spec".
-//     return err.ViaField("spec")
-//   }
+//
+//	if err := foo.Spec.Validate(); err != nil {
+//	  // Augment any field paths with the context that they were accessed
+//	  // via "spec".
+//	  return err.ViaField("spec")
+//	}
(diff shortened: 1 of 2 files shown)


apis/field_error.go
```diff
@@ -251,6 +256,16 @@ func (fe *FieldError) normalized() []*FieldError {
 	return errors
 }
 
+// GetErrors returns the value of the nested errors.
+func (fe *FieldError) GetErrors() []FieldError {
```
mattmoor, line 260: I don't love the name `GetErrors()`.  Maybe `Expand()`?

Looking at precedent, Hashicorp's `go-multierror` uses `WrappedErrors()`: https://pkg.go.dev/github.com/hashicorp/go-multierror#Error.WrappedErrors

tcnghia: Done. I name it `WrappedErrors()` instead.

apis/field_error.go
```diff
@@ -251,6 +256,16 @@ func (fe *FieldError) normalized() []*FieldError {
 	return errors
 }
 
+// WrapErrors returns the value of the nested errors.
+func (fe *FieldError) WrapErrors() []FieldError {
```
mattmoor, line 260: I think the Hashicorp library called with `WrappedErrors` not `WrapErrors`, which reads differently.

tcnghia: Done. I meant to rename to `WrappedErrors`, thanks for catching.

apis/field_error.go
```diff
@@ -251,6 +256,16 @@ func (fe *FieldError) normalized() []*FieldError {
 	return errors
 }
 
+// WrappedErrors returns the value of the nested errors.
+func (fe *FieldError) WrappedErrors() []FieldError {
+	if fe == nil {
+		return nil
+	}
+	results := make([]FieldError, 0, len(fe.errors))
+	results = append(results, fe.errors...)
```
mattmoor, line 265: Are you sure the errors here are already flattened? 🤔

tcnghia: They are not flattened. That is done in `normalized()`. I've added a change to expose that as well so that we can access in a flattened way.

mattmoor: I think we may just want to call `normalize()` and possibly return that? 🤔

tcnghia: Make sense. I've changed `WrappedErrors = merge(normalized())`.

mattmoor (approved):
/lgtm
/approve
🎉

PR #2619, opened 2022-10-27
Cleanup: Turn multiple warnings into multiple warnings.

mattmoor:
:broom: Previously we lacked a public method for turning our `apis.FieldError` multi-error into a list of constituent error messages, so when we turned things into a webhook warning we simply used the combined serialization of all of the warnings.  Thanks to Nghia's recent change, we can now access the list of warnings to return as a list of errors.

This also has the nice property that we lower the risk of potential message truncation due to the assorted limits imposed by both upstream and the warning-header medium that's used to send these to clients.

/cc @vaikas @tcnghia @evankanderson 

/kind cleanup

**Release Note**

```release-note
Multiple apis.FieldError style "warnings" will now be returned as separate warnings from resource validation webhooks.
```

**Docs**

```docs

```

### DIFF
--- a/webhook/resourcesemantics/validation/validation_admit.go
+++ b/webhook/resourcesemantics/validation/validation_admit.go
@@ -84,7 +84,10 @@ func (ac *reconciler) Admit(ctx context.Context, request *admissionv1.AdmissionR
 		// below all overwrite `resp`, but the `defer` affords us one final
 		// crack at things.
 		defer func() {
-			resp.Warnings = []string{warnings.Error()}
+			resp.Warnings = make([]string, 0, len(warnings))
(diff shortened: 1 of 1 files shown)


tcnghia (approved):

mattmoor:
@evankanderson I'm not aware of any APIs it changes today since this only applied to warnings.

re: tests

The existing coverage for this looked good, but I can add an explicit test for multiple errors, if we want one, but it looks like a do have a bunch of tests in `validation_deprecated_test.go`

dprotaso:
There's a linting error here - can you add a comment to ignore it?

mattmoor:
https://github.com/knative/pkg/pull/2627

PR #2620, opened 2022-10-28
ListPointer flag should also apply to filtered informers

dprotaso:
Missed in https://github.com/knative/pkg/pull/2613

### DIFF
--- a/codegen/cmd/injection-gen/generators/filtered_informer.go
+++ b/codegen/cmd/injection-gen/generators/filtered_informer.go
@@ -41,6 +41,7 @@ type filteredInjectionGenerator struct {
 	injectionClientSetPackage   string
 	clientSetPackage            string
 	listerPkg                   string
+	listerHasPointerElem        bool
 }
 
 var _ generator.Generator = (*filteredInjectionGenerator)(nil)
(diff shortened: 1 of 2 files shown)


dprotaso:
cc @a7i

/assign @AngeloDanducci @nak3

codegen/cmd/injection-gen/generators/filtered_informer.go
```diff
@@ -246,7 +248,7 @@ func (w *wrapper) List(selector {{ .labelsSelector|raw }}) (ret []*{{ .type|raw
 		return nil, err
 	}
 	for idx := range lo.Items {
-		ret = append(ret, &lo.Items[idx])
+		ret = append(ret, {{if not .listerHasPointerElem}}&{{end}}lo.Items[idx])
```
nader-ziada, line 251: what happens to the append if the `not` condition is not true

dprotaso: Depending on the condition it will be either

`&lo.Items[idx]` 
or
`lo.Items[idx]` 

We had to do this to support code generation for istio types

nader-ziada: sounds good

PR #2621, opened 2022-10-28
CmpOpts overrides should be passed through

dprotaso:
This allows us to pass options to ignore unexpected fields in structs 

/assign @AngeloDanducci @nak3 

/cc @a7i


nader-ziada:
unit tests failed, but seems like not related to this change

/retest

PR #2623, opened 2022-10-29
add support for filtered informer in secret reconciler

nader-ziada:
# Changes

- to allow for using a filtered informer with the secret reconciler


nader-ziada:
to be used by  https://github.com/knative-sandbox/control-protocol/pull/225

PR #2639, opened 2022-11-15
Min K8s version is now 1.24

dprotaso:
- bump k8s deps to the latest patch
- bump min version of k8s to 1.24

/assign @psschwei


psschwei (commented):
/lgtm
/approve

Issue #2642, opened 2022-11-19
queue-proxy nil pointer dereference


mbrancato:
## What version of Knative?
1.8.0

## Expected Behavior

No panic

## Actual Behavior

```
2022/11/19 04:10:08 http: panic serving 10.32.40.190:37370: runtime error: invalid memory address or nil pointer dereference
goroutine 153447 [running]:
net/http.(*conn).serve.func1()
	net/http/server.go:1825 +0xbf
panic({0xf1c8c0, 0x1932df0})
	runtime/panic.go:844 +0x258
knative.dev/pkg/network/handlers.(*Drainer).resetTimer(0xc0001882a0)
	knative.dev/pkg@v0.0.0-20221011175852-714b7630a836/network/handlers/drain.go:199 +0x78
knative.dev/pkg/network/handlers.(*Drainer).ServeHTTP(0xc0001882a0, {0x1216e60, 0xc01a24bb20}, 0xc00f393500)
	knative.dev/pkg@v0.0.0-20221011175852-714b7630a836/network/handlers/drain.go:112 +0x139
golang.org/x/net/http2/h2c.h2cHandler.ServeHTTP({{0x120c4c0?, 0xc0001882a0?}, 0xc0000bb300?}, {0x1216e60, 0xc01a24bb20}, 0xc00f393500)
	golang.org/x/net@v0.0.0-20220722155237-a158d28d115b/http2/h2c/h2c.go:113 +0x49f
net/http.serverHandler.ServeHTTP({0x12113c0?}, {0x1216e60, 0xc01a24bb20}, 0xc00f393500)
	net/http/server.go:2916 +0x43b
net/http.(*conn).serve(0xc00b391860, {0x1217fc8, 0xc0001167e0})
	net/http/server.go:1966 +0x5d7
created by net/http.(*Server).Serve
	net/http/server.go:3071 +0x4db
```

## Steps to Reproduce the Problem

No really sure, I noticed this panic in my logs. Seems to happen during a pod startup.

PR #2645, opened 2022-11-22
Fix nil pointer deref in drain when it's reset

dprotaso:
Fixes: https://github.com/knative/pkg/issues/2642


dprotaso:
/retest

TestDialTLSWithBackoffSuccess flake

dprotaso:
I realize the change was introduced in 1.7+

PR #2657, opened 2022-11-30
Multi-controller leadership electors

mbaynton:
# Changes

This is an approach to constructing and running one leadership elector that promotes and demotes more than one controller/reconciler in tandem based on a single lease. Such capability is a must when it is desirable for several related reconcilers to run from the same process.

There are definitely cleaner ways to do this, but I'm not sure there are cleaner ways that don't break BC. For example currently an undecorated context will still result in construction and running of an `unopposedElector` when the controller is run, so that any `LeaderAware` reconcilers get promote()d. So to start running a controller without (re)running its elector, we have to decorate the context explicitly with `ErrElectorBuiltManually`.

Relates to bug https://github.com/knative/serving/issues/13447
This solution would probably also be a minor enhancement.

/kind bug
/kind enhancement

Still needs more unit and manual testing.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -465,21 +465,23 @@ func (c *Impl) RunContext(ctx context.Context, threadiness int) error {
 	if la, ok := c.Reconciler.(reconciler.LeaderAware); ok {
 		// Build and execute an elector.
 		le, err := kle.BuildElector(ctx, la, c.Name, c.MaybeEnqueueBucketKey)
-		if err != nil {
-			return err
-		}
-		if ib, ok := le.(kle.ElectorWithInitialBuckets); ok {
-			for _, b := range ib.InitialBuckets() {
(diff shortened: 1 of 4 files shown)


mbaynton:
Any suggestions for anything else that would be helpful to advance this?

psschwei:
See Dave's comments: https://github.com/knative/serving/issues/13447#issuecomment-1377447601

dprotaso:
Closing in favour of https://github.com/knative/serving/pull/13585

As mentioned in [the serving issue ](https://github.com/knative/serving/issues/13447#issuecomment-1377447601)
> The elector code is a bit of a foot cannon (https://github.com/knative/pkg/pull/2483) and I don't want to enable niche use cases by default.

I think it's worth capturing this requirement in an issue in case we do a redesign/extension of the leaderelection/controller libs

PR #2660, opened 2022-12-08
make sure ChildName produces a valid dns subdomain

### LINKED ISSUE #2551: kmeta/ChildName() function needs more finer trimming.
https://github.com/knative/pkg/blob/1ad02a5662d29355fbd23d13cce08b2677ece901/kmeta/names.go#L60

Sometimes resource generation does not work without validation of `.` as well as `-`.

It may be a good idea to modify the terminating character if it is not an Alphanumeric character.

related to https://github.com/knative-sandbox/net-contour/issues/785

### LINKED ISSUE #2659: kmeta package: ChildName function produces invalid resource names
## Expected Behavior
When I use [ChildName](https://github.com/knative/pkg/blob/main/kmeta/names.go) to create a truncated resource name, I expect the name produced to be a valid resource name in kubernetes.

## Actual Behavior
The function will produce an invalid resource name ending with `.`

I've set up a go playground example with my input that cause the problem: https://go.dev/play/p/HB42UPAWp7h

## Additional Info
This came up as an error in the net-contour controller that looks like this:
```
{"severity":"ERROR","timestamp":"2022-12-08T19:13:48.437648502Z","logger":"net-contour-controller","caller":"controller/controller.go:566","message":"Reconcile error","commit":"041f9e3","knative.dev/controller":"knative.dev.net-contour.pkg.reconciler.contour.Reconciler","knative.dev/kind":"networking.internal.knative.dev.Ingress","knative.dev/traceid":"ec12c38a-f1ac-4581-aae5-2350b60211a2","knative.dev/key":"cody/foo-java","duration":"18.23483ms","error":"HTTPProxy.projectcontour.io \"foo-java-contour-5f549ae3e6f584a5f33d069a0650c0d8foo-java.cody.\" is invalid: metadata.name: Invalid value: \"foo-java-contour-5f549ae3e6f584a5f33d069a0650c0d8foo-java.cody.\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","stacktrace":"knative.dev/pkg/controller.(*Impl).handleErr\n\tknative.dev/pkg@v0.0.0-20221011175852-714b7630a836/controller/controller.go:566\nknative.dev/pkg/controller.(*Impl).processNextWorkItem\n\tknative.dev/pkg@v0.0.0-20221011175852-714b7630a836/controller/controller.go:543\nknative.dev/pkg/controller.(*Impl).RunContext.func3\n\tknative.dev/pkg@v0.0.0-20221011175852-714b7630a836/controller/controller.go:491"}
```

net-contour uses ChildName here: https://github.com/knative-sandbox/net-contour/blob/main/pkg/reconciler/contour/resources/httpproxy.go#L275

KauzClay:
# Changes

- :bug: Fix bug
- Add check to remove `.` from end of name

/kind bug

Fixes #2659 
Fixes #2551 

**Release Note**
N/A

**Docs**
N/A

### DIFF
--- a/kmeta/names.go
+++ b/kmeta/names.go
@@ -19,7 +19,7 @@ package kmeta
 import (
 	"crypto/md5" //nolint:gosec // No strong cryptography needed.
 	"fmt"
-	"strings"
+	"regexp"
 )
 
 // The longest name supported by the K8s is 63.
@@ -29,6 +29,8 @@ const (
 	md5Len  = 32
 	head    = longest - md5Len // How much to truncate to fit the hash.
 )
+
+var isAlphanumeric = regexp.MustCompile(`^[a-zA-Z0-9]*$`)
 
 // ChildName generates a name for the resource based upon the parent resource and suffix.
 // If the concatenated name is longer than K8s permits the name is hashed and truncated to permit
@@ -55,12 +57,19 @@ func ChildName(parent, suffix string) string {
 			if d := longest - len(ret); d > 0 {
 				ret += suffix[:d]
 			}
-			// If due to trimming above we're terminating the string with a `-`,
-			// remove it.
-			return strings.TrimRight(ret, "-")
+			return makeValidName(ret)
 		}
 		//nolint:gosec // No strong cryptography needed.
 		n = fmt.Sprintf("%s%x", parent[:head-len(suffix)], md5.Sum([]byte(parent)))
 	}
 	return n + suffix
+}
+
+// If due to trimming above we're terminating the string with a non-alphanumeric
+// character, remove it.
+func makeValidName(n string) string {
+	for i := len(n) - 1; !isAlphanumeric.MatchString(string(n[i])); i-- {
+		n = n[:len(n)-1]
+	}
+	return n
 }
--- a/kmeta/names_test.go
+++ b/kmeta/names_test.go
@@ -17,10 +17,12 @@ limitations under the license.
 package kmeta
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 func TestChildName(t *testing.T) {
@@ -56,11 +58,20 @@ func TestChildName(t *testing.T) {
 		parent: "aaaa",
 		suffix: strings.Repeat("b---a", 20),
 		want:   "aaaa7a3f7966594e3f0849720eced8212c18b---ab---ab---ab---ab---ab",
-	}}
+	}, {
+		parent: strings.Repeat("a", 17),
+		suffix: "a.-.-.-.-.-.-.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+		want:   "aaaaaaaaaaaaaaaaaa1eb10dc911444f8434af83b7225442da",
+	},
+	}
 
 	for _, test := range tests {
 		t.Run(test.parent+"-"+test.suffix, func(t *testing.T) {
-			if got, want := ChildName(test.parent, test.suffix), test.want; got != want {
+			got, want := ChildName(test.parent, test.suffix), test.want
+			if errs := validation.IsDNS1123Subdomain(got); len(errs) != 0 {
+				t.Errorf(fmt.Sprintf("Invalid DNS1123 Subdomain %63s\n\n Errors: %v", got, errs))
+			}
+			if got != want {
 				t.Errorf("%s-%s: got: %63s want: %63s\ndiff:%s", test.parent, test.suffix, got, want, cmp.Diff(want, got))
 			}
 		})


kmeta/names_test.go
```diff
@@ -56,7 +56,12 @@ func TestChildName(t *testing.T) {
 		parent: "aaaa",
 		suffix: strings.Repeat("b---a", 20),
 		want:   "aaaa7a3f7966594e3f0849720eced8212c18b---ab---ab---ab---ab---ab",
-	}}
+	}, {
+		parent: strings.Repeat("a", 17),
+		suffix: "a.-.-.-.-.-.-.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+		want:   "aaaaaaaaaaaaaaaaaa1eb10dc911444f8434af83b7225442da",
```
evankanderson, line 62: It looks like we have no tests with capital letters. I suspect that's because we don't want names with capital letters.

Can you add a test on 63 or so that checks `IsDNS1123Subdomain` on the strings as well as a double-check?

(This will help if we add fuzzing tests later.)

evankanderson (commented):
Nice find!

A couple small efficiency and correctness suggestions now that I'm looking at the code.

kmeta/names_test.go
KauzClay: ah okay, let me make sure I understand...

so in addition to checking that want == got, we add an extra check to make sure got is a InDNS1123Subdomain?

kmeta/names.go
```diff
@@ -55,9 +55,13 @@ func ChildName(parent, suffix string) string {
 			if d := longest - len(ret); d > 0 {
 				ret += suffix[:d]
 			}
-			// If due to trimming above we're terminating the string with a `-`,
+			// If due to trimming above we're terminating the string with a non-alphanumeric character,
 			// remove it.
-			return strings.TrimRight(ret, "-")
+			for i := len(ret) - 1; !regexp.MustCompile(`^[a-zA-Z0-9]*$`).MatchString(string(ret[i])); i-- {
```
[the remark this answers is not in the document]
KauzClay: okay, so I pulled things out into a function that:
* removes "-"
* removes "."
* checks for capital letters and just to lowercases them 

Does that line up with what you're imagining?

kmeta/names.go
```diff
@@ -55,12 +56,20 @@ func ChildName(parent, suffix string) string {
...
 		//nolint:gosec // No strong cryptography needed.
 		n = fmt.Sprintf("%s%x", parent[:head-len(suffix)], md5.Sum([]byte(parent)))
 	}
 	return n + suffix
 }
+
+// If due to trimming above we're terminating the string with a non-alphanumeric character,
+// remove it. Also make sure things are lowercase
+func makeValidName(n string) string {
+	match := regexp.MustCompile(`^[a-zA-Z0-9]*$`)
```
evankanderson, line 70: Capital letters are still not allowed in valid names. (Sorry, I think I slipped this into another comment earlier.)

This also means you don't need `strings.ToLower`.

kmeta/names_test.go
```diff
@@ -56,11 +58,25 @@ func TestChildName(t *testing.T) {
...
+	}, {
+		parent: strings.Repeat("A", 17),
+		suffix: "A.-.-.-.-.-.-.AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
```
evankanderson, line 67: I think kubernetes resource names (which is implicitly the input to this function) will already be forced to lowercase.

I had meant that the _input_ to the function was constrained, and the lack of capital letters in the test was evidence of this.

Sorry for the confusion!

kmeta/names_test.go
```diff
@@ -56,11 +58,25 @@ func TestChildName(t *testing.T) {
...
+		want:   "aaaaaaaaaaaaaaaaae982551a43f641a37bf40c8feccd95f3a",
+	},
+	}
 
 	for _, test := range tests {
 		t.Run(test.parent+"-"+test.suffix, func(t *testing.T) {
-			if got, want := ChildName(test.parent, test.suffix), test.want; got != want {
+			got, want := ChildName(test.parent, test.suffix), test.want
+			errs := validation.IsDNS1123Subdomain(got)
+			if len(errs) != 0 {
```
evankanderson, line 76: Try this (a standard Go pattern) for collapsing these two lines:

```suggestion
			if errs := validation.IsDNS1123Subdomain(got); len(errs) != 0 {
```

One thing to watch out for in this pattern is that moving the declaration of `errs` into the `if` statement makes `errs` scoped to the control block (the `if` statement). This can result in variable shadowing in some cases, though it's not relevant here.

evankanderson (commented):
I think you ended up doing more than I meant, sorry about that!

kmeta/names.go
```diff
@@ -55,12 +55,20 @@ func ChildName(parent, suffix string) string {
...
 		//nolint:gosec // No strong cryptography needed.
 		n = fmt.Sprintf("%s%x", parent[:head-len(suffix)], md5.Sum([]byte(parent)))
 	}
 	return n + suffix
 }
+
+// If due to trimming above we're terminating the string with a non-alphanumeric
+// character, remove it.
+func makeValidName(n string) string {
+	match := regexp.MustCompile(`^[a-zA-Z0-9]*$`)
```
vagababov, line 69: Also you probably need regexp compiled once, no need to do it per name?

kmeta/names_test.go
KauzClay: no worries, thanks for clarifying!

kmeta/names.go
KauzClay: ah good point, I can pull it out

dprotaso:
> I'm not sure what to make of this failing downstream unit test on "eventing-natss". It doesn't seem related to the changes in this PR
> 
> https://github.com/knative/pkg/actions/runs/3659935995/jobs/6186498978

natts is broken and it seems like people aren't maintaining it - we should probably remove it. But for this PR you can ignore it

dprotaso:
/approve

leaving LGTM for others

evankanderson (commented):
/lgtm

PR #2662, opened 2022-12-16
Allow overriding webhook secret data keys

skonto:
# Changes

- Downstream we are using [knative operator's webhook](https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/cmd/operator/main.go#L30) that is passed in shared main for creation but does not utilize certificates from pkg. OLM manages the cert lifecycle and is generated due to webhook configuration at the [csv](https://github.com/openshift-knative/serverless-operator/blob/main/olm-catalog/serverless-operator/manifests/serverless-operator.clusterserviceversion.yaml#L865). The generated secret has the following data:
```
data:
  olmCAKey: ...
  tls.crt: ..
  tls.key: ... 
```
  We would like to be able to override the default keys so in general any external webhook that extends 
  webhook.AdmissionController or webhook.ConversionController can be used.
- The caCert  is not used when constructing a new Webhook so no need to override. Webhooks usually fetch it in order inject it at the webhook configuration via the webhook's reconciler logic.
- Making generic the setup of existing keys eg in MakeSecret is out of scope.

/kind enhancement

### DIFF
--- a/webhook/certificates/resources/secret.go
+++ b/webhook/certificates/resources/secret.go
@@ -26,9 +26,11 @@ import (
 
 const (
 	// ServerKey is the name of the key associated with the secret's private key.
-	ServerKey = "server-key.pem"
+	ServerKey    = "server-key.pem"
+	ServerKeyEnv = "KNATIVE_SECRET_WEBHOOK_SERVER_KEY"
 	// ServerCert is the name of the key associated with the secret's public key.
-	ServerCert = "server-cert.pem"
+	ServerCert    = "server-cert.pem"
+	ServerCertEnv = "KNATIVE_SECRET_WEBHOOK_SERVER_CERT"
 	// CACert is the name of the key associated with the certificate of the CA for
 	// the keypair.
 	CACert = "ca-cert.pem"
--- a/webhook/certificates/resources/secret.go
+++ b/webhook/certificates/resources/secret.go
@@ -27,12 +27,8 @@ import (
 const (
 	// ServerKey is the name of the key associated with the secret's private key.
 	ServerKey = "server-key.pem"
-	// ServerKeyEnv is the env var name for the webhook secret's key eg. `tls.key`.
-	ServerKeyEnv = "KNATIVE_WEBHOOK_SERVER_KEY"
 	// ServerCert is the name of the key associated with the secret's public key.
 	ServerCert = "server-cert.pem"
-	// ServerCertEnv is the env var name for the webhook secret's ca data key eg. `tls.crt`.
-	ServerCertEnv = "KNATIVE_WEBHOOK_SERVER_CERT"
 	// CACert is the name of the key associated with the certificate of the CA for
 	// the keypair.
 	CACert = "ca-cert.pem"
--- a/webhook/certificates/resources/secret.go
+++ b/webhook/certificates/resources/secret.go
@@ -18,6 +18,7 @@ package resources
 
 import (
 	"context"
+	"os"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


skonto:
@dprotaso gentle ping

webhook/certificates/resources/secret.go
```diff
@@ -26,9 +26,11 @@ import (
 
 const (
 	// ServerKey is the name of the key associated with the secret's private key.
-	ServerKey = "server-key.pem"
+	ServerKey    = "server-key.pem"
+	ServerKeyEnv = "KNATIVE_SECRET_WEBHOOK_SERVER_KEY"
```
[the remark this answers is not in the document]
hhk7734: How about `WEBHOOK_CERTS_SECRET_SERVER_KEY/CERT`? Cause I use `WEBHOOK_CERTS_SECRET_NAME` in #2685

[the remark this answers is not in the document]
skonto: I like the KNATIVE prefix tbh to separate from other env vars that could co-exist. Other than that its up to @dprotaso no strong preference.

webhook/certificates/resources/secret.go
```diff
@@ -27,8 +28,12 @@ import (
 const (
 	// ServerKey is the name of the key associated with the secret's private key.
 	ServerKey = "server-key.pem"
+	// ServerKeyEnv is the env var name for the webhook secret's key eg. `tls.key`.
+	ServerKeyEnv = "KNATIVE_WEBHOOK_SERVER_KEY"
 	// ServerCert is the name of the key associated with the secret's public key.
 	ServerCert = "server-cert.pem"
+	// ServerCertEnv is the env var name for the webhook secret's ca data key eg. `tls.crt`.
+	ServerCertEnv = "KNATIVE_WEBHOOK_SERVER_CERT"
```
dprotaso, line 36: I realize now the env vars don't influence the certificates & keys being created. Thus I don't think these overrides should be in this package.


This package is for _creating_ and _updating_ certificates

webhook/webhook.go
```diff
@@ -148,13 +148,13 @@ func New(
 					logger.Errorw("failed to fetch secret", zap.Error(err))
 					return nil, nil
 				}
-
-				serverKey, ok := secret.Data[certresources.ServerKey]
+				sKey, sCert := certresources.GetSecretDataKeyNamesOrDefault()
```
dprotaso, line 151: Are you creating custom webhooks binaries where you drop the certificate controller? 

If so it might be better to add these overrides to the `webhook.Options` and when they're empty default them to the keys in the `certificates/resources` package.

webhook/certificates/resources/secret.go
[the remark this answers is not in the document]
dprotaso: We don't have `KNATIVE_` prefixes for other environment variables ie. `SYSTEM_DOMAIN` so I suggest dropping the prefix so that we remain consistent

webhook/certificates/resources/secret.go
skonto: Ok will do the webhook options approach thanks.

webhook/webhook.go
skonto: Makes sense the env var thingy is a bit too intrusive.

skonto:
@dprotaso gentle ping this is ready for review.

webhook/webhook.go
```diff
@@ -58,6 +58,14 @@ type Options struct {
 	// If no SecretName is provided, then the webhook serves without TLS.
 	SecretName string
 
+	// ServerKeyEnv is the name for the webhook secret's data key eg. `tls.key`.
+	// Default value is `server-key.pem` if no value is passed.
+	ServerKey string
+
+	// ServerCertEnv is the name for the webhook secret's ca data key eg. `tls.crt`.
+	// Default value is `server-cert.pem` if no value is passed.
+	ServerCert string
```
dprotaso, line 67: Comment doesn't match var name

Maybe `SecretCertifcateName` ?

webhook/certificates/resources/secret.go
```diff
@@ -59,3 +59,17 @@ func MakeSecretInternal(ctx context.Context, name, namespace, serviceName string
...
+	serverCert = ServerCert
+
+	if sKey != "" {
+		serverKey = sKey
+	}
+	if sCert != "" {
+		serverCert = sCert
+	}
+	return serverKey, serverCert
+}
```
dprotaso, line 75: seems like this could be a private helper in the `webhook` package

webhook/webhook_integration_test.go
```diff
@@ -105,8 +105,69 @@ func TestMissingContentType(t *testing.T) {
 	metricstest.CheckStatsNotReported(t, requestCountName, requestLatenciesName)
 }
 
+func TestMissingContentTypeCustomSecret(t *testing.T) {
+	defaultOptions := newCustomOptions()
+	certresources.MakeSecret = customSecretWithOverrides
```
dprotaso, line 110: This might cause flakiness when running tests in parallel.

We should just have this test perform it's own setup (ie. create it's own cert etc)

Also the test name and assertions is a bit misleading `TestMissingContentTypeCustomSecret` - we really want to assert that the right certificate is presented. Thus we should skip stuff that's not relevant

skonto: I copied the logic in [certificates_test.go](certresources.MakeSecret = certresources.MakeSecretInternal) so probably that is unsafe too?

skonto: Anyway I updated the test to be more compact and have its own setup.

skonto:
@dprotaso gentle ping.

PR #2663, opened 2023-01-04
Simplify the test/upgrade package

### LINKED ISSUE #2421: Simplify the test/upgrade package
This whole package is really hard to review and I think part of it is because we're mixing two loggers.  Related I think we can simplify the need for this PR if we map continous tests to golang tests differently.

Take serving for example which has the following structure

```
test/upgrade.TestServingUpgrades
test/upgrade.TestServingUpgrades/InstallingBase/*
test/upgrade.TestServingUpgrades/ContinualTests
test/upgrade.TestServingUpgrades/ContinualTests/SetupProbeTest
test/upgrade.TestServingUpgrades/PreUpgradeTests/*
test/upgrade.TestServingUpgrades/UpgradeWith/*
test/upgrade.TestServingUpgrades/PostUpgradeTests/*
test/upgrade.TestServingUpgrades/DowngradeWith/*
test/upgrade.TestServingUpgrades/PostDowngradeTests/*
test/upgrade.TestServingUpgrades/VerifyContinualTests
test/upgrade.TestServingUpgrades/VerifyContinualTests/ProbeTest
```

The continous probe test is 'split' across two golang tests instances - `ContinualTests/SetupProbeTest`  & `VerifyContinualTests/ProbeTest`.  Because of this split we don't have access to a test logger during the background probing. Thus I believe that's why we have a separate logger.

The alternative would be to run the continual tests in a single `t.Run()` - then the test logger is accessible during the life of the entire 'logical' test
ie.
```
test/upgrade.TestServingUpgrades/Continual/ProbeTest
```

In this golang test instance we would run setup, the background probing and verify.

_Originally posted by @dprotaso in https://github.com/knative/pkg/pull/2293#pullrequestreview-866991548_

mgencur:
# Changes

* Remove the "Verify" phase of continual tests. Both Setup and Verify phases now run within a single t.Run. This is a solution based on the suggestions in the linked issue above.
* Use t.Parallel() to run the following two sets of tests in parallel:
   - Continual tests
   - Upgrade, Post-upgrade tests, Downgrade, Post-downgrade tests
* Remove the complicated Logger implementation
* Simplify StopEvent, it is a simple struct{} now. The required testing.T and logger are passed through BackgroundContext
* Remove the need to send a signal back via StopEvent.Finished channel. As both Setup and Verify now run both
in the same t.Run the Verify phase is now blocking until the simplified stop event is sent from the top-level test at the end.
* Remove deprecated Configuration.Log in favor of Configuration.LogConfig

The results of this changes and how they affect tests can be found in experimental PR for Serving: https://github.com/knative/serving/pull/13568
That PR describes test structure for different cases (all tests passed, failures in setup, failures in verify, etc.)
This change will allow for log streaming from system namespaces as manifested in the PR above as well.
Together with https://github.com/knative/pkg/pull/2591, this PR will help gather all related logs from system namespaces and test namespaces and print them within the test scope.

The successful test run looks like this:
```
PASS test/upgrade.TestServingUpgrades/InstallingBase/CreateTestNamespace (0.26s)
PASS test/upgrade.TestServingUpgrades/InstallingBase (0.26s)
PASS test/upgrade.TestServingUpgrades/PreUpgradeTests/ServicePreUpgradeTest (5.80s)
PASS test/upgrade.TestServingUpgrades/PreUpgradeTests/BYORevisionPreUpgradeTest (6.54s)
PASS test/upgrade.TestServingUpgrades/PreUpgradeTests/ServicePreUpgradeAndScaleToZeroTest (30.84s)
PASS test/upgrade.TestServingUpgrades/PreUpgradeTests/InitialScalePreUpgradeTest (65.71s)
PASS test/upgrade.TestServingUpgrades/PreUpgradeTests (0.00s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/UpgradeWith/ServingHead (54.88s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/UpgradeWith (54.88s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/PostUpgradeFromScaleToZeroTest (10.29s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/CRDStoredVersionPostUpgradeTest (2.46s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/InitialScalePostUpgradeTest (0.36s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/BYORevisionPostUpgradeTest (0.38s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/CreateNewServicePostUpgradeTest (3.88s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/ServicePostUpgradeTest (5.95s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests (12.76s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/DowngradeWith/ServingLatestRelease (48.45s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/DowngradeWith (48.45s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostDowngradeTests/ServicePostDowngradeTest (5.02s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostDowngradeTests/CreateNewServicePostDowngradeTest (7.08s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostDowngradeTests (0.00s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade (129.12s)
PASS test/upgrade.TestServingUpgrades/Parallel/ProbeTest (132.13s)
PASS test/upgrade.TestServingUpgrades/Parallel/AutoscaleSustainingTest (145.18s)
PASS test/upgrade.TestServingUpgrades/Parallel/AutoscaleSustainingWithTBCTest (143.17s)
PASS test/upgrade.TestServingUpgrades/Parallel (31.71s)
PASS test/upgrade.TestServingUpgrades (227.30s)
PASS test/upgrade
```

/kind enhancement

Fixes #2421

**Release Note**

```release-note

```

**Docs**

```docs

```


mgencur:
There's a unit test failing:
```
--- FAIL: TestSuiteExecuteWithComplete (0.02s)
    functions_test.go:140: expected "Serving - probing functionality..." is not in: INFO	🏃 Running upgrade test suite...
```
From the logs it looks like this is caused by thread starvation:
The Serving and Eventing continual tests run in parallel and while the Eventing test logs this several times:`DEBUG	Eventing continual test - probing functionality...`, the Serving test doesn't.
The Serving test is started and then immediately stopped:
```
INFO	Running Serving continual test
DEBUG	Serving received a stop event
INFO	Stopping and verify of Serving continual test
```
I've been able to reproduce this when setting explicitly `export GOMAXPROCS=2`. It works nicely with  GOMAXPROCS=3 and higher.
Also, I was not able to reproduce this when running Serving upgrade test suite using this new style. Either the VM has more than 2 cores or there are blocking calls in Serving upgrades tests which are not here in the unit tests ((syscalls, channel reads/writes, mutex locks, etc) which would yield control of test execution to the other thread.

mgencur:
Seems that unit tests can sometimes hang when using `GOMAXPROCS=2`. I was able to reproduce locally, it goes away with `GOMAXPROCS=3` again. I will check why this happens.

mgencur:
So, I think it is as simple as this with the current implementation: The `GOMAXPROCS` must be set to at least `number-of-parallel-continual-tests + 1` to prevent deadlocks.

In unit tests we have at most 2 continual tests running in parallel so `GOMAXPROCS=3` is enough. The problem is more general though and I was able to reproduce the problem even with 8 parallel continual tests and `GOMAXPROCS=8`, setting `GOMAXPROCS=9` fixes that again.

Right now, I don't have a good workaround for this. Maybe it is acceptable to request certain GOMAXPROCS? Not sure...

mgencur:
The failure in downstream knative/eventing tests is being resolved here: https://github.com/knative/eventing/pull/6663

test/upgrade/background_verification_test.go
```diff
@@ -51,42 +39,14 @@ func TestSkipAtBackgroundVerification(t *testing.T) {
...
 					},
 				),
-				upgrade.NewBackgroundVerification("ShouldNotBeSkipped",
-					func(c upgrade.Context) {
-						c.Log.Info("Setup 2")
-					},
-					func(c upgrade.Context) {
-						c.Log.Info("Verify 2")
-					},
-				),
```
dprotaso, line 89: Are we dropping this because it's covered in another test?

test/upgrade/types.go
```diff
@@ -81,45 +81,29 @@ type Context struct {
...
-// to indicate that is should stop it's operations and validate results using
-// passed T. User should use Finished channel to signalize upgrade suite that
-// all stop & verify operations are finished and it is safe to end tests.
-type StopEvent struct {
-	T        *testing.T
-	Finished chan<- struct{}
-	name     string
-	logger   *zap.SugaredLogger
+	T    *testing.T
+	Log  *zap.SugaredLogger
```
dprotaso, line 89: Should each operation get a logger that's backed by the testing.T - ideally it's not a global test object but the one scoped under `t.Run(name, func(t *testing.T))`?

https://pkg.go.dev/go.uber.org/zap/zaptest#NewLogger

Then I don't think you have to worry about printing on failures.

dprotaso (commented):
I think having `Parallel` in the test name is confusing. Since it gave me the impression that some serial steps were being executed concurrently. I think in the go playground example I made we didn't need this. Is it necessary for some reason?

ie. from your example `UpgradeWith` isn't parallel with `PostUpgradeTests`
```
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/UpgradeWith/ServingHead (54.88s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/UpgradeWith (54.88s)
PASS test/upgrade.TestServingUpgrades/Parallel/UpgradeDowngrade/PostUpgradeTests/PostUpgradeFromScaleToZeroTest (10.29s)
```

I think something like `Continual` or even `Background` gave a better sense of what tests were running concurrently while the upgrade was happening.

test/upgrade/background_verification_test.go
mgencur: Two reasons:
1) It's covered by other tests, for example [TestSuiteExecuteEmpty](https://github.com/knative/pkg/blob/main/test/upgrade/functions_test.go#L100) tests successful pass through the handler/BackgroundOperation. And [TestFailAtBackgroundVerification](https://github.com/knative/pkg/blob/main/test/upgrade/background_verification_test.go#L131) uses the NewBackgroundVerification API which is a simple wrapper though (so no extensive testing needed for that).
2) It's also removed to keep just one "continual" test so that it can pass in Github action that runs the unit tests with just 2 logical CPUs. And I don't see an easy way to increase GOMAXPROCS because it uses a shared github action.
From my earlier comments you can see that with 2 CPUs the test suite can only run on continual test in parallel with the main flow.

mgencur:
> I think having Parallel in the test name is confusing. Since it gave me the impression that some serial steps were being executed concurrently. I think in the go playground example I made we didn't need this. Is it necessary for some reason?

Yes it is necessary because when someone is calling `s.Execute()` from their test function it would run the parallel parts at the end of their top-level test, and this might be after they want to do some verifications after `s.Execute()`. It has the same role as `Main` in your [original go playground example](https://go.dev/play/p/5dagczP8er8).

The idea was that the direct children of "Parallel" are running in parallel (i.e. individual continual tests and "UpgradeDowngrade" itself). And children of "UpgradeDowngrade" might again run serially within UpgradeDowngrade.

If the above is confusing I could use the "Main" from your example, in that case the flow would look like this:
```
--- PASS: TestUpgrade (12.00s)
    --- PASS: TestUpgrade/Main (8.00s)
        --- PASS: TestUpgrade/Main/Base (2.00s)
        --- PASS: TestUpgrade/Main/PreUpgrade (2.00s)
        --- PASS: TestUpgrade/Main/UpgradeDowngrade (4.00s)
            --- PASS: TestUpgrade/Main/UpgradeDowngrade/Upgrade (2.00s)
            --- PASS: TestUpgrade/Main/UpgradeDowngrade/Downgrade (2.00s)
        --- PASS: TestUpgrade/Main/Continual_Test_1 (6.00s)
        --- PASS: TestUpgrade/Main/Continual_Test_0 (6.00s)
```
instead of the confusing
```
--- PASS: TestUpgrade (12.00s)
    --- PASS: TestUpgrade/Base (2.00s)
    --- PASS: TestUpgrade/PreUpgrade (2.00s)
    --- PASS: TestUpgrade/Parallel (4.00s)
        --- PASS: TestUpgrade/Parallel/UpgradeDowngrade (4.00s)
            --- PASS: TestUpgrade/Parallel/UpgradeDowngrade/Upgrade (2.00s)
            --- PASS: TestUpgrade/Parallel/UpgradeDowngrade/Downgrade (2.00s)
        --- PASS: TestUpgrade/Parallel/Continual_Test_1 (6.00s)
        --- PASS: TestUpgrade/Parallel/Continual_Test_0 (6.00s)
```

I think the individual background/continual tests have to run at the same level as UpgradeDowngrade because we have the Setup part which need to to run first. If I enclosed the individual continual tests within another "Background" testing.T it would run the setup phase in parallel with UpgradeDowngrade.

test/upgrade/types.go
mgencur: Great point! This was actually a bug in my code because it was not printing the logs even on failure.
Pushing another commit now that fixes it. I've tested this new change in [serving run](https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_serving/13568/upgrade-tests_serving_main/1611279118299566080/build-log.txt) and it now properly logs stuff like `INFO	3) 🔄 Starting continual tests. 3 tests are registered.` and similar.

dprotaso:
I think it's possible to drop that `Main` `Parallel` wrapper

https://go.dev/play/p/rCLcJx68GJV

mgencur:
@dprotaso No. Imagine for example the [TestSuiteExecuteWithComplete](https://github.com/mgencur/pkg/blob/simplify_logging/test/upgrade/functions_test.go#L112) test which runs `suite.Execute(c)`. The "Continuous" and "Flow" tests from your last example (more precisely, their parallel parts) will only run at then end of the enclosing test `TestSuiteExecuteWithComplete` which is actually after a few other assertions that are done in the test after running `suite.Execute(c)`. The parallel parts won't run within the scope of "Execute" method but will overflow to the enclosing test. This is why we need another `t.Run("...")` wrapper to make the parallel parts run within the `Execute` function. In my current solution in this PR, the "Parallel" test has this role. I gave a few example structures that support this. Is it clearer? Or I'm missing something  :)

dprotaso:
Thanks for the clarification. 

I'd probably just change

`Parallel` => `Run`
`UpgradeDowngrade` => `Steps`

mgencur:
> I'd probably just change

> Parallel => Run
> UpgradeDowngrade => Steps

OK. Done.

PR #2670, opened 2023-01-23
Make SetDefaults of Destination duck type nil safer

creydr:
To use `SetDefaults` on the `Destination` type, the caller must check that the destination is not nil currently:

e.g.
```
if subscriber != nil {
    subscriber.SetDefaults(ctx)
}
```

Some other `SetDefaults` implementations, check this internally to prevent panics. E.g.: https://github.com/knative/eventing/blob/7e899fd166deb40c885d3fc4fb9e2901375a7bff/pkg/apis/duck/v1/delivery_defaults.go#L22-L24
This makes it easier to directly call `SetDefaults` without surrounding it with `nil` checks.

This PR addresses it and includes the "nil check" into the `SetDefaults` methods of a `Destination`

### DIFF
--- a/apis/duck/v1/destination.go
+++ b/apis/duck/v1/destination.go
@@ -73,6 +73,10 @@ func (d *Destination) GetRef() *KReference {
 }
 
 func (d *Destination) SetDefaults(ctx context.Context) {
+	if d == nil {
+		return
+	}
+
 	if d.Ref != nil && d.Ref.Namespace == "" {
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm

pierDipi:
Can we add a unit test ?

vaikas:
Thanks!
/lgtm
/approve

PR #2671, opened 2023-01-26
Add health checks

skonto:
# Changes

- Sets up a handler for readiness/liveness probes to be enabled on demand in sharedMain. Moves work here from:
https://github.com/knative/serving/pull/13563
- Checks if start up work has been done.


skonto:
@dprotaso @evankanderson gentle ping for review.

evankanderson:
We should not have the same problem as the controller-runtime issue, because we always start controllers, even if they don't win the leader election.

I agree that the check should start up and run independently of the controller startup, but it seems like we shouldn't return healthy until informer caches have been filled (which, for example, can be blocked by incorrect RBAC permissions).

evankanderson:
I'd also advocate for having health endpoints always enabled.  If you don't want to use them, they should be minimal overhead.

Ideally, we could provide this same server to the Prometheus metrics implementation and share a port, but I don't want to force this behind refactoring that big ball of mud.

skonto:
> We should not have the same problem as the controller-runtime issue, because we always start controllers, even if they don't win the leader election.

We do but they dont own any bucket right? That technically means they are not ready no?

skonto:
Will update the PR.

network/health_check.go
```diff
@@ -0,0 +1,90 @@
...
+		// start the web server on port and accept requests
+		log.Printf("Probes server listening on port %s", port)
+
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			log.Fatal(err)
+		}
+	}()
+}
+
+func newHealthCheck(sigCtx context.Context) func() error {
```
evankanderson, line 58: Any thoughts about how we might let users customize this later?

network/health_check.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+// healthHandler handles responding to kubelet probes with a provided health check.
+type healthHandler struct {
+	HealthCheck func() error
+}
+
+func (h *healthHandler) handle(w http.ResponseWriter, r *http.Request) {
+	if IsKubeletProbe(r) {
```
evankanderson, line 80: Checking `IsKubeletProbe` here keeps this health checker from being used for e.g. proxy or apiserver health checks.

I know we may not need them today, but I'm wondering if we should be passing the request to the health checking function rather than doing this check here.  i.e. `func(http.Request) error`.  This might be a half-baked idea, though.

network/health_check.go
```diff
@@ -0,0 +1,90 @@
...
+		if err := h.HealthCheck(); err != nil {
+			log.Println("Healthcheck failed: ", err.Error())
```
evankanderson, line 82: Should this extract the logger from the context and use that?

evankanderson (commented):
Thanks for doing this! I'm happy to build this up incrementally.

It feels to me like it might make sense to have a registration mechanism; something like `AddReadinessCheck()` and `AddLivenessCheck()`.

I'm also wondering if these interfaces should live in `sharedmain` instead of `network`, but I think we could work that out now or later.

I am going to insist on switching the usage of `log` for getting the `logger ` from the Context, though.

network/health_check.go
skonto: We could refactor this to allow pass a func in ServeHealthProbes(ctx context.Context) or context.

network/health_check.go
```diff
@@ -0,0 +1,155 @@
...
+
+type addUserReadinessProbeKey struct{}
+
+// WithUserReadinessProbe signals to ServeHealthProbes that it should set explicitly a readiness probe.
+func WithUserReadinessProbe(ctx context.Context) context.Context {
+	return context.WithValue(ctx, addUserReadinessProbeKey{}, struct{}{})
+}
+
+// UserReadinessProbe checks if user has explicitly requested to set a readiness probe in the related context.
+func UserReadinessProbe(ctx context.Context) bool {
```
skonto, line 127: We can extend with a custom url in the future to change the default path too.

skonto:
@evankanderson pls review. Updates:
a) Added a mechanism to register only specific probe types.
b) made the health handler easy to override
c) added unit tests

Not sure how queue proxy could utilize this because drainer is used as part of a handler chain. The scope of this for now is for regular deployments of Knative control plane but probably we can refactor further.

evankanderson:
Yeah, I'm not deeply concerned about using this for `queue-proxy` or `activator`, which will both be a little funny, but I just didn't want to over-specialize and then have to walk it back.

At the same time, it would be nice if we could use this for e.g. eventing brokers, and have them fail readiness or liveness if they can't reach their storage backends.  That probably needs that we need something like an ability to register health checks in the `injection.ControllerConstructor`, which probably means adding the healthcheck registration(s) to the context before calling the constructors.

evankanderson:
(Reviewing now)

network/health_check.go
evankanderson: I'm struggling to figure out why you'd want to disable these on the server.  I understand why you might want to remove them from the manifest, but what's the rationale for not registering them with the HTTPServer?

network/health_check.go
```diff
@@ -0,0 +1,155 @@
...
+		mux.HandleFunc("/readiness", handler.ReadinessProbeRequestHandle)
+	}
+	if UserLivenessProbe(ctx) {
+		mux.HandleFunc("/health", handler.LivenessProbeRequestHandle)
+	}
+	// Set both probes if user does not want to explicitly set only one.
+	if !UserReadinessProbe(ctx) && !UserLivenessProbe(ctx) {
+		mux.HandleFunc("/readiness", handler.ReadinessProbeRequestHandle)
+		mux.HandleFunc("/health", handler.LivenessProbeRequestHandle)
+	}
```
evankanderson, line 78: This feels complicated -- could it simplify to something like:

```golang
readiness := getReadinessHandlerOrDefault(ctx)
liveness := getLivenessHandlerOrDefault(ctx)

mux := http.NewServeMux()
mux.HandleFunc("/readiness", readiness)
mux.HandleFunc("/liveness", liveness)
return mux
```

And then the `...OrDefault(ctx)` calls would get the correct handler function.

network/health_check.go
```diff
@@ -0,0 +1,155 @@
...
+// HealthCheckHandler allows to set up a handler for probes. User can override the default handler
+// with a custom one.
+type HealthCheckHandler struct {
+	ReadinessProbeRequestHandle http.HandlerFunc
+	LivenessProbeRequestHandle  http.HandlerFunc
+}
+
+func newDefaultProbesHandle(sigCtx context.Context) http.HandlerFunc {
+	logger := logging.FromContext(sigCtx)
+	once := sync.Once{}
```
evankanderson, line 91: Why do we need this `once`?

network/health_check.go
```diff
@@ -0,0 +1,155 @@
...
+				once.Do(func() {
+					logger.Info("Signal context canceled")
+				})
+				return errors.New("received SIGTERM from kubelet")
+			default:
+				return nil
+			}
+		}
+
+		if IsKubeletProbe(r) {
```
evankanderson, line 106: I'm still wondering if we need this `IsKubeletProbe` check.

injection/sharedmain/main.go
```diff
@@ -324,6 +330,18 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
...
 
+type healthProbesDisabledKey struct{}
+
+// WithHealthProbesDisabled signals to MainWithContext that it should disable default probes (readiness and liveness).
+func WithHealthProbesDisabled(ctx context.Context) context.Context {
+	return context.WithValue(ctx, healthProbesDisabledKey{}, struct{}{})
+}
+
+// HealthProbesDisabled checks if default health checks are disabled in the related context.
+func HealthProbesDisabled(ctx context.Context) bool {
```
evankanderson, line 341: Does this need to be an exported method, or can it be internal-only?

evankanderson (commented):
I _like_ the idea of enabling health probes -- is there any reason not to have them on all the time (it might simplify the code)?

I'd also sort of like to be able to say:

```
func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
  ...
  checker := func (r http.Request) error {
    // pretend we have appropriate locking
    return impl.connectionError
  }

  // internally, calls getReadinessRegisterFunc(ctx).Register(checker)
  injection.AddReadiness(ctx, checker)

  return impl
}
```

Of course, `checker` could also be `impl.IsHealthy` or the like.

I'm okay with either a `func (http.Request) error` or a `func () error` signature -- whichever seems less of a pain in the common case.


/hold
/lgtm

If you like this interface,

injection/sharedmain/main.go
skonto: Should be private.

network/health_check.go
skonto: We want to reject a request that is not coming from kubelet. The validation is weak but it seemed to me it is better than nothing.

skonto:
> I like the idea of enabling health probes -- is there any reason not to have them on all the time (it might simplify the code)?

The probes are on by default, the default ones are used.

network/health_check.go
skonto: Ok let me work on the interface and see what I can come up with.

network/health_check.go
skonto: Activator has the same approach too for its healthcheck. So since the probe handler could be part of a handler chain (we might want to make it generic) we want the msg to be printed once. Of course I can simplify it for now if we expect SIGTERM is managed only by one handler once.

skonto:
```
I'd also sort of like to be able to say:

func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
  ...
  checker := func (r http.Request) error {
    // pretend we have appropriate locking
    return impl.connectionError
  }

  // internally, calls getReadinessRegisterFunc(ctx).Register(checker)
  injection.AddReadiness(ctx, checker)

  return impl
}
```

Now you can say:

```
ctx = WithOverrideHealthCheckHandler(ctx, &HealthCheckHandler{
			ReadinessProbeRequestHandle: testHandler(),
			LivenessProbeRequestHandle:  testHandler(),
		})
networkpkg.ServeHealthProbes(ctx)
```

skonto:
@evankanderson I simplified the impl. Pls review.

skonto:
> 
{"level":"error","ts":1675945698.819293,"logger":"fallback","caller":"consumergroup/reconciler.go:302","msg":"Returned an error","targetMethod":"FinalizeKind","error":"unable to delete the consumer group my.group.id: kafka server: The client is not authorized to send this request type","stacktrace":"knative.dev/eventing-kafka-broker/control-plane/pkg/client/internals/kafka/injection/reconciler/eventing/v1alpha1/consumergroup.(*reconcilerImpl).Reconcile\n\t/home/runner/work/pkg/pkg/downstream/control-plane/pkg/client/internals/kafka/injection/reconciler/eventing/v1alpha1/consumergroup/reconciler.go:302\nknative.dev/pkg/reconciler/testing.(*TableRow).Test\n\t/home/runner/work/pkg/pkg/downstream/vendor/knative.dev/pkg/reconciler/testing/table.go:160\nknative.dev/pkg/reconciler/testing.TableTest.Test.func1\n\t/home/runner/work/pkg/pkg/downstream/vendor/knative.dev/pkg/reconciler/testing/table.go:389\ntesting.tRunner\n\t/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1439"}

skonto:
@evankanderson gentle ping.

skonto:
@evankanderson pls review

evankanderson (commented):
/lgtm
/approve

Looks good for now! Sorry about the delays in reviewing; I had Friday as a day off.

injection/health_check.go
```diff
@@ -0,0 +1,114 @@
...
+		go func() {
+			<-ctx.Done()
+			_ = server.Shutdown(ctx)
+		}()
+
+		// start the web server on port and accept requests
+		logger.Infof("Probes server listening on port %s", port)
+
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.Fatal(err)
```
dprotaso, line 50: I don't think we should fatal -  we should return an error

injection/health_check.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// ServeHealthProbes sets up liveness and readiness probes.
+// If user sets no probes explicitly via the context then defaults are added.
+func ServeHealthProbes(ctx context.Context) {
+	logger := logging.FromContext(ctx)
+	port := os.Getenv("KNATIVE_HEALTH_PROBES_PORT")
+	if port == "" {
+		port = "8080"
+	}
```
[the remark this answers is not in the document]
skonto: We could.

injection/sharedmain/main.go
```diff
@@ -313,6 +313,11 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 		return controller.StartAll(ctx, controllers...)
 	})
 
+	// Setup default health checks to catch issues with cache sync etc.
+	if !healthProbesDisabled(ctx) {
+		injection.ServeHealthProbes(ctx)
```
dprotaso, line 318: I think we should make this part of the `errgroup` so that if it fails to start it should shut everything down

injection/health_check.go
skonto: Ok we can just emit the error but eventually we will fail since probes are failing.

injection/sharedmain/main.go
skonto: Ok makes sense we can also avoid some of the go routine nesting.

evankanderson:
Yes, I hadn't even realized there was still a hold.

I'll typically use a hold if I want to give the author a choice between merge now and change based on optional feedback.

skonto:
@evankanderson @dprotaso pls take a look. Addressed the latest comment.
For the use of the PR check here: https://github.com/knative/serving/pull/13563.

One scenario not addressed here is what default-domain job uses:

```
	// Start an HTTP Server
	h := netprobe.NewHandler(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.WriteHeader(http.StatusOK)
	}))
	//nolinlt:gosec https://github.com/knative/serving/issues/13439
	server := http.Server{Addr: ":8080", Handler: h, ReadHeaderTimeout: time.Minute}
	go server.ListenAndServe()
```

This is a special case as we need to modify the path, have only a common handler and use only the "/" as net-kourier probes that path only, any other will not work. This is left for a separate PR.

evankanderson:
I think it's fine to add the option to change the path (register additional paths?) as a follow-up.

/lgtm

skonto:
@dprotaso is it ok to unhold?

PR #2672, opened 2023-02-03
update OWNERS files to use teams with active members

dprotaso:
- Eventing & Serving Writers can now approve more packages (in order to share the ownership)
- API packages require approval from Serving/Eventing WG Leads
- Drop individual user names as entries - in favour of using `-writer` and `-reviewer` teams


dprotaso:
/hold

Would like to get buy in from eventing folks as they'll start getting more notifications for reviews

/assign @psschwei @pierDipi

controller/OWNERS
```diff
@@ -1,7 +1,7 @@
 # The OWNERS file is used by prow to automatically merge approved PRs.
 
 approvers:
-- pkg-controller-writers
+- serving-writers
 
 reviewers:
-- pkg-controller-reviewers
+- serving-reviewers
```
dprotaso, line 7: I'm hoping prow is smart enough to rotate through this list of folks for reviews and approvals vs. the larger pool at the top OWNERS files

psschwei (commented):
looks good to me, but will let @pierDipi give the final stamp for eventing side

apis/OWNERS
```diff
@@ -1,9 +1,12 @@
 # The OWNERS file is used by prow to automatically merge approved PRs.
 
 approvers:
-- mattmoor
-- vaikas
-- n3wscott
+- technical-oversight-committee
+- serving-wg-leads
```
pierDipi, line 6: Why isn't eventing-wg-leads here too?

pierDipi (commented):
This makes a lot of sense to me, thanks for doing it!

I left just one question

apis/OWNERS
dprotaso: let me add it

Issue #2677, opened 2023-02-10
reconcilerImpl.updateStatus method performance impacted by state difference calculation used in debug mode only


rafalbigaj:
/area API
/kind bug

## Expected Behavior

The existing and desired resource state difference should be calculated only when debug logging is enabled.

The state comparison using `kmp.SafeDiff(existing.Status, desired.Status)` in the `reconcilerImpl.updateStatus` method is performed for debug logging purposes only and should not affect the reconciler performance in the production mode. The state difference should be calculated only when debug log level is enabled.

## Actual Behavior

The `reconcilerImpl.updateStatus` method contribution to overall reconciler CPU consumption happens to be extremely high for resources of significant size (> 30kB). Attached logs show the example of 66.5% CPU time spend on calculating state diff using `kmp.SafeDiff`, which is used only for debug logging.

| CPU Time | Code |
| --- | --- |
| 66.5%  | github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1alpha1/run.(*reconcilerImpl).updateStatus.func1 at github.com/tektoncd/pipeline@v0.41.0/pkg/client/injection/reconciler/pipeline/v1alpha1/run/reconciler.go:330 |
| 66.5% | knative.dev/pkg/kmp.SafeDiff at knative.dev/pkg@v0.0.0-20221011175852-714b7630a836/kmp/diff.go:50 |

![image](https://user-images.githubusercontent.com/824894/218104206-a369f74c-30e8-45d1-8d51-aeabc5d663f3.png)

## Steps to Reproduce the Problem

1. Create hundreds of resources with state of significant size (> 30kB).
2. Start resources controller with CPU profiler.
3. Observer `reconcilerImpl.updateStatus` method CPU consumption.

PR #2678, opened 2023-02-10
fix: `reconcilerImpl.updateStatus` calculates state difference in debug mode only

rafalbigaj:
The state comparison using `kmp.SafeDiff(existing.Status, desired.Status)` in the `reconcilerImpl.updateStatus` method is costly and performed for debug logging purposes.

# Changes

-  :bug: improve reconciler `updateStatus` performance by eliminating the existing and desired state calculation in none debug mode

/kind performance

Fixes https://github.com/knative/pkg/issues/2677

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/reconciler.go
@@ -24,6 +24,7 @@ import (
 	fmt "fmt"
 
 	zap "go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	clientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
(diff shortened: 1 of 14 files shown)


davidhadas:
Why fix previous versions of the API?

psschwei (commented):
I think in general the change suggested here makes sense, as there's no reason to run the diff if we're not going to do anything with it.
 
cc @dprotaso for thoughts (as he knows this area much better than me)

davidhadas:
@dprotaso, 

Using the context as a bearer of the logger, configs, etc., is a normal pattern across API boundaries and between processes. But disputable when used inside a package instead of passing the actual parameters.  

In cases like this, where we need the logger in the caller and callee, should we add a logger as a call parameter?

dprotaso:
Thanks for the optimization

/lgtm
/approve

dprotaso:
> @dprotaso, @psschwei
> 
> Using the context as a bearer of the logger, configs, etc., is a normal pattern across API boundaries and between processes. But disputable when used inside a package instead of passing the actual parameters.
> 

I generally don't like how Knative abuses `ctx` to supply dynamic arguments. It's not great for discoverability and type safety. 

> In cases like this, where we need the logger in the caller and callee, should we add a logger as a call parameter?

I'd rather not be strict here given this is part of a code generator and isn't linted etc. Editing the template is also annoying.

Generally though I'm thinking it would make sense to create a bag of properties in a struct (ie. `ReconcileContext`) and then call Reconcile. The goal being to group objects with a similar lifetime.

psschwei:
@dprotaso any objection to cherry picking this into v1.9? Tekton is asking if we could: https://github.com/tektoncd/pipeline/pull/6194#issuecomment-1437546097

dprotaso:
@psschwei nope

/cherrypick release-1.9

PR #2688, opened 2023-02-21
Allow reconcilers to listen to leader promotion events 

### LINKED ISSUE #2675: Allow reconcilers to listen to leader promotion events
The current approach doesn't allow controllers to hook 
additional logic when a reconciler is promoted for buckets,
only when it is demoted with `DemoteFunc` passed through
the `controller.OptionsFn` function.

In eventing, we have components running in the background
(aka detached to resource created/update/delete),
which should only take action when a given reconciler
is the leader ([the autoscaler](https://github.com/knative/eventing/blob/main/pkg/scheduler/statefulset/autoscaler.go)).

## Additional Info

Similarly to `DemoteFunc`, we can add a `PromoteFunc func(b Bucket)`
to `controller.Options` that would be called in addition to the existing
defined `PromoteFunc` as follow:

```golang
			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) error {
  
				promoteFunc(bkt) // dispatch promoted bucket before enqueue-ing objects

				all, err := lister.List(labels.Everything())
				if err != nil {
					return err
				}
				for _, elt := range all {
					if promoteFilterFunc != nil {
						if ok := promoteFilterFunc(elt); !ok {
							continue
						}
					}
					enq(bkt, types.NamespacedName{
						Namespace: elt.GetNamespace(),
						Name:      elt.GetName(),
					})
				}
				return nil
			},
```

/kind feature

pierDipi:
The current approach doesn't allow controllers to hook 
additional logic when a reconciler is promoted for buckets,
only when it is demoted with `DemoteFunc` passed through
the `controller.OptionsFn` function.

In eventing, we have components running in the background
(aka detached to resource created/update/delete),
which should only take action when a given reconciler
is the leader ([the autoscaler](https://github.com/knative/eventing/blob/main/pkg/scheduler/statefulset/autoscaler.go)).

Fixes #2675 

**Release Note**

```release-note
Allow reconcilers to listen to leader promotion events 
```

### DIFF
--- a/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/controller.go
+++ b/client/injection/apiextensions/reconciler/apiextensions/v1/customresourcedefinition/controller.go
@@ -64,10 +64,15 @@ func NewImpl(ctx context.Context, r Interface, optionsFns ...controller.OptionsF
 	lister := customresourcedefinitionInformer.Lister()
 
 	var promoteFilterFunc func(obj interface{}) bool
+	var promoteFunc = func(bkt reconciler.Bucket) {}
 
 	rec := &reconcilerImpl{
 		LeaderAwareFuncs: reconciler.LeaderAwareFuncs{
 			PromoteFunc: func(bkt reconciler.Bucket, enq func(reconciler.Bucket, types.NamespacedName)) error {
+
+				// Signal promotion event
+				promoteFunc(bkt)
+
 				all, err := lister.List(labels.Everything())
 				if err != nil {
 					return err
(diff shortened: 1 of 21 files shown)


pierDipi:
/cc @dprotaso 

I couldn't find any unit test for this area, any ideas?

dprotaso:
> I couldn't find any unit test for this area, any ideas?

Yeah - we at least have coverage that the code generation can compile and we have downstream tests to catch breaking changes.

PR #2690, opened 2023-02-22
Make ZipkinTracingEnabled flag thread safe

mgencur:
# Changes

- Introduce a new variable TracingEnabled of type atomic.Bool
- Deprecate the old ZipkinTracingEnabled
- Set the TracingEnabled to true automatically when SetupZipkinTracing is called. This simplifies usage in multiple places as it's not required to manually set the flag. Currently, it's being set on multiple places, these might be removed after integrating this change:
  - https://github.com/knative/eventing/blob/main/test/upgrade/prober/verify.go#L63
  - https://github.com/knative-sandbox/reconciler-test/blob/main/pkg/milestone/emitter_tracing.go#L51
  - https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/pkg/tracing/zipkin.go#L34 

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


cardil (approved):
/lgtm

PR #2697, opened 2023-02-27
bump golang version in downstream tests

dprotaso:
Fixes the operator downstream test which had their go version bumped to go1.19

### DIFF
--- a/.github/workflows/knative-downstream.yaml
+++ b/.github/workflows/knative-downstream.yaml
@@ -50,10 +50,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Set up Go 1.18.x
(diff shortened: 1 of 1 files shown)


dprotaso:
cc @pierDipi looks like your api docs generating tool doesn't work with go1.20  https://github.com/ahmetb/gen-crd-api-reference-docs

We'll want to bump go to 1.20 for the next release

houshengbo:
@dprotaso Do we target to upgrade the go version to 1.20? Or just 1.19 will be enough?

dprotaso:
> Do we target to upgrade the go version to 1.20? Or just 1.19 will be enough?

We should use go1.20 in our toolchain. Whether or not your repo (operator) requires go1.19 or go1.20 in your go.mod is up to you.

PR #2708, opened 2023-03-27
add NewServerTLS

### LINKED ISSUE #2699: Support creating TLS servers with mutual trust
/area networking

Add func NewTrustedServer(addr string, h http.Handler) *http.Server in knative.dev/pkg/network/h2c.go

davidhadas:
Add another public function to pkg h2c that allows configuring the server's tls.Config  

Fix #2699

### DIFF
--- a/network/h2c.go
+++ b/network/h2c.go
@@ -27,6 +27,18 @@ import (
 	"golang.org/x/net/http2/h2c"
 )
 
+// NewServerTLS is same as NewServer but allows configuring the server's TLSConfig
+func NewServerTLS(addr string, h http.Handler, tlsConfig *tls.Config) *http.Server {
+	//nolint:gosec
+	h1s := &http.Server{
+		Addr:      addr,
+		Handler:   h2c.NewHandler(h, &http2.Server{}),
+		TLSConfig: tlsConfig,
+	}
+
+	return h1s
+}
+
 // NewServer returns a new HTTP Server with HTTP2 handler.
 func NewServer(addr string, h http.Handler) *http.Server {
 	h1s := &http.Server{


network/h2c.go
```diff
@@ -27,6 +27,18 @@ import (
...
+	//nolint:gosec
+	h1s := &http.Server{
+		Addr:      addr,
+		Handler:   h2c.NewHandler(h, &http2.Server{}),
+		TLSConfig: tlsConfig,
+	}
+
+	return h1s
+}
+
```
evankanderson, line 41: Consider using the functional options pattern:

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

This would look like:

```golang
type ServerOption func(*http.Server)

func WithTLSConfig(tlsConfig *tls.Config) ServerOption {
  return func(s *http.Server) {
    s.TLSConfig = tlsConfig
  }
}

func NewServer(addr string, h http.Handler, opts ...ServerOption) *http.Server {
  h1s := ...

  for _, opt := range opts {
    opt(h1s)
  }

  return h1s
}
```

Among other things, it will ensure that the `NewServerTLS` and other implementations also implement the slowloris attack mitigation on line 47.

Also, it will avoid creating many `NewServerFoo` implementations, at the host of writing:

```
srv = NewServer(":8080", myHandler, WithTLSConfig(myTLS))
```

dprotaso:
`TLSConfig` is a public property on the `http.Server` - the helper here isn't necessary because end-users can just configure TLS directly

```go
server := pkg.NewServer(...)
server.TLSConfig = ...
```

davidhadas:
> `TLSConfig` is a public property on the `http.Server` - the helper here isn't necessary because end-users can just configure TLS directly
> 
> ```go
> server := pkg.NewServer(...)
> server.TLSConfig = ...
> ```

That is a very good point :) 
Closing this PR

PR #2718, opened 2023-04-07
add admissionRequest in ctx

### LINKED ISSUE #2656: Consider putting the admission review on the webhook context
> The use case is that I need to parse again the AdmissionRequest from Request.Body to perform further checks in the validating/mutating phase.

Would it make more sense for the `AdmissionRequest` go object to be on the context - instead of having the raw request body and parsing it again yourself?

_Originally posted by @dprotaso in https://github.com/knative/pkg/issues/2583#issuecomment-1331507771_

rahulii:
# Changes

-  :broom: This PR associated the admissionv1.AdmissionRequest object the webhook received with the context.
-  adds a unit test

/kind <kind>

Fixes #2656 

**Release Note**

**Docs**

### DIFF
--- a/apis/contexts.go
+++ b/apis/contexts.go
@@ -20,9 +20,9 @@ import (
 	"context"
 	"net/http"
 
+	admissionv1 "k8s.io/api/admission/v1"
 	authenticationv1 "k8s.io/api/authentication/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	admissionv1 "k8s.io/api/admission/v1"
 )
 
 // This is attached to contexts passed to webhook interfaces when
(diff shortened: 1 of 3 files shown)


apis/contexts.go
```diff
@@ -245,6 +246,7 @@ func IsDryRun(ctx context.Context) bool {
 // This is attached to contexts passed to webhook interfaces with
 // additional context from the HTTP request.
 type httpReq struct{}
+type AdmissionReq struct{}
```
dprotaso, line 249: don't expore this
```suggestion
type admissionRequest struct{}
```

dprotaso (changes requested):
We should remove the HTTP request from the context in favour of the AdmissionRequest

rahulii:
> We should remove the HTTP request from the context in favour of the AdmissionRequest

https://github.com/knative/pkg/blob/main/webhook/conversion.go#L59
converion webhook also uses this function - we would not be able to remove it I guess unless we also support `Request *ConversionRequest`

dprotaso:
> conversion webhook also uses this function - we would not be able to remove it I guess unless we also support Request *ConversionRequest

Yeah let's do that and add helpers for `ConversionReview`

PR #2721, opened 2023-04-10
Make minimum TLS version configurable for webhooks

izabelacg:
# Changes

- :gift: make TLS minimum version configurable via environment variables for webhooks
- changes the TLS minimum version to 1.3

/kind enhancement

TLS 1.3 comes with numerous enhancements, such as a quicker TLS handshake and more secure cipher suites, which make it simpler to use. Additionally, the introduction of Zero Round-Trip Time (0-RTT) key exchanges further simplifies the TLS handshake. 

**Release Note**

```release-note
- Added an additional option when creating a webhook, which allows to configure the minimum TLS version acceptable to communicate with the API server.
- The default TLS version was also changed from TLS 1.2 to TLS 1.3.
```

**Docs**
N/A

### DIFF
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -79,6 +79,6 @@ func TLSMinVersionFromEnv(defaultTLSMinVersion uint16) uint16 {
 	case "":
 		return defaultTLSMinVersion
 	default:
-		panic(fmt.Sprintf("the environment variable %q has to be either '1.2' or '1.3'", tlsMinVersion))
+		panic(fmt.Sprintf("the environment variable %q has to be either '1.2' or '1.3'", tlsMinVersionEnvKey))
 	}
 }
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -31,7 +31,7 @@ const (
 
 	secretNameEnvKey = "WEBHOOK_SECRET_NAME" //nolint:gosec // This is not a hardcoded credential
 
-	tlsMinVersionEnvKey = "TLS_MIN_VERSION"
+	tlsMinVersionEnvKey = "WEBHOOK_TLS_MIN_VERSION"
 )
 
 // PortFromEnv returns the webhook port set by portEnvKey, or default port if env var is not set.
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -17,6 +17,7 @@ limitations under the License.
 package webhook
 
 import (
+	"crypto/tls"
 	"fmt"
 	"os"
 	"strconv"
@@ -29,6 +30,8 @@ const (
 	webhookNameEnvKey = "WEBHOOK_NAME"
 
 	secretNameEnvKey = "WEBHOOK_SECRET_NAME" //nolint:gosec // This is not a hardcoded credential
+
+	tlsMinVersionEnvKey = "TLS_MIN_VERSION"
 )
 
 // PortFromEnv returns the webhook port set by portEnvKey, or default port if env var is not set.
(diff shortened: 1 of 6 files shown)


dprotaso:
@ReToCode @nak3 do you have any qualms with bumping the min TLS version for our webhooks to 1.3?

dprotaso:
Also @izabelacg any chance you can test this change in a WIP serving PR. Curious to see it run against a few k8s version.

nak3:
> @ReToCode @nak3 do you have any qualms with bumping the min TLS version for our webhooks to 1.3?

We will confirm it once the PR was merged into serving but I have no qualm for now.
With said, could you share the context why bumping the default TLS minimum version to 1.3?

dprotaso:
It's been stable since 2018

I found these informative (essentially there's a perf & security gain)
https://kinsta.com/blog/tls-1-3/
https://blog.cloudflare.com/rfc-8446-aka-tls-1-3/

nak3:
Thank you. Okay, so it is not because TLS 1.2 has an issue/vulnerability but just gain the power of 1.3. I have no objection for it.

webhook/env.go
```diff
@@ -29,6 +30,8 @@ const (
 	webhookNameEnvKey = "WEBHOOK_NAME"
 
 	secretNameEnvKey = "WEBHOOK_SECRET_NAME" //nolint:gosec // This is not a hardcoded credential
+
+	tlsMinVersionEnvKey = "TLS_MIN_VERSION"
```
dprotaso, line 34: Let's be consistent here

```suggestion
	tlsMinVersionEnvKey = "WEBHOOK_TLS_MIN_VERSION"
```

webhook/webhook.go
```diff
@@ -136,7 +138,7 @@ func New(
 		secretInformer := kubeinformerfactory.Get(ctx).Core().V1().Secrets()
 
 		webhook.tlsConfig = &tls.Config{
-			MinVersion: tls.VersionTLS12,
+			MinVersion: TLSMinVersionFromEnv(tls.VersionTLS13),
```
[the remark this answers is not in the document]
izabelacg: Made changes, if you can review them pls. Thanks

webhook/env.go
```diff
@@ -66,3 +69,16 @@ func SecretNameFromEnv(defaultSecretName string) string {
...
+func TLSMinVersionFromEnv(defaultTLSMinVersion uint16) uint16 {
+	switch tlsMinVersion := os.Getenv(tlsMinVersionEnvKey); tlsMinVersion {
+	case "1.2":
+		return tls.VersionTLS12
+	case "1.3":
+		return tls.VersionTLS13
+	case "":
+		return defaultTLSMinVersion
+	default:
+		panic(fmt.Sprintf("the environment variable %q has to be either '1.2' or '1.3'", tlsMinVersion))
```
dprotaso, line 82:
```suggestion
		panic(fmt.Sprintf("the environment variable %q has to be either '1.2' or '1.3'", tlsMinVersionEnvKey))
```

webhook/webhook.go
```diff
@@ -119,6 +123,16 @@ func New(
...
+	webhookTLSMinVersion := uint16(tls.VersionTLS13)
+	switch opts.TLSMinVersion {
+	case 0:
+		// No-op, the default value is already set
+	case tls.VersionTLS13, tls.VersionTLS12:
+		webhookTLSMinVersion = opts.TLSMinVersion
+	default:
+		return nil, fmt.Errorf("unsupported TLS version: %d", opts.TLSMinVersion)
+	}
+
```
dprotaso, line 135: probably simpler to do something like

```go
if opt.TLSMinVersion == 0 {
    opts.TLSMinVersion = TLSMinVersionFromEnv(defaultTLSMinVersion)
}
```

izabelacg: My concern was that someone would configure the minimum version to VersionTLS11, for example. Do you think this is a problem or there is no need to verify "unsupported versions"?

dprotaso: yeah good point

PR #2420, opened 2022-02-03
Allow overriding webhook controller options

pierDipi:
Creating 2 defaulting webhooks w1 and w2 inside the same pod
for 2 different mutating webhook configuration manifests
doesn't work since the lease name is based on the hardcoded
queue name DefaultingWebhook. So, we get a single pod competing for
the same lease in 2 separate reconcilers for 2 separate webhooks.

This PR allows overriding `ControllerOptions` for webhooks
so that callers can work around the issue while maintaining
backward compatibility.

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

# Changes

- Allow overriding webhook controller options

/kind bug

Fixes #2418 

**Release Note**

```release-note
Allow overriding webhook controller options
```

PR #2731, opened 2023-04-25
fix: use the controller name when creating the lease name

hectorj2f:
# Changes

When defining more than a single validating/defaulting/conversion webhooks with the `sharedmain` contructors, I detected that only one of the webhooks was mutated and fully reconciled. 
After looking at the code, I found the controllers of the same type (e.g. defaulting) were competing for the lease. The main reason comes from the lease name creation which doesn't take into account the name of the controller itself. It uses a component name and a hardcoded queueName.
As a consequence, if I have two validating webhooks I'll only get one lease and the reconciliation of the second validating webhook will exit [here](https://github.com/knative/pkg/blob/main/webhook/resourcesemantics/validation/reconcile_config.go#L81).

Therefore I propose to reuse the name of the controllers or the path (for the conversion webhooks), and thus allow declaring more than a single webhook for the same `sharedmain.Main...` execution.

`{component-name}.{webhook-type}.{webhook-name}-%02d-of-%02d`

/kind bug

Fixes #

**Release Note**

```release-note
webhooks: use controller name to construct the lease name.
```

**Docs**

```docs

```

### DIFF
--- a/webhook/resourcesemantics/conversion/controller.go
+++ b/webhook/resourcesemantics/conversion/controller.go
@@ -18,6 +18,8 @@ package conversion
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -117,7 +119,8 @@ func NewConversionController(
 		crdLister:    crdInformer.Lister(),
 	}
 
-	const queueName = "ConversionWebhook"
+	queueName := fmt.Sprintf("ConversionWebhook.%s", path)
+	queueName = strings.ReplaceAll(queueName, "/", ".")
 	logger := logging.FromContext(ctx)
 	c := controller.NewContext(ctx, r, controller.ControllerOptions{WorkQueueName: queueName, Logger: logger.Named(queueName)})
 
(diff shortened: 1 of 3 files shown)


hectorj2f:
@pierDipi Let's see what we can do to overcome this concern.

Generally, I believe there is already a leakage in terms of leases with the current version. We have to often delete the leases during the re-installation. 
Anyway I'd like to find a solution for this problem. I assume none has defined more than a single webhook, therefore I was wondering if we could store in the context the list of registered webhooks. If the list of webhook is greater than 1 then we could follow this new approach, otherwise we could stick to the old naming.

pierDipi:
How about something like this? https://github.com/knative/pkg/pull/2420

> I assume none has defined more than a single webhook

We do this, this is our "hack": https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/hack/patches/defaulting-webhook-queue-name.patch (that I'd like to get rid of)

hectorj2f:
@pierDipi  Indeed It looks similar to what I proposed. However it looks like the queueName remains the same all the time.

pierDipi:
> However it looks like the queueName remains the same all the time.

is that a problem for you?

hectorj2f:
@pierDipi The queueName is creating conflicts as done in my PR. The controllers are competing regarding the lease name which is always the same if you define multiple webhooks of the same type (conversion, defaulting, validation). That is why I initially proposed in my changes to make the queue name to align with the controller.Name.

hectorj2f:
closing this PR as fixed in https://github.com/knative/pkg/pull/2738!

Issue #2739, opened 2023-05-15
Drop dynamic wrapped client generation


dprotaso:
/assign @dprotaso 

We have lots of code generation to support using a dynamic client driving a typed kubernetes interface. This was introduced in https://github.com/knative/pkg/pull/2210 It's not clear if this is really being used anywhere.

What we do know is 
- it's not fully implemented (bunch of panics in the code)
- it causes friction when updating k8s deps
- it prevents downstream implementations from bumping their k8s version to a higher one 
 -- https://github.com/knative/pkg/issues/2725

I propose we drop this logic from `knative/pkg`

The only use of the dynamic injection  I found was https://github.com/tektoncd/triggers/pull/1207 but they don't seem to use the wrapped clients

dprotaso:
cc @mattmoor in case you have more references to usages

mattmoor:
We use this downstream to inject clients that exhibit different behavior.

cc @wlynch who I know spent a bunch of time looking at improving things a few weeks ago to streamline updates.

dprotaso:
> exhibit different behavior.

In what way?

> We use this downstream...

I'm thinking of making the generation optional - so you can just generate these yourself downstream

mattmoor:
@dprotaso I think for downstream codegen to work, the clients would need to be in their own go module, so that folks can `go mod -replace` with additional options.

dprotaso:
I don't understand. I'm suggesting downstream add a generation script like so

https://github.com/knative/pkg/blob/db8a35330281c41c7e8e90df6059c23a36af0643/hack/update-codegen.sh#L46-L54

But add a flag `--force-dynamic-clients` or something to get the extra dynamic stuff

dprotaso:
So looking at all the downstream packages that depend on `injection.Dynamic` I only see registration of dynamic clients and informers (due to codegen/cargo culting with no actual use). 

The only downstream dependency that seems to setup the context with a dynamic client is tekton triggers

```go
cache/github.com/tektoncd/triggers@v0.24.0/cmd/eventlistenersink/main.go:	ctx = injection.Dynamic.SetupDynamic(ctx)
```

But then they setup the `Default` Informers and clients overriding everything. _So `injection.Dynamic` doesn't really have any usage._

Also based on https://github.com/knative/pkg/pull/2210 

[...]

PR #2742, opened 2023-05-17
Drop dynamic wrapper injection code generation

dprotaso:
Part of https://github.com/knative/pkg/issues/2739

This removes the codegen logic that generates informers and clients that use the dynamic.Client underneath. These aren't being used anywhere.


pierDipi (approved):
/lgtm
/approve

PR #2750, opened 2023-05-24
Drop the unused injection.Dynamic interface

dprotaso:
Fixes: https://github.com/knative/pkg/issues/2739


dprotaso:
/hold 

Eventing hasn't updated their k8s deps - will wait for that

dprotaso:
@pierDipi what's left are some eventing repos - but it looks like the updates are trickling in slowly

/hold cancel
/assign @pierDipi

PR #2754, opened 2023-06-12
Generate kresource duck type injection

pierDipi:
In eventing, we need to track and act on arbitrary resources, so we need the genduck for KResource


matzew (approved):
/lgtm
/approve

Issue #2759, opened 2023-06-21
Third party libraries are not latest version


nak3:
/area test-and-release
/kind question
/kind bug
/kind cleanup

## Expected Behavior

* All deps in `go.mod` except for k8s libs should be the latest versions (if no reason to stay the old version).

## Actual Behavior

* Some deps in `go.mod` are not updated.
* For example:
  * https://pkg.go.dev/google.golang.org/protobuf?tab=versions ... v1.28.1 which is old.
  * https://pkg.go.dev/go.uber.org/zap?tab=versions ... v1.19.1 which is old.
  * ...etc

## Additional Info

* I haven't found any security issue but our org has a security survey with the question "Using the latest third party libraries?".
* It might be an issue of `hack/update-deps.sh`.

dprotaso:
We should setup dependabot for this repo and exclude knative and k8s deps from being bumped

nak3:
It seems that adding dependabot will work well :tada: 

* dependant bot: https://github.com/nak3/net-kourier/commit/34974b31bb3f1483079af159036452ee7d38ff97
* Result - https://github.com/nak3/net-kourier/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

Maybe we should add it to net-kourier or a few repos first then add it to the bot as in common knobot - https://github.com/knative-sandbox/knobots (after 1.11 cut and agreemet with the TOC or leads.)

pierDipi:
Regarding dependabot see prior art and thoughts: https://github.com/knative/pkg/pull/1942

nak3:
Ah, thank you!

nak3:
The following command line bumps all thrid-party libs except for k8s.io and knative.dev.

```sh
source vendor/knative.dev/hack/library.sh

for domain in `ls vendor --ignore={k8s.io,knative.dev}` ; do
  echo "Bump $domain"
  go_update_deps --upgrade --domain ${domain}
done
```

Although we can add it to common update deps (maybe adding it to https://github.com/knative-sandbox/knobots/blob/main/actions/update-deps/entrypoint.sh ) but some repos wants to exclude some specific libs - e.g. Istio does wants to keep `istio.io` lib version.
So, I am thinking that we should add each repo rather than common action. Please let me know if you have any idea.

nak3:
Above command does not work perfectly... (I think buoy's issue, though...)

[...]

PR #2768, opened 2023-07-26
Add dependabot and auto update deps

nak3:
As per title, this is same with https://github.com/knative-sandbox/net-gateway-api/pull/498.

Fix https://github.com/knative/pkg/issues/2759

### DIFF
--- a/.github/workflows/knative-dependabot.yaml
+++ b/.github/workflows/knative-dependabot.yaml
@@ -1,0 +1,36 @@
+name: depndabot - update-deps.sh
+
+on:
+  pull_request:
+
+permissions:
+  contents: write
+
+jobs:
+  generate_some_code:
+    name: update deps
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
+
+      - name: Run ./hack/update-deps.sh
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: ./hack/update-deps.sh
+
+      - name: git push
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: |
+          if ! git diff --exit-code --quiet
+          then
+            git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git config --local user.name "github-actions[bot]"
+            git add .
+            git commit -m "Run ./hack/update-deps.sh"
+            git push
+          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/knative-dependabot.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
+
+      - name: Run ./hack/update-deps.sh
+        working-directory: ./src/github.com/${{ github.repository }}
+        run: ./hack/update-deps.sh
```
dprotaso, line 24: You'll want to run `update-codegen` as well

nak3: Okay, `update-codegen.sh` includes `update-deps.sh` so replaced with `update-codegen.sh`

.github/workflows/knative-dependabot.yaml
```diff
@@ -0,0 +1,36 @@
+name: depndabot - update-codegen.sh
```
kvmware, line 1: nitpick: "dependabot" spelling

kvmware, line 1:
```suggestion
name: dependabot - update-codegen.sh
```

.github/workflows/knative-dependabot.yaml
```diff
@@ -0,0 +1,36 @@
...
+    name: update deps
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          path: ./src/github.com/${{ github.repository }}
+          fetch-depth: 0
```
kvmware, line 20: Curious on these settings some seem to be the defaults and why do you want all history?

.github/workflows/knative-dependabot.yaml
dprotaso: We'll want to run `update-codegen.sh` _after_ `update-deps.sh` - updates in dependencies could change the generated code (ie. generating clients)

.github/workflows/knative-dependabot.yaml
nak3: Just used with same value in `.github/workflows/knative-downstream.yaml`. It seems that it was added for some reason https://github.com/knative/pkg/pull/2265#discussion_r702107869
If we remove it, I think we should use consistent value in all action.

.github/workflows/knative-dependabot.yaml
nak3: Okay, updated. Just note, the client codes (`k8s.io/*` and `knative.dev/*` are ignored by dependabot so it is not supposed to be bumped automatically.

PR #2776, opened 2023-07-29
Fix github action for dependabot

nak3:
This patch fixes the current github action for dependabot.

/cc @kvmware @dprotaso


nak3:
I verified by https://github.com/nak3/pkg/pull/4

PR #2836, opened 2023-09-25
Return Audience in URIResolver.AddressableFromDestinationV1() too

creydr:
# Changes

- :gift: `URIResolver.AddressableFromDestinationV1()` returns now the Audience of the Destination/target too

### DIFF
--- a/resolver/addressable_resolver.go
+++ b/resolver/addressable_resolver.go
@@ -163,8 +163,9 @@ func (r *URIResolver) AddressableFromDestinationV1(ctx context.Context, dest duc
 			return nil, fmt.Errorf("URI is not absolute (both scheme and host should be non-empty): %q", dest.URI.String())
 		}
 		return &duckv1.Addressable{
-			URL:     dest.URI,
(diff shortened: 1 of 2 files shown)


resolver/addressable_resolver.go
```diff
@@ -258,6 +261,11 @@ func (r *URIResolver) addressableFromDestinationRef(ctx context.Context, dest du
 		addr.CACerts = dest.CACerts
 	}
 
+	if dest.Audience != nil && *dest.Audience != "" {
+		// destinations audience takes preference
+		addr.Audience = dest.Audience
+	}
```
pierDipi, line 267: Shouldn't these 2 logic should be similar but they aren't? as in they both prioritize `dest` or they prioritize `addr`

pierDipi: The TLS one seems wrong to me

pierDipi: https://github.com/knative/pkg/issues/2837

pierDipi (approved):
/lgtm
/approve

PR #2842, opened 2023-09-27
Use `DialTLSContextFunc` instead of `tls.Config` for `NewProxyAutoTLSTransport`

nak3:
Current `NewProxyAutoTLSTransport` takes `TLSConfig`. It is easy to use but impossible to customize to verify custom SAN.
So, this patch replaces it with `DialTLSContext`.

#### Alternative

I considered a downstream may be using `NewProxyAutoTLSTransport` with `TLSConfig`. But `NewProxyAutoTLSTransport` was added for internal encryption https://github.com/knative/pkg/pull/2479 1 year ago. If an user wants `NewProxyAutoTLSTransport` to take `TLSConfig` we can add it later.

See also https://github.com/knative/serving/pull/14452

### DIFF
--- a/network/h2c.go
+++ b/network/h2c.go
@@ -59,13 +59,11 @@ func newH2CTransport(disableCompression bool) http.RoundTripper {
 
 // newH2Transport constructs a neew H2 transport. That transport will handles HTTPS traffic
 // with TLS config.
-func newH2Transport(disableCompression bool, tlsConf *tls.Config) http.RoundTripper {
+func newH2Transport(disableCompression bool, tlsContext DialTLSContextFunc) http.RoundTripper {
 	return &http2.Transport{
 		DisableCompression: disableCompression,
-		DialTLS: func(netw, addr string, tlsConf *tls.Config) (net.Conn, error) {
-			return DialTLSWithBackOff(context.Background(),
-				netw, addr, tlsConf)
+		DialTLSContext: func(ctx context.Context, network, addr string, cfg *tls.Config) (net.Conn, error) {
+			return tlsContext(ctx, network, addr)
 		},
-		TLSClientConfig: tlsConf,
 	}
 }
--- a/network/transports.go
+++ b/network/transports.go
@@ -127,7 +127,9 @@ func newHTTPTransport(disableKeepAlives, disableCompression bool, maxIdle, maxId
 	return transport
 }
 
-func newHTTPSTransport(disableKeepAlives, disableCompression bool, maxIdle, maxIdlePerHost int, tlsConf *tls.Config) http.RoundTripper {
+type DialTLSContextFunc func(ctx context.Context, network, addr string) (net.Conn, error)
+
+func newHTTPSTransport(disableKeepAlives, disableCompression bool, maxIdle, maxIdlePerHost int, tlsContext DialTLSContextFunc) http.RoundTripper {
 	transport := http.DefaultTransport.(*http.Transport).Clone()
 	transport.DialContext = DialWithBackOff
 	transport.DisableKeepAlives = disableKeepAlives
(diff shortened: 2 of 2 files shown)


network/transports.go
```diff
@@ -149,10 +151,10 @@ func NewProberTransport() http.RoundTripper {
 }
 
 // NewProxyAutoTLSTransport is same with NewProxyAutoTransport but it has tls.Config to create HTTPS request.
```
KauzClay, line 153: should this comment be updated to say dialcontext instead of tls.config?

nak3:
> So by switching to use the DialTLSContext, it will let us insert some custom logic to handle setting up the TLS connection, such as verifying a custom SAN?

Yes, that's right.

> But just for my understanding, why is the GetCertificate function on the TLSConfig not enough for this case? My understanding was you could use that to get SAN info from the client hello, and in there you could return the correct cert. Perhaps I'm mixing things up?

We want the client to validate if the server certificate is correctly returned from the expected namespace or not - i.e. prevent MITM attacks.
_"you could use that to get SAN info from the client hello, and in there you could return the correct cert."_ allows the server to return the correct SAN, yes, but the client does not validate even if the certificate is returned by an impostor.

KauzClay:
> We want the client to validate if the server certificate matches the certificate of the expected namespace - i.e., to prevent MITM attacks.
"you could use that to get SAN info from the client hello, and in there you could return the correct cert." allows the server to return a correct certificate, yes, but the client does not validate the SAN even if an impostor returns a certificate by using another namespace's certificate.

Okay I see, I was mixing up client and server activities. Thanks for taking the time to explain that, I appreciate it 👍 

Seems like a good change to me, but perhaps you want Dave or Reto to see it as well?

KauzClay:
I learned how to use these labels lol

It looks good to me, but I'll put it on hold. @nak3 feel free to unhold, or keep it held for Reto or someone else to look at.


/lgtm
/hold

nak3:
@dprotaso Could you please take a look at this and https://github.com/knative/serving/pull/14452 ?
Then if you are alright, override `Downstream Knative / Unit Test (knative/serving)` ?

network/h2c.go
```diff
@@ -59,13 +59,11 @@ func newH2CTransport(disableCompression bool) http.RoundTripper {
...
 // with TLS config.
-func newH2Transport(disableCompression bool, tlsConf *tls.Config) http.RoundTripper {
+func newH2Transport(disableCompression bool, tlsContext DialTLSContextFunc) http.RoundTripper {
 	return &http2.Transport{
 		DisableCompression: disableCompression,
-		DialTLS: func(netw, addr string, tlsConf *tls.Config) (net.Conn, error) {
-			return DialTLSWithBackOff(context.Background(),
-				netw, addr, tlsConf)
+		DialTLSContext: func(ctx context.Context, network, addr string, cfg *tls.Config) (net.Conn, error) {
+			return tlsContext(ctx, network, addr)
```
dprotaso, line 66: Any reason for dropping `DialTLSWithBackOff` ?

dprotaso (commented):
It might be simpler to plumb in a `tls.Config` with `GetConfigForClient` set?

network/h2c.go
nak3: `DialWithBackOff` is called downstream https://github.com/knative/serving/pull/14452.
If we keep `DialWithBackOff` and call it here, we need to bring `tlsConf`  as an extra argument of `newH2Transport()`. Is it better to call `DialTLSWithBackOff` here?

nak3:
> It might be simpler to plumb in a tls.Config with GetConfigForClient set?

`GetConfigForClient` is only available to verify client certificates on server side[1]. We want to verify server certificate so it seems not available here.

[1] https://pkg.go.dev/crypto/tls#Config

> 	// GetConfigForClient, if not nil, is called after a ClientHello is
>	// received from a client.

PR #2844, opened 2023-09-29
Add `SinkAudience` to `SourceStatus`

creydr:
Add `SinkAudience` to `SourceStatus` so that we store the Audience of the resolved sink's destination.


pierDipi (approved):
/lgtm
/approve

PR #2847, opened 2023-10-02
Add rest.Config to fake context

creydr:
When we use `testing.SetupFakeContext`, we create a rest.Config, but don't add this to the context. Some reconcilers might need this config in the context.
This PR adds the rest.Config to the context (as it is created anyhow).

# Changes

- :broom: Add rest.Config to fake context from `SetupFaceContextWithCancel`


pierDipi (approved):
/lgtm
/approve

PR #2855, opened 2023-10-09
Fix potential `--kubeconfig` redefined error

dsimansk:
# Changes

- Fix potential redefined error with kubeconfig flag

The `--kubeconfing` flag is a popular one. I've run into a case with `sigs.k8s.io/controller-runtime` & `/pkg` when vendoring together there's `flag redefined` error thrown. Depending on the order of package inits.

This change should make `/pkg` safer to use alongside other K8s libs.

### DIFF
--- a/environment/client_config.go
+++ b/environment/client_config.go
@@ -44,8 +44,12 @@ func (c *ClientConfig) InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.ServerURL, "server", "",
 		"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 
-	fs.StringVar(&c.Kubeconfig, "kubeconfig", os.Getenv("KUBECONFIG"),
-		"Path to a kubeconfig. Only required if out-of-cluster.")
+	if f := fs.Lookup("kubeconfig"); f != nil {
+		c.Kubeconfig = f.Value.String()
+	} else {
+		fs.StringVar(&c.Kubeconfig, "kubeconfig", os.Getenv("KUBECONFIG"),
+			"Path to a kubeconfig. Only required if out-of-cluster.")
+	}
 
 	fs.IntVar(&c.Burst, "kube-api-burst", int(envVarOrDefault("KUBE_API_BURST", 0)), "Maximum burst for throttle.")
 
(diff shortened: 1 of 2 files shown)


dsimansk:
Failed downstream test is unrelated. The `eventing-rabbitmq` repo needs to be fixed because of the `k8s.io` version bump.
```
Downstream Knative / Unit Test (knative-extensions/eventing-rabbitmq) (pull_request)
```

pierDipi (approved):
/lgtm
/approve

environment/client_config.go
```diff
@@ -44,8 +44,12 @@ func (c *ClientConfig) InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(&c.ServerURL, "server", "",
 		"The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 
-	fs.StringVar(&c.Kubeconfig, "kubeconfig", os.Getenv("KUBECONFIG"),
-		"Path to a kubeconfig. Only required if out-of-cluster.")
+	if f := fs.Lookup("kubeconfig"); f != nil {
+		c.Kubeconfig = f.Value.String()
```
dprotaso, line 48: usually we call  InitFlags method then a `flags.Parse` gets called

Given we're setting c.Kubeconfig here - i don't think it will update after flags.Parse is called. 

So this config will have a different value

dsimansk: Yes, indeed `c.Kubeconfig` will set as default value of first defined `--kubeconfig`. Not reflecting values provided as args and `flags.Parse()` call.

dsimansk: Would you prefer a revert here? There are several places in `knative/pkg` that call flags.Parse() and could be refactored to isolated flatset, but that's probably not worth it.

```
flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
```

dprotaso: > Would you prefer a revert here?

I think so - otherwise it's a subtle bug that I will be difficult to discover. 

> There are several places in knative/pkg that call flags.Parse() and could be refactored to isolated flatset, but that's probably not worth it.

Yeah a lot of the test code had that - hence why I pulled out this resource into a separate package that doesn't call Parse

I don't have a good solution for you to mix this with kubebuilder. You may want to initialize your own flagset and configure each configuration yourself

PR #2856, opened 2023-10-10
Migrate knative.dev/hack/shell to knative.dev/pkg/test/shell

mgencur:
# Changes

- Copy knative.dev/hack/shell to knative.dev/pkg/test/upgrade/shell and adjust tests which depend on specific paths.
- Change shell.NewExecutor function to have this signature:
```
       func NewExecutor(t TestingT, loc ProjectLocation, opts... Option) Executor
```

Example usage can be seen in https://github.com/knative/serving/pull/14495 . This way the logs from shell scripts are printed within test scope.

- TODO: Mark the shell package in knative.dev/hack/shell as deprecated (via separate PR)

/kind <kind>

Fixes https://github.com/knative/serving/issues/14461

https://github.com/knative/serving/pull/14495 for knative/serving is also required to fully fix it.

**Release Note**

```release-note

```

**Docs**

```docs

```


test/shell/types.go
```diff
@@ -0,0 +1,94 @@
...
+	Environ []string
+}
+
+// testingWriter writes to the given testing.TB.
+type testingWriter struct {
+	t testing.TB
+
+	// If true, the test will be marked as failed if this testingWriter is
+	// ever used.
+	markFailed bool
```
cardil, line 56: This is no-go. The scripts often write messages to STDERR (in unix world usually debugging output goes to STDERR). The sole fact the script/tool writes something to STDERR doesn't mean we should fail the test.

test/shell/types.go
```diff
@@ -0,0 +1,94 @@
...
+type ExecutorConfig struct {
+	ProjectLocation
+	Streams
+	Labels
+	Environ []string
+}
+
+// testingWriter writes to the given testing.TB.
+type testingWriter struct {
+	t testing.TB
```
cardil, line 52: Let's use our own interface. We just need one method:

```go
type TestingT interface {
  Logf(format string, args ...any)
}
```

test/shell/executor.go
```diff
@@ -0,0 +1,213 @@
...
+func NewExecutor(config ExecutorConfig) Executor {
+	configureDefaultValues(&config)
+	return &streamingExecutor{
+		ExecutorConfig: config,
+	}
+}
+
+// TestingTStreams returns Streams which writes to t.Log and marks
+// the test as failed if anything is written to Streams.Err.
+func TestingTStreams(t testing.TB) Streams {
```
cardil, line 49:
```suggestion
func testingTStreams(t TestingT) Streams {
```

test/shell/executor.go
```diff
@@ -0,0 +1,213 @@
...
+	defaultLabelOut = "[OUT]"
+	defaultLabelErr = "[ERR]"
+	executeMode     = 0700
+)
+
+// ErrNoProjectLocation is returned if user didnt provided the project location.
+var ErrNoProjectLocation = errors.New("project location isn't provided")
+
+// NewExecutor creates a new executor from given config.
+func NewExecutor(config ExecutorConfig) Executor {
```
cardil, line 40:
```suggestion
func NewExecutor(t TestingT, config ExecutorConfig) Executor {
```

I would prefer to create an API that can't be misused. We probably want to force to pass the `TestingT`, we shouldn't allow people just switch without this and continue to output to `os.Stdout`, and `os.Stderr`.

test/shell/types.go
```diff
@@ -0,0 +1,94 @@
...
+// and executed.
+type Function struct {
+	Script
+	FunctionName string
+}
+
+// ExecutorConfig holds a executor configuration options.
+type ExecutorConfig struct {
+	ProjectLocation
+	Streams
```
cardil, line 45:
```suggestion

```

cardil (changes requested):
Thanks, @mgencur! This looks promising.

There are a couple of places that require changes.

**TLDR**: We should force passing `TestingT`, and we can't fail in case of _STDERR_ output.

test/shell/executor.go
mgencur: The API `shell.NewExecutor` is general. It doesn't have to be used only in upgrade tests and doesn't always have to log to testing.T. Somebody can use it to log to normal stdout and stderr. That's why I keep it optional. 
I will fix using this in a few repositories, at least Serving, Eventing, EventingKafkaBroker. I don't think it's a concern that somebody will migrate to the new package but will not use this new out/err. It's a narrow set of repositories.

test/shell/types.go
mgencur: OK. Just curious how does this help?

test/shell/types.go
mgencur: OK

test/shell/executor.go
cardil: This shell executor is only allowed to be used in testing, see: knative/hack#32

And if so, it doesn't make sense to log to stdout/stderr in tests, as knative/serving#14461 clearly shows.

This is why I wanted a different API, fixing the project location, at the same time. I propose to have:
```go
type Option func(*ExecutorConfig)
func NewExecutor(t TestingT, loc ProjectLocation, opts... Option) Executor
```

Such API clearly says which params are required, and also allows customization if needed.

mgencur: Wow. So this package is deprecated from the beginning? I think we're trying to do too many things with this fix. I like the previous API. From the example above it's not clear how ExecutorConfig is passed to the NewExecutor and it's increasing the num of parameters. 
There would be further improvements needed: 
* actually move the "shell" package under test/pkg/upgrade because that's the only place where it can be used according to the validator code
* remove "Streams" from the ExecutorConfig because it doesn't make sense anymore

Since this API is deprecated I don't see a point in improving it much. The goal is to fix the bug with logging.

mgencur: OK. I pushed the changes. I left `Streams` within ExecutorConfig because it is the place where the stream configuration is held, it's later used by the executor.

test/shell/types.go
cardil: Helps with testing of functions, as they accept any object that suffices the interface.

This is just good practice in Golang. Here's example article on this https://www.thoughtworks.com/insights/blog/programming-languages/mistakes-to-avoid-when-coming-from-an-object-oriented-language

cardil (approved):

cardil (commented):
/lgtm

It look

cardil (approved):
It looks great.

/lgtm

cardil (approved):
It looks great.

/lgtm

cardil:
Thanks @mgencur!

PR #2858, opened 2023-10-10
Add Reconciler for corev1.Service

creydr:
For eventing-istio we need to reconcile services. Instead of generating this reconciler on our own, we could use one from pkg.
This PR adds a reconciler for a kubernetes corev1 Service.

# Changes

- :gift: Add Reconciler for corev1.Service


pierDipi (approved):
/lgtm
/approve

PR #2861, opened 2023-10-11
Don't migrate storage version if CRD has one storage version

### LINKED ISSUE #2845: Storage Version Migrator doing empty patches for no reason
When a CRD has a single storage version in it's status. The storage version migrator should skip performing migrations (empty PATCHes)

Currently it seems like we don't check for this and always perform migrations when we don't need to.

kauana:
# Changes

- :bug: Fixed an issue where the [storage version migrator ](https://github.com/knative/pkg/blob/efe9f4a7b3f33a9db1a355b5375136b3a8c0fb1d/apiextensions/storageversion/migrator.go#L57)was running on CRDs that only had one storage version (related [comment](https://github.com/knative/serving/issues/14029#issuecomment-1741550974) for context).

Fixes #2845

### DIFF
--- a/apiextensions/storageversion/migrator.go
+++ b/apiextensions/storageversion/migrator.go
@@ -68,6 +68,11 @@ func (m *Migrator) Migrate(ctx context.Context, gr schema.GroupResource) error {
 		return fmt.Errorf("unable to determine storage version for %s", gr)
 	}
 
+	// don't migrate storage version if CRD has a single valid storage in its status
+	if len(crd.Status.StoredVersions) == 1 && crd.Status.StoredVersions[0] == version {
+		return nil
+	}
+
 	if err := m.migrateResources(ctx, gr.WithVersion(version)); err != nil {
 		return err
 	}
--- a/apiextensions/storageversion/migrator_test.go
+++ b/apiextensions/storageversion/migrator_test.go
@@ -19,6 +19,7 @@ package storageversion
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -63,21 +64,6 @@ var (
 			},
 		},
 	}
-
-	singleVersionFakeCRD = &apix.CustomResourceDefinition{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: fakeGR.String(),
-		},
-		Spec: apix.CustomResourceDefinitionSpec{
-			Group: fakeGK.Group,
-			Versions: []apix.CustomResourceDefinitionVersion{
-				{Name: "v1", Served: true, Storage: true},
-			},
-		},
-		Status: apix.CustomResourceDefinitionStatus{
-			StoredVersions: []string{"v1"},
-		},
-	}
 )
 
 func TestMigrate(t *testing.T) {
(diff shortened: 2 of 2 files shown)


apiextensions/storageversion/migrator.go
```diff
@@ -62,6 +62,11 @@ func (m *Migrator) Migrate(ctx context.Context, gr schema.GroupResource) error {
 		return fmt.Errorf("unable to fetch crd %s - %w", gr, err)
 	}
 
+	// don't migrate storage version if CRD has a single storage in its status
+	if len(crd.Status.StoredVersions) == 1 {
+		return nil
+	}
+
```
dprotaso, line 69: I think we need a stronger check and assert the storage version in the status is length 1 but it also is the version that's expected in the spec

dprotaso: The reason why I suggest this is because I believe when you switch the storage version in the spec it won't show up in your status until certain resources are being stored using that version - that only happens when new resources are created and older ones and updated

dprotaso (changes requested):
unit tests please :)

apiextensions/storageversion/migrator.go
kauana: Got it. I've updated the check to ensure that the version in the status and spec are the same. 

I do have a question about the tests though. If the CRD doesn't need to migrate versions, I just return. What is the best way to test this behavior? Currently I'm checking that Migrate didn't fail and that the storage version is correct, but that's a bit icky. Should I be returning an errors instead if the CRD version doesn't need to migrate?

dprotaso: K8s has a fake client that we are using in the test you could

1. Go through all the actions and ensure there are no patches
https://github.com/kubernetes/client-go/blob/bc7002f524b2527ceede8092e3e5232c005a42fb/testing/fake.go#L214

2. Prepend a Reactor and when there's a patch fail the test
https://github.com/kubernetes/client-go/blob/bc7002f524b2527ceede8092e3e5232c005a42fb/testing/fake.go#L103 

The first might be more straight forward

apiextensions/storageversion/migrator_test.go
```diff
@@ -63,6 +63,21 @@ var (
 			},
 		},
 	}
+
+	singleVersionFakeCRD = &apix.CustomResourceDefinition{
```
dprotaso, line 67: nit: just pull this into `TestMigrate_SingleStoredVersion` if it's only used there

apiextensions/storageversion/migrator_test.go
```diff
@@ -88,6 +103,22 @@ func TestMigrate(t *testing.T) {
...
+	cclient := apixFake.NewSimpleClientset(singleVersionFakeCRD)
+	m := NewMigrator(dclient, cclient)
+
+	if err := m.Migrate(context.Background(), fakeGR); err != nil {
+		t.Fatal("Migrate() =", err)
+	}
+
+	assertPatches(t, cclient.Actions(),
+		// patch resource definition dropping non-storage version
+		crdStorageVersionPatch(singleVersionFakeCRD.Name, "v1"),
```
dprotaso, line 118: assume if the storage version remains the same we don't need to patch the CRD status

apiextensions/storageversion/migrator_test.go
```diff
@@ -88,6 +89,44 @@ func TestMigrate(t *testing.T) {
...
+	}
+
+	// setup
+	dclient := dynamicFake.NewSimpleDynamicClient(runtime.NewScheme())
+	cclient := apixFake.NewSimpleClientset(singleVersionFakeCRD)
+
+	cclient.Fake.PrependReactor("patch", "customresourcedefinitions", func(action k8stesting.Action) (handled bool, ret runtime.Object, err error) {
+		pa, ok := action.(k8stesting.PatchAction)
+		if !ok {
+			return true, nil, fmt.Errorf("not a patch action: %#v", action)
```
dprotaso, line 115: We want to return `false` since we aren't handling this action - but given you're filtering using the `patch` verb in the `Prepend` call I would recommend just dropping this if block cause you're always going to get `PatchAction` type here

kauana: Good catch. I've removed the unnecessary check too.

apiextensions/storageversion/migrator.go
kauana: Thank you!

PR #2871, opened 2023-10-20
[CVE-2023-44487] Disable http2 for webhooks

skonto:
# Changes

- Disables http2 by default for webhooks. Allows to set it on demand via options.
- Similar to https://github.com/prometheus-operator/prometheus-operator/pull/6028.

Fixes # CVE-2023-44487 (partially)

**Release Note**

```release-note
Due to CVE-2023-44487 webhooks will use http1 by default, until the issue is fixed properly.
```

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -256,6 +256,7 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
 		QuietPeriod: wh.Options.GracePeriod,
 	}
 
+	// If TLSNextProto is not nil, HTTP/2 support is not enabled automatically.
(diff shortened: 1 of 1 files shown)


skonto:
cc @dprotaso

webhook/webhook.go
```diff
@@ -245,12 +256,18 @@ func (wh *Webhook) Run(stop <-chan struct{}) error {
 		QuietPeriod: wh.Options.GracePeriod,
 	}
 
+	nextProto := map[string]func(*http.Server, *tls.Conn, http.Handler){}
```
dprotaso, line 259: can we add a comment that
// If TLSNextProto is not nil, HTTP/2 support is not enabled automatically.

PR #2884, opened 2023-11-02
fix closure variable in filtered factory

### LINKED ISSUE #2883: Using filteredInformers with more than one label causes to filters with the wrong LabelSelector
# Problem description

Having a controller that is passing more than one label to the secret's filteredInformerFactory like this:

```golang
	labels := []string{
		networking.ServingCertName + certificates.SystemInternalTLSSecretLabelPostfix,
		networking.ServingCertName + certificates.KnativeCertIssuerSecretLabelPostfix,
	}
	ctx := filteredFactory.WithSelectors(signals.NewContext(), labels...)
```

to use it as

```golang
ctx = filteredFactory.WithSelectors(ctx, labelName)
secretInformer := getSecretInformer(ctx)
```

Will pass through `filtered_factory.go` `withInformerFactory` using this code

```golang
for _, selector := range labelSelectors {
		opts := []informers.SharedInformerOption{}
		if injection.HasNamespaceScope(ctx) {
			opts = append(opts, informers.WithNamespace(injection.GetNamespaceScope(ctx)))
		}
		opts = append(opts, informers.WithTweakListOptions(func(l *v1.ListOptions) {
			l.LabelSelector = selector
		}))
		ctx = context.WithValue(ctx, Key{Selector: selector},
			informers.NewSharedInformerFactoryWithOptions(c, controller.GetResyncPeriod(ctx), opts...))
	}
```

`selector` is used in a func() using a closure. So if this method is called with two labelSelectors it always passed only the last one to `informers.NewSharedInformerFactoryWithOptions`.

/assign

ReToCode:
# Changes
- Copy the value of `selector` in `withInformerFactory` within the for loop to avoid issues in func()

/kind bug

Fixes #2883

/hold verifying this in Serving

### DIFF
--- a/client/injection/apiextensions/informers/factory/filtered/fake/fake_filtered_factory.go
+++ b/client/injection/apiextensions/informers/factory/filtered/fake/fake_filtered_factory.go
@@ -45,14 +45,15 @@ func withInformerFactory(ctx context.Context) context.Context {
 	}
 	labelSelectors := untyped.([]string)
 	for _, selector := range labelSelectors {
+		selectorVal := selector
 		opts := []externalversions.SharedInformerOption{}
 		if injection.HasNamespaceScope(ctx) {
 			opts = append(opts, externalversions.WithNamespace(injection.GetNamespaceScope(ctx)))
 		}
 		opts = append(opts, externalversions.WithTweakListOptions(func(l *v1.ListOptions) {
-			l.LabelSelector = selector
+			l.LabelSelector = selectorVal
 		}))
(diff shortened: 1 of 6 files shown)


ReToCode:
/unhold

@dprotaso PTAL.

skonto:
LGTM should not these generators create some unit tests?

dprotaso:
/approve

> LGTM should not these generators have some unit tests?

Ideally.

We should also be replacing this generated code with generics. But I think when we do that we should consider refactoring/tweaking some of the controller framework we built.

PR #2904, opened 2023-11-28
Bump K8s dependencies to v0.28.4 and set K8s min-version to 1.27

### LINKED ISSUE #2902: Bump K8s min version to v1.27
https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md#upcoming-releases

We should bump our client libs to v0.28.x now

ReToCode:
# Changes
- Bump K8s dependencies to v0.28.4 and set K8s min-version to 1.27

Fixes #2902

**Release Note**

```release-note
The minimal required kubernetes version to use `knative/pkg` is now v1.27.0
```

/assign @dprotaso


ReToCode:
According to the discussion in [slack](https://cloud-native.slack.com/archives/C04LY4M2G49/p1701163419512209), we can directly use `go run` instead of using the K8s script. We already did this a few lines below, not sure why the script was called in the first place. The last commit was tested with:

```bash
find apis/duck -name "zz_generated.deepcopy.go" -delete 
rm -rf client/injection
hack/update-codegen.sh
git diff
# shows no differences
```

dprotaso:
hey sorry @ReToCode a update-deps merged - can you rebase?

ReToCode:
Not sure what is happening with the codegen:

```
Found diffs in: third_party/VENDOR-LICENSE/golang.org/x/sys/unix/LICENSE
fatal: ambiguous argument 'third_party/VENDOR-LICENSE/golang.org/x/sys/unix/LICENSE': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Error: Please run ./hack/update-codegen.sh.

knative/pkg is out of date. Please run hack/update-codegen.sh
Error: Process completed with exit code 1.
```

When I run it locally:

```
Diffing /Users/rlehmann/code/knative/pkg against freshly generated codegen
/Users/rlehmann/code/knative/pkg up to date.
```

ReToCode:
@dprotaso I'm still unable to reproduce. The codegen is fine in ubuntu in container and also on a GH codespace:

```
# Local ubuntu container
go version
go version go1.18.1 linux/arm64
...
=== Removing empty go.work.sum
Diffing /work against freshly generated codegen
/work up to date.

# GH codespace
go version
go version go1.21.4 linux/amd64
...
=== Removing empty go.work.sum
Diffing /workspaces/pkg against freshly generated codegen
/workspaces/pkg up to date.
```

GH action is using go 1.21, so no idea why this is different.
But as you are able to reproduce this, can you push the fix to my branch?

dprotaso:
I think you need to run this on amd64 not arm - the license tool isn't actually reporting licenses for all architectures

dprotaso:
/hold
/lgtm

@ReToCode unsure about the failures but if those were expected (and there are changes downstream) then feel free to drop the hold

ReToCode:
>  I think you need to run this on amd64 not arm - the license tool isn't actually reporting licenses for all architectures

Not sure, the GH actions env was actually `go version go1.21.4 linux/amd64`. Anyway, its ok now.

> @ReToCode unsure about the failures but if those were expected (and there are changes downstream) then feel free to drop the hold

I'm not sure what's the best approach here. Each downstream repo has to make some changes. I see these options:

a) update their `hack/update-codegen.sh` scripts to still call the deprecated `generate-groups.sh` but with `exec` to avoid the permission issues for now.
b) use `go run` to call all the generators, e.g. here an example for Serving: https://github.com/knative/serving/compare/main...ReToCode:serving:bump-k8s-deps?expand=1#diff-149dfe7bb29d1191dceae3a52915e750e64b7f87257a5fb309c29d3056e2a95dR70
c) Migrate to use `k8s.io/code-generator/kube_codegen.sh`

All of them have pros and cons. b) has the downside that we do not cleanup the old resources. It seems like the k8s scripts do delete old files.

Opinions? @dprotaso @skonto @pierDipi @matzew @mgencur @cardil (context: k8s 1.28 made changes to the generator scripts).

dprotaso:
I'd suggest just using `go run` and not wait - it's what gateway API is doing

dprotaso:
> Not sure, the GH actions env was actually `go version go1.21.4 linux/amd64`. Anyway, its ok now.

Your container env was using `arm` arch

ReToCode:
> I'd suggest just using go run and not wait - it's what gateway API is doing

@dprotaso I also like that approach, but is it ok to lose the cleanup?

```
All of them have pros and cons. b) has the downside that we do not cleanup the old resources. It seems like the k8s scripts do delete old files.
```

ReToCode:
As the release is very close, we'll just keep the old generators and just bump K8s deps [here](https://github.com/knative/pkg/pull/2928). I opened https://github.com/knative/pkg/issues/2927 to tackle the generation change separatly.

PR #2905, opened 2023-11-28
lower log level for AdmissionController.Admit validation  failures

### LINKED ISSUE #2008: Consider lowering log level for AdmissionController.Admit failures
/area monitoring
/kind question

## Expected Behavior
Consider logging validation errors at a lower level (`info`) to signal that the webhook is working as expected. 

## Actual Behavior
The AdmissionController.Admit code logs validation error on CRD's with `level==error`. Validation errors are a normal function of the webhook, so logging them as `level==error` is perhaps to aggressive and misleading as the webhook is operating normally when validations fail this way.

## Additional Info
example of logging statement: https://github.com/knative/pkg/blob/6045ed499615ecca39d92a8277489de1fbe3dcff/webhook/resourcesemantics/validation/validation_admit.go#L177

xiangpingjiang:
# Changes

- bug: Fix bug

/kind <kind>

Fixes #2008

**Release Note**

```release-note
lower log level for AdmissionController.Admit validation  failures
```

**Docs**

### DIFF
--- a/webhook/resourcesemantics/validation/validation_admit.go
+++ b/webhook/resourcesemantics/validation/validation_admit.go
@@ -180,7 +180,7 @@ func validate(ctx context.Context, resource resourcesemantics.GenericCRD, req *a
 	}
 
 	if result := resource.Validate(ctx); result != nil {
-		logger.Errorw("Failed the resource specific validation", zap.Error(err))
+		logger.Infow("Failed the resource specific validation", zap.Error(result))
 		// While we have the strong typing of apis.FieldError, partition the
 		// returned error into the error-level diagnostics and warning-level
 		// diagnostics, so that the admission response can embed things into


webhook/resourcesemantics/validation/validation_admit.go
```diff
@@ -180,7 +180,7 @@ func validate(ctx context.Context, resource resourcesemantics.GenericCRD, req *a
 	}
 
 	if result := resource.Validate(ctx); result != nil {
-		logger.Errorw("Failed the resource specific validation", zap.Error(err))
+		logger.Infow("Failed the resource specific validation", zap.Error(result))
```
skonto, line 183: I am wondering if we actually need to keep that logging statement given the comment [here](https://github.com/knative/pkg/pull/1172/files#r402569907).
cc @dprotaso

dprotaso: I'd be open to that change in a new PR

PR #2931, opened 2024-01-10
webhook: add options to disable resource_namespace tag in metrics

zhouhaibing089:
To add some context, historically, `resource_name` was removed from this tag list due to its high potential of causing high metrics cardinality. See [knative/pkg#1464][1] for more information.

While that's great, but it might not be sufficient for large scale use cases where namespaces can be super dynamic (with generateName, too) or grows fase enough. There is an issue report from
[tektoncd/pipeline#3171][2] which talks about this.

This proposal makes it possible to disable `resource_namespace` tag via an option function. The default behavior is not changed, so no user impact if any of existing users rely on this tag. There is no API contract change either due to the beauty of variadic functions.

Now downstream projects can consume this by override `StatsReporter` in webhook context options with their own preference. As a caveat here, if downstream project does choose to override `StatsReporter`, the default `ReportMetrics` function shouldn't be called by default as they may now have a different set of tag keys to report. As such, this function is only called if the default `StatsReporter` is used.

[1]: https://github.com/knative/pkg/pull/1464
[2]: https://github.com/tektoncd/pipeline/issues/3171

```release-note
webhook: add options to disable resource_namespace tag in metrics
```

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -290,7 +290,12 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	var wh *webhook.Webhook
 	if len(webhooks) > 0 {
 		// Register webhook metrics
-		webhook.RegisterMetrics()
+		opts := webhook.GetOptions(ctx)
+		if opts != nil {
+			webhook.RegisterMetrics(opts.StatsReporterOptions...)
+		} else {
+			webhook.RegisterMetrics()
+		}
 
 		wh, err = webhook.New(ctx, webhooks)
 		if err != nil {
--- a/webhook/stats_reporter.go
+++ b/webhook/stats_reporter.go
@@ -26,6 +26,7 @@ import (
 	"go.opencensus.io/tag"
 	admissionv1 "k8s.io/api/admission/v1"
 	apixv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/metrics"
 )
 
@@ -65,19 +66,83 @@ var (
 	resultCodeKey        = tag.MustNewKey("result_code")
 )
 
+type admissionToValue func(*admissionv1.AdmissionRequest, *admissionv1.AdmissionResponse) string
+type conversionToValue func(*apixv1.ConversionRequest, *apixv1.ConversionResponse) string
+
+var (
+	allAdmissionTags = map[tag.Key]admissionToValue{
+		requestOperationKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return string(req.Operation)
+		},
+		kindGroupKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Kind.Group
+		},
+		kindVersionKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Kind.Version
+		},
+		kindKindKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Kind.Kind
+		},
+		resourceGroupKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Resource.Group
+		},
+		resourceVersionKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Resource.Version
+		},
+		resourceResourceKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Resource.Resource
+		},
+		resourceNamespaceKey: func(req *admissionv1.AdmissionRequest, _ *admissionv1.AdmissionResponse) string {
+			return req.Namespace
+		},
+		admissionAllowedKey: func(_ *admissionv1.AdmissionRequest, resp *admissionv1.AdmissionResponse) string {
+			return strconv.FormatBool(resp.Allowed)
+		},
+	}
+	allConversionTags = map[tag.Key]conversionToValue{
+		desiredAPIVersionKey: func(req *apixv1.ConversionRequest, _ *apixv1.ConversionResponse) string {
+			return req.DesiredAPIVersion
+		},
+		resultStatusKey: func(_ *apixv1.ConversionRequest, resp *apixv1.ConversionResponse) string {
+			return resp.Result.Status
+		},
+		resultReasonKey: func(_ *apixv1.ConversionRequest, resp *apixv1.ConversionResponse) string {
+			return string(resp.Result.Reason)
+		},
+		resultCodeKey: func(_ *apixv1.ConversionRequest, resp *apixv1.ConversionResponse) string {
+			return strconv.Itoa(int(resp.Result.Code))
+		},
+	}
+)
+
 // StatsReporter reports webhook metrics
 type StatsReporter interface {
 	ReportAdmissionRequest(request *admissionv1.AdmissionRequest, response *admissionv1.AdmissionResponse, d time.Duration) error
 	ReportConversionRequest(request *apixv1.ConversionRequest, response *apixv1.ConversionResponse, d time.Duration) error
 }
 
+type statsReporterOptions struct {
+	tagsToExclude sets.Set[string]
+}
+
+type StatsReporterOption func(_ *statsReporterOptions)
+
+func WithoutTags(tags ...string) StatsReporterOption {
+	return func(opts *statsReporterOptions) {
+		opts.tagsToExclude.Insert(tags...)
+	}
+}
+
 // reporter implements StatsReporter interface
 type reporter struct {
 	ctx context.Context
+
+	admissionTags  map[tag.Key]admissionToValue
+	conversionTags map[tag.Key]conversionToValue
 }
 
 // NewStatsReporter creates a reporter for webhook metrics
-func NewStatsReporter() (StatsReporter, error) {
+func NewStatsReporter(opts ...StatsReporterOption) (StatsReporter, error) {
 	ctx, err := tag.New(
 		context.Background(),
 	)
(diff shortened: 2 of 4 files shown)


dprotaso:
/hold we have a release in the next week

Will revisit this after

zhouhaibing089:
/retest

(The failed tests reported seem to be fine locally)

webhook/webhook.go
```diff
@@ -144,6 +144,7 @@ func New(
 	logger := logging.FromContext(ctx)
 
 	if opts.StatsReporter == nil {
+		RegisterMetrics()
```
zhouhaibing089, line 147: If there are multiple webhooks, this function might end up with being called multiple times and that won't work (which failed the unit test).

dprotaso:
> (The failed tests reported seem to be fine locally)

They're consistent in CI - try using a docker container with ubuntu?

zhouhaibing089:
Kindly ping, may I get some attention here?

skonto:
Hi @zhouhaibing089 sorry for the delay, just to clarify here we have two changes:
a) avoid resourceNamespaceKey if someone sets a custom reporter using `WithoutResourceNamespace()` (in Webhook options)
b) register metrics only when the default reporter is used.

I think it would be more flexible if we can exclude an arbitrary tag key instead of having if statements like:

```
if r.opts.resourceNamespace {
	mutators = append(mutators, tag.Insert(resourceNamespaceKey, req.Namespace))
}
```
We could use:
```
type options struct {
	tagsToExclude    []tag.Key
}
```
wdyth?

zhouhaibing089:
> I think it would be more flexible if we can exclude an arbitrary tag key..

Thanks. I completely agree (PR updated with your suggestion).

webhook/stats_reporter.go
```diff
@@ -71,37 +71,77 @@ type StatsReporter interface {
...
-		tag.Insert(kindKindKey, req.Kind.Kind),
-		tag.Insert(resourceGroupKey, req.Resource.Group),
-		tag.Insert(resourceVersionKey, req.Resource.Version),
-		tag.Insert(resourceResourceKey, req.Resource.Resource),
-		tag.Insert(resourceNamespaceKey, req.Namespace),
-		tag.Insert(admissionAllowedKey, strconv.FormatBool(resp.Allowed)),
-	)
+	mutators := []tag.Mutator{}
+
+	if _, ok := r.opts.tagsToExclude[requestOperationKey.Name()]; !ok {
```
skonto, line 116: Could we make this more compact? For example:

```
"k8s.io/apimachinery/pkg/util/sets"
...
	keySet := sets.Set[string]{}
	tagforKey := map[string]struct {
		key   tag.Key
		value string
	}{
		requestOperationKey.Name():  {key: requestOperationKey, value: string(req.Operation)},
		kindGroupKey.Name():         {key: kindGroupKey, value: req.Kind.Group},
		kindVersionKey.Name():       {key: kindVersionKey, value: req.Kind.Version},
		kindKindKey.Name():          {key: kindKindKey, value: req.Kind.Kind},
		resourceGroupKey.Name():     {key: resourceGroupKey, value: req.Resource.Group},
		resourceVersionKey.Name():   {key: resourceVersionKey, value: req.Resource.Version},
		resourceResourceKey.Name():  {key: resourceResourceKey, value: req.Resource.Resource},
		resourceNamespaceKey.Name(): {key: resourceNamespaceKey, value: req.Namespace},
		admissionAllowedKey.Name():  {key: admissionAllowedKey, value: strconv.FormatBool(resp.Allowed)},
	}
	for key := range tagforKey {
		keySet.Insert(key)
	}
	keysToKeep := keySet.Difference(r.opts.tagsKeysToExclude)
	for key := range keysToKeep {
		mutators = append(mutators, tag.Insert(tagforKey[key].key, tagforKey[key].value))
	}

```
that means you need to adjust:

```
type options struct {
	tagsKeysToExclude sets.Set[string]
}
```

webhook/webhook_test.go
```diff
@@ -32,20 +32,30 @@ import (
 )
 
 func newDefaultOptions() Options {
+	statsReporter, err := NewStatsReporter()
```
skonto, line 35: We need to make sure that web_integration_test suite works without passing a StatsReporter as previously. Right now if I comment out this part we have a race condition. I strongly recommend to move the change 
```
if opts.StatsReporter == nil {
		RegisterMetrics()
```
to a new PR.

webhook/stats_reporter.go
zhouhaibing089: Good suggestions. I updated it again as almost exactly like you suggested.

webhook/webhook_test.go
zhouhaibing089: I removed those changes as you requested, but allow me explaining why I included them in the first place:

knative/pkg consumers still won't be able to use a customized `StatsReporter` that excludes certain tags, as double metrics registration won't work.

I can send a followup to address the data race.

skonto: @zhouhaibing089 hi,
I am not sure about what is the use case we are targeting but I see two things to cover:

a) default reporter with excluded tags
b) custom reporter with excluded tags

Could we just have (main):
```
	if len(webhooks) > 0 {
		opts := GetOptions(ctx)
		// Register webhook metrics
		webhook.RegisterMetrics(opts...)
```
so with this PR to support a) and then later decide about b)?  wdyth?

zhouhaibing089: Sorry for my late response.

I'm looking for a way to exclude tags, either "a" or "b" would be fine to me, and "a" seems to be more appealing. I think what you suggested definitely makes it easier for downstream. Let me try that way.

(Sorry I tended to amend my commits when there are comments that need to be addressed, I'll try append going forwards in this repo.)

webhook/stats_reporter.go
```diff
@@ -65,43 +65,111 @@ var (
...
-		tag.Insert(admissionAllowedKey, strconv.FormatBool(resp.Allowed)),
-	)
+	mutators := []tag.Mutator{}
+
+	for key, f := range admissionTags {
+		if _, ok := r.opts.tagsToExclude[key.Name()]; ok {
+			continue
+		}
+		mutators = append(mutators, tag.Insert(key, f(req, resp)))
+	}
```
dprotaso, line 170: Can we compute this list once so we're not doing this on every request but instead it's computed when the `StatsReporter` is initialized.

webhook/stats_reporter.go
```diff
@@ -65,43 +65,111 @@ var (
...
-		tag.Insert(kindGroupKey, req.Kind.Group),
-		tag.Insert(kindVersionKey, req.Kind.Version),
-		tag.Insert(kindKindKey, req.Kind.Kind),
-		tag.Insert(resourceGroupKey, req.Resource.Group),
-		tag.Insert(resourceVersionKey, req.Resource.Version),
-		tag.Insert(resourceResourceKey, req.Resource.Resource),
-		tag.Insert(resourceNamespaceKey, req.Namespace),
-		tag.Insert(admissionAllowedKey, strconv.FormatBool(resp.Allowed)),
-	)
+	mutators := []tag.Mutator{}
```
dprotaso, line 163: When we init this let's use the capacity of our mutators list

webhook/stats_reporter.go
```diff
@@ -114,13 +182,16 @@ func (r *reporter) ReportAdmissionRequest(req *admissionv1.AdmissionRequest, res
...
 // Captures req count metric, recording the count and the duration
 func (r *reporter) ReportConversionRequest(req *apixv1.ConversionRequest, resp *apixv1.ConversionResponse, d time.Duration) error {
-	ctx, err := tag.New(
-		r.ctx,
-		tag.Insert(desiredAPIVersionKey, req.DesiredAPIVersion),
-		tag.Insert(resultStatusKey, resp.Result.Status),
-		tag.Insert(resultReasonKey, string(resp.Result.Reason)),
-		tag.Insert(resultCodeKey, strconv.Itoa(int(resp.Result.Code))),
-	)
+	mutators := []tag.Mutator{}
```
dprotaso, line 185: Same feedback here as above

1. Precompute the list of keys we care about
2. Ensure the mutator list is initialized to the right capacity

webhook/stats_reporter.go
```diff
@@ -65,43 +65,111 @@ var (
...
+)
+
 // StatsReporter reports webhook metrics
 type StatsReporter interface {
 	ReportAdmissionRequest(request *admissionv1.AdmissionRequest, response *admissionv1.AdmissionResponse, d time.Duration) error
 	ReportConversionRequest(request *apixv1.ConversionRequest, response *apixv1.ConversionResponse, d time.Duration) error
 }
 
+type options struct {
+	tagsToExclude map[string]struct{}
```
dprotaso, line 124: For sets we've been using `sets.Set[string]` from k8s util package

webhook/stats_reporter.go
```diff
@@ -65,43 +65,111 @@ var (
...
+}
+
+type Option func(_ *options)
+
+func WithoutTag(tag string) Option {
```
dprotaso, line 129: Thoughts on allow multiple entries here?

```suggestion
func WithoutTags(tag ...string) Option {
```

dprotaso:
Overall looks great thanks for the changes - some minor stuff

webhook/stats_reporter.go
zhouhaibing089: If I understand this correctly, this can't be pre-computed, as this is where tag values are actually computed (per request).

webhook/stats_reporter.go
zhouhaibing089: I think that's a great idea!

webhook/stats_reporter.go
zhouhaibing089: Make sense! I moved this computation into `NewStatsReporter`.

webhook/stats_reporter.go
zhouhaibing089: Updated.

webhook/stats_reporter.go
zhouhaibing089: This is updated.

webhook/stats_reporter.go
zhouhaibing089: Great. I was trying to minimize the dependency, but if it is already part of the dependency, I'll use it.

webhook/webhook_test.go
zhouhaibing089: I pushed a second commit to add `StatsReporterOptions` in `webhook.Options`.

dprotaso (commented):
thanks for the changes - just some minor `nil` checks left

webhook/webhook.go
```diff
@@ -144,7 +147,7 @@ func New(
 	logger := logging.FromContext(ctx)
 
 	if opts.StatsReporter == nil {
-		reporter, err := NewStatsReporter()
+		reporter, err := NewStatsReporter(opts.StatsReporterOptions...)
```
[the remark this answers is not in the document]
zhouhaibing089: I think this should be [fine][1] even when `opts.StatsReporterOptions` is not set explicitly.

[1]: https://go.dev/play/p/bXynueunGFY

injection/sharedmain/main.go
```diff
@@ -290,7 +290,8 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	var wh *webhook.Webhook
 	if len(webhooks) > 0 {
 		// Register webhook metrics
-		webhook.RegisterMetrics()
+		opts := webhook.GetOptions(ctx)
```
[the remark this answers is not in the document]
zhouhaibing089: Thanks. Done.

dprotaso:
/lgtm
/approve

thanks @zhouhaibing089 🎉

PR #2964, opened 2024-02-20
Bump min k8s version to v1.28

izabelacg:
Fixes #2958 

# Changes

- bump min k8s version to v1.28
- bump client-go dependencies to v0.29.2
- remove k8s.io/client-go/informers/flowcontrol/v1alpha1
  - upstream change: https://github.com/kubernetes/client-go/commit/def2075eacb10d31a36acc337f6235c7c33b18e7
- remove k8s.io/client-go/informers/policy/v1beta1/podsecuritypolicy
   - upstream change: https://github.com/kubernetes/client-go/commit/1e11b3a8ca1cfd5fb196f05de2a6d77379d22fea
- remove k8s.io/client-go/informers/networking/v1alpha1/clustercidr
   - upstream change: https://github.com/kubernetes/client-go/commit/656ec39ea1f5267c82b9146de6856e75686e5591

PR #2966, opened 2024-02-21
drop toolchain directive

dprotaso:
I missed that we bumped some non-dependency properties in the go.mod file in https://github.com/knative/pkg/pull/2964

We shouldn't need to specify the `toolchain` directive - this oddly this breaks our go-licenses tool.

Bumping go version to v1.21 should be fine

see failure: https://github.com/knative-extensions/net-http01/pull/593

/assign @izabelacg @ReToCode @skonto

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,8 +2,6 @@ module knative.dev/pkg
 
 go 1.21
 
-toolchain go1.21.1
-
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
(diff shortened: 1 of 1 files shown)


dprotaso:
client failures are expected - k8s client-go had breaking changes

ReToCode (approved):
/lgtm
/approve

We really need to remove that licence-tool. `go xxx` will always re-introduce that directive when the version changes.

PR #2969, opened 2024-02-21
drop eventing-rabbitmq downstream testing

dprotaso:
Part of https://github.com/knative/eventing/issues/7670

/assign @izabelacg 
/cc @pierDipi


dprotaso:
@pierDipi unsure if there are other eventing integrations that don't need to be tested anymore

PR #2976, opened 2024-02-28
Provide option to exclude a registered informer from starting

skonto:
# Changes

- Adds a context based predicate to filter out registered informers that should not start.
- Needed by https://github.com/knative/serving/pull/14955 (integrating net-certmanager in Serving)
- Here is how to use the filtering mechanism:
```
// informer to skip from starting, it implies that we have generated code for that informer as usual
"knative.dev/serving/pkg/net-certmanager/client/certmanager/injection/informers/acme/v1/challenge"

func main () {
       ....
        if shouldEnableNetCertManagerController(ctx, client) {
		ctors = append(ctors, certificate.NewController) // add the net-certmanager controller
	} else {
		// Remove all non required informers
		ctx = injection.WithInformerExcludingFilter(ctx, excludeNetCertManagerInformers)
	}
	sharedmain.MainWithConfig(ctx, "controller", cfg, ctors...)
}

func excludeNetCertManagerInformers(ctx context.Context, inf controller.Informer) (bool, string) {
	switch {
	case v1certificate.Get(ctx).Informer() == inf:
		return true, "CertificateInformer"
	case ...
	}
	return false, ""
}
```

This works because injection via an init func stores instances of [cache.SharedIndexInformer](https://github.com/knative/pkg/blob/main/injection/clients/namespacedkube/informers/core/v1/secret/secret.go#L39), it calls `inf.Informer()`,  which are unique at the k8s client go side (one per watched resource type). Thus we can always get back the same instance (`inf.Informer()`) and compare.
* Tested in https://github.com/knative/serving/pull/14955

Output in controller logs without any flag set:

```
2024/03/08 11:42:15 Registering 6 clients
2024/03/08 11:42:15 Registering 6 informer factories
2024/03/08 11:42:15 Registering 20 informers
2024/03/08 11:42:15 Registering 9 controllers
{"level":"info","ts":1709898135.6646473,"logger":"fallback","caller":"injection/injection.go:68","msg":"Excluding informer ChallengeInformer"}
{"level":"info","ts":1709898135.6646998,"logger":"fallback","caller":"injection/injection.go:68","msg":"Excluding informer CertificateInformer"}
{"level":"info","ts":1709898135.6647108,"logger":"fallback","caller":"injection/injection.go:68","msg":"Excluding informer CertificateRequestInformer"}
{"level":"info","ts":1709898135.664719,"logger":"fallback","caller":"injection/injection.go:68","msg":"Excluding informer ClusterIssuerInformer"}
{"level":"info","ts":1709898135.6647272,"logger":"fallback","caller":"injection/injection.go:68","msg":"Excluding informer IssuerInformer"}
....
{"level":"info","ts":1709898135.7881877,"logger":"fallback","caller":"injection/injection.go:76","msg":"Starting 15 informers..."}

```

In config-network cm we set  `external-domain-tls: enabled`, install cert-manager and then delete the previous controller pod:

```
2024/03/08 11:47:35 Registering 6 clients
2024/03/08 11:47:35 Registering 6 informer factories
2024/03/08 11:47:35 Registering 20 informers
2024/03/08 11:47:35 Registering 10 controllers
...
{"level":"info","ts":1709898455.9024231,"logger":"fallback","caller":"injection/injection.go:76","msg":"Starting 20 informers..."}

```
All related informers are started.

* It could be an option to avoid even registering the informers but that requires further changes at the codegen part which is more intrusive imho.

/kind enhancement

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/injection/injection.go
+++ b/injection/injection.go
@@ -59,7 +59,7 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Contex
 
 	ctx, informers := Default.SetupInformers(ctx, cfg)
 
-	var includedInformers []controller.Informer
+	includedInformers := make([]controller.Informer, 0)
 
 	for _, inf := range informers {
 		if p := getExcludeInformerPredicate(ctx); p != nil && (*p)(ctx, inf) {
--- a/injection/injection.go
+++ b/injection/injection.go
@@ -80,6 +80,7 @@ type ExcludeInformerPredicate func(ctx context.Context, inf controller.Informer)
 type excludeInformerPredicateKey struct{}
 
 // WithExcludeInformerPredicate sets the predicate to exclude informers from being started.
+// If predicate returns true the informer will not be started.
 func WithExcludeInformerPredicate(ctx context.Context, predicate ExcludeInformerPredicate) context.Context {
 	return context.WithValue(ctx, excludeInformerPredicateKey{}, predicate)
 }
--- a/injection/injection.go
+++ b/injection/injection.go
@@ -59,10 +59,36 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Contex
 
 	ctx, informers := Default.SetupInformers(ctx, cfg)
 
+	var includedInformers []controller.Informer
+
+	for _, inf := range informers {
+		if p := getExcludeInformerPredicate(ctx); p != nil && (*p)(ctx, inf) {
+			continue
+		}
+		includedInformers = append(includedInformers, inf)
+	}
+
 	return ctx, func() {
-		logging.FromContext(ctx).Info("Starting informers...")
-		if err := controller.StartInformers(ctx.Done(), informers...); err != nil {
+		logging.FromContext(ctx).Infof("Starting %d informers...", len(includedInformers))
+		if err := controller.StartInformers(ctx.Done(), includedInformers...); err != nil {
 			logging.FromContext(ctx).Fatalw("Failed to start informers", zap.Error(err))
 		}
 	}
+}
+
+type ExcludeInformerPredicate func(ctx context.Context, inf controller.Informer) bool
+type excludeInformerPredicateKey struct{}
+
+// WithExcludeInformerPredicate sets the predicate to exclude informers from being started.
+func WithExcludeInformerPredicate(ctx context.Context, predicate ExcludeInformerPredicate) context.Context {
+	return context.WithValue(ctx, excludeInformerPredicateKey{}, predicate)
+}
+
+func getExcludeInformerPredicate(ctx context.Context) *ExcludeInformerPredicate {
+	untyped := ctx.Value(excludeInformerPredicateKey{})
+	if untyped == nil {
+		return nil
+	}
+	e := untyped.(ExcludeInformerPredicate)
+	return &e
 }
--- a/injection/injection.go
+++ b/injection/injection.go
@@ -62,8 +62,12 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Contex
 	includedInformers := make([]controller.Informer, 0)
 
 	for _, inf := range informers {
-		if p := getExcludeInformerPredicate(ctx); p != nil && (*p)(ctx, inf) {
-			continue
+		if p := getInformerExcludingFilter(ctx); p != nil {
+			skip, name := (*p)(ctx, inf)
+			if skip {
+				logging.FromContext(ctx).Infof("Excluding informer %s", name)
+				continue
+			}
 		}
 		includedInformers = append(includedInformers, inf)
 	}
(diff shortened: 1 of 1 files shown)


injection/injection.go
```diff
@@ -59,10 +59,37 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Contex
 
 	ctx, informers := Default.SetupInformers(ctx, cfg)
 
+	includedInformers := make([]controller.Informer, 0)
+
+	for _, inf := range informers {
+		if p := getExcludeInformerPredicate(ctx); p != nil && (*p)(ctx, inf) {
+			continue
```
ReToCode, line 66: maybe log when an informer is ignored?

injection/injection.go
```diff
@@ -59,10 +59,37 @@ func EnableInjectionOrDie(ctx context.Context, cfg *rest.Config) (context.Contex
...
 		}
 	}
 }
+
+type ExcludeInformerPredicate func(ctx context.Context, inf controller.Informer) bool
+type excludeInformerPredicateKey struct{}
+
+// WithExcludeInformerPredicate sets the predicate to exclude informers from being started.
+// If predicate returns true the informer will not be started.
+func WithExcludeInformerPredicate(ctx context.Context, predicate ExcludeInformerPredicate) context.Context {
```
ReToCode, line 84: Maybe "WithInformerExcludingFilter" or something in that direction?

ReToCode:
Nice :)

skonto:
@dprotaso gentle ping, could you pls review?

injection/injection.go
skonto: fixed

injection/injection.go
skonto: fixed

skonto:
@dprotaso gentle ping, could you pls review?

pierDipi (commented):
How do you see a non-started informer to be started?

pierDipi:
I wonder if in that case using manually creating informers is easier?

```
infX := factory.Get(ctx).ForResource(...)
infY := factory.Get(ctx).ForResource(...)
..  := controller.StartInformers(ctx, infX, infY)

// use infX and infY
```

ReToCode:
> I wonder if in that case using manually creating informers is easier?

I think we (as in the controller in Serving) do not know the informers, as they are added dynamically based on what packages you import (they self-register in their `init()` functions).

pierDipi:
I think you do, you need to avoid importing the injection informers for optional resources

ReToCode:
> I think you do, you need to avoid importing the injection informers for optional resources and just import the factory package and the concrete informers (if really needed)

AFAIK, the Serving controller itself holds multiple "subcontrollers". The parent one starts all the informers (via sharedmain) that are on the context (so it does not know which ones upfront). The subcontrollers add them to the context indirectly based on the init functions. So IMHO we would need to change quite a lot to do that. I mean here we just want to ignore a specific one when the CRDs are not installed. Maybe @skonto has something to add?

pierDipi:
Example on the serving PR:

```golang
import "knative.dev/serving/pkg/net-certmanager/client/certmanager/injection/informers/factory/factory"
// ...

infCertificates := factory.Get(ctx).Certmanager().V1().Certificates()
...  := controller.StartInformers(ctx, infCertificates)

// ...
```

pierDipi:
You need to *avoid* importing packages like to opt-out of injection:

```golang
"knative.dev/serving/pkg/net-certmanager/client/certmanager/injection/informers/certmanager/v1/certificate/certificate"

**Comment by skonto** [comment]
> I wonder if in that case using manually creating informers is easier?

I though about that I didn't want to re-write too much the main function for the [controller](https://github.com/knative/serving/pull/14955/files#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554d) we already have that for autoscaler, activator and wanted to provide a feature.

> You need to avoid importing packages like to opt-out of injection for that informer:

Yes because of the init thingy though I want to be able to start something on demand so I need to have it injected if I dont create it dynamically.

**Comment by skonto** [comment]
> How do you see a non-started informer to be started?

It is not, it is simply going to be excluded and never started.

**Comment by pierDipi** [comment]
> > I wonder if in that case using manually creating informers is easier?
> 
> I though about that I didn't want to re-write too much the main function for the [controller](https://github.com/knative/serving/pull/14955/files#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554d) we already have that for autoscaler, activator and wanted to provide a feature.
> 

With the solution I'm suggesting, the main function will remain to be the one that is currently on `main`, so another added benefit because with the PR you will need to maintain your own flags now.

> > You need to avoid importing packages like to opt-out of injection for that informer:
> 
> Yes because of the init thingy though I want to be able to start something on demand so I need to have it injected if I dont create it dynamically.

isn't this possible with the alternative solution? 
It's actually more dynamic and even possible to remove the requirement of "restart controller" when enabling TLS

**Comment by skonto** [comment]
> With the solution I'm suggesting, the main function will remain to be the one that is currently on main, so another added benefit because with the PR you will need to maintain your own flags now.

What I am saying unless you expand main then you don't report the started informers correctly because you use the factory to create them on demand within the targeted controller. See https://github.com/knative/pkg/blob/main/injection/sharedmain/main.go#L230. That is what I meant create it as a feature, go through main.  Btw you need to wait for the cache to be populated after startup which is a different UX.
If we don't mind about this deviation then I am fine. 


> It's actually more dynamic and even possible to remove the requirement of "restart controller" when enabling TLS

The idea of dynamic informers to avoid restarts has been around for sometime.

 I don't mind closing this and just creating/starting the informer in the corresponding controller but let's wait for @dprotaso and his take on this one.

**Comment by dprotaso** [comment]
It would be good to add flexibility to our codegen that does the injection 

eg. If can add an option to make this `init` optional and export the `WithInformer` function 

https://github.com/knative/pkg/blob/2c15a6fd07afbf402f8aa64799c4bd66cb8e22c4/codegen/cmd/injection-gen/generators/informer.go#L109-L120

Then in our mains we could decide to register the informer like so

```go
injection.Default.RegisterInformer(WithInformer)
```

skonto:
I wanted some quick fix to unblock stuff, but maybe doing a more long term solution is more beneficial. I will take a look
on the codegen approach.

skonto:
closing in favor of #2989

PR #2989, opened 2024-03-14
Optionally generate an init func for an informer

skonto:
# Changes

- Allows to kip the generation of the init func for an informer
- Based on the suggestion here https://github.com/knative/pkg/pull/2976#issuecomment-1994401541
- Replacement for https://github.com/knative/pkg/pull/2976
- Tested locally and here: https://github.com/knative/serving/pull/14955.
- It can be used as follows:
```
 Knative Injection (for cert-manager)
${KNATIVE_CODEGEN_PKG}/hack/generate-knative.sh "injection" \
  knative.dev/serving/pkg/net-certmanager/client/certmanager github.com/cert-manager/cert-manager/pkg/apis \
  "certmanager:v1 acme:v1" \
  --skipInitFuncForInformer \
  --go-header-file ${REPO_ROOT_DIR}/hack/boilerplate/boilerplate.go.txt
```

and then on demand register the right informer:
```
if shouldEnableNetCertManagerController(ctx, client) {
		injection.Default.RegisterInformer(challenge.WithInformer)
		injection.Default.RegisterInformer(v1certificate.WithInformer)
		injection.Default.RegisterInformer(certificaterequest.WithInformer)
		injection.Default.RegisterInformer(clusterissuer.WithInformer)
		injection.Default.RegisterInformer(issuer.WithInformer)
		ctors = append(ctors, certificate.NewController) // add the net-certmanager controller
	}
```
- We could have used a tag to skip per type but these are external types. In the feature we can do this per type for our 
types eg. `+skipInitInformer` and thus pass this directly to the InformerGenerator.
- We could expand it to filteredInformers if we want.

/assign @dprotaso 

/kind enhancement

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/codegen/cmd/injection-gen/args/args.go
+++ b/codegen/cmd/injection-gen/args/args.go
@@ -30,6 +30,7 @@ type CustomArgs struct {
 	ListersPackage                   string
 	ForceKinds                       string
 	ListerHasPointerElem             bool
+	DisableInformerInit              bool
 }
 
 // NewDefaults returns default arguments for the generator.
@@ -50,7 +51,7 @@ func (ca *CustomArgs) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&ca.ListerHasPointerElem, "lister-has-pointer-elem", false, "")
 	fs.MarkDeprecated("lister-has-pointer-elem", "this flag has no effect")
 
-	fs.Bool("skipInitFuncForInformer", false, "Skip the init function for informer")
+	fs.BoolVar(&ca.DisableInformerInit, "disable-informer-init", false, "disable generating the init function for the informer")
 }
 
 // Validate checks the given arguments.
--- a/codegen/cmd/injection-gen/args/args.go
+++ b/codegen/cmd/injection-gen/args/args.go
@@ -49,6 +49,8 @@ func (ca *CustomArgs) AddFlags(fs *pflag.FlagSet) {
 
 	fs.BoolVar(&ca.ListerHasPointerElem, "lister-has-pointer-elem", false, "")
 	fs.MarkDeprecated("lister-has-pointer-elem", "this flag has no effect")
+
+	fs.Bool("skipInitFuncForInformer", false, "Skip the init function for informer")
 }
 
 // Validate checks the given arguments.
--- a/codegen/cmd/injection-gen/generators/informer.go
+++ b/codegen/cmd/injection-gen/generators/informer.go
@@ -18,17 +18,15 @@ package generators
 
 import (
 	"io"
-	"os"
-	"strings"
 
 	clientgentypes "k8s.io/code-generator/cmd/client-gen/types"
 	"k8s.io/gengo/generator"
 	"k8s.io/gengo/namer"
 	"k8s.io/gengo/types"
 	"k8s.io/klog/v2"
-)
 
-const SkipInitFuncForInformerEnv = "SKIP_INIT_FUNC_FOR_INFORMER"
+	"github.com/spf13/pflag"
+)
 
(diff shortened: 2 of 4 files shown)


codegen/cmd/injection-gen/main.go
```diff
@@ -36,13 +37,17 @@ func main() {
 	genericArgs.AddFlags(pflag.CommandLine)
 	customArgs.AddFlags(pflag.CommandLine)
 	flag.Set("logtostderr", "true")
+	skipInitFuncForInformer := flag.Bool("skipInitFuncForInformer", false, "Skip the init function for informer")
```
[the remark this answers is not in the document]
skonto: Cool, I was not sure if I should touch those.

codegen/cmd/injection-gen/generators/informer.go
```diff
@@ -100,15 +104,21 @@ func (g *injectionGenerator) GenerateType(c *generator.Context, t *types.Type, w
 		}),
 	}
 
-	sw.Do(injectionInformer, m)
+	if os.Getenv(SkipInitFuncForInformerEnv) == "true" {
+		sw.Do(strings.ReplaceAll(injectionInformer, "withInformer", "WithInformer"), m)
+	} else {
+		sw.Do(initFunc+injectionInformer, m)
+	}
```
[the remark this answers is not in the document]
skonto: Ok didn't know it was using go templates behind the scenes, not it makes sense.

skonto:
@dprotaso gentle ping

codegen/cmd/injection-gen/args/args.go
```diff
@@ -49,6 +49,8 @@ func (ca *CustomArgs) AddFlags(fs *pflag.FlagSet) {
 
 	fs.BoolVar(&ca.ListerHasPointerElem, "lister-has-pointer-elem", false, "")
 	fs.MarkDeprecated("lister-has-pointer-elem", "this flag has no effect")
+
+	fs.Bool("skipInitFuncForInformer", false, "Skip the init function for informer")
```
dprotaso, line 53: We should wire this to the `CustomArgs` structs like we do for the other flags

Also can we follow the flag naming convention we are using  which is `snake-case`

I'd probably call this `disable-informer-init`

skonto: @dprotaso ok done

skonto:
@dprotaso pls take another look, thanks for the review!

codegen/cmd/injection-gen/generators/informer.go
```diff
@@ -78,6 +80,8 @@ func (g *injectionGenerator) GenerateType(c *generator.Context, t *types.Type, w
 
 	klog.V(5).Info("processing type ", t)
 
+	disableInformerInit, _ := pflag.CommandLine.GetBool("disable-informer-init")
```
dprotaso, line 83: we should be pulling this from the custom args struct and the funneling it into this generator

https://github.com/knative/pkg/blob/b8b7ca1f790479f3a6ea48084824357b9608d87e/codegen/cmd/injection-gen/generators/packages.go#L41

https://github.com/knative/pkg/blob/b8b7ca1f790479f3a6ea48084824357b9608d87e/codegen/cmd/injection-gen/generators/packages.go#L97

otherwise it's like we're reaching for globals.

The logic in packages is a bit verbose :/

skonto: It is ok I will adjust it.

skonto: @dprotaso done.

skonto:
@dprotaso ready, could you stamp it?

dprotaso:
/lgtm
/approve

thanks @skonto

PR #2997, opened 2024-03-25
drop special logic when handling kubelet probe header for istio mTLS

izabelacg:
# Changes

- Deprecates k-kubelet-probe header

Part of https://github.com/knative/serving/issues/14981

### DIFF
--- a/network/network.go
+++ b/network/network.go
@@ -59,11 +59,12 @@ const (
 	// KubeletProbeHeaderName is the header name to augment the probes, because
 	// Istio with mTLS rewrites probes, but their probes pass a different
 	// user-agent.
+	//
+	// Deprecated: use knative.dev/networking/pkg/http/header.UserAgentKey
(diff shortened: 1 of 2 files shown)


dprotaso:
This is a breaking change and not something we want.

The serving issue is about removing the bit of code where we set the header on the deployment readiness/liveness etc.

dprotaso:
Circling back - I misread the PR (early morning reviews are bad for me)

The changes in the PR are good but I think we should merge them after a later release

See details here: https://github.com/knative/networking/pull/950#discussion_r1539974433

/hold

dprotaso:
/retest
unrelated failure

```
{Failed  === RUN   TestNamespaceStream
    stream_test.go:316: Log was still produced, trying again...
    stream_test.go:323: No watching should have happened context canceled
--- FAIL: TestNamespaceStream (0.41s)
}
```

PR #3005, opened 2024-04-02
webhook: register metrics only with default stats reporter

zhouhaibing089:
This is a followup from [knative/pkg#2931][1]. RegisterMetrics may register unwanted metrics if the default stats reporter is not used.

[1]: https://github.com/knative/pkg/pull/2931

/kind cleanup

```release-note
None
```

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -25,6 +25,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"sync"
 	"time"
 
 	// Injection stuff
@@ -38,6 +39,10 @@ import (
 	admissionv1 "k8s.io/api/admission/v1"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/system"
+)
+
+var (
+	metricsOnce sync.Once
 )
 
 // Options contains the configuration for the webhook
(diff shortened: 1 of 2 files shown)


webhook/webhook.go
```diff
@@ -147,6 +152,8 @@ func New(
 	logger := logging.FromContext(ctx)
 
 	if opts.StatsReporter == nil {
+		// Register webhook metrics
+		metricsOnce.Do(func() { RegisterMetrics(opts.StatsReporterOptions...) })
```
dprotaso, line 156: Before `RegisterMetrics` would always be called.  Now it's only being invokved when the `opts.StatsReporter` isnt' set.

This seems like a breaking change

zhouhaibing089: I agree with your assessment. It might break users who specify `StatsReporter` yet still rely on `RegisterMetrics` to register views properly.

However I consider this as an unpleasant situation when users want to use a different `StatsReporter` but register the same view with slightly different tags.

To be fair, this is not something that I need anymore since the default StatsReporter can be customized already.

As it is today, `webhook.NewStatsReporter()` and `webhook.RegisterMetrics()` seems to be independent interfaces while they actually share the same data structure (tracks the same metrics tags, and relies on the same `opts.StatsReporterOptions` to initialize).

dprotaso: @skonto do you have any opinions?

skonto: I think a migration path would be to have a flag `disableDefaultMetrics` with default being `false`.

PR #3006, opened 2024-04-03
Allow skipping conversion crd update

skonto:
# Changes

- Downstream we need skip the crd update as this is done by OLM. See https://github.com/openshift-knative/serverless-operator/pull/2597
- We would like to be able to delegate crd updates and only keep the reconciler for validation.

/kind enhancement

### DIFF
--- a/webhook/resourcesemantics/conversion/controller.go
+++ b/webhook/resourcesemantics/conversion/controller.go
@@ -131,6 +131,8 @@ func newController(ctx context.Context, optsFunc ...OptionFunc) *controller.Impl
 		client:       client,
 		secretLister: secretInformer.Lister(),
 		crdLister:    crdInformer.Lister(),
+
+		skipReconcile: shouldSkipReconcile(ctx),
 	}
 
 	logger := logging.FromContext(ctx)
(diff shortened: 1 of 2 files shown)


dprotaso:
Why not just delete the conversion webhook controller from your binary?

skonto:
> Why not just delete the conversion webhook controller from your binary?

We keep this aligned with the knative/operator https://github.com/knative/operator/blob/main/config/crd/bases/operator.knative.dev_knativeservings.yaml#L2470
Not sure what is coming next (if any major change) like from v1beta1 to v1 for example. :shrug:

skonto:
@dprotaso wdyth about my last comment?

dprotaso:
I guess I'm still confused are you looking to skip conversion or skip the reconciliation that updates the cacert and service path?

If it's the CA cert then don't we have the knobs(options) to change the default secret and properties?


> We keep this aligned with the knative/operator https://github.com/knative/operator/blob/main/config/crd/bases/operator.knative.dev_knativeservings.yaml#L2470

I don't think you need to set that - it will default to what the API server supports - which is probably `v1` now? v1beta1 and v1 are equivalent (or it was when I last looked)

dprotaso:
I'm going to close this out - i think downstream it just makes sense to drop the certificate reconciler from the webhook controller

skonto:
> I guess I'm still confused are you looking to skip conversion webhook callbacks or skip the reconciliation that updates the cacert and service path?

I wanted to skip the update of the cacert as this is done from another operator.

> v1beta1 and v1 are equivalent (or it was when I last looked)
> I'm going to close this out - i think downstream it just makes sense to drop the certificate reconciler from the webhook controller

I wanted to keep this around in case there is a migration to v1, probably not?

PR #3027, opened 2024-04-29
Add composite ConfigStore to combine multiple `ConfigStore`

pierDipi:
In some repositories, we used multiple ConfigStores to store feature flags, for example, in eventing kafka broker, we use the Eventing core feature flags store and the EKB feature flags store

# Changes

- Add composite ConfigStore to combine multiple `ConfigStore`

### DIFF
--- a/reconciler/configstore.go
+++ b/reconciler/configstore.go
@@ -20,6 +20,20 @@ import "context"
 
 // ConfigStore is used to attach the frozen configuration to the context.
 type ConfigStore interface {
-	// ConfigStore is used to attach the frozen configuration to the context.
+	// ToContext is used to attach the frozen configuration to the context.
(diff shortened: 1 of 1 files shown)


skonto:
Do you see 
```
	for _, s := range stores {
		ctx = s.ToContext(ctx)
	}
```
being used often in different reconcilers?

pierDipi:
This can be reused everywhere we have multiple config stores (config stores being mainly configmaps), a service example could be: config-features and config-network in the case one reconciler depends on both

skonto:
> in the case one reconciler depends on both

Not sure if we have a scenario in Serving but ok I see how this can be useful.

pierDipi:
/unhold

@dprotaso I'm merging this as I don't see objections here

PR #3032, opened 2024-05-07
Update AuthStatus to have list of ServiceAccountNames

creydr:
Some components can have multiple identities for sending events (e.g. the Parallel). Therefor the AuthStatus should support exposing multiple IDs.


pierDipi (approved):
/lgtm
/approve

PR #3056, opened 2024-06-14
Add AuthenticatableType duck type

creydr:
Similar to what we have for Addressables, this PR adds the duck type `AuthenticatableType` for objects, which have a [`AuthStatus`](https://github.com/knative/pkg/blob/15e6cdf2f3868c9d3b1058d50cb91f1bdbe4e8af/apis/duck/v1/auth_types.go#L21-L30).

In addition it adds a resolver similar to the addressable_resolver to work with this type.


pierDipi (approved):
/lgtm
/approve

PR #3065, opened 2024-06-21
drop test-infra code

dprotaso:
This code was pulled into `knative/pkg` from `knative/test-infra` long ago

This code isn't need or living in `knative/toolbox` repo.

/assign @upodroid 
/assign @cardil 
/cc @pierDipi


dprotaso:
Diff is too large

/override "style / Golang / Lint"

PR #3066, opened 2024-06-21
[logstream] test polling should use a different context

dprotaso:
Fixes: https://github.com/knative/pkg/issues/3063

/assign @izabelacg 

Like 302-303 kills the context passed to logstream. This context was being used when polling which isn't correct

```go
	// Kill the context.
	cancel()
```


dprotaso:
different failure - `TestMetricsExport/OpenCensus`

/retest

PR #3070, opened 2024-06-26
Fix rand source in name

skonto:
# Changes

- Fixes issue with rand source concurrent access. See https://github.com/knative/serving/pull/15354

### DIFF
--- a/test/helpers/name.go
+++ b/test/helpers/name.go
@@ -19,6 +19,7 @@ package helpers
 import (
 	"math/rand"
 	"strings"
+	"sync"
 	"time"
 	"unicode"
 )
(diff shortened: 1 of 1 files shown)


test/helpers/name.go
```diff
@@ -75,9 +79,11 @@ func AppendRandomString(prefix string) string {
 // RandomString will generate a random string.
 func RandomString() string {
 	suffix := make([]byte, randSuffixLen)
+	nameMu.Lock()
 	for i := range suffix {
 		suffix[i] = letterBytes[nameRand.Intn(len(letterBytes))]
 	}
+	nameMu.Unlock()
```
dprotaso, line 86: minor tweak - let's just do a defer unlock

```suggestion
	nameMu.Lock()
	defer nameMu.Unlock()
	for i := range suffix {
		suffix[i] = letterBytes[nameRand.Intn(len(letterBytes))]
	}

```

dprotaso:
actually let's get this merged - the change can be done in a follow up

/lgtm
/approve

PR #3084, opened 2024-08-11
bump k8s deps to v0.30.3

dprotaso:
- **bump k8s version**
- **vendor k8s code from 1.29**


codegen/cmd/injection-gen/generators/namer.go
```diff
@@ -0,0 +1,58 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
dprotaso, line 2: Don't want to change this boilerplate cause I copied it from `k8s.io/code-generator` repo

dprotaso:
`verify / Verify Deps and Codegen` - complains go.mod file should have `1.22.0` but we had `1.22` before. I don't recall if pinning the patch is bad or not

dprotaso:
/override "style / Golang / Boilerplate Check (go)"

ReToCode (approved):
/lgtm
/approve

go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/pkg
 
-go 1.22
+go 1.22.0
```
dsimansk, line 3: @dprotaso is there any specific reason for the go patch version in `go` directive? Asking for the silent error in `update-deps` run[1]. Or any other potential side effects?

[1] https://github.com/knative-extensions/knobots/actions/runs/10352426965/job/28653073330#step:6:430

dprotaso: @dsimansk the go toolchain was forcing it - I couldn't figure out how to let it remain 1.22

dsimansk: Yep, it's force through k8s.io modules. There's top level go.work with `go 1.22.0` that's reflected through toolchain `go.mod` upadate behavior to every importing module as well.

I guess, we won't be able to get rid off for a while. 

https://github.com/kubernetes/kubernetes/blob/release-1.30/staging/src/k8s.io/client-go/go.mod#L5

PR #3095, opened 2024-09-23
:broom: Give users the ability to disable namespace ownership of webhook configurations

jonathan-innis:
# Changes

Allow users to configure `DisableNamespaceOwnership` in webhook options. This ensures that users that have other manifest managers of ValidatingWebhookConfigurations and MutatingWebhookConfigurations don't hit conflicts when ownership is overridden.

This has specific detrimental impact when using ArgoCD which [checks for ownerReferences before determining whether it should prune a resource](https://github.com/argoproj/gitops-engine/blob/master/pkg/cache/cluster.go#L1167). This has [caused issues for `kubernetes-sigs/karpenter`](https://github.com/aws/karpenter-provider-aws/issues/6982), which uses `knative/pkg` for webhook management.

I've tested this change in the `sigs.k8s.io/karpenter` codebase with [this commit](https://github.com/jonathan-innis/karpenter/commit/0922f28fae7718a3f46912c859f15a67a5ed757d) and it's completely solved our Argo interaction issues.

/kind cleanup

Fixes https://github.com/knative/serving/issues/15483

**Release Note**

```release-note
Allow passing `DisableNamespaceOwnership` in `webhook.Options` to remove webhook ownerReferences
```

**Docs**

```docs

```

### DIFF
--- a/webhook/configmaps/configmaps.go
+++ b/webhook/configmaps/configmaps.go
@@ -58,7 +58,8 @@ type reconciler struct {
 	vwhlister    admissionlisters.ValidatingWebhookConfigurationLister
 	secretlister corelisters.SecretLister
 
-	secretName string
+	secretName                string
+	disableNamespaceOwnership bool
 }
 
 var (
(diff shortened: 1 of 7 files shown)


dprotaso:
Instead of adding such a knob I'm wondering if we should remove this functionality 

@ReToCode @pierDipi @skonto curious about your thoughts?

dprotaso:
I'm going to approve this to unblock - but I think we should revisit the practice of the setting the owner reference here

/lgtm
/approve

PR #3096, opened 2024-09-24
Use istio.sidecar.inject label instead of annotation

ReToCode:
Partially https://github.com/knative/serving/issues/15521

## Proposed Changes
* Use istio.sidecar.inject label instead of annotation

/assign @skonto


dprotaso:
actually might need a follow up depending on the answer to this question:

https://github.com/knative-extensions/net-istio/pull/1369#issuecomment-2371351757

PR #3103, opened 2024-10-14
Add `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` env-var

thesuperzapper:
# Changes

- Added a new environment variable `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` to set the `DisableNamespaceOwnership` webhook option added by https://github.com/knative/pkg/pull/3095
- This allows the downstream apps (like Knative Serving) to make use of the ability to not have the Namespace own the webhook configurations, which breaks ArgoCD: https://github.com/knative/serving/issues/15483

/kind enhancement

Related https://github.com/knative/serving/issues/15483

**Release Note**

```release-note
Added `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` env-var to set `DisableNamespaceOwnership` in `webhook.Options`
```

**Docs**

```docs

```

### DIFF
--- a/webhook/env.go
+++ b/webhook/env.go
@@ -32,6 +32,8 @@ const (
 	secretNameEnvKey = "WEBHOOK_SECRET_NAME" //nolint:gosec // This is not a hardcoded credential
 
 	tlsMinVersionEnvKey = "WEBHOOK_TLS_MIN_VERSION"
+
+	disableNamespaceOwnershipEnvKey = "WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP"
 )
 
 // PortFromEnv returns the webhook port set by portEnvKey, or default port if env var is not set.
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@dprotaso this is a follow up to https://github.com/knative/pkg/pull/3095 so that we can actually use it in KNative Serving and fix ArgoCD.

webhook/env.go
```diff
@@ -82,3 +84,15 @@ func TLSMinVersionFromEnv(defaultTLSMinVersion uint16) uint16 {
 		panic(fmt.Sprintf("the environment variable %q has to be either '1.2' or '1.3'", tlsMinVersionEnvKey))
 	}
 }
+
+func DisableNamespaceOwnershipFromEnv() *bool {
+	disableNamespaceOwnership := os.Getenv(disableNamespaceOwnershipEnvKey)
+	if disableNamespaceOwnership == "" {
+		return nil
```
dprotaso, line 91: Can we default this as `false` and have the return type be `bool`

thesuperzapper: @dprotaso the behavior I was intending was that `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` overrides the `DisableNamespaceOwnership` webhook option __only__  when the env-var is set.

To do this, I need to be able to distinguish between an unset env-var and a `false` one:

https://github.com/knative/pkg/pull/3103/files#diff-71e3f02004a234858e10c42bf73e303ba86e5000ad4a1a95ccf5f6d94447c5e8R169-R174

dprotaso:
/hold 

see: https://github.com/knative/serving/issues/15483#issuecomment-2420022338

thesuperzapper:
@dprotaso I was saying in https://github.com/knative/serving/issues/15483#issuecomment-2420275511, we should make both changes (that is, set the controller flag to false, AND let users disable the owner references fully).

As long as you agree on the way this PR does the disabling, it should be safe to merge.

Also, it would be a real shame if KServe/Kubeflow/deployKF had to run a forked version of kNative for this simple change that doesn't change any default behavior.

dprotaso:
/lgtm
/approve

I'll merge this in to unblock for now - but I don't consider this functionality stable. We may make edits to the mechanism in a future date especially if ArgoCD merges the PR mentioned in the serving issue.

/cherry-pick release-1.16

dprotaso:
OpenCensus flake

/retest

PR #3107, opened 2024-10-17
Fix `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` env-var

thesuperzapper:
# Changes

- Fixes the implementation from https://github.com/knative/pkg/pull/3103
- Previously, the env-var would not properly apply for knative-serving, because the controller (which adds the owner references) is initialized before the webhook starts (so we need to move the env-var override to the controller constructor).

/kind bug

Related https://github.com/knative/serving/issues/15483

**Release Note**

```release-note
Fixed `WEBHOOK_DISABLE_NAMESPACE_OWNERSHIP` env-var
```

**Docs**

```docs

```

### DIFF
--- a/webhook/configmaps/controller.go
+++ b/webhook/configmaps/controller.go
@@ -46,6 +46,12 @@ func NewAdmissionController(
 	secretInformer := secretinformer.Get(ctx)
 	options := webhook.GetOptions(ctx)
 
+	// if this environment variable is set, it overrides the value in the Options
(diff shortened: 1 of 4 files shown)


thesuperzapper:
@dprotaso Sorry for the quick follow-up, but the implementation in https://github.com/knative/pkg/pull/3103 did not quite work in my testing. This PR fixes it, and I have tested it on the deployKF fork of Knative Serving and it works.

__NOTE:__ this PR is based on https://github.com/knative/pkg/pull/3103 so both need to be cherry-picked into any release branches.

PR #3133, opened 2025-01-09
run dependabot workflow only on dependabot branches

dprotaso:
/assign @skonto 

Got the idea from - https://github.com/dependabot/dependabot-core/issues/935#issuecomment-698481919


dprotaso:
eventing downstream seems broken and unrelated to the PR

PR #3146, opened 2025-02-06
Update Knative injection-gen to use gengov2

dprotaso:
- **refactor codegen into different packages and use gengo v2**
- **updated go.mod drops gengo v1**
- **tweak scripts to support gengov2 framework**


dprotaso:
/override "style / Golang / Boilerplate Check (go)"

n3wscott (approved):
It looks ok based on what I see the v2 gengo needs and the upstream tests passing.

codegen/cmd/injection-gen/namer/namer.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Knative Authors.
+Copyright 2016 The Kubernetes Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
```
skonto, line 4: I guess we are ok to have Kubernetes Authors?

client/injection/kube/informers/admissionregistration/v1/validatingadmissionpolicy/filtered/validatingadmissionpolicy.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2022 The Knative Authors
```
skonto, line 2: is 2022 correct?

hack/generate-knative.sh
```diff
@@ -82,11 +82,15 @@ if grep -qw "injection" <<<"${GENS}"; then
   # Clear old injection
   rm -rf ${OUTPUT_PKG}
 
+  MODULE_NAME=$(go list -m)
```
skonto, line 85: This is just `knative.dev/pkg` it seems. Do we need this?

dprotaso: this script is used in other repos to generate their injection code - so it's not always knative.dev/pkg

client/injection/kube/informers/admissionregistration/v1/validatingadmissionpolicy/filtered/validatingadmissionpolicy.go
dprotaso: let me update the boilerplate

hack/generate-knative.sh
skonto: Ah correct ok

PR #3164, opened 2025-03-31
Add a flag in tracing config to use service instead of pod name

pomyslowynick:
### Add a flag in tracing config to use service instead of pod name

While the OTel migration is in progress it would be very helpful to have a flag which allows users toggling the `ServingService` to be used as the `service.name` in traces. 
This is related to [this issue](https://github.com/knative/serving/issues/12969) and would still be relevant after the migration to OTel.

From my own experience, `service.name` is used to group the traces together and when it's set to the pod name it's impossible to keep them all together.

This flag is a `noop` for existing configurations, since it defaults to `false` and will be used in [another PR](https://github.com/knative/serving/pull/15822)
I've tested it in a live cluster for true/false/undefined values in the config map, in all the cases it worked as expected.

/kind enhancement

**Release Note**

```release-note
You can now use the `useServingService` flag in the tracing config to use the `ServingService` as the `service.name` in traces. Current default is to use the `ServingPod` as the `service.name`, and it remains the same if the flag is not set.
```

### DIFF
--- a/tracing/config/tracing.go
+++ b/tracing/config/tracing.go
@@ -31,11 +31,12 @@ const (
 	// ConfigName is the name of the configmap
 	ConfigName = "config-tracing"
 
-	enableKey         = "enable"
-	backendKey        = "backend"
-	zipkinEndpointKey = "zipkin-endpoint"
-	debugKey          = "debug"
-	sampleRateKey     = "sample-rate"
+	enableKey            = "enable"
+	backendKey           = "backend"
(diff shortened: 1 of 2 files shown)


dprotaso:
/hold

I want to revisit this after the next release (Apr 22) - I think for the subsequent release we should do real work to migrate to OTel. Supposedly they are removing OpenCensus support in the OTel collector.

dprotaso:
Closing this out - OTel work is in progress and I'll make sure that we use the 'right' service name in Serving.

PR #3178, opened 2025-05-13
group golang.org/x dependencies

dprotaso:
this groups the golang.org/x/ dependencies which frequently conflict with each other when their versions are bumped


dprotaso:
cc @Fedosin

PR #3184, opened 2025-06-10
bump min k8s version to 1.32

dprotaso:
This bumps the min version of k8s to 1.32 (using client-go libs v0.33).

Later commits will also address linter fixes (which were triggered by bumping go version to 1.24)

### DIFF
--- a/apis/duck/v1/source_types.go
+++ b/apis/duck/v1/source_types.go
@@ -223,6 +223,7 @@ func validateExtensionName(key string) *apis.FieldError {
 	}
 
 	for _, c := range key {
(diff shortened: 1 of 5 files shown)


dprotaso:
looks like go-fuzz was archived so k8s community created their own fork

I'm guessing we'll have breaking changes downstream

dprotaso:
/hold 

since we bumped the go version to 1.24 there are some linters that are being triggered - let me address those

dprotaso:
ok this is ready

dsimansk (approved):
/approve
/lgtm

PR #3185, opened 2025-06-11
pull configmap parsing into separate package

dprotaso:
We want to parse configmaps but not don't want a heavy dependency on k/k libs

Part of https://github.com/knative/serving/issues/9957

This will matter as a part of the OTel migration - since we will want to parse the config in the queue-proxy


dprotaso:
TestMetricsExport flake

/test unit-tests_pkg_main

PR #3186, opened 2025-06-17
refactor webhook testing

dprotaso:


dprotaso:
We can ignore the downstream tests - that's an unrelated error from bumping k8s deps that needs to be fixed.


/assign @dsimansk @skonto @Fedosin

PR #3189, opened 2025-06-25
[webhook] OTel changes

dprotaso:
This stacks onto the shared main PR - https://github.com/knative/pkg/pull/3190

You can review the webhook changes here

- **webhook changes for observability**
- **go mod**
- **vendor**


dprotaso:
This is ready for a review cc @evankanderson @Cali0707 

Based on the feedback from the OTel folks I removed the package variable instruments.  This makes unit tests easier to write and FYI the instruments are de-duped by the otel-go sdk when multiple webhooks are created (excluding the pointers in our webhook struct)

webhook/admission.go
```diff
@@ -120,12 +137,23 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
...
 		var patchType string
 		if reviewResponse.PatchType != nil {
 			patchType = string(*reviewResponse.PatchType)
 		}
 
+		allowedAttr := AdmissionAllowed.With(reviewResponse.Allowed)
+		attrs = append(attrs, allowedAttr)
+		labeler.Add(allowedAttr)
+
+		wh.metrics.recordHandlerDuration(ctx,
```
skonto, line 152: We report time for bad requests too?

webhook/webhook.go
```diff
@@ -70,17 +75,14 @@ type Options struct {
...
-
-	// StatsReporter reports metrics about the webhook.
-	// This will be automatically initialized by the constructor if left uninitialized.
-	StatsReporter StatsReporter
-
 	// GracePeriod is how long to wait after failing readiness probes
 	// before shutting down.
 	GracePeriod time.Duration
 
+	// // WithMeterProvider specifies a meter provider to use for creating a meter.
```
skonto: remove double comments?

webhook/webhook_test.go
```diff
@@ -65,6 +65,8 @@ func newNonRunningTestWebhook(t *testing.T, options Options, acs ...interface{})
 		stopCb()
 	}
 
+	// acs = slices.DeleteFunc(acs, func(a any) bool { return a == nil })
```
skonto: remove this line?

webhook/admission.go
dprotaso: yeah

webhook/webhook_test.go
dprotaso: done

webhook/webhook.go
dprotaso: done

webhook/admission.go
evankanderson: Bad requests may have a different time distribution than "good" requests, which can be a useful hint when debugging, so I'm +1 on separating these.

webhook/admission.go
```diff
@@ -100,6 +104,19 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
 		}
 		r.Body = io.NopCloser(&bodyBuffer)
 
+		labeler, _ := otelhttp.LabelerFromContext(r.Context())
```
evankanderson, line 106: If the second result is `false`, then:

>  In this case it is safe to use the Labeler but any attributes added to it will not be used.

Do we want / need to call `otelhttp.ContextWithLabeler` to set the labeler if this was false?

webhook/admission.go
```diff
@@ -69,15 +72,17 @@ type StatelessAdmissionController interface {
...
 	result := apierrors.NewBadRequest(fmt.Sprintf(reason, args...)).Status()
 	return &admissionv1.AdmissionResponse{
 		Result:  &result,
 		Allowed: false,
 	}
 }
 
-func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c AdmissionController, synced <-chan struct{}) http.HandlerFunc {
+func admissionHandler(wh *Webhook, c AdmissionController, synced <-chan struct{}) http.HandlerFunc {
+	webhookTypeAttr := WebhookType.With(WebhookTypeAdmission)
```
evankanderson: Why do this here, rather than on line 115?

webhook/admission.go
```diff
@@ -100,6 +104,19 @@ func admissionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Admi
...
 		r.Body = io.NopCloser(&bodyBuffer)
 
+		labeler, _ := otelhttp.LabelerFromContext(r.Context())
+
+		attrs := []attribute.KeyValue{
+			AdmissionOperation.With(string(review.Request.Operation)),
+			AdmissionGroup.With(review.Request.Kind.Group),
+			AdmissionVersion.With(review.Request.Kind.Version),
+			AdmissionKind.With(review.Request.Kind.Kind),
+			AdmissionSubresource.With(review.Request.SubResource),
```
evankanderson: Can you order this the same as the logger, e.g. (one possible order, group/version are not logged but are in metrics):

```suggestion
			AdmissionKind.With(review.Request.Kind.Kind),
			AdmissionGroup.With(review.Request.Kind.Group),
			AdmissionVersion.With(review.Request.Kind.Version),
			AdmissionOperation.With(string(review.Request.Operation)),
			AdmissionSubresource.With(review.Request.SubResource),
```

webhook/admission_integration_test.go
```diff
@@ -676,6 +686,30 @@ func TestAdmissionValidResponseForRequestBody(t *testing.T) {
...
+			"", // any metric
+			WebhookType.With(WebhookTypeAdmission),
+			AdmissionOperation.With("CREATE"),
+			AdmissionGroup.With("pkg.knative.dev"),
+			AdmissionVersion.With("v1alpha1"),
+			AdmissionKind.With("Resource"),
+			AdmissionSubresource.With(""),
+			AdmissionAllowed.With(allowed),
+		),
+	)
```
evankanderson: Do we get a new set of metrics for each test?  If not, it seems like the tests should pass for every case after the first `allowed` value without actually checking if the metrics were exported.  (Or I've missed how the reset happens.)

webhook/conversion.go
```diff
@@ -66,14 +76,21 @@ func conversionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Con
 			Response: c.Convert(ctx, review.Request),
 		}
 
+		wh.metrics.recordHandlerDuration(ctx, time.Since(ttStart),
```
evankanderson: This doesn't seem to include the `kind`?

webhook/conversion.go
```diff
@@ -66,14 +76,21 @@ func conversionHandler(rootLogger *zap.SugaredLogger, stats StatsReporter, c Con
...
+			metric.WithAttributes(
+				versionAttr,
+				webhookTypeAttr,
+				ConversionResultStatus.With(strings.ToLower(response.Response.Result.Status)),
+			),
+		)
+
+		labeler.Add(
+			ConversionResultStatus.With(strings.ToLower(response.Response.Result.Status)),
+		)
```
evankanderson: Why do this _after_ the `recordHandlerDuration` call?

webhook/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	// WebhookType is an attribute that specifies whether the type of webhook is an admission
+	// eg. (defaulting/validation) or conversion
+	WebhookType = attributekey.String("kn.webhook.type")
+
+	AdmissionOperation   = attributekey.String("kn.webhook.admission.operation")
+	AdmissionGroup       = attributekey.String("kn.webhook.admission.group")
+	AdmissionVersion     = attributekey.String("kn.webhook.admission.version")
+	AdmissionKind        = attributekey.String("kn.webhook.admission.kind")
+	AdmissionSubresource = attributekey.String("kn.webhook.admission.subresource")
+	AdmissionAllowed     = attributekey.Bool("kn.webhook.admission.result.allowed")
```
evankanderson: Do we need `admission` in these keys?  If not, they could cover all kinds of webhook metrics (conversion, defaulting, etc), with fewer definitions.  (e.g. the admission and conversion status could use the same attribute)

evankanderson (commented):
/approve

One concern about use of `LabelerFromContext` without looking at the second value in `admission.go`, but this seems like (not counting libraries), this somewhat simplifies our overall code.

webhook/admission.go
dprotaso: shouldn't need to since - the `otelhttp` middleware sets it.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/45af9252727c122174f7adab90f6c24ab04f1a6b/instrumentation/net/http/otelhttp/handler.go#L176

dprotaso: If someone's using the handler without that middleware then the labeler is a `noop`

webhook/admission.go
dprotaso: No reason - this value doesn't change so i though we could just declare it once.

webhook/admission_integration_test.go
dprotaso: > Do we get a new set of metrics for each test? 

We do - each test creates a manual reader so there's no need for a reset

webhook/conversion.go
dprotaso: It's part of the `attrs` declared in 109

dprotaso: oh sorry looked at the wrong file

evankanderson: (This may have been an existing gap)

webhook/admission.go
evankanderson: Unless you declare it statically outside the method, it feels like it would be just as efficient to put it next to the other labels.  (Unless I misunderstand the go optimizer, which is possible.  I also suspect that optimization here is probably not a big win.)

webhook/admission.go
evankanderson: Yeah, I didn't know if that was a concern.  You can add a comment here that this relies on that middleware setting the labeler context as another option, since this looks like ignoring an error value.

webhook/conversion.go
dprotaso: Ok - we don't know the kind here but we can add it through the labeler in the nested convert methods. Let me do that

evankanderson: (If something is too hard / non-blocking, feel free to defer it, but it does seem odd to have metrics on what versions were converted, but not which kinds.)

webhook/conversion.go
dprotaso: `recordHandlerDuration` will add the duration to our custom histogram metric that measures the call to the conversion controller.

The labeler on the other hand will add attributes that otelhttp middelware reports. The the order of these don't really mater in my mind

evankanderson: So the labeler labels won't have any relation to the metric attributes?

webhook/metrics.go
dprotaso: It's possible but a lot of the fields don't apply to conversion webhooks (maybe just two).  

What are your concerns?

webhook/conversion.go
dprotaso: ok - i added it via the labeler in the resourcesemantics/conversion handler

webhook/metrics.go
evankanderson: Having to type a lot of long strings into Prometheus queries.  (Yes, it's more human laziness than technical cost, which should be fairly minimal with a reasonable interning implementation.)

webhook/conversion.go
dprotaso: > So the labeler labels won't have any relation to the metric attributes?

There is no relationship. What we should probably do is actually get the attribute keys from the label when recording the metric - that way we'll have the exact same attributes (though there's a slice alloc)

dprotaso: ok - i moved this invocation upward and `record` will now fetch the labeler attributes

webhook/admission.go
dprotaso: done

webhook/admission.go
dprotaso: i'll keep it simple and change it

webhook/metrics.go
dprotaso: ok - i did a pass take a look - i think it worked out

dprotaso:
rebased

dprotaso:
/hold cancel

@evankanderson if you have any follow up comments let me know and i can do them in a follow up PR next week.

PR #3190, opened 2025-06-25
[injection/sharedmain] OTel Support

dprotaso:
This stacks onto https://github.com/knative/pkg/pull/3188


injection/sharedmain/main.go
```diff
@@ -482,3 +521,45 @@ func ControllersAndWebhooksFromCtors(ctx context.Context,
...
+		return val
+	}
+
+	// CRI-O and containerd set HOSTNAME to the Pod Name.
+	// This seems to be K8s behaviour - originating from
+	// older docker behaviour
+	if val := os.Getenv("HOSTNAME"); val != "" {
+		return val
+	}
+	return ""
```
evankanderson: Do we need the detailed comments here, or can we just do:

```suggestion
	return cmp.Or(os.Getenv("POD_NAME"), os.Getenv("HOSTNAME"), "")
```

injection/sharedmain/main.go
```diff
@@ -372,14 +378,53 @@ func SetupLoggerOrDie(ctx context.Context, component string) (*zap.SugaredLogger
...
 	if err != nil {
 		logger.Fatal("Error loading observability configuration: ", err)
 	}
-	observabilityConfigMap := observabilityConfig.GetConfigMap()
-	metrics.ConfigMapWatcher(ctx, component, SecretFetcher(ctx), logger)(&observabilityConfigMap)
-	profilingHandler.UpdateFromConfigMap(&observabilityConfigMap)
+
+	pprof.UpdateFromConfig(cfg.Runtime)
+
+	resource := getResource(component)
```
evankanderson: Do we want `getResource` to be in `observability` for sharing with non-`sharedmain` binaries?

injection/sharedmain/main.go
```diff
@@ -372,14 +378,53 @@ func SetupLoggerOrDie(ctx context.Context, component string) (*zap.SugaredLogger
...
+	tracerProvider, err := tracing.NewTracerProvider(
+		ctx,
+		cfg.Tracing,
+		trace.WithResource(resource),
+	)
+	if err != nil {
+		logger.Fatalw("Failed to setup trace provider", zap.Error(err))
+	}
+
+	otel.SetTextMapPropagator(tracing.DefaultTextMapPropagator())
```
evankanderson, line 420: This feels like it should be part of `tracing.NewTracerProvider`, even though it's actually a global spooky-action-at-a-distance manipulation.  WDYT?

injection/sharedmain/main.go
```diff
@@ -424,30 +469,24 @@ func WatchLoggingConfigOrDie(ctx context.Context, cmw *cminformer.InformedWatche
...
-	// even if you scope down the Lister to a given namespace after requesting it. Instead,
-	// we package up a function from kubeclient.
-	// TODO(evankanderson): If this direct request to the apiserver on each TLS connection
-	// to the opencensus agent is too much load, switch to a cached Secret.
-	return func(name string) (*corev1.Secret, error) {
-		return kubeclient.Get(ctx).CoreV1().Secrets(system.Namespace()).Get(ctx, name, metav1.GetOptions{})
+	if _, err := client.Get(ctx, cmName, metav1.GetOptions{}); err == nil {
+		cmw.Watch(cmName, observers...)
+	} else if !apierrors.IsNotFound(err) {
+		logger.Fatalw("Error reading ConfigMap "+cmName, zap.Error(err))
```
evankanderson, line 485: I'm assuming we don't have a way to notice if the ConfigMap shows up later?

injection/sharedmain/pprof.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type pprofServer struct {
+	*runtime.ProfilingServer
+	log *zap.SugaredLogger
+}
+
+func newProfilingServer(logger *zap.SugaredLogger) *pprofServer {
+	s := runtime.NewProfilingServer()
+	s.SetEnabled(false)
```
evankanderson: This feels weird, since it's also part of `runtime.NewProfilingServer()`, IIRC.

evankanderson (commented):
A few comments about usage, but this makes sense to me overall.

injection/sharedmain/main.go
dprotaso: We can do both - I think the comments are handy because the `HOSTNAME`  comment is useful because I only discovered it in the past week ;)

injection/sharedmain/main.go
dprotaso: > tracing.NewTracerProvider

That doesn't actually set the global tracer in NewTracerProvider. We do that in sharedmain after setting the SetTextMapPropagator. 

I prefer the observability package to _not_ set any OTel globals.

injection/sharedmain/main.go
dprotaso: This file could use a cleanup - I've just changed the 'Watch' here but the diff shows a larger change.

Generally, I think the controllers should _halt_ and wait if a config map is not present. I could imagine a split brain scenario occurring. 

I'd rather not make those edits in this PR.

injection/sharedmain/pprof.go
dprotaso: You're thinking about the old profiling server here - https://github.com/knative/pkg/tree/main/profiling

The new one I added in observability isn't 'aware' of parsing config maps. I'm trying to remove dependencies on corev1 types (reason. https://github.com/knative/serving/issues/9957)

So this acts as a shim between sharedmain watch and the profilng server

injection/sharedmain/main.go
evankanderson: That's fine.  I'm slighly worried about "halt and catch fire" as a behavior when a configmap we ship has been removed.  The idea of shipping them has mostly been about documentation, not about requiring them to have specific contents (or even be present) for things to function.

injection/sharedmain/main.go
evankanderson: There are cases where `POD_NAME` is not set by Kubernetes, I take it?

injection/sharedmain/main.go
```diff
@@ -482,3 +521,45 @@ func ControllersAndWebhooksFromCtors(ctx context.Context,
...
+	// CRI-O and containerd set HOSTNAME to the Pod Name.
+	// This seems to be K8s behaviour - originating from
+	// older docker behaviour
+	if val := os.Getenv("HOSTNAME"); val != "" {
+		return val
+	}
+	return ""
+}
+
+func getResource(component string) *resource.Resource {
```
evankanderson: I asked above, but does it make sense to put `GetResource` (and `GetPodName`) into the `observability` package?  It looks like it only has Knative dependencies `system` and `changeset`, and external dependencies on `semconv` and `resource`, and I can see other binaries wanting to use the same pattern.

injection/sharedmain/pprof.go
evankanderson: `NewProfilingHandler` calls `enabled.Store(false)`:

https://github.com/knative/pkg/blob/main/observability/runtime/pprof.go#L60

I think this felt slightly weird to me at the time (why not do it in `NewProfilingServer`?), but I think this sets it false twice.

injection/sharedmain/main.go
dprotaso: done

dprotaso:
I'm marking this ready - we can merge this and it will 'break' metrics/tracing @ HEAD - but I think that's worth it since it'll be easier to review and make follow up changes.

Plus metrics/tracing would have already been 'broken' downstream if we merged pkg changes

observability/resource/default.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"knative.dev/pkg/changeset"
+	"knative.dev/pkg/system"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.34.0"
+)
+
+// New returns a default OpenTelemetry Resource enriched
```
Cali0707: nit:
```suggestion
// Default returns a default OpenTelemetry Resource enriched
```

observability/resource/default.go
```diff
@@ -0,0 +1,58 @@
...
+		semconv.ServiceVersion(changeset.Get()),
+	}
+
+	if pn := system.PodName(); pn != "" {
+		attrs = append(attrs, semconv.K8SPodName(pn))
+	}
+
+	// Ignore the error because it complains about semconv
+	// schema version differences
+	resource, _ := resource.Merge(
```
dprotaso: there shouldn't be an error here anymore since the telemetry sdk attributes and resource now use the same semconv version.


But before if the semconv packages were different it would return this error with a resource without a schema
https://github.com/open-telemetry/opentelemetry-go/blob/957f4417b7719eeff17f051228520176fd2d5b78/sdk/resource/resource.go#L218

There was a bug in v1.36.0 where they didn't match up which was odd.

dprotaso:
/hold

Actually I want to make another related change

system/env.go
```diff
@@ -17,15 +17,17 @@ limitations under the License.
...
 	"os"
 )
 
 const (
 	// NamespaceEnvKey is the environment variable that specifies the system namespace.
 	NamespaceEnvKey = "SYSTEM_NAMESPACE"
+
 	// ResourceLabelEnvKey is the environment variable that specifies the system resource
-	// label.
+	// label. This label should be used to limit the
```
Cali0707: I think part of this comment went missing

system/env.go
```diff
@@ -60,3 +62,28 @@ import (
...
+//	- name: MY_POD_NAME
+//	  valueFrom:
+//	    fieldRef:
+//	      fieldPath: metadata.name
+//
+// As a fallback it will read HOSTNAME. This is undocumented
+// Kubernetes behaviour that podman, cri-o and containerd have
+// inherited from docker.
+//
+// If none of tehse env-vars is set PodName will return an
```
Cali0707:
```suggestion
// If none of these env-vars is set PodName will return an
```

Cali0707 (commented):
Just a few nits, but overall lgtm

observability/resource/default.go
dprotaso: I added a test and now we panic if there is an error.

We can always change this in the future if we find it annoying

dprotaso:
Talking to the OTel folks - views are an 'end-user' override. Unlike OpenCensus they are not mandatory to get metrics export to work.

So it seems more fitting to remove the global package I created and just make it a sharedmain option that controller authors can change.

Cali0707 (approved):
/lgtm
/hold in case @dprotaso wants more feedback from others

evankanderson (commented):
I'm okay with this as-is, but I had a couple comments that didn't get sent this morning.

injection/sharedmain/pprof.go
dprotaso: Oh - I see your point. Yeah I can drop the duplicate call.

injection/sharedmain/main.go
dprotaso: FYI POD_NAME is a knative convention using downward API that we use

injection/sharedmain/main.go
dprotaso: this was moved to `observability/resource`

evankanderson (commented):
/lgtm

PR #3199, opened 2025-06-30
When shutting down the OTel providers have a ten second timeout.

dprotaso:
I noticed a misconfigured metrics URL would prevent the process
from quiting because it was trying to flush metrics forever.


evankanderson:
/lgtm
/approve

I would even argue for a shorter shutdown (2-5s, and/or use a configurable timeout), but 10s seems plenty.

PR #3201, opened 2025-07-02
[controller] OTel instrumentation

dprotaso:
This changes the controller package to drop the use of the OpenCensus stats reporter.

As highlighted in the design document we had an overlap in our controller custom metrics and the ones provided by client-go workqueue package. For now we are removing the custom metrics and just going to rely on the workqueue metrics.

Breakdown of changes
- **add workqueue metrics interface**
- **incorporate k8s client-go tool metrics - request latency & result**
- **drop controller custom stats reporter**


observability/metrics/k8s/tools.go
```diff
@@ -0,0 +1,145 @@
...
+type result struct {
+	cp *ClientProvider
+}
+
+func (r *result) Increment(ctx context.Context, code string, method string, host string) {
+	codeInt := 200
+
+	if code == "200" {
+		// happy path - noop
+	} else if c, err := strconv.ParseInt(code, 10, 64); err != nil && c >= 100 && c < 600 {
```
evankanderson: If code is <100 or >=600, then it will be recorded as 200.  Maybe you want a special sentinel value in that case?

evankanderson (commented):
/approve

One concern in the client-go metrics, but this overall makes sense, and I like re-using the existing k8s metrics infrastructure, even if the workqueue implementation feels verbose (I like the tools version better).

observability/metrics/k8s/tools.go
dprotaso: good catch

observability/metrics/k8s/tools.go
```diff
@@ -0,0 +1,151 @@
...
+func (r *result) Increment(ctx context.Context, code string, method string, host string) {
+	var attrs attribute.Set
+
+	if code == "200" {
+		attrs = attribute.NewSet(
+			semconv.ServerAddressKey.String(host),
+			semconv.HTTPRequestMethodKey.String(strings.ToUpper(method)),
+			semconv.HTTPResponseStatusCodeKey.Int(200),
+		)
+	} else if c, err := strconv.Atoi(code); err != nil {
```
Cali0707, line 134: @dprotaso is there any value to having a specific code path just for the `code == "200"` case? 

As far as I can tell, the only difference between this block and the next is that we are manually setting the int value of the response status code in the attribute set, instead of using the parsed value. IMO, this would be a bit simpler to read if we just drop the special case for `200`.

Cali0707 (commented):
/approve

Overall, this looks great - just one small nit about how we are getting the response status code for the attributes

observability/metrics/k8s/tools.go
dprotaso: > is there any value to having a specific code path just for the code == "200" case?

I was trying to avoid the string parsing on the 'happy' path - if you feel it's too much we can revert an revisit if there are problems in the future

Cali0707: That makes sense, let's keep it as is then. Maybe add a comment for why?

dprotaso: done

Cali0707 (approved):
/lgtm

dprotaso:
/retest

Soon this will be gone -  `metrics: TestMetricsExport/OpenCensus`

PR #3202, opened 2025-07-02
Drop OpenCensus use in various test packages

dprotaso:
- **drop opencensus from test/logging and test/spoof**
- **run ./hack/update-dep.sh**

Merge https://github.com/knative/pkg/pull/3189 first


test/logging/logging.go
```diff
@@ -24,66 +24,29 @@ import (
...
-}
-
 // GetEmitableSpan starts and returns a trace.Span with a name that
 // is used by the ExportSpan method to emit the span.
 //
 //nolint:spancheck
-func GetEmitableSpan(ctx context.Context, metricName string) *trace.Span {
-	_, span := trace.StartSpan(ctx, emitableSpanNamePrefix+metricName)
+func GetEmitableSpan(ctx context.Context, metricName string) trace.Span {
+	_, span := noopTracer.Start(ctx, metricName)
```
evankanderson: Why is this always `noopTracer`?  Is that because this is in the testing package?

Maybe put a comment on line 36 about the why of `noopTracer`.

evankanderson (commented):
/approve

One small question, but I'm happy to LGTM when it's time.

test/logging/logging.go
dprotaso: > Why is this always noopTracer? Is that because this is in the testing package?

We had tracing in our tests WAAAY back. I don't think we use it for anything anymore so I was thinking of removing this but there's quite a few places this method is used downstream and don't think it's worth breaking right now.

So I figured I'd make this a noop.

I can create a follow up issue if you prefer

dprotaso: I'll just leave a regular tracer for now.

We should figure out what utility this has in a follow up.

Cali0707: +1 on figuring what the utility of this is as a follow up

PR #3203, opened 2025-07-02
[injection/sharedmain] Fix hanging shutdown due to pprof server

dprotaso:
shutdown must be called explicitly since pprof listen is part of the errgroup

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -259,7 +259,6 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	rest.SetDefaultWarningHandler(&logging.WarningHandler{Logger: logger})
 
 	pprof := newProfilingServer(logger.Named("pprof"))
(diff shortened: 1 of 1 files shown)


injection/sharedmain/main.go
```diff
@@ -340,6 +339,8 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	// returns an error.
 	<-egCtx.Done()
 
+	pprof.Shutdown(context.Background())
```
Cali0707, line 342: should we have some kind of timeout on this context to ensure that we stop eventually here?

dprotaso: We didn't have it before. I think it's fine because I would expect pprof to only have active requests when debugging when someone is debugging.

Cali0707: Ah okay 👍

Cali0707 (approved):
/lgtm
/approve

PR #3204, opened 2025-07-03
[observability/resource] fix service name calculation

dprotaso:
OTel SDK didn't behave as I thought - so we change the order we merge attributes

### DIFF
--- a/observability/resource/default.go
+++ b/observability/resource/default.go
@@ -37,12 +37,9 @@ const otelServiceNameKey = "OTEL_SERVICE_NAME"
 // - PodName using system.PodName
 // - ServiceVersion with changeset.Get
 func Default(serviceName string) *resource.Resource {
-	// We do this since a downstream user might want to
-	// change the service name but not change the attributes
-	// Currently OTel env detectors only override the service name
-	// if both OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES
-	// are set
-	if name := os.Getenv("OTEL_SERVICE_NAME"); name != "" {
+	// If the OTEL_SERVICE_NAME is set then let this override
+	// our own serviceName
+	if name := os.Getenv(otelServiceNameKey); name != "" {
 		serviceName = name
 	}
 


observability/resource/default.go
```diff
@@ -33,10 +37,20 @@ import (
 // - PodName using system.PodName
 // - ServiceVersion with changeset.Get
 func Default(serviceName string) *resource.Resource {
+
```
Cali0707: @dprotaso could you fix the lint error here?

observability/resource/default.go
```diff
@@ -33,10 +37,20 @@ import (
...
+	// We do this since a downstream user might want to
+	// change the service name but not change the attributes
+	// Currently OTel env detectors only override the service name
+	// if both OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES
+	// are set
+	if name := os.Getenv("OTEL_SERVICE_NAME"); name != "" {
+		serviceName = name
+	}
```
Cali0707: Just to make sure I'm understanding this properly, the issue is that if the user sets `OTEL_SERVICE_NAME` currently, that won't be picked up on by the sdk unless `OTEL_RESOURCE_ATTRIBUTES` is also set? So, we need to do it ourselves here?

observability/resource/default.go
dprotaso: done

observability/resource/default.go
dprotaso: Sorry I messed this comment up and I just fixed it now.

Since we had `resource.Default()` last in our Merge a 'detector' would override the service name we supply with a [default name that doesn't really work](https://github.com/open-telemetry/opentelemetry-go/blob/78f873f083be4a11166318403695bc636ef9861b/sdk/resource/builtin.go#L97).


But I _also_ want the ability for the env var to override our supplied default so I need to have this check.

dprotaso:
@Cali0707 PTAL

Cali0707 (approved):
/lgtm
/approve

PR #3205, opened 2025-07-04
wire zap into the client-go logger

dprotaso:
There's a whole whack of logging from the client-go lib that doesn't end up in our zap json output. This configures `klog` the k8s logger with the zap logger to accomplish this.

/assign @Cali0707 @evankanderson @skonto


injection/sharedmain/main.go
```diff
@@ -255,6 +256,8 @@ func MainWithConfig(ctx context.Context, component string, cfg *rest.Config, cto
 	defer logger.Sync()
 	ctx = logging.WithLogger(ctx, logger)
 
+	klog.SetLogger(zapr.NewLogger(logger.Desugar()))
```
Cali0707, line 260: I think this is a typo:
```suggestion
	klog.SetLogger(zap.NewLogger(logger.Desugar()))
```

dprotaso: weird my import wasn't part of the commit - it should be zapr - let me push

Cali0707 (approved):
/lgtm
/approve

PR #3206, opened 2025-07-04
[controller/observability] adjust metrics reporting for the two lane queue

dprotaso:
`k8s.io/client-go` workqueue package will report metrics for named queues. Our two lane queue was reporting three different metrics named `name-slow`, `name-fast`, `name-consumer` 

The metrics from these queues don't really paint the entire picture of how long a resource is queued in the two lane queue.

This change duplicates/forks some k8s code so that we can properly report 'accurate' metrics for our two lane queue.

/hold for review

/assign @Cali0707 @evankanderson @skonto

### DIFF
--- a/controller/queue_metrics.go
+++ b/controller/queue_metrics.go
@@ -28,11 +28,14 @@ import (
 // Copyright The Kubernetes Authors.
 // Forked from: https://github.com/kubernetes/client-go/blob/release-1.33/util/workqueue/metrics.go
 
-// Unfortunately k8s package doesn't expose this variable so for now
-// we'll create our own and potentially upstream some changes in the future.
-var globalMetricsProvider workqueue.MetricsProvider = k8s.NewNoopWorkqueueMetricsProvider()
+var (
+	noopProvider = k8s.NewNoopWorkqueueMetricsProvider()
 
-var setGlobalMetricsProviderOnce sync.Once
+	// Unfortunately k8s package doesn't expose this variable so for now
+	// we'll create our own and potentially upstream some changes in the future.
+	globalMetricsProvider        workqueue.MetricsProvider = noopProvider
+	setGlobalMetricsProviderOnce sync.Once
+)
 
 func SetMetricsProvider(metricsProvider workqueue.MetricsProvider) {
 	setGlobalMetricsProviderOnce.Do(func() {
@@ -109,6 +112,9 @@ func (m *queueMetrics) done(item any) {
 }
 
 func (m *queueMetrics) updateUnfinishedWork() {
+	if m == nil {
+		return
+	}
 	// Note that a summary metric would be better for this, but prometheus
 	// doesn't seem to have non-hacky ways to reset the summary metrics.
 	var total float64
--- a/controller/queue_metrics.go
+++ b/controller/queue_metrics.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2025 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/clock"
+	"knative.dev/pkg/observability/metrics/k8s"
+)
+
+// Copyright The Kubernetes Authors.
+// Forked from: https://github.com/kubernetes/client-go/blob/release-1.33/util/workqueue/metrics.go
+
+// Unfortunately k8s package doesn't expose this variable so for now
+// we'll create our own and potentially upstream some changes in the future.
+var globalMetricsProvider workqueue.MetricsProvider = k8s.NewNoopWorkqueueMetricsProvider()
+
+var setGlobalMetricsProviderOnce sync.Once
+
+func SetMetricsProvider(metricsProvider workqueue.MetricsProvider) {
+	setGlobalMetricsProviderOnce.Do(func() {
+		globalMetricsProvider = metricsProvider
+	})
+}
+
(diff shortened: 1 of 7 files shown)


dprotaso:
Note - there's one more metric I need to add - but I think the PR is ready for review in lieu of that

controller/two_lane_queue.go
```diff
@@ -16,51 +16,89 @@ limitations under the License.
...
+
 // Creates a new twoLaneQueue.
-func newTwoLaneWorkQueue(name string, rl workqueue.TypedRateLimiter[any]) *twoLaneQueue {
+func newTwoLaneWorkQueue(name string, rl workqueue.TypedRateLimiter[any]) *twoLaneRateLimitingQueue {
+	mp := globalMetricsProvider
+
 	tlq := &twoLaneQueue{
-		TypedRateLimitingInterface: workqueue.NewNamedRateLimitingQueue(
-			rl,
-			name+"-fast",
```
dprotaso, line 44: dropping the name will mean no metrics will be reported

controller/two_lane_queue_test.go
```diff
@@ -55,14 +55,6 @@ func TestRateLimit(t *testing.T) {
 	}
 
 	q := newTwoLaneWorkQueue("live-in-the-limited-lane", rl)
-	// Verify the slow lane has the proper RL.
-	q.SlowLane().AddRateLimited("1")
```
dprotaso, line 59: Looking a the controller.go code we never added things to the slow lane using AddRateLimited. The changes in this PR changed the slow queue type to a regular one (no rate limiting - since it's not needed)

I think this makes sense given if something is reconciled and fails it should be requeued on the fast lane

controller/queue_metrics.go
```diff
@@ -0,0 +1,129 @@
...
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if startTime, exists := m.processingStartTimes[item]; exists {
+		m.workDuration.Observe(m.sinceInSeconds(startTime))
+		delete(m.processingStartTimes, item)
+	}
+}
+
+func (m *queueMetrics) updateUnfinishedWork() {
```
Cali0707, line 114: @dprotaso according to the linter this function is unused currently, is this related to the metric you still need to add?

controller/queue_metrics.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (m *queueMetrics) get(item any) {
+	if m == nil {
+		return
+	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.depth.Dec()
```
evankanderson, line 91: Why does `get` decrement the counters?  It seems like a method comment could be helpful, because the names vs behavior make me suspect I'd hold this wrong.

Maybe this is `start`?

evankanderson: Is there any risk that `get`/`start` will get called before `done` (from separate goroutines)?

Another option here would be to enqueue something on the channel that included the start time as well as the object, but I haven't thought about the other implications (that would avoid needing the map updates under this lock).

controller/queue_metrics.go
evankanderson: If we have a counter for firished work, couldn't `adds` - `finished` be used to calculate outstanding work?

controller/two_lane_queue.go
```diff
@@ -16,51 +16,89 @@ limitations under the License.
...
+	q *twoLaneQueue
+	workqueue.TypedRateLimitingInterface[any]
+}
+
+var _ workqueue.TypedInterface[any] = (*twoLaneQueue)(nil)
+
 // Creates a new twoLaneQueue.
-func newTwoLaneWorkQueue(name string, rl workqueue.TypedRateLimiter[any]) *twoLaneQueue {
+func newTwoLaneWorkQueue(name string, rl workqueue.TypedRateLimiter[any]) *twoLaneRateLimitingQueue {
+	mp := globalMetricsProvider
```
evankanderson, line 55: This needs to be called _after_ `SetMetricsProvider` for that to be useful, correct?  Is there a way to ensure that?

controller/two_lane_queue_test.go
evankanderson: Is it worth adding a comment to explain why we don't need rate limiting on the slow queue?  Is that because the consumer queue is limited, and we'll only add to the consumer queue from the slow queue when the fast queue is empty?

evankanderson (commented):
A few questions, mostly because I remember this code being a bit tricky the first time around.

I'm not sure that I'm that bothered by the three named sets of metrics, but we would need to document what they meant, so simplifying things to two sets of metrics seems fine, too.

Will we end up with separate queueing and processing latency for "fast lane" and "slow lane" watches?

controller/queue_metrics.go
dprotaso: > Why does get decrement the counters?

This is copied from:  https://github.com/kubernetes/client-go/blob/660181394a7e0dd678325f4b3f1041590b23d056/util/workqueue/metrics.go#L104

This is decremented because `depth` is just counting the items in the queue and it doesn't include the items being processed.

dprotaso: > Is there any risk that get/start will get called before done (from separate goroutines)?

No - I don't believe so since processNextWorkItem will call `Get()` & `Done()`

It's odd because the original k8s code didn't have a lock - and I was getting data races when testing running our tests.

Controllers have multiple threads so you have multiple go routines calling Get - I would expect locks are necessary.

dprotaso: > Another option here would be to enqueue something on the channel that included the start time as well as the object, but I haven't thought about the other implications (that would avoid needing the map updates under this lock).

I think for now I'd prefer to keep the semantics the same and adjust performance in a follow up PR.

controller/two_lane_queue.go
dprotaso: No - this is the same behaviour as calling `workqueue.SetProvider`

controller/two_lane_queue_test.go
dprotaso: > Is it worth adding a comment to explain why we don't need rate limiting on the slow queue? 

Honestly, we just never used it. I can add a comment.

> Is that because the consumer queue is limited, and we'll only add to the consumer queue from the slow queue when the fast queue is empty?

Yeah I believe so.

controller/queue_metrics.go
dprotaso: I think I'd prefer to keep these metrics in line with the workqueue package so I wouldn't want to deviate.

evankanderson (commented):
/lgtm
/approve

PR #3207, opened 2025-07-05
[observability/trace] allow for stdout exporter for tracing - to allow for debugging

dprotaso:
Setting up a collector or jaeger for manual testing is a bit much  - looking at the stdout is sometimes easier.

Thought this would be useful to have


observability/tracing/provider.go
```diff
@@ -162,6 +169,10 @@ func sampleFor(cfg Config) (sdktrace.Sampler, error) {
 		return nil, nil
 	}
 
+	if cfg.Protocol == ProtocolStdout {
```
evankanderson, line 172: I think this needs a comment, as ignoring the other sampling parameters is surprising.

PR #3209, opened 2025-07-06
[webhook] enrich webhook traces

dprotaso:
A big suprised is that otelhttp doesn't take labeler attributes and add them to a span.

- **don't trace kubelet probes**
- **instrument span status**
- **add attributes to span**
- **format the span name to something useful METHOD /path**
- **add a clarifying comment**


webhook/admission.go
```diff
@@ -94,17 +96,24 @@ func admissionHandler(wh *Webhook, c AdmissionController, synced <-chan struct{}
...
+		defer func() {
+			span.SetAttributes(labeler.Get()...)
+		}()
+
 		var review admissionv1.AdmissionReview
 		bodyBuffer := bytes.Buffer{}
 		if err := json.NewDecoder(io.TeeReader(r.Body, &bodyBuffer)).Decode(&review); err != nil {
-			http.Error(w, fmt.Sprint("could not decode body:", err), http.StatusBadRequest)
+			err := fmt.Sprint("could not decode body:", err)
+			span.SetStatus(codes.Error, err)
```
dprotaso: One thing to note is that `otelhttp` doesn't mark 4xx status as errors on the span.

I'm unsure if that's a recommendation on what's an error or not. But I'm inclined to set the status to these as errors

evankanderson:
Did you use `otelhttp.NewHandler` to wrap the existing handlers?  (Sorry, looking at this on phone at the moment)

dprotaso:
> Did you use `otelhttp.NewHandler` to wrap the existing handlers? (Sorry, looking at this on phone at the moment)

Yup - that's already in place in HEAD - https://github.com/knative/pkg/blob/ff153ade2580d5201f1dafc0f9e704ca753cd729/webhook/webhook.go#L274

webhook/admission.go
Cali0707: my preference would also be to set these as errors

webhook/admission.go
```diff
@@ -94,17 +96,24 @@ func admissionHandler(wh *Webhook, c AdmissionController, synced <-chan struct{}
 		logger := wh.Logger
 		logger.Infof("Webhook ServeHTTP request=%#v", r)
 
+		span := trace.SpanFromContext(r.Context())
+		// otelhttp middleware creates the labeler
+		labeler, _ := otelhttp.LabelerFromContext(r.Context())
+
+		defer func() {
+			span.SetAttributes(labeler.Get()...)
+		}()
```
Cali0707: Could we add a comment clarifying that this is because otel does not set the attributes from the labeler on the span?

webhook/conversion.go
```diff
@@ -48,20 +51,30 @@ func conversionHandler(wh *Webhook, c ConversionController) http.HandlerFunc {
 		logger := wh.Logger
 		logger.Infof("Webhook ServeHTTP request=%#v", r)
 
+		span := trace.SpanFromContext(r.Context())
+		// otelhttp middleware creates the labeler
+		labeler, _ := otelhttp.LabelerFromContext(r.Context())
+
+		defer func() {
+			span.SetAttributes(labeler.Get()...)
+		}()
```
Cali0707: Could we add a comment clarifying that this is because otel does not set the attributes from the labeler on the span?

webhook/admission.go
```diff
@@ -94,17 +96,24 @@ func admissionHandler(wh *Webhook, c AdmissionController, synced <-chan struct{}
...
+
 		var review admissionv1.AdmissionReview
 		bodyBuffer := bytes.Buffer{}
 		if err := json.NewDecoder(io.TeeReader(r.Body, &bodyBuffer)).Decode(&review); err != nil {
-			http.Error(w, fmt.Sprint("could not decode body:", err), http.StatusBadRequest)
+			err := fmt.Sprint("could not decode body:", err)
```
Cali0707: I find using the `err` variable to contain a `string` and not an `error` to be confusing, would it be possible to rename the variable or to use `fmt.Errorf()` (along with `fmt.Error()` where strings are needed)?

webhook/conversion.go
```diff
@@ -48,20 +51,30 @@ func conversionHandler(wh *Webhook, c ConversionController) http.HandlerFunc {
...
+
 		var review apixv1.ConversionReview
 		if err := json.NewDecoder(r.Body).Decode(&review); err != nil {
-			http.Error(w, fmt.Sprint("could not decode body:", err), http.StatusBadRequest)
+			err := fmt.Sprint("could not decode body:", err)
```
Cali0707: Same comment about the err variable

webhook/conversion.go
```diff
@@ -48,20 +51,30 @@ func conversionHandler(wh *Webhook, c ConversionController) http.HandlerFunc {
...
+			span.SetStatus(codes.Error, err)
+			http.Error(w, err, http.StatusBadRequest)
 			return
 		}
 
 		gv, err := parseAPIVersion(review.Request.DesiredAPIVersion)
 		if err != nil {
-			http.Error(w, fmt.Sprint("could parse desired api version:", err), http.StatusBadRequest)
+			err := fmt.Sprint("could parse desired api version:", err)
```
Cali0707: Same comment about the err variable

Cali0707 (changes requested):

webhook/conversion.go
dprotaso: done

Cali0707 (approved):
/lgtm
/approve

PR #3211, opened 2025-07-07
fix: revert string parsing in configmap to not trim spaces

Cali0707:
In #3200 we switched the configmap parser to trim spaces in strings. Unfortunately, this has broken the parsing of configmaps with multi-line embedded yamls in downstream repos like eventing (see failures in https://github.com/knative/eventing/pull/8618). This PR reverts the change

# Changes

- Do not trim spaces in configmap string parsing

### DIFF
--- a/configmap/parser/parse.go
+++ b/configmap/parser/parse.go
@@ -19,7 +19,6 @@ package parser
 import (
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
 )
 
@@ -72,7 +71,7 @@ func parse[T parseable](s string) (T, error) {
 
 	switch any(zero).(type) {
 	case string:
-		val = strings.TrimSpace(s)
+		val = s
 	case int16:
 		val, err = strconv.ParseInt(s, 10, 16)
 		val = int16(val.(int64))


dprotaso:
Can you clarify the problem more?

The trim'ing just trims leading and trailing whitespace

Cali0707:
> Can you clarify the problem more?
> 
> The trim'ing just trims leading and trailing whitespace

Yeah, the issue we are seeing in eventing is that there are configmaps with data that looks like:
```go
"channel-template-spec": "\tapiVersion: Foo/v1\n\tkind: Bar\n"
```
And since we trim the whitespace the "value" becomes:
```go
"apiVersion: Foo/v1\tkind: Bar\n"
```
Resulting in an error that `yaml: line 2: mapping values are not allowed in this context`, when we try and convert the result from yaml to JSON (see https://github.com/knative/eventing/blob/main/pkg/reconciler/broker/config.go#L65)

Essentially, the problem is that if we try to use `cm.AsString` with a string that contains multiple lines of yaml, the subsequent parsing/manipulation we do of that now breaks

dprotaso:
This honestly might be more of an issue with the fixture string in the test

eg. https://go.dev/play/p/wH7mgI5UXBO

dprotaso:
I think I'd prefer eventing just make their own parser func to handle this YAML scenario.

The alternative would be to have all this extra validation everywhere. We'd have to trim the strings for whitespace then check if it's empty.

Cali0707:
> I think I'd prefer eventing just make their own parser func to handle this YAML scenario.
> 
> The alternative would be to have all this extra validation everywhere. We'd have to trim the strings for whitespace then check if it's empty.

That's fair, I'll close this then

/close

evankanderson (commented):
/lgtm
/approve

PR #3212, opened 2025-07-07
add json struct tags to the observability configs

dprotaso:
we'll be able to serialize the config to the queue proxy

/assign @evankanderson @Cali0707 @skonto


dprotaso:
> You could also use the default JSON serialization, but this is fine.

The bigger thing was having `omitempty`

PR #3215, opened 2025-07-09
drop dependency on metrics package

dprotaso:
Profiling test package doesn't need a reference to the metrics package anymore. This indirectly keeps opencensus dep around


dprotaso:
/assign @skonto @evankanderson @Cali0707 @dsimansk @Fedosin 

We're still pulling in opencensus into serving through this usage

evankanderson (approved):

PR #3226, opened 2025-08-04
Fix otel exported names

### LINKED ISSUE #3224: knative 1.19 links against interim / broken release of otel-go SDK
The otel-go SDK is in a bit of an interim state (sorry!) and is producing different metric names than it did previously. This breaks continuity with previous metrics, and the new metrics are UTF-8 which is not supported in some situations. We haven't yet finished adding the code to make the metric name translation mode configurable, so this leaves users in an unfixable state without downgrading knative.  Downgrading the otel-go sdk to 0.58 should fix this issue, and I'd recommend pinning to that version until we finish our work (https://github.com/open-telemetry/opentelemetry-go/issues/7083)

( @dashpole @arthursens )

dsimansk:
# Changes

- Fix otel exported names

Fix #3224 

Adding hold to prevent accidental merge before reviews.
/hold


dsimansk:
@dprotaso PTAL whenever you're back, thanks! 

Based on comment from: https://github.com/knative/pkg/issues/3224

dprotaso:
Based on advice from dashpole I added the following options to ensure the names will be consistent with the future default of `otelprom`

```
otelprom.WithoutCounterSuffixes(),
otelprom.WithoutUnits()
```

Cali0707 (approved):
/lgtm

dprotaso:
whoops
/hold cancel

PR #3227, opened 2025-08-04
drop opencensus packages

dprotaso:


dprotaso:
This is good to merge and I'd rather not block on the downstream test being red.

Summary:
- operator - needs a deps bump and downstream will go green. Blocked on some eventing upgrade test
- kafka-broker - tests are flakey - i tried to push some changes to unblock the dep updated but I don't know enough to fix it

/assign @Cali0707 @evankanderson

evankanderson:
/lgtm

We're at a good point in the release cycle to have time to fix breakages.

I hope this doesn't have impacts on outside-Knative consumers of `pkg`.

dprotaso:
> I hope this doesn't have impacts on outside-Knative consumers of pkg.

I reached out to Tekton and they were happy about our OTel support. They actually added OTel tracing downstream vs trying to upstream it into knative.dev/pkg.

I gave some notice to sigstore but no one replied to my messages so unsure if they cared.

Either way the alternative is there so the migration can happen

PR #3235, opened 2025-08-15
[release-1.19] Fix otel exported names (#3226/#3235)

dprotaso:
Cherry-pick of https://github.com/knative/pkg/pull/3226 and https://github.com/knative/pkg/pull/3235 into release-1.19 branch

Part of https://github.com/knative/pkg/issues/3224

PR #3236, opened 2025-08-17
Fix prometheus exporter names

dprotaso:
There was a typo in a comment https://github.com/knative/pkg/issues/3224

This change will update the OTel prom exporter metric names to match
what will become the default in the future. If N is the current
SDK version the N+2 is when the new default will apply.

In the N+1 release - a new TranslationStrategy option will be introduced and we
will want to set the strategy to 'UnderscoreEscapingWithSuffixes'.

Then in N+2 we can drop that option as it will become the default
(or leave it explicitly).

### DIFF
--- a/observability/metrics/prometheus_enabled.go
+++ b/observability/metrics/prometheus_enabled.go
@@ -35,10 +35,7 @@ func init() {
 }
 
 func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownFunc, error) {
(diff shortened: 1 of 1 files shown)


dprotaso:
/assign @Cali0707 

cc @evankanderson

evankanderson (commented):
/lgtm
/approve

dprotaso:
I updated the cherry-pick here: https://github.com/knative/pkg/pull/3235

PR #3245, opened 2025-09-03
Bump OTel dependencies and semconv

dprotaso:
- **bump OTel deps**
- **bump OTel semconv to v1.37.0**
- **Use OTel TranslationStrategy options for prometheus**

The default strategy remains the same `UnderscoreEscapingWithSuffixes` - it is just explicit now

Fixes: https://github.com/knative/pkg/issues/3224

Note: I don't plan on cherry-picking this back to our 1.19 release branch


dprotaso:
cc @ywwg @dashpole - for a sanity check

observability/metrics/prometheus_enabled.go
```diff
@@ -23,19 +23,16 @@ import (
...
 
-func init() {
-	//nolint:staticcheck
-	model.NameValidationScheme = model.LegacyValidation
-}
-
 func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownFunc, error) {
-	r, err := otelprom.New()
+	r, err := otelprom.New(
+		otelprom.WithTranslationStrategy(otlptranslator.UnderscoreEscapingWithSuffixes),
```
ywwg, line 34: This looks great.  Would it be possible to make it configurable in the future? We do want to enable users across the ecosystem to use otel-native names if they so choose.

ywwg (approved):

observability/metrics/prometheus_enabled.go
dprotaso: > We do want to enable users across the ecosystem to use otel-native names if they so choose.

I think for now I'll leave this as the default for prom pull. If we get requests we can add other the strategies.  Users can still push OTel metric names

PR #3246, opened 2025-09-03
[observability] Support setting metric temporality when exporting

dprotaso:
This option should unblock this user: https://github.com/knative/eventing/issues/8686


dprotaso:
Closing this out - see https://github.com/knative/pkg/issues/3256#issuecomment-3285883925

PR #3247, opened 2025-09-06
Add threadiness in log line

jeongukjae:
# Changes

- :broom: Surface the threadiness value of controller

/kind enhancement

**why**

knative/pkg already supports controlling concurrency of the workers in controller (look at below codes), but it doesn't give back any feedback except it actually runs concurrently. This small change will help users to quickly catch how concurrently the controller runs.

https://github.com/knative/pkg/blob/528bde37b646f02c95ed92ee682c0e48c98504b5/injection/sharedmain/main.go#L205-L211


dprotaso:
thanks for the change!

PR #3255, opened 2025-09-09
Make webhook owner references not a controller

dprotaso:
Part of https://github.com/knative/serving/issues/15483

We setup namespace as the owner of the webhooks in order to clean up
instances where a user deletes the namespace.

This has issues with ArgoCD which we've included a workaround.
But an upcoming ArgoCD PR will handle owner references properly if
it is not a 'controller=true' reference.

This PR changes the owner reference to not be a controlling one.
We still block owner deletion because that was the original intent
of adding the owner reference.


dprotaso:
bump @aliok

dprotaso:
bump @aliok

dprotaso:
bump @dsimansk @nader-ziada

PR #3268, opened 2025-10-06
Log context cancellation as debug instead of error

thiagomedina:
# Changes

- Change context cancellation logging from ERROR to DEBUG

/kind enhancement

Fixes https://github.com/knative/serving/issues/13746

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/network/error_handler.go
+++ b/network/error_handler.go
@@ -17,6 +17,8 @@ limitations under the License.
 package network
 
 import (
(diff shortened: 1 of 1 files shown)


dprotaso:
@thiagomedina we have automation that bumps dependencies daily so this change will get into serving tomorrow

PR #3269, opened 2025-10-07
create a knative semconv package to simplify otel dep bumps

dprotaso:
Instead of manually bumping the semconv version across many downstream repos I created this package that will do the versioning for us.

If a key or function is missing we should feel free to add more here


Cali0707 (approved):
/lgtm
/approve

Thanks @dprotaso - this should really improve how we keep semconv in sync across repos

PR #3275, opened 2025-10-18
Add support for Server-Side Apply finalizer management

### LINKED ISSUE #2128: Investigate server-side apply to allow cooperating controllers
Refs: 
https://kubernetes.io/docs/reference/using-api/server-side-apply/
https://github.com/knative-sandbox/net-contour/issues/347#issuecomment-838613107

This would allow customization of the resources we create

/assign @dprotaso

enarha:
This improves performance and load on the k8s API by decreasing the number of re-tries, especially on objects managed by multiple controllers. The feature is opt-in and the current use of merge patch is still the default behavior.
We introduce three controller options:
- UseServerSideApplyForFinalizers used to enable that functionality.
- FinalizerFieldManager to specify finalizer field manager, required when SSA is enabled.
- ForceApplyFinalizers to set the force option on the apply patch (default disabled).

# Changes

- :gift: Add support for Server-Side Apply finalizer management
This improves performance and load on the k8s API by decreasing the number of re-tries, especially on objects managed by multiple controllers. The feature is opt-in and the current use of merge patch is still the default behavior.
We introduce three controller options:
- UseServerSideApplyForFinalizers used to enable that functionality.
- FinalizerFieldManager to specify finalizer field manager, required when SSA is enabled.
- ForceApplyFinalizers to set the force option on the apply patch (default disabled).

/kind api-change

Fixes #2128

**Release Note**

```release-note
Added support Server-Side Apply finalizer management. The feature is opt-in and the current use of merge patch is still the default behavior. Here is an example of controller options which enable and configure that functionality:

controller.Options{
    FinalizerName: "my-finalizer",
    UseServerSideApplyForFinalizers: true,
    FinalizerFieldManager: "my-finalizer-manager"
}

`UseServerSideApplyForFinalizers` enables the use Server-Side Apply for the finalizer management. If enabled, `FinalizerFieldManager` must also be provided. Use unique finalizer name and finalizer field manager to avoid conflicts. If you need to force the finalizer update, you can also add the `ForceApplyFinalizers` option and set it to `true`. That's usually not a good idea and different controllers should use different finalizer name and field manager.

IMPORTANT: 
If your controller is managing other fields using Server-Side Apply as part of its reconciliation loop (e.g. annotations), make sure you are using different filed manager for each (e.g. "my-controller/finalizers", "my-controller/annotations") to avoid accidental removal.
When switching from Merge Patch strategy to Server-Side Apply, objects caught in flight (existing on the cluster during transition) will need extra attention. Finalizer set using Merge Patch can't be removed using Server-Side Apply because filed manager (used with Server-Side Apply) does not own that field. In that case the finalizer must be removed manually.
```

**Docs**

```docs

```

### DIFF
--- a/client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
+++ b/client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
@@ -32,6 +32,7 @@ import (
 	types "k8s.io/apimachinery/pkg/types"
 	sets "k8s.io/apimachinery/pkg/util/sets"
 	kubernetes "k8s.io/client-go/kubernetes"
+	scheme "k8s.io/client-go/kubernetes/scheme"
 	admissionregistrationv1 "k8s.io/client-go/listers/admissionregistration/v1"
 	record "k8s.io/client-go/tools/record"
 	controller "knative.dev/pkg/controller"
@@ -97,6 +98,15 @@ type reconcilerImpl struct {
 
 	// finalizerName is the name of the finalizer to reconcile.
 	finalizerName string
+
+	// useServerSideApplyForFinalizers configures whether to use server-side apply for finalizer management
+	useServerSideApplyForFinalizers bool
+
+	// finalizerFieldManager is the field manager name for server-side apply of finalizers
+	finalizerFieldManager string
+
+	// forceApplyFinalizers configures whether to force server-side apply for finalizers
+	forceApplyFinalizers bool
 }
 
 // Check that our Reconciler implements controller.Reconciler.
@@ -150,6 +160,14 @@ func NewReconciler(ctx context.Context, logger *zap.SugaredLogger, client kubern
 		}
 		if opts.DemoteFunc != nil {
 			rec.DemoteFunc = opts.DemoteFunc
+		}
+		if opts.UseServerSideApplyForFinalizers {
+			if opts.FinalizerFieldManager == "" {
+				logger.Fatal("FinalizerFieldManager must be provided when UseServerSideApplyForFinalizers is enabled")
+			}
+			rec.useServerSideApplyForFinalizers = true
+			rec.finalizerFieldManager = opts.FinalizerFieldManager
+			rec.forceApplyFinalizers = opts.ForceApplyFinalizers
 		}
 	}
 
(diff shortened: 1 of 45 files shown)


dprotaso:
/hold

pkg release-1.20 branch was cut today. Let's revisit this after the 1.20 release next week

enarha:
Can this PR get some attention please. I'll be happy to elaborate if you have any questions. Thank you.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+
+		patchOpts := metav1.PatchOptions{
+			FieldManager: r.finalizerFieldManager,
+			Force:        &r.forceApplyFinalizers,
+		}
+
+		updated, err := patcher.Patch(ctx, resource.Name, types.ApplyPatchType, patch, patchOpts)
+		if err != nil {
+			r.Recorder.Eventf(resource, corev1.EventTypeWarning, "FinalizerUpdateFailed",
+				"Failed to remove finalizer for %q via server-side apply: %v", resource.Name, err)
```
nader-ziada: Should use %w for error wrapping

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+	} else {
+		if !existingFinalizers.Has(r.finalizerName) {
+			// Nothing to do.
+			return resource, nil
+		}
+		// For removal, we apply an empty configuration for our finalizer field manager.
+		// This effectively removes our finalizer while preserving others.
+		gvks, _, err := scheme.Scheme.ObjectKinds(resource)
+		if err != nil || len(gvks) == 0 {
+			return resource, fmt.Errorf("failed to determine GVK for resource: %v", err)
```
nader-ziada: Should use %w for error wrapping

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+	// Check if we need to do anything
+	existingFinalizers := sets.New[string](resource.Finalizers...)
+
+	if desiredFinalizers.Has(r.finalizerName) {
+		if existingFinalizers.Has(r.finalizerName) {
+			// Nothing to do.
+			return resource, nil
+		}
+		// Apply configuration with only our finalizer to add it.
+		gvks, _, err := scheme.Scheme.ObjectKinds(resource)
```
nader-ziada: This GVK lookup logic is duplicated below in the else condition

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+		if err != nil || len(gvks) == 0 {
+			return resource, fmt.Errorf("failed to determine GVK for resource: %v", err)
```
nader-ziada: Should use %w for error wrapping

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+			"apiVersion": gvk.GroupVersion().String(),
+			"kind":       gvk.Kind,
+			"metadata": map[string]interface{}{
+				"name":       resource.Name,
+				"uid":        resource.UID,
+				"finalizers": []string{r.finalizerName},
+			},
+		}
+
+		patch, err := json.Marshal(applyConfig)
```
nader-ziada: This patch operation logic is duplicated below as well

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
```diff
@@ -272,6 +290,109 @@ func (r *reconcilerImpl) Reconcile(ctx context.Context, key string) error {
...
+		if err != nil {
+			return resource, err
+		}
+
+		patcher := r.Client.AdmissionregistrationV1().MutatingWebhookConfigurations()
+
+		patchOpts := metav1.PatchOptions{
+			FieldManager: r.finalizerFieldManager,
+			Force:        &r.forceApplyFinalizers,
```
nader-ziada: Is force necessary here?

dprotaso:
Testing this downstream with - https://github.com/knative/serving/pull/16217

dprotaso:
Downstream test ran successfully. I'm good to merge once Nader's feedback is addressed.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Indeed it has. It could be called before the if clause, but then it should be called on each reconciliation cycle. Instead we call it just twice in the lifetime of the object, once to set it and once to remove it, while the reconciliation loop is called 10-20 or even more times, at least on the controllers I work on. That's why I think less calls is worth the duplication, but I'll be happy to change it you think otherwise.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Indeed. It could be moved after the else clause with the only penalty of having less accurate log message "failed to add/remove finalizer" and "Added/removed finalizer" compared to "failed to update finalizer" and "updated finalizer". Personally I prefer the former, but I'll update so it's also inline with the merge patch method.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: By default it's not set as the value is set to false. We expose the ForceApplyFinalizers controller option though, so the controller implementation can set that to true. Force apply is often discussed in the context of server-side apply, so I decided to provide that flexibility to the controller implementation.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Nevermind, I moved it after the else block.

enarha: Updated.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Agree, that makes sense as we return error.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Updated. Much cleaner now.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: Here we do not return the error, just print it, so the simpler %v makes more sense and it is also consistent with the original code (merge patch). I updated the case where we return the error to use %w which allows unwrapping. If you still think I should update this one, I'll do.

client/injection/kube/reconciler/admissionregistration/v1/mutatingwebhookconfiguration/reconciler.go
enarha: done

nader-ziada:
thanks for the updates @enarha 

@dprotaso good from my side, you will need to approve it if you are good

PR #2092, opened 2021-04-09
resolve cluster domain using env CLUSTER_DOMAIN as a fallback

n3wscott:
# Changes

- :gift: Resolve cluster domain with `CLUSTER_DOMAIN` as a fallback.

/kind enhancement

This will help with tests that are attempting to resolve cluster local urls from outside the cluster, that do not have access to `/etc/resolv.conf` directly. 

**Release Note**

```release-note
network.GetServiceHostname and network.GetClusterDomainName now respect the env var CLUSTER_DOMAIN.
```

PR #3283, opened 2025-10-23
Edit CLUSTER_DOMAIN environment variable support

prashanthjos:
## Description

This PR changes support for the `CLUSTER_DOMAIN` environment variable to allow explicit configuration of the cluster domain, taking precedence over `/etc/resolv.conf` parsing.

## Motivation

When using custom service mesh configurations (e.g., Istio with custom domains), the internal DNS domain may differ from the Kubernetes cluster domain. Currently, Knative derives the cluster domain by parsing `/etc/resolv.conf`, which doesn't work in these scenarios.

For example, a service mesh might use `mesh.net` internally while Kubernetes uses `cluster.local`. This PR enables users to explicitly set the cluster domain via environment variable.

## Changes
network/domain.go : Modified getClusterDomainName() to check `CLUSTER_DOMAIN` environment variable **first**, then fall back to /etc/resolv.conf parsing
- Maintains full backward compatibility - if env var is not set, behavior is identical to current implementation

**Release Note**

Edit CLUSTER_DOMAIN environment variable support to allow explicit configuration of cluster domain for custom service mesh configurations.

### DIFF
--- a/network/domain_test.go
+++ b/network/domain_test.go
@@ -50,11 +50,6 @@ options ndots:5
 		env:        "abc.com",
 		want:       "abc.com",
 	}, {
-		name:       "env variable with custom domain",
-		resolvConf: ``,
-		env:        "custom.local",
-		want:       "custom.local",
-	}, {
 		name: "all good with trailing dot",
 		resolvConf: `
 nameserver 1.1.1.1


dprotaso:
We have a release next week and `knative.dev/pkg` release-1.120 has already been cut. Will revisit after.
/hold

dprotaso:
> When using custom service mesh configurations (e.g., Istio with custom domains), the internal DNS domain may differ from the Kubernetes cluster domain. 

Do you have an example on this can be setup with Istio?

prashanthjos:
When using Istio with a custom mesh domain, we configured a ServiceEntry to route all internal service-to-service traffic through the Istio proxy instead of using Kubernetes default DNS resolution.

```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: mesh-internal-services
  namespace: istio-system
spec:
  # All traffic to *.mesh.net is routed through the Istio proxy
  hosts:
    - "*.mesh.net"
  
  # Traffic is routed to the loopback address where Istio proxy intercepts it
  # via iptables rules configured by istio-init
  addresses:
    - "127.0.0.6"  # Example loopback IP for mesh traffic
  
  location: MESH_INTERNAL
  resolution: STATIC
  
  endpoints:
    # The Istio proxy intercepts and routes based on mTLS identity
    - address: "127.0.0.6"
```

With Default Kubernetes DNS all the services are accessed via <service>.<namespace>.svc.cluster.local, but however with custom mesh DNS through `ServiceEntry` services are accessed via <service>.<namespace>.svc.mesh.net.

Istio Interception: When one service calls another service in our mesh

- Istio's istio-init container modifies iptables rules in each pod (this is during initialization of Pod)
- Traffic to *.mesh.net is intercepted at 127.0.0.6 (loopback)
- The Envoy proxy handles mTLS, routing, policies, and telemetry
- Traffic is then forwarded to the actual service

Without the CLUSTER_DOMAIN environment variable, Knative components would:

- Parse resolv.conf and get cluster.local as the domain
- Generate URLs like: http://my-service.default.svc.cluster.local
- Traffic would bypass Istio because Kubernetes DNS resolution goes directly to pod IPs

We have required VS and DR (generated by our custom implementations of the Knative network spec) in place so that Knative components work as expected.

We also created an [issue](https://github.com/knative-extensions/net-istio/issues/1465#issuecomment-3349647129) to see how can use net-istio so that we don't write our own custom implementation but that is a long shot for us.

prashanthjos:
> We have a release next week and `knative.dev/pkg` release-1.120 has already been cut. Will revisit after. /hold

@dprotaso can we expect this to be merged next week?

prashanthjos:
@dprotaso please do let me know if you have any questions. I would be happy to try any alternatives too before we can merge this PR.

dprotaso:
> Maintains full backward compatibility - if env var is not set, behavior is identical to current implementation

Technically this isn't accurate - setting the env var now will take precedence over `/etc/resolve.conf`

Though this env var is being used in tests (https://github.com/knative/pkg/pull/2092) so it's probably ok to tweak this.

network/domain_test.go
```diff
@@ -36,19 +36,24 @@ options ndots:5
...
-		name:       "all good from env",
+		name:       "env variable used when resolv.conf is empty",
 		resolvConf: ``,
 		env:        "abc.com",
 		want:       "abc.com",
+	}, {
+		name:       "env variable with custom domain",
+		resolvConf: ``,
+		env:        "custom.local",
+		want:       "custom.local",
```
dprotaso: This test seems unnecessary - I say we drop it

prashanthjos: No worries, I removed the test!

PR #3286, opened 2025-11-03
bump k8s to v1.34 libs

dprotaso:
As per release schedule - we bump to v1.34 libs since it supports K8s 1.33 and 1.34

Our K8s target for 1.21 is K8s v1.33

https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md


dprotaso:
/assign @nader-ziada @linkvt @Fedosin 

FYI @matzew @creydr @Cali0707 @twoGiants - looks like the bump introduces some breaking change to eventing
FYI @dsimansk @lkingland for kn-plugin-admin

dprotaso:
Actually I think those jobs are optional in this repo

PR #3287, opened 2025-11-03
run golangci-lint run to run gofumpt on various files

dprotaso:
Unclear why go linter is complaining about this now (https://github.com/knative/pkg/pull/3286)

I ran `golangci-lint run --fix`


nader-ziada:
/lgtm

I'm surprised it didn't complain about these before

PR #3289, opened 2025-11-13
Reduce noisy "object has been modified" logs/events

linkvt:
See https://github.com/knative/serving/issues/15746

# Changes

- :broom: Reduce noisy logs/events containing "object has been modified" messages
  - Events are not recorded anymore and ERROR log messages were downgraded to DEBUG level.

/kind enhancement

# Release Note

```release-note
Reduce noisy logs/events containing "object has been modified" which are expected to happen from time to time during normal operation.
- Events were removed and ERROR log messages were downgraded to DEBUG level.
```

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -533,10 +533,14 @@ func (c *Impl) processNextWorkItem() bool {
 }
 
 func (c *Impl) handleErr(logger *zap.SugaredLogger, err error, key types.NamespacedName, startTime time.Time) {
-	if IsSkipKey(err) {
+	// Check if we should skip this key or if the queue is shutting down.
+	// We check shutdown here since controller Run might have exited by now
+	// (since while this item was being processed, queue.Len==0).
+	if IsSkipKey(err) || c.workQueue.ShuttingDown() {
 		c.workQueue.Forget(key)
 		return
 	}
+
 	if ok, delay := IsRequeueKey(err); ok {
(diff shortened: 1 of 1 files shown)


linkvt:
Failed CI should be resolved as soon as https://github.com/knative/client/pull/2101 is merged and propagated to the `kn-plugin-admin` repo.

controller/controller.go
```diff
@@ -542,6 +543,13 @@ func (c *Impl) handleErr(logger *zap.SugaredLogger, err error, key types.Namespa
 		return
 	}
 
+	// Conflict errors are expected, requeue to retry
+	if apierrors.IsConflict(err) {
+		logger.Debugw("Reconcile conflict", zap.Duration("duration", time.Since(startTime)))
+		c.workQueue.AddRateLimited(key)
```
dprotaso, line 553: Do we want to guard `!c.workQueue.ShuttingDown()` ? here

linkvt: Thanks that makes sense. I moved the shutdown check to the top, it shouldn't really affect the logic as the `AddRateLimited` functions (which call `AddAfter`) effectively also check that again, so it's more about avoiding useless error logs during shutdown.

https://github.com/knative/pkg/blob/745990698d3c3884a3e08c2fc5f813f8acf458ca/vendor/k8s.io/client-go/util/workqueue/delaying_queue.go#L249-L253

PR #3297, opened 2025-11-26
[wip] Set Connection: close header when draining HTTP/1

dprotaso:
Follow up to https://github.com/knative/pkg/pull/3295

This will trigger the http server to close the connection https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/http/server.go;l=1389-1391

I'm going to hold this PR and see if we need it


dprotaso:
Closing this out - I haven't seen webhook EOFs since I removed the disabling of keep alives

PR #3298, opened 2025-11-30
[wip] make use of new go1.24 so we don't need x/net/h2c package

dprotaso:
- **drop use of old h2c package**
- **run ./hack/update-dep.sh**

### DIFF
--- a/network/server.go
+++ b/network/server.go
@@ -24,6 +24,7 @@ import (
 func NewServer(addr string, h http.Handler) *http.Server {
 	var protocols http.Protocols
 	protocols.SetHTTP1(true)
+	protocols.SetHTTP2(true)
 	protocols.SetUnencryptedHTTP2(true)
 
 	return &http.Server{


network/server.go
```diff
@@ -0,0 +1,38 @@
...
+	protocols.SetHTTP2(true)
+	protocols.SetUnencryptedHTTP2(true)
+
+	return &http.Server{
+		Addr:      addr,
+		Handler:   h,
+		Protocols: &protocols,
+
+		// https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6
+		ReadHeaderTimeout: time.Minute,
```
skonto, line 36: Do we care about `MaxHeaderBytes` to make it configurable?

dprotaso: Unsure - circling back on this PR I realize we can't just remove the h2c upgrade flow because that can break existing workloads.

In theory we need 

1. Add support for h2 prior knowledge
2. deprecated h2c upgrade flow
3. then eventually drop h2c upgrade flow in some future release

dprotaso:
Going to close this out for now

PR #3300, opened 2025-12-09
Add TLSMaxVersion, TLSCipherSuites, and TLSCurvePreferences to webhook.Options for minimal tls customization

### LINKED ISSUE #3299: Expose TLS configuration for webhook servers to support platform-managed TLS policies
**Summary**
The knative.dev/pkg/webhook package currently does not expose the tlsConfig field ([webhook.go#L141](https://github.com/knative/pkg/blob/main/webhook/webhook.go#L141)), preventing downstream platforms from dynamically configuring TLS settings beyond the basic WEBHOOK_TLS_MIN_VERSION environment variable. This limitation blocks integration with centralized TLS policy management systems required for enterprise Kubernetes distributions.

**Motivation**
Modern Kubernetes platforms increasingly require centralized TLS policy management to:

- Prepare for Post-Quantum Cryptography (PQC) transitions requiring TLS 1.3+ and specific cipher suites
- Comply with organizational security policies that mandate specific cipher configurations
- Support custom TLS profiles beyond simple version selection (e.g., FIPS mode, industry-specific compliance)
- Enable dynamic security policy updates without rebuilding/redeploying components

**Current Limitations**
https://github.com/knative/pkg/blob/main/webhook/webhook.go#L141
```
tlsConfig: &tls.Config{
    MinVersion: tlsMinVersion,  // Only configurable via env var
    // No cipher suite configuration
    // No way to inject custom tls.Config
}
```

The existing implementation is missing:
- Cipher suite configuration
- Support for custom TLS profiles (Old/Intermediate/Modern/Custom as defined in [Mozilla's SSL Configuration Generator](https://ssl-config.mozilla.org/))`

**Proposed Solution:**
Add TLSConfig functional option
```
// webhook.go
type Options struct {
    // ... existing fields ...
    
    // TLSConfig allows customization of the webhook server's TLS configuration.
    // If nil, a default configuration with MinVersion from WEBHOOK_TLS_MIN_VERSION is used.
    // When provided, this completely overrides the default TLS configuration.
    TLSConfig *tls.Config
}
```

jkhelil:
This change adds minimal TLS configuration API for webhooks : TLSMaxVersion, TLSCipherSuites, and TLSCurvePreferences fields to webhook.Options for granular TLS control

- TLSMaxVersion: enforce Modern profile
- TLSCipherSuites: custom cipher suites
- TLSCurvePreferences: elliptic curve configuration

Changes:
- Add TLSMaxVersion, TLSCipherSuites, and TLSCurvePreferences fields to webhook.Options 
- Added unit tests for all new fields
- Added documentation  for webhook tls (README)
- Maintains backward compatibility with existing TLSMinVersion usage

- :gift: Add new feature
- :bug: Fix bug

/kind enhancement

Fixes #3299

**Release Note**

```release-note
Enhanced webhook TLS configuration with support for max version, custom cipher suites, and curve preferences.
```

**Docs**

```docs

```

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -52,6 +52,21 @@ type Options struct {
 	// TLS 1.3 is the minimum version if not specified otherwise.
 	TLSMinVersion uint16
 
+	// TLSMaxVersion contains the maximum TLS version that is acceptable.
+	// If not set (0), the maximum version supported by the implementation will be used.
+	// This is useful for enforcing Modern profile (TLS 1.3 only) by setting both
+	// TLSMinVersion and TLSMaxVersion to tls.VersionTLS13.
+	TLSMaxVersion uint16
+
+	// TLSCipherSuites specifies the list of enabled cipher suites.
+	// If empty, a default list of secure cipher suites will be used.
+	// Note: Cipher suites are not configurable in TLS 1.3; they are determined by the implementation.
+	TLSCipherSuites []uint16
+
+	// TLSCurvePreferences specifies the elliptic curves that will be used in an ECDHE handshake.
+	// If empty, the default curves will be used.
+	TLSCurvePreferences []tls.CurveID
+
 	// ServiceName is the service name of the webhook.
 	ServiceName string
 
@@ -191,7 +206,10 @@ func New(
 
 		//nolint:gosec // operator configures TLS min version (default is 1.3)
 		webhook.tlsConfig = &tls.Config{
-			MinVersion: opts.TLSMinVersion,
+			MinVersion:       opts.TLSMinVersion,
+			MaxVersion:       opts.TLSMaxVersion,
+			CipherSuites:     opts.TLSCipherSuites,
+			CurvePreferences: opts.TLSCurvePreferences,
 
 			// If we return (nil, error) the client sees - 'tls: internal error"
 			// If we return (nil, nil) the client sees - 'tls: no certificates configured'
--- a/webhook/webhook_test.go
+++ b/webhook/webhook_test.go
@@ -124,3 +124,158 @@ func TestTLSMinVersionWebhookOption(t *testing.T) {
 		}
 	})
 }
+
+func TestTLSMaxVersionWebhookOption(t *testing.T) {
+	opts := newDefaultOptions()
+	t.Run("when TLSMaxVersion is not configured, default is used", func(t *testing.T) {
+		wh, err := newAdmissionControllerWebhook(t, opts)
+		if err != nil {
+			t.Fatal("Unexpected error", err)
+		}
+
+		if wh.tlsConfig != nil && wh.tlsConfig.MaxVersion != 0 {
+			t.Errorf("Expected MaxVersion to be 0 (default), got %d", wh.tlsConfig.MaxVersion)
+		}
+	})
+	t.Run("when TLSMaxVersion is configured to TLS 1.3", func(t *testing.T) {
+		opts.TLSMaxVersion = tls.VersionTLS13
+		wh, err := newAdmissionControllerWebhook(t, opts)
+		if err != nil {
+			t.Fatal("Unexpected error", err)
+		}
+		if wh.tlsConfig == nil {
+			t.Fatal("Expected tlsConfig to be set")
+		}
+		if wh.tlsConfig.MaxVersion != tls.VersionTLS13 {
+			t.Errorf("Expected MaxVersion to be TLS 1.3, got %d", wh.tlsConfig.MaxVersion)
+		}
+	})
+	t.Run("when both TLSMinVersion and TLSMaxVersion are TLS 1.3 (Modern profile)", func(t *testing.T) {
+		opts.TLSMinVersion = tls.VersionTLS13
+		opts.TLSMaxVersion = tls.VersionTLS13
+		wh, err := newAdmissionControllerWebhook(t, opts)
+		if err != nil {
+			t.Fatal("Unexpected error", err)
+		}
+		if wh.tlsConfig == nil {
+			t.Fatal("Expected tlsConfig to be set")
+		}
+		if wh.tlsConfig.MinVersion != tls.VersionTLS13 {
+			t.Errorf("Expected MinVersion to be TLS 1.3, got %d", wh.tlsConfig.MinVersion)
+		}
+		if wh.tlsConfig.MaxVersion != tls.VersionTLS13 {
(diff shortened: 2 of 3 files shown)


vdemeester:
cc @dprotaso

webhook/webhook.go
```diff
@@ -52,6 +52,12 @@ type Options struct {
 	// TLS 1.3 is the minimum version if not specified otherwise.
 	TLSMinVersion uint16
 
+	// TLSConfig allows for full customization of the TLS configuration used by the webhook server.
+	// Note: When providing TLSConfig, ensure MinVersion is set appropriately for security.
+	// If MinVersion is not set in the provided TLSConfig, it will default to Go's default
+	// (currently TLS 1.2), which may be less secure than the webhook's default of TLS 1.3.
+	TLSConfig *tls.Config
```
twoGiants: **First point** I want to mention.

Not sure if providing access to the full `tls.Config` API is necessary. I prefer minimal API where possible, but I am not a TLS expert and maybe that is the call here BUT hear me out :smiley_cat: : 

How about expanding the current API to support:
```go
type Options struct {
    TLSMinVersion           uint16        // Existing
    TLSMaxVersion           uint16        // NEW - for Modern profile (TLS 1.3 only)
    TLSCipherSuites         []uint16      // NEW
    TLSCurvePreferences     []tls.CurveID // NEW
    TLSPreferServerCiphers  bool          // NEW - for Old profile
}
``` 

Then all you mentioned in the issue would be covered. But for future needs/changes another code change would be needed => which I personally prefer over a bigger API.

**Second point** I want to mention.

If it is decided to go the route of keeping the current `TLSMinVersion` AND adding the full `tls.Config` like in the PR => I would prefer to deprecating the current `TLSMinVersion` and then switch entirely to the `tls.Config`. Then the min version validation for >= 1.2 must be done on the provided `tls.Config` so that the current functionality remains preserved. Deprecation comments then be added here and in the documentation and in the code and we probably need an issue so we don't forget to remove it in the next or later versions. Also I would then enforce the user to be able to use only one of the options => either set `TLSMinVersion` or `tls.Config` but not both.

What do the others say? @vdemeester @dprotaso @creydr @Cali0707

webhook/webhook.go
```diff
@@ -189,38 +195,68 @@ func New(
...
-					logger.Warn("server cert missing")
-					return nil, nil
-				}
-				cert, err := tls.X509KeyPair(serverCert, serverKey)
-				if err != nil {
-					return nil, err
-				}
-				return &cert, nil
-			},
+			webhook.tlsConfig.GetCertificate = getCertificate
```
twoGiants: Maybe bug: Now if the user provides a `GetCertificate` function with his tls.Config it will be overwritten => this is not documented. Is this desired?

webhook/webhook.go
```diff
@@ -189,38 +195,68 @@ func New(
...
+			if err != nil {
+				return nil, err
+			}
+			return &cert, nil
+		}
+
+		// Use custom TLSConfig if provided, otherwise create default config
+		if opts.TLSConfig != nil {
+			// Check if both custom TLSConfig and environment variable are set
+			if envTLSVersion := TLSMinVersionFromEnv(0); envTLSVersion != 0 {
```
twoGiants: With the new configuration the `TLSMinVersion` can remain unset or even be set to some old insecure versions. Currently this is not possible => only TLS 1.2 and 1.3 are allowed and the default is TLS 1.3. There is a warning but the setup continues => I would prefer to return an error if it's lower than 1.2.

And as mentioned above => I would prefer to use one way OR the other and enforce the user to do so too.

webhook/webhook.go
```diff
@@ -189,38 +195,68 @@ func New(
...
+					zap.Uint16("TLSConfig.MinVersion", opts.TLSConfig.MinVersion),
+					zap.Uint16("EnvMinVersion", envTLSVersion))
+			}
+
+			// Clone the provided TLSConfig to avoid modifying the original
+			webhook.tlsConfig = opts.TLSConfig.Clone()
 
+			// Warn if MinVersion is not set in the provided TLSConfig
+			if webhook.tlsConfig.MinVersion == 0 {
+				logger.Warnw("TLSConfig.MinVersion is not set, will use Go's default (TLS 1.2). Consider setting MinVersion explicitly for better security.",
```
twoGiants: The default is 1.3.

webhook/webhook_test.go
```diff
@@ -124,3 +124,112 @@ func TestTLSMinVersionWebhookOption(t *testing.T) {
 		}
 	})
 }
+
+func TestCustomTLSConfigWebhookOption(t *testing.T) {
```
twoGiants: Missing testcase for tls.Config without TLSMinVerision or a lower than 1.2 version => which should not be possible.

webhook/webhook.go
vdemeester: I tend to agree with @twoGiants here 👼🏼

jkhelil: @twoGiants Thank you for the thoughtful review. I appreciate your preference for a minimal API, and I share that hesitation.
I've observed two patterns in the ecosystem:

1. Full api: net/http exposes complete TLSConfig https://pkg.go.dev/net/http#Server
2. Selected fields: Kubernetes apiserver exposes specific fields  https://github.com/kubernetes/apiserver/blob/master/pkg/server/options/serving.go#L66

After reflection, I'm leaning toward the minimal API approach  for the following reasons:
- We can keep the existing TLSMinVersion field without deprecation
- Reduces breaking changes and migration burden
- WEBHOOK_TLS_MIN_VERSION env var continues working as-is(prevents breaking TLS 1.2 clients)
- Users won't  accidentally override certificate loading logic
- Smaller API surface to review and maintain

and it makes the pr much more simple

webhook/webhook.go
jkhelil: With the recent update, this is nolonger relevant

webhook/webhook.go
jkhelil: With the latest changes, this concern is no longer relevant.

webhook/webhook.go
jkhelil: With the latest changes, this concern is no longer relevant.

webhook/webhook_test.go
jkhelil: With the latest changes, this concern is no longer relevant.

dprotaso:
Change looks good - just see the linter warnings about extra whitespace

PR #3308, opened 2026-01-13
Gracefully handle missing shipped ConfigMaps at startup

### LINKED ISSUE #3195: Consider changing sharedmain behaviour about panic'ing when a config map is missing
That's fine.  I'm slighly worried about "halt and catch fire" as a behavior when a configmap we ship has been removed.  The idea of shipping them has mostly been about documentation, not about requiring them to have specific contents (or even be present) for things to function.

_Originally posted by @evankanderson in https://github.com/knative/pkg/pull/3190#discussion_r2169001114_

anjalii-28:
This change updates sharedmain to handle missing shipped ConfigMaps more gracefully
at startup.

Behavior:
-> Missing shipped ConfigMaps are treated as empty (defaults used) instead of panicking
- A warning is logged when the ConfigMap is not found
- Watchers remain active and pick up the ConfigMap if it is created later
- No changes to existing observability restart semantics

Fixes #3195

Testing:
- Unit tests in pkg/injection/sharedmain pass
- Local failures in pkg/test/upgrade/shell are due to macOS Bash 3.2,  CI runs Bash ≥4


configmap/informer/informed_watcher.go
```diff
@@ -199,14 +204,39 @@ func (i *InformedWatcher) registerCallbackAndStartInformer(addConfigMapEvent fun
...
-			if _, ok := i.defaults[k]; ok && k8serrors.IsNotFound(err) {
-				// It is defaulted, so it is OK that it doesn't exist.
+			if k8serrors.IsNotFound(err) {
+				if _, ok := i.defaults[k]; ok {
+					// It is defaulted, so it is OK that it doesn't exist.
+					return nil
+				}
+				// ConfigMap doesn't exist and is not defaulted. Log a warning but don't fail.
+				// The watcher will pick up the ConfigMap if it's created later.
+				log.Printf("WARNING: ConfigMap %q in namespace %q not found, using defaults and watching for creation", k, i.Namespace)
```
dprotaso, line 239: shouldn't use `log` package. we use a zap logger we'll probably want a new constructor that accepts a logger if we want these errors

injection/sharedmain/main.go
```diff
@@ -486,11 +486,16 @@ func SetupConfigMapWatchOrDie(ctx context.Context, logger *zap.SugaredLogger) *c
...
-	} else if !apierrors.IsNotFound(err) {
-		logger.Fatalw("Error reading ConfigMap "+logging.ConfigMapName(), zap.Error(err))
+		cmw.Watch(cmName, logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else if apierrors.IsNotFound(err) {
+		// ConfigMap doesn't exist, but we still register a watcher so updates are picked up if it's created later.
+		logger.Warnw("ConfigMap "+cmName+" not found, using defaults and watching for creation", zap.String("configmap", cmName))
+		cmw.Watch(cmName, logging.UpdateLevelFromConfigMap(logger, atomicLevel, component))
+	} else {
+		logger.Fatalw("Error reading ConfigMap "+cmName, zap.Error(err))
 	}
```
dprotaso, line 499: Seems like we can simply call `Watch` now? and not have these `if` clauses here

injection/sharedmain/main.go
```diff
@@ -513,7 +518,11 @@ func WatchObservabilityConfigOrDie(
 
 	if _, err := client.Get(ctx, cmName, metav1.GetOptions{}); err == nil {
 		cmw.Watch(cmName, observers...)
-	} else if !apierrors.IsNotFound(err) {
+	} else if apierrors.IsNotFound(err) {
+		// ConfigMap doesn't exist, but we still register a watcher so updates are picked up if it's created later.
+		logger.Warnw("ConfigMap "+cmName+" not found, using defaults and watching for creation", zap.String("configmap", cmName))
+		cmw.Watch(cmName, observers...)
+	} else {
```
dprotaso, line 525: Likewise we can simplify this and just call `Watch` and not worry about fetching the config maps?

dprotaso: Also one thing we could do for the observability config map is maybe kill the main context thus triggering a shutdown of the component. Then Kubernetes would restart the pod.

configmap/informer/informed_watcher.go
```diff
@@ -171,6 +172,10 @@ func (i *InformedWatcher) Start(stopCh <-chan struct{}) error {
 		return errors.New("error waiting for ConfigMap informer to sync")
 	}
 
+	// Check which ConfigMaps exist and mark missing ones (without defaults) as done in the synced callback
+	// so we don't wait for them indefinitely
+	i.markMissingConfigMapsAsDone(s)
```
dprotaso, line 177: This behaviour should probably be opt-in for now using an option. Someone might be relying `checkObservedResourcesExist` to return the error and we don't want to break that.

configmap/informer/informed_watcher.go
```diff
@@ -199,14 +204,39 @@ func (i *InformedWatcher) registerCallbackAndStartInformer(addConfigMapEvent fun
...
+	// so we don't wait for them indefinitely
+	i.ForEach(func(k string, _ []configmap.Observer) error {
+		if _, err := i.informer.Lister().ConfigMaps(i.Namespace).Get(k); err != nil {
+			if k8serrors.IsNotFound(err) {
+				if _, ok := i.defaults[k]; !ok {
+					// ConfigMap doesn't exist and is not defaulted. Mark it as done
+					// so we don't wait for it. The watcher will pick it up if it's created later.
+					s.MarkKeyAsDone(k)
+				}
+			}
```
dprotaso, line 220: should probably return an error if `Get` returns something besides `IsNotFound` no?

configmap/informer/informed_watcher_test.go
```diff
@@ -269,11 +266,9 @@ func TestWatchMissingFailsOnStart(t *testing.T) {
 			defer close(stopCh)
 			err := cmw.Start(stopCh)
 
-			switch {
-			case err == nil:
-				t.Fatal("Failed to start InformedWatcher =", err)
-			case err.Error() != tc.expectErr:
-				t.Fatal("Unexpected error =", err)
+			// Missing ConfigMaps should not cause Start to fail
+			if err != nil {
```
dprotaso, line 270: revert this - we shouldn't change the default behaviour

dprotaso: We should add a separate test for the new behaviour

injection/sharedmain/main_test.go
```diff
@@ -121,3 +135,148 @@ func TestWithObservabilityConfig(t *testing.T) {
...
+	kc := fakekubeclientset.NewSimpleClientset()
+	ctx := logtesting.TestContextWithLogger(t)
+	// Set up context with the fake client using the same key as the injection system
+	ctx = context.WithValue(ctx, kubeclient.Key{}, kc)
+
+	// Create watcher directly (similar to SetupConfigMapWatchOrDie but with direct client)
+	cmw := cminformer.NewInformedWatcher(kc, system.Namespace())
+
+	// Track if the update handler was invoked
+	var handlerInvoked sync.WaitGroup
```
dprotaso, line 208: It would be simpler to just have a channel that returns `ConfigMaps`

injection/sharedmain/main_test.go
```diff
@@ -121,3 +135,148 @@ func TestWithObservabilityConfig(t *testing.T) {
...
+	// Start the watcher
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+
+	if err := cmw.Start(stopCh); err != nil {
+		t.Fatalf("cmw.Start() should succeed even with missing ConfigMap, got error: %v", err)
+	}
+
+	// Give the watcher time to fully start and sync
+	time.Sleep(500 * time.Millisecond)
```
dprotaso, line 233: We should probably use the config map watchers `Wait` method here

injection/sharedmain/main_test.go
dprotaso: Then you could drop `handlerMutex` as well

injection/sharedmain/main_test.go
```diff
@@ -121,3 +135,148 @@ func TestWithObservabilityConfig(t *testing.T) {
...
+		handlerInvoked.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// Handler was invoked, verify the ConfigMap
+		handlerMutex.Lock()
+		defer handlerMutex.Unlock()
+		if receivedConfigMap == nil {
```
dprotaso, line 263: We use `cmp.Diff` to make this simpler - see other examples in the codebase

PR #3316, opened 2026-02-09
Bump k8s libs to 1.35

creydr:
Next knative release (1.22) will require 1.34 as the min version (see [Release-Schedule](https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md)). So bumping to 1.35 libs since it supports k8s 1.34 and 1.35.

### DIFF
--- a/apis/duck/typed.go
+++ b/apis/duck/typed.go
@@ -54,14 +54,13 @@ func (dif *TypedInformerFactory) Get(ctx context.Context, gvr schema.GroupVersio
 	}
 
 	listObj := dif.Type.GetListType()
-	lw := &cache.ListWatch{
+	rawLW := &cache.ListWatch{
 		ListWithContextFunc:  asStructuredLister(dif.Client.Resource(gvr).List, listObj),
 		WatchFuncWithContext: AsStructuredWatcher(dif.Client.Resource(gvr).Watch, dif.Type),
 	}
-	// Wrap with detection for WatchList semantics support (K8s 1.35+)
-	// This is backward compatible - for K8s <1.35, the wrapper just passes through
-	wrappedLW := wrapListWatchWithWatchListDetection(lw, dif.Client)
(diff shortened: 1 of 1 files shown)


dprotaso:
FYI - unit tests failed

creydr:
> FYI - unit tests failed

I addressed it in 4785202bc1a65a8f9a1a44e66910b09b280c97d6. Can you check on this?

dprotaso:
@creydr when we pull this into serving what's the fix? Cause the unit tests are failing there too with this change

dprotaso:
@creydr see the net-contour downstream test - it seems like you deleted something that it uses from the packages?

cc @dsimansk for the client - something with your work file causes issues here

go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/pkg
 
-go 1.24.0
```
dprotaso, line 3: curious did you bump this manually?

creydr: no. Came after `go mod tidy`

creydr: k8s 1.35 seems to be on 1.25 too: https://github.com/kubernetes/kubernetes/blob/de51841dbf70f7cd9f78f3397bfc6a0097a6f048/go.mod#L9

creydr:
> @creydr see the net-contour downstream test - it seems like you deleted something that it uses from the packages?

Seems contour is [pinned](https://github.com/knative-extensions/net-contour/blob/51e6edb315c4a2f68ded2a0de4893165d5928dd8/go.mod#L10) to k8s 1.33. I tried to make the fix backwards compatible

dprotaso:
Oh ok - I can unpin contour and do a manual bump

dprotaso:
I unpinned contour (https://github.com/knative-extensions/net-contour/pull/1243) - can you see if it's possible to drop your work around -

PR #3317, opened 2026-02-09
Bump k8s min version to 1.34

creydr:
Next knative release (1.22) will require 1.34 as the min version (see [Release-Schedule](https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md))

### DIFF
--- a/version/version_test.go
+++ b/version/version_test.go
@@ -40,16 +40,16 @@ func TestVersionCheck(t *testing.T) {
 		wantError       bool
 	}{{
 		name:          "greater version (patch)",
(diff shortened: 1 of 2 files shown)


version/version_test.go
```diff
@@ -63,13 +63,13 @@ func TestVersionCheck(t *testing.T) {
 		actualVersion: &testVersioner{version: "v1.34.0"},
```
dprotaso: this needs to be v1.35

creydr: oh yes. Updated

PR #3321, opened 2026-02-17
ignore deprecated client warnings

dprotaso:


linkvt:
/lgtm

I guess we should look into migrating away from custom controllers at some point in time?

PR #3322, opened 2026-02-24
Add TLS support to Prometheus metrics server

jkhelil:
This PR adds native TLS/HTTPS support to the Prometheus metrics server in `knative.dev/pkg/observability/metrics/prometheus`, enabling secure metrics collection without requiring additional sidecars or proxies.

### Problem
Many production environments require encrypted metrics endpoints for compliance and security, but the current Prometheus server only supports plain HTTP. Existing solutions have significant drawbacks:

1. **kube-rbac-proxy limitations:**
   - Deprecated and unmaintained (archived by Brancz in 2023)
   - Requires sidecar container for each component (15+ containers in multi-component operators)
   - Increased memory overhead and operational complexity
   - Additional maintenance burden for certificate rotation

2. **Post-Quantum Cryptography (PQC) Readiness:**
   - Organizations need to configure TLS settings (MinVersion, MaxVersion, CipherSuites) to prepare for PQC migration
   - Current solution provides no programmatic control over TLS configuration
   
### Solution
Native TLS support in the Prometheus server provides:
- Zero additional containers or sidecars
- Minimal memory footprint
- Built-in support for all Knative components
- Full programmatic control over TLS configuration (PQC-ready)
- Simple certificate file-based configuration for basic use cases
- Backward compatible (plain HTTP when TLS not configured)

# Changes

- Add certFile and keyFile on the server (populated from env). 
Env vars: METRICS_PROMETHEUS_TLS_CERT, METRICS_PROMETHEUS_TLS_KEY for cert/key paths; TLS settings (MinVersion, MaxVersion, CipherSuites, CurvePreferences) via METRICS_PROMETHEUS_TLS_* using knative.dev/pkg/network/tls.DefaultConfigFromEnv("METRICS_PROMETHEUS_").

- Add mTLS (client certificate verification) via two new env vars:
METRICS_PROMETHEUS_TLS_CLIENT_AUTH: controls client cert policy (none, optional, require).
METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE: path to the CA bundle used to verify client certificates (e.g. the 

- API: ListenAndServe() (plain HTTP), ListenAndServeTLS(certFile, keyFile) (TLS), and Serve() (use TLS when both cert and key are set, otherwise plain HTTP).

- Validation: In validate(), require both cert and key set or both unset; return a clear error otherwise (no silent fallback to plain HTTP when only one is set). Also validates mTLS configuration: clientAuth must be one of none, optional, or require; setting mTLS env vars without TLS enabled is an error; optional or require without a client CA file is an error.

- Godoc updated for TLS and mTLS behavior and env vars.

- observability/metrics/prometheus_enabled.go
Call server.Serve() so TLS vs plain is chosen from config (e.g. when using sharedmain and env-based config). Check NewServer error before starting the serve goroutine to prevent nil-pointer panics. Ignore http.ErrServerClosed when logging server errors so graceful shutdown is not reported as an error.

- observability/metrics/prometheus/server_test.go
Tests for TLS from env (cert/key + METRICS_PROMETHEUS_TLS_MIN_VERSION and related vars). 

# Configuration
- Cert/key (required for TLS): METRICS_PROMETHEUS_TLS_CERT, METRICS_PROMETHEUS_TLS_KEY (file paths). Both must be set or both unset.

- TLS parameters (optional, via network/tls): METRICS_PROMETHEUS_TLS_MIN_VERSION, METRICS_PROMETHEUS_TLS_MAX_VERSION, METRICS_PROMETHEUS_TLS_CIPHER_SUITES, METRICS_PROMETHEUS_TLS_CURVE_PREFERENCES. Defaults (e.g. TLS 1.3) apply when not set.

- Client authentication (optional, requires TLS): METRICS_PROMETHEUS_TLS_CLIENT_AUTH (none, optional, require), METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE (path to CA bundle for verifying client certs). optional or require requires METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE to be set. Setting either without TLS cert/key is a validation error.

- Host/port: Existing METRICS_PROMETHEUS_HOST, METRICS_PROMETHEUS_PORT unchanged.
   

/kind enhancement

Fixes #

**Release Note**

```release-note
The Prometheus metrics server in knative.dev/pkg now supports native HTTPS and mTLS. TLS is configured via environment variables: set METRICS_PROMETHEUS_TLS_CERT and METRICS_PROMETHEUS_TLS_KEY for certificate and key file paths, and optionally METRICS_PROMETHEUS_TLS_MIN_VERSION, METRICS_PROMETHEUS_TLS_MAX_VERSION, METRICS_PROMETHEUS_TLS_CIPHER_SUITES, and METRICS_PROMETHEUS_TLS_CURVE_PREFERENCES (using knative.dev/pkg/network/tls) for Post-Quantum Cryptography (PQC)–ready tuning. For mutual TLS (mTLS), set METRICS_PROMETHEUS_TLS_CLIENT_AUTH (none, optional, or require) and METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE (path to the CA bundle used to verify client certificates). This enables operators to enforce client certificate authentication on the metrics endpoint,
```

**Docs**

```docs

```

### DIFF
--- a/observability/metrics/prometheus/server.go
+++ b/observability/metrics/prometheus/server.go
@@ -18,30 +18,41 @@ package prometheus
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net"
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	knativetls "knative.dev/pkg/network/tls"
 )
 
 const (
 	defaultPrometheusPort            = "9090"
-	defaultPrometheusReportingPeriod = 5
 	maxPrometheusPort                = 65535
 	minPrometheusPort                = 1024
 	defaultPrometheusHost            = "" // IPv4 and IPv6
 	prometheusPortEnvName            = "METRICS_PROMETHEUS_PORT"
 	prometheusHostEnvName            = "METRICS_PROMETHEUS_HOST"
+	prometheusTLSCertEnvName         = "METRICS_PROMETHEUS_TLS_CERT"
+	prometheusTLSKeyEnvName          = "METRICS_PROMETHEUS_TLS_KEY"
+	prometheusTLSClientAuthEnvName   = "METRICS_PROMETHEUS_TLS_CLIENT_AUTH"
+	prometheusTLSClientCAFileEnvName = "METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE"
+	// used with network/tls.DefaultConfigFromEnv. E.g. METRICS_PROMETHEUS_TLS_MIN_VERSION.
+	prometheusTLSEnvPrefix = "METRICS_PROMETHEUS_"
 )
 
 type ServerOption func(*options)
 
 type Server struct {
-	http *http.Server
+	http     *http.Server
+	certFile string
+	keyFile  string
 }
 
 func NewServer(opts ...ServerOption) (*Server, error) {
@@ -56,9 +67,25 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 
 	envOverride(&o.host, prometheusHostEnvName)
 	envOverride(&o.port, prometheusPortEnvName)
+	envOverride(&o.certFile, prometheusTLSCertEnvName)
+	envOverride(&o.keyFile, prometheusTLSKeyEnvName)
+	envOverride(&o.clientAuth, prometheusTLSClientAuthEnvName)
+	envOverride(&o.clientCAFile, prometheusTLSClientCAFileEnvName)
 
 	if err := validate(&o); err != nil {
 		return nil, err
+	}
+
+	var tlsConfig *tls.Config
+	if o.certFile != "" && o.keyFile != "" {
+		cfg, err := knativetls.DefaultConfigFromEnv(prometheusTLSEnvPrefix)
+		if err != nil {
+			return nil, err
+		}
+		if err := applyPrometheusClientAuth(cfg, &o); err != nil {
+			return nil, err
+		}
+		tlsConfig = cfg
 	}
 
 	mux := http.NewServeMux()
@@ -68,16 +95,34 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 
 	return &Server{
 		http: &http.Server{
-			Addr:    addr,
-			Handler: mux,
+			Addr:      addr,
+			Handler:   mux,
+			TLSConfig: tlsConfig,
 			// https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6
 			ReadHeaderTimeout: 5 * time.Second,
 		},
+		certFile: o.certFile,
+		keyFile:  o.keyFile,
 	}, nil
 }
 
-func (s *Server) ListenAndServe() {
-	s.http.ListenAndServe()
+// ListenAndServe starts the metrics server on plain HTTP.
+func (s *Server) ListenAndServe() error {
+	return s.http.ListenAndServe()
+}
+
+// ListenAndServeTLS starts the metrics server on TLS (HTTPS) using the given certificate and key files.
+func (s *Server) ListenAndServeTLS(certFile, keyFile string) error {
+	return s.http.ListenAndServeTLS(certFile, keyFile)
+}
+
+// Serve starts the metrics server, choosing TLS or plain HTTP based on the server configuration.
+// If both METRICS_PROMETHEUS_TLS_CERT and METRICS_PROMETHEUS_TLS_KEY are set, it calls ListenAndServeTLS
+func (s *Server) Serve() error {
+	if s.certFile != "" && s.keyFile != "" {
+		return s.http.ListenAndServeTLS(s.certFile, s.keyFile)
+	}
+	return s.http.ListenAndServe()
 }
 
 func (s *Server) Shutdown(ctx context.Context) error {
@@ -85,8 +130,12 @@ func (s *Server) Shutdown(ctx context.Context) error {
 }
 
 type options struct {
-	host string
-	port string
+	host         string
+	port         string
+	certFile     string
+	keyFile      string
+	clientAuth   string
+	clientCAFile string
 }
 
 func WithHost(host string) ServerOption {
@@ -113,6 +162,33 @@ func validate(o *options) error {
 			port, minPrometheusPort, maxPrometheusPort)
 	}
 
+	if (o.certFile != "" && o.keyFile == "") || (o.certFile == "" && o.keyFile != "") {
+		return fmt.Errorf("both %s and %s must be set or neither", prometheusTLSCertEnvName, prometheusTLSKeyEnvName)
+	}
+
+	tlsEnabled := o.certFile != "" && o.keyFile != ""
+	auth := strings.TrimSpace(strings.ToLower(o.clientAuth))
+
+	if auth != "" && auth != "none" && auth != "optional" && auth != "require" {
+		return fmt.Errorf("invalid %s %q: must be %q, %q, or %q",
+			prometheusTLSClientAuthEnvName, o.clientAuth, "none", "optional", "require")
+	}
+
+	if !tlsEnabled && ((auth != "" && auth != "none") || o.clientCAFile != "") {
+		return fmt.Errorf("%s and %s require TLS to be enabled (%s and %s must be set)",
+			prometheusTLSClientAuthEnvName, prometheusTLSClientCAFileEnvName, prometheusTLSCertEnvName, prometheusTLSKeyEnvName)
+	}
+
+	if tlsEnabled && (auth == "optional" || auth == "require") && strings.TrimSpace(o.clientCAFile) == "" {
+		return fmt.Errorf("%s must be set when %s is %q (client certs cannot be validated without a CA)",
+			prometheusTLSClientCAFileEnvName, prometheusTLSClientAuthEnvName, auth)
+	}
+
+	if tlsEnabled && (auth == "" || auth == "none") && strings.TrimSpace(o.clientCAFile) != "" {
+		return fmt.Errorf("%s is set but %s is %q; set %s to %q or %q to use client certificate verification",
+			prometheusTLSClientCAFileEnvName, prometheusTLSClientAuthEnvName, auth, prometheusTLSClientAuthEnvName, "optional", "require")
+	}
+
 	return nil
 }
 
@@ -121,4 +197,37 @@ func envOverride(target *string, envName string) {
 	if val != "" {
 		*target = val
 	}
+}
+
+// applyPrometheusClientAuth configures mTLS (client certificate verification) on cfg.
+// o.clientAuth and o.clientCAFile are populated from env vars; validate() has already checked them.
+func applyPrometheusClientAuth(cfg *tls.Config, o *options) error {
+	v := strings.TrimSpace(strings.ToLower(o.clientAuth))
+	if v == "" || v == "none" {
+		return nil
+	}
+
+	var clientAuth tls.ClientAuthType
+	switch v {
+	case "optional":
+		clientAuth = tls.VerifyClientCertIfGiven
+	case "require":
+		clientAuth = tls.RequireAndVerifyClientCert
+	}
+
+	caFile := strings.TrimSpace(o.clientCAFile)
+	if caFile != "" {
+		pem, err := os.ReadFile(caFile)
+		if err != nil {
+			return fmt.Errorf("reading %s: %w", prometheusTLSClientCAFileEnvName, err)
+		}
+		pool := x509.NewCertPool()
+		if !pool.AppendCertsFromPEM(pem) {
+			return fmt.Errorf("no valid CA certificates found in %s", prometheusTLSClientCAFileEnvName)
+		}
+		cfg.ClientCAs = pool
+	}
+
+	cfg.ClientAuth = clientAuth
+	return nil
 }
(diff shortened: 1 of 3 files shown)


observability/metrics/prometheus/server.go
```diff
@@ -36,12 +37,16 @@ const (
 	defaultPrometheusHost            = "" // IPv4 and IPv6
 	prometheusPortEnvName            = "METRICS_PROMETHEUS_PORT"
 	prometheusHostEnvName            = "METRICS_PROMETHEUS_HOST"
+	prometheusTLSCertEnvName         = "METRICS_TLS_CERT"
```
dprotaso:
```suggestion
	prometheusTLSCertEnvName         = ""METRICS_PROMETHEUS_TLS_CERT"
```

observability/metrics/prometheus/server.go
```diff
@@ -36,12 +37,16 @@ const (
...
+	prometheusTLSKeyEnvName          = "METRICS_TLS_KEY"
```
dprotaso:
```suggestion
	prometheusTLSKeyEnvName          = "METRICS_PROMETHEUS_TLS_KEY"
```

observability/metrics/prometheus/server.go
```diff
@@ -122,3 +153,21 @@ func envOverride(target *string, envName string) {
...
+}
+
+// TLSConfigFromContext retrieves the TLS configuration from the context.
+// Returns nil if no TLS configuration was set.
+func TLSConfigFromContext(ctx context.Context) *tls.Config {
+	if cfg, ok := ctx.Value(tlsConfigKey{}).(*tls.Config); ok {
+		return cfg
+	}
+	return nil
+}
```
dprotaso, line 233: I'd say remove this - since we have the server options. I generally am mixed on passing so much config through the context (even though we do this a lot in the project)

```suggestion
```

observability/metrics/prometheus/server.go
```diff
@@ -68,25 +75,33 @@ func NewServer(opts ...ServerOption) (*Server, error) {
...
+		keyFile:  o.keyFile,
 	}, nil
 }
 
-func (s *Server) ListenAndServe() {
-	s.http.ListenAndServe()
+func (s *Server) ListenAndServe() error {
+	if s.http.TLSConfig != nil || (s.certFile != "" && s.keyFile != "") {
+		return s.http.ListenAndServeTLS(s.certFile, s.keyFile)
+	}
```
dprotaso: Is there a case when a TLS config would not be used with `ListenAndServeTLS` ?

observability/metrics/prometheus_enabled.go
```diff
@@ -54,10 +54,17 @@ func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownF
 		)
 	}
 
+	// Check for TLSConfig in context (highest priority)
+	if tlsConfig := prometheus.TLSConfigFromContext(ctx); tlsConfig != nil {
+		opts = append(opts, prometheus.WithTLSConfig(tlsConfig))
+	}
+
```
dprotaso: Drop context check here.

If anything we should add options to `Config` if env vars overrides aren't enough - but it seems like they should be since we namespaced them

dprotaso (commented):
You'll need to gofumpt the files you are touching (https://github.com/mvdan/gofumpt) to satisfy the linter

observability/metrics/prometheus/server.go
```diff
@@ -68,25 +75,33 @@ func NewServer(opts ...ServerOption) (*Server, error) {
...
 			ReadHeaderTimeout: 5 * time.Second,
 		},
+		certFile: o.certFile,
+		keyFile:  o.keyFile,
 	}, nil
 }
 
-func (s *Server) ListenAndServe() {
-	s.http.ListenAndServe()
+func (s *Server) ListenAndServe() error {
```
dprotaso, line 110: We should add appropriate godoc to highlight that with the tls and cert/key options this will listen on TLS

observability/metrics/prometheus/server.go
```diff
@@ -68,25 +75,33 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 
 	return &Server{
 		http: &http.Server{
-			Addr:    addr,
-			Handler: mux,
-			// https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6
```
twoGiants, line 73: I would probably leave this comment here.

observability/metrics/prometheus_enabled.go
```diff
@@ -54,10 +54,17 @@ func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownF
...
 	server, err := prometheus.NewServer(opts...)
```
twoGiants, line 59: Existing error: missing error check. If `NewServer` fails, server is nil but the go routine below executes anyway and panics.

observability/metrics/prometheus/server.go
```diff
@@ -68,25 +75,33 @@ func NewServer(opts ...ServerOption) (*Server, error) {
...
 		},
+		certFile: o.certFile,
+		keyFile:  o.keyFile,
 	}, nil
 }
 
-func (s *Server) ListenAndServe() {
-	s.http.ListenAndServe()
+func (s *Server) ListenAndServe() error {
+	if s.http.TLSConfig != nil || (s.certFile != "" && s.keyFile != "") {
```
twoGiants: This condition will silently fallback to `s.http.ListenAndServe()` if `certFile` or `keyFile` are not set. Don't think that is intended, I'd rather return an error that both must be set or none.

You could validate in the [`validate`](https://github.com/knative/pkg/blob/57e664f0ea404e8087c47f2826b3afaeb8e1959e/observability/metrics/prometheus/server.go#L104) function where other options are validated.

observability/metrics/prometheus/server.go
twoGiants: nit

This one is unused. Can be removed.

observability/metrics/prometheus_enabled.go
```diff
@@ -54,10 +54,17 @@ func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownF
...
 
 	go func() {
-		server.ListenAndServe()
+		if err := server.ListenAndServe(); err != nil {
```
twoGiants: Filter for `http.ErrServerClosed` which is returned on graceful shutdown which is not an error. See [injection/sharedmain/main.go](https://github.com/knative/pkg/blob/57e664f0ea404e8087c47f2826b3afaeb8e1959e/injection/sharedmain/main.go#L353) for example.


```suggestion
		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```

observability/metrics/prometheus/server.go
twoGiants, line 110: `ListenAndServe` is a misleading name it actually also serves TLS traffic. Lets add a docs comment to the method signature stating that it can also serve TLS if TLS config or certificates are provided.

--- 
Random thoughts: _I would actually prefer to separate `ListenAndServe` and `ListenAndServeTLS` like it is done by `net/http` own API design. The caller should make the choice which method to use depending on the options it's having. But that would have bigger implications on the refactoring => we would need to pull out the env var checks to the caller and **I don't think it's worth it here**._

dprotaso: > I don't think it's worth it here.

Yeah that's the big issue - it pushes the ownership on the caller to pick TLS. The original intent was to have `ListenAndServe` mimic the `http.Server`

What we _could_ do is have `ListenAndServe` - `ListenAndServeTLS` and a new method that auto picks TLS if the settings are present. eg. `ListenAndServeAutoTLS` ? Though unsure about the name is best fits

twoGiants: > What we could do is have ListenAndServe - ListenAndServeTLS and a new method that auto picks TLS if the settings are present.

Yes, that would be quite close to what I was talking about but it means expanding the public API by 2 public methods. I wouldn't mind, it's a good idea :smile_cat:

observability/metrics/prometheus_enabled.go
dprotaso: @jkhelil mind adding a fix for this?

dprotaso:
Added a comment - https://github.com/knative/pkg/pull/3322#discussion_r2926206599

observability/metrics/prometheus/server.go
```diff
@@ -71,22 +98,44 @@ func NewServer(opts ...ServerOption) (*Server, error) {
 			Addr:    addr,
 			Handler: mux,
 			// https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6
```
twoGiants, line 101: This comment is for line 102.

observability/metrics/prometheus_enabled.go
```diff
@@ -20,16 +20,18 @@ package metrics
...
+	"net/http"
 
 	"github.com/prometheus/otlptranslator"
 	otelprom "go.opentelemetry.io/otel/exporters/prometheus"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"knative.dev/pkg/observability/metrics/prometheus"
 )
 
-func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownFunc, error) {
+func buildPrometheus(ctx context.Context, cfg Config) (sdkmetric.Reader, shutdownFunc, error) {
```
twoGiants: nit: Can stay blank if it's not used or use it below in line 66 and get the logger from ctx and log an error.



```suggestion
func buildPrometheus(_ context.Context, cfg Config) (sdkmetric.Reader, shutdownFunc, error) {
```

observability/metrics/prometheus/server.go
```diff
@@ -18,30 +18,41 @@ package prometheus
...
+	defaultPrometheusPort          = "9090"
+	maxPrometheusPort              = 65535
+	minPrometheusPort              = 1024
+	defaultPrometheusHost          = "" // IPv4 and IPv6
+	prometheusPortEnvName          = "METRICS_PROMETHEUS_PORT"
+	prometheusHostEnvName          = "METRICS_PROMETHEUS_HOST"
+	prometheusTLSCertEnvName       = "METRICS_PROMETHEUS_TLS_CERT"
+	prometheusTLSKeyEnvName        = "METRICS_PROMETHEUS_TLS_KEY"
+	prometheusTLSClientAuthEnvName = "METRICS_PROMETHEUS_TLS_CLIENT_AUTH"
+	prometheusTLSClientCAFileEnv   = "METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE"
```
twoGiants:
```suggestion
	prometheusTLSClientCAFileEnvName   = "METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE"
```

observability/metrics/prometheus/server.go
```diff
@@ -113,6 +162,28 @@ func validate(o *options) error {
...
+	if !tlsEnabled && ((auth != "" && auth != "none") || o.clientCAFile != "") {
+		return fmt.Errorf("%s and %s require TLS to be enabled (%s and %s must be set)",
+			prometheusTLSClientAuthEnvName, prometheusTLSClientCAFileEnv, prometheusTLSCertEnvName, prometheusTLSKeyEnvName)
+	}
+
+	if tlsEnabled && (auth == "optional" || auth == "require") && strings.TrimSpace(o.clientCAFile) == "" {
+		return fmt.Errorf("%s must be set when %s is %q (client certs cannot be validated without a CA)",
+			prometheusTLSClientCAFileEnv, prometheusTLSClientAuthEnvName, auth)
+	}
+
```
twoGiants, line 191: `validate()` allows: TLS cert+key set, clientAuth="none", clientCAFile="/some/path". Then `applyPrometheusClientAuth()` returns early for "none" and the CA file is never loaded.

Operator sets a CA file path but it does nothing which is confusing. Either reject the combination in validation or warn.

I see two scenarios where this could be an issue:
- Sets `METRICS_PROMETHEUS_TLS_CLIENT_CA_FILE=/etc/tls/ca.pem` but forgets to set `METRICS_PROMETHEUS_TLS_CLIENT_AUTH` => expects `mTLS` gets plain `TLS`.
- Sets METRICS_PROMETHEUS_TLS_CLIENT_AUTH=none plus a CA file by mistake and the CA file does nothing, but no error signals the misconfiguration.

This would fix it:

```suggestion
if tlsEnabled && (auth == "" || auth == "none") && strings.TrimSpace(o.clientCAFile) != "" {
    return fmt.Errorf("%s is set but %s is %q; set %s to %q or %q to use client certificate verification",
        prometheusTLSClientCAFileEnv, prometheusTLSClientAuthEnvName, auth, prometheusTLSClientAuthEnvName, "optional", "require")
}
```

observability/metrics/prometheus/server_test.go
```diff
@@ -72,3 +81,249 @@ func TestNewServerFailure(t *testing.T) {
...
+			t.Fatal("expected NewServer to fail with invalid client auth value")
+		}
+		if !strings.Contains(err.Error(), prometheusTLSClientAuthEnvName) {
+			t.Errorf("expected error to mention %s; got %v", prometheusTLSClientAuthEnvName, err)
+		}
+	})
+
+	t.Run("optional with valid client CA file sets ClientAuth and ClientCAs", func(t *testing.T) {
+		caFile := createTempCACertFile(t)
+		t.Cleanup(func() { _ = os.Remove(caFile) })
```
twoGiants: Move this line into `createTempCACertFile`, then you don't have to call it here and below in line 231. Cleanup runs when the test that owns `t` finishes, so it should be fine.

PR #3328, opened 2026-03-04
Replace NewConfigFromEnv with DefaultConfigFromEnv

linkvt:
DefaultConfigFromEnv replaces NewConfigFromEnv by returning a full default tls.Config with overrides from env vars. This avoids specifying e.g. the TLS MinVersion explicitely.
I also simplified the webhook TLS config handling a bit to use the default we provide as a base.

This code wasn't tested yet in consuming repos like serving/...

/kind enhancement

### DIFF
--- a/tls/config.go
+++ b/tls/config.go
@@ -33,22 +33,14 @@ const (
 	CurvePreferencesEnvKey = "TLS_CURVE_PREFERENCES"
 )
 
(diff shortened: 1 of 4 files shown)


linkvt:
/cc @twoGiants 
In case you're interested in this as you're also involved in the other PRs :)

PR #3331, opened 2026-03-09
Move tls package to network/tls, keep aliases for backward compatibility

Fedosin:
# Changes

The TLS configuration package is moved from `tls`/ to `network/tls/` to co-locate it with the rest of the networking code. The old `tls/` package now re-exports all public symbols as deprecated aliases so that existing consumers continue to compile without changes. The webhook package is updated to import from the new location directly.

/kind api-change

Fixes #

**Release Note**

```release-note
The knative.dev/pkg/tls package has been relocated to knative.dev/pkg/network/tls. The old import path continues to work via deprecated aliases but will be removed in a future release. Consumers should update their imports to knative.dev/pkg/network/tls.
```

**Docs**

```docs

```


Fedosin:
Related request: https://github.com/knative/pkg/pull/3324#issuecomment-4019929378

twoGiants:
/cc @dprotaso 

Hi Dave! Can you take a look? Thank you!

Mike moved the implementation as you requested.

Fedosin:
We have to keep deprecated aliases in `tls/` for now. Otherwise tests will fail

dprotaso:
/lgtm
/approve

Once we move over to the new usage can we delete the old package?

PR #3333, opened 2026-03-12
remove deprecated TLS package

dprotaso:
This isn't a breaking change since we've never published this package in a release

PR #3337, opened 2026-03-18
[release-1.21] Add network/tls package for TLS configuration

Fedosin:
# Changes

Backport of the following PRs from main to release-1.21:

- #3324 - feat: add shared tls package for reading TLS config from environment
- #3328 - Replace NewConfigFromEnv with DefaultConfigFromEnv
- #3331 - Move tls package to network/tls, keep aliases for backward compatibility
- #3333 - remove deprecated TLS package

/kind enhancement

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


dprotaso:
What's the intent here are we hoping to pull this into specific repos which we release? If so what bug are we fixing?

Or do you just want these changes in a release branch for other reasons?

dprotaso:
To clarify not looking to block this - just looking for clarity on what's the subsequent plan

Fedosin:
@dprotaso right, I have to add more clarity here... In OpenShift 4.22 it will be mandatory for all components to support configurable TLS options (min/max version, cipher suites, curve preferences). Since OpenShift 4.22 will ship Knative 1.21, we need this TLS configuration support on the release-1.21 branch so that the Knative components can comply with that requirement.

dprotaso:
ok

/lgtm
/approve

PR #3350, opened 2026-05-02
export MinimumVersion() accessor

psschwei:
# Changes

Expose the minimum Kubernetes version Knative is built and tested against via a new exported function, so tools that need the target version before a cluster exists can read it directly instead of hardcoding and hand-bumping it each release.

xref knative-extensions/kn-plugin-quickstart#644

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -36,6 +36,12 @@ const (
 	defaultMinimumVersion = "v1.34.0"
 )
 
+// MinimumVersion returns the minimum Kubernetes version Knative is built
+// and tested against. The returned value has a leading "v" (e.g. "v1.34.0").
+func MinimumVersion() string {
+	return defaultMinimumVersion
+}
+
 func getMinimumVersion() string {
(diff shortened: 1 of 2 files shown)


Cali0707 (approved):
/lgtm
/approve

Cali0707 (approved):
/lgtm
/approve

version/version.go
```diff
@@ -36,6 +36,12 @@ const (
 	defaultMinimumVersion = "v1.34.0"
 )
 
+// MinimumVersion returns the minimum Kubernetes version Knative is built
+// and tested against. The returned value has a leading "v" (e.g. "v1.34.0").
+func MinimumVersion() string {
```
dprotaso, line 41: This is a bit confusing since we also have `getMinimumVersion` below.

Maybe we should just expose `defaultMinimumVersion` as `DefaultKubernetesMinVersion` const

and expose `getMinimumVersion` as `KubernetesMinimumVersion()`

PR #3356, opened 2026-05-21
Add configurable metric attributes deny list

creydr:
Adds a `metrics-attributes-deny` config key to the observability ConfigMap. When set, the listed attribute keys (comma-separated) are stripped from all metric instruments via an OTel SDK View using `attribute.NewDenyKeysFilter`.

This prevents unbounded metric cardinality from attributes like `cloudevents.type` or `messaging.destination.name` that can cause OOM in production.

The deny filter is automatically applied in `sharedmain.SetupObservabilityOrDie`, so all components using `sharedmain` benefit without any component-specific code.

**Example configuration:**
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-observability
data:
  metrics-attributes-deny: "cloudevents.type, messaging.destination.name"
```

### DIFF
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -400,8 +400,8 @@ func SetupObservabilityOrDie(
 	resource := resource.Default(component)
 
 	views := OTelViews(ctx)
-	if len(cfg.Metrics.AttributesDenyList) > 0 {
-		views = append(views, metrics.MetricAttributesDenyFilter(cfg.Metrics.AttributesDenyList))
+	if denyList := cfg.Metrics.AttributesDenyList(); len(denyList) > 0 {
+		views = append(views, metrics.MetricAttributesDenyFilter(denyList))
 	}
 
 	meterProvider, err := metrics.NewMeterProvider(
--- a/injection/sharedmain/main.go
+++ b/injection/sharedmain/main.go
@@ -399,10 +399,15 @@ func SetupObservabilityOrDie(
 
 	resource := resource.Default(component)
 
+	views := OTelViews(ctx)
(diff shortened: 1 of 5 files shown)


dprotaso:
Do you think we should just remove metric attributes with high cardinality? 

cc @Cali0707

creydr:
> Do you think we should just remove metric attributes with high cardinality?
> 
> cc @Cali0707

I thought that this way the user can decide which they want (and also would be a "non-breaking" change that way :thinking: )

Cali0707:
> I thought that this way the user can decide which they want (and also would be a "non-breaking" change that way 🤔 )

@dprotaso I agree with this ^

As an example in many systems `cloudevents.type` would be a well defined, bounded metric (and useful/part of the recommended metrics from the cloudevents community). However in some if they are being generated at runtime they can be unbounded and cause problems.

dprotaso:
I think my concern is that the defaults don't scale. I think I'd rather have the defaults scale and allow folks to opt into things that potentially wouldn't scale.

I did similar changes in serving and disabled metrics that had high cardinality. Eventually I'll be able to add them back when OTel supports removing labels.

But in the same pattern I could allow users to opt into high cardinality metrics.


It's also straight forward to allow custom observability options in downstream repositories - eg. we do this in serving already to support a different set of settings for queue proxy metrics

https://github.com/knative/serving/blob/main/pkg/observability/config.go#L54

dsimansk:
@creydr are you still looking to get this merged?

maschmid:
I suppose if we're OK with the deny list mechanism in principle, while also wanting a more conservative defaults, we could just set some reasonable default for the "metrics-attributes-deny" list.

PR #3358, opened 2026-05-27
Fix memory leaks in expired matchers

dsimansk:
# Changes

- Fix memory leak in metrics' expired matchers

Memory leaks observed in @maschmid long running soak tests:

After pruning expired inexact matchers, the code checked `len(s)` (the exact matchers map) instead of `len(ms)` (the inexact matchers map), and then deleted from i.exact instead of `i.inexact`. This meant empty inexact matcher entries were never cleaned up, causing a memory leak for every label-selector-based tracker reference.

When removing inexact matchers for a deleted object, the code called delete(i.exact, ref) instead of `delete(i.inexact, ref)`, so empty inexact entries accumulated indefinitely and could also incorrectly remove unrelated exact entries.

**Release Note**

```release-note
Fix memory leak in metrics' expired matchers
```

PTAL, it's a follow-up to other memory leak in networking hanging idle connections. 

/cc @dprotaso

### DIFF
--- a/tracker/enqueue.go
+++ b/tracker/enqueue.go
@@ -277,8 +277,8 @@ func (i *impl) GetObservers(obj interface{}) []types.NamespacedName {
 				keys = append(keys, key)
 			}
 		}
-		if len(s) == 0 {
-			delete(i.exact, ref)
+		if len(ms) == 0 {
+			delete(i.inexact, ref)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


linkvt (commented):
/lgtm
/hold in case Dave should also take a look, feel free to unhold

dprotaso:
I don't generally like the tests accessing the internals - but I see the style is just complementing what's already there so I don't want to gate the PR on that

/hold cancel

PR #3361, opened 2026-05-28
[release-1.21] Fix compilation

dprotaso:
newer go versions found this bug and made it a compiler error


evankanderson (approved):
/lgtm
/approve

PR #3368, opened 2026-06-03
[WIP] feat: support multi secret namespace watch

### LINKED ISSUE #3367: informers for multiple namespaces
currently the informers are only cluster wide or namespace specific.

I want to have an option like in controller-runtime to watch specific namespaces. 

wide cluster access to secrets for example is problematic because of security concerns   

I want it for the kafka-controller specifically but this is a general issue.

I already forked the kafka controller and override  the default secret informer from this pacakge, so maybe this is something you would consider adding here.

what i basicly do is:

```
	subs := make([]informers.SharedInformerFactory, 0, len(namespaces))
	for _, ns := range namespaces {
		subs = append(subs, informers.NewSharedInformerFactoryWithOptions(
			client, resync, informers.WithNamespace(ns),
		))
	}
```
and then only for secrets create some wrap informer which maps the namespace to the dedicated namespaced informer

YoniRomm:
# Changes

Created a factory that wraps multiple informers when there is a need to watch multiple specific namespaces, overcoming the core Go library limitation which only allows an informer to watch either cluster-wide or a single namespace.

- :gift: Add new feature

/kind enhancement

Fixes #3367

**Release Note**

```release-note
Added a multi-namespace informer factory to support watching multiple specific namespaces simultaneously.

### DIFF
--- a/injection/context.go
+++ b/injection/context.go
@@ -22,9 +22,13 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-// nsKey is the key that namespaces are associated with on
+// nsKey is the key that a single namespace is associated with on
 // contexts returned by WithNamespaceScope.
 type nsKey struct{}
+
+// nsScopesKey is the key that multiple namespaces are associated with on
+// contexts returned by WithNamespaceScopes.
+type nsScopesKey struct{}
 
 // WithNamespaceScope associates a namespace scoping with the
 // provided context, which will scope the informers produced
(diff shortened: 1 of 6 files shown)


dprotaso:
hmmm... I was under the impression this would leverage some mulit-namespace watch in client-go but I see now we're just creating mulitple per-namespace shared informer factories.

Let me think about this

YoniRomm:
@dprotaso what do you think ?

PR #3387, opened 2026-08-27
chore: Upgrade k8s.io core libs to 1.36 (0.36.4)

dsimansk:
# Changes

- chore: Upgrade k8s.io core libs to 1.36/0.36.4
-->
```release-note
chore: Upgrade k8s.io core libs to 1.36/0.36.4
```

/cc @dprotaso 

/hold


shashankxrm:
I've been looking into #3383 as the upstream prerequisite for Tekton's server-side sharded List/Watch work (tektoncd/pipeline#10610). It looks like this Kubernetes 0.36 / Go 1.26 upgrade is the prerequisite for exposing ListOptions.ShardSelector.

Once this upgrade is settled, I can work on the informer injection plumbing in #3383 as a follow-up. Just mentioning the dependency here for context.
@vdemeester

PR #3390, opened 2026-09-14
Replace two lane queue with controller-runtime priorityqueue

alvaroaleman:
This change replaces the two lane queue with a small wrapper around the controller-runtime priorityqueue. Having a single queue for both guarantees correct deduplication and avoid metrics issues like describes in https://github.com/knative/pkg/pull/3253. It is also much simpler.

/assign @dprotaso 

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


controller/controller.go
```diff
@@ -274,8 +274,8 @@ func (c *Impl) EnqueueSlowKey(key types.NamespacedName) {
 	c.workQueue.AddSlow(key)
 
 	if logger := c.logger.Desugar(); logger.Core().Enabled(zapcore.DebugLevel) {
-		logger.Debug(fmt.Sprintf("Adding to the slow queue %s (depth(total/slow): %d/%d)",
-			safeKey(key), c.workQueue.Len(), c.workQueue.SlowLen()),
+		logger.Debug(fmt.Sprintf("Adding to the slow queue %s (depth: %d)",
```
alvaroaleman, line 277: Not being able to log the lenght of the slow queue is the one behavioral change we can not avoid

